PR #36, opened 2026-05-04
May discard a card, if you do bug

bolav:
There is a bug in the "if you do" on the may discard.

Here is a test to prove the bug, and a solution made by Claude. Don't know if this is the right solution.


wingedsheep:
Fixed by adding a IfYouDo effect

```kotlin
    triggeredAbility {
        trigger = Triggers.Valiant
        effect = MayEffect(
            effect = IfYouDoEffect(
                action = EffectPatterns.discardCards(1),
                ifYouDo = Effects.DrawCards(1)
            ),
            description_override = "You may discard a card. If you do, draw a card."
        )
    }
 ```

bolav:
Nice. Looks much better, and can also be used on Melded Moxite.

PR #51, opened 2026-05-08
Add Final Fantasy set

bolav:
Start on Final Fantasy set, and add the three cards needed for Landfall deck.


wingedsheep:
Included in https://github.com/wingedsheep/argentum-engine/commit/abd50df93fc02cea7578cabe2c50e6b94d87d93a

PR #52, opened 2026-05-08
Ordeal of Nylea (FDN) — fix counter placement and sacrifice trigger

bolav:
Fix two bugs found during implementation:

1. AddCountersExecutor used the stateless context.resolveTarget() overload, which returns null for EffectTarget.EnchantedCreature (needs AttachedToComponent lookup). Switch to context.resolveTarget(effect.target, state).

2. The ConditionalEffect checking for 3+ counters used EntityReference.AffectedEntity, which is only populated during layer/projection — null during trigger resolution. Changed to EntityReference.Triggering, which AttachmentTriggerDetector sets to the enchanted creature when EnchantedCreatureAttacks fires.


wingedsheep:
Added and fixed bugs in https://github.com/wingedsheep/argentum-engine/commit/a249feb4546dafccadfbe5d00e0b3a810a08963b

PR #64, opened 2026-05-09
Add Blue Sun's Twilight (ONE) with ManaValueAtMostX targeting support

bolav:
Introduces CardPredicate.ManaValueAtMostX for targeting restrictions that depend on the spell's X value. At cast-time validation the predicate reads xValue from the cast action via PredicateContext; at resolution it reads from SpellOnStackComponent. During legal-action enumeration (X unknown) it conservatively allows all targets.

xValue is now threaded through TargetValidator.validateTargets → validatePermanentTarget / validateGraveyardTarget → PredicateContext, and populated from EffectContext in PredicateContext.fromEffectContext.

It's still a bit wonky:

```
❯ I still can target creatures with a too low X value, but it now says: "Target does not match filter",
   and then it get's cancelled when casting.

⏺ That's the expected behavior given the engine's architecture. The server correctly rejects the illegal
   combination with a clear error — the cast doesn't go through and no state changes. The UI showing all
   creatures as selectable is an architectural limitation: legal targets are computed when the game
  presents available actions, before the player has chosen X, so the engine has no X value to filter by
  yet.

  To properly filter the UI you'd need a two-step cast flow for X spells — announce the spell + pay mana
   (which determines X), then the server computes filtered legal targets and presents a second targeting
   decision. That's a meaningful engine change to the casting pipeline. The current behavior (server
  rejects + "Target does not match filter" error) is the practical middle ground.

  Want me to implement the two-step flow, or is the server-side rejection acceptable for now?
```

image

image

### DIFF
--- a/backlog/magezero-coverage.md
+++ b/backlog/magezero-coverage.md
@@ -69,7 +69,7 @@ one enchantment (Unholy Annex) are double-faced / modal DFCs and depend on DFC s
 - [x] Tolarian Terror ×3
 
 ### Sorceries (1)
-- [ ] Blue Sun's Twilight ×1
+- [x] Blue Sun's Twilight ×1
 
 ### Instants (25)
 - [x] Consider ×4 *(Innistrad: Midnight Hunt)*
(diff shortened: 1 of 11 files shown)


wingedsheep:
Accidentally implemented the same card

PR #67, opened 2026-05-09
Add Earthbending Lesson (TLA) with Earthbend mechanic

bolav:
Earthbend N: animate target land you control as a 0/0 creature-land with haste (permanent), put N +1/+1 counters on it, and grant it the EARTHBENDED internal keyword so that when it dies or is exiled from the battlefield it returns to the battlefield tapped.

Engine changes:
- Keyword.EARTHBEND (display) + Keyword.EARTHBENDED (internal return marker)
- Effects.Earthbend(amount, target) facade composing AnimateLand + GrantKeyword(HASTE)
  + AddCounters + GrantKeyword(EARTHBENDED)
- DeathAndLeaveTriggerDetector.detectEarthbendReturnTriggers fires on BATTLEFIELD→GRAVEYARD and BATTLEFIELD→EXILE; uses fromZone = event.toZone so the return is skipped if the card is moved out of that zone before the trigger resolves (e.g. exiling the card from the graveyard)
- TriggerDetector wires the new detector into both the death and exile paths

### DIFF
--- a/manual-scenarios/cards/e/earthbending-lesson.json
+++ b/manual-scenarios/cards/e/earthbending-lesson.json
@@ -1,0 +1,42 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Earthbending Lesson", "Earthbending Lesson"],
+    "battlefield": [
+      {"name": "Forest"},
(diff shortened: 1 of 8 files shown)


wingedsheep:
Implemented in https://github.com/wingedsheep/argentum-engine/commit/b18055ffa540e371a6c123043a540f0788117b09

PR #75, opened 2026-05-10
Add Badgermole Cub (TLA) with AdditionalManaFromCreatureTap mechanic

bolav:
New static ability: AdditionalManaFromCreatureTap(color, amount)
- Fires whenever the controller taps a creature for mana (not a land).
- Triggered mana ability — resolves immediately without the stack.
- Implemented by adding resolveAdditionalManaOnCreatureTap() to ActivateAbilityHandler, following the same pattern as the existing resolveAdditionalManaOnLandTap (Lavaleaper) and resolveAdditionalManaOnTap (Elvish Guidance). Only fires when the mana source is a creature.

Card: Badgermole Cub {1}{G} 2/2 — ETB earthbend 1 (targets a land you control), plus the creature-tap mana bonus.

Tests cover: earthbend 1 animates a land into 1/1, +1 {G} on creature tap, no bonus on land tap, no bonus without the Cub.

Mana resolver does not understand that the creatures can give extra mana.


wingedsheep:
Included in https://github.com/wingedsheep/argentum-engine/pull/78

PR #137, opened 2026-05-13
Add Jackal, Genius Geneticist

nymann:
## Summary

Adds **Jackal, Genius Geneticist** ({G}{U} Legendary Creature — Human Scientist Villain, 1/1, Trample) from Marvel's Spider-Man.

> Whenever you cast a creature spell with mana value equal to Jackal's power, copy that spell, except the copy isn't legendary. Then put a +1/+1 counter on Jackal. (The copy becomes a token.)

Two engine adjustments were needed to make the trigger work end-to-end:

- `CopyTargetSpellExecutor` now copies permanent spells (creatures, artifacts, …) — it previously bailed out because they have no `spellEffect`. The `spellEffect` requirement is deferred to the paths that actually need it.
- `StackResolver.enterPermanentOnBattlefield` tags the resulting permanent with `TokenComponent` when it resolved from a `putSpellCopy` spell-copy (CR 707.10f — a copy of a permanent spell becomes a token). Clone-style "enters as a copy" effects (Mockingbird) are distinguished by carrying `originalCardComponent` on their `CopyOfComponent`, which preserves the revert-on-leave rule (CR 400.7 / 707.2).

`CopyTargetSpellEffect.stripSupertypes` is now documented as a no-target-path-only flag.

Also includes:
- A new engine test for the existing Layer 7b CDA used by `*`-power cards (`CharacteristicDefiningPowerEqualToGreatestManaValueAmongPermanentsYouControlTest`) — verifies the basic case and the rule-corner case (an opponent's higher-MV permanent must not count).
- `docs/RULES.md` — an evidence-based catalogue of currently supported rules, intended as planner input.

## Test plan

- [x] `:rules-engine:test` (full)
- [x] `:mtg-sdk:test` (full)
- [x] `:game-server:test` (full — 2044 tests, all green)
- [x] `JackalGeniusGeneticistScenarioTest` — card definition, trigger fires on matching MV (token copy + counter), no trigger on non-matching MV
- [x] `CharacteristicDefiningPowerEqualToGreatestManaValueAmongPermanentsYouControlTest`
- [x] `MockingbirdScenarioTest`, `MischievousQuanarScenarioTest`, all storm/copy-spell scenarios — no regressions

---
Reviewed and forwarded from nymann/argentum-engine#25 (original author: @nymann). The fork PR was titled "Add Kraven, Proud Predator" but in fact added Jackal, Genius Geneticist plus an engine test for the `*`-power CDA mechanic that Kraven would use; Kraven the card is not part of this PR.


wingedsheep:
Reviewed and applied the two important findings as suggested fixups on branch [`review/pr-137-removeLegendary`](https://github.com/wingedsheep/argentum-engine/tree/review/pr-137-removeLegendary) (commit [`8d72615`](https://github.com/wingedsheep/argentum-engine/commit/8d72615235), on top of this PR with `origin/main` merged in to resolve the backlog conflict).

## What was good
- Card composes existing primitives (`Effects.Composite(CopyTargetSpell + AddCounters)` with a `Compare` trigger condition) — no one-off effect type.
- `CopyOfComponent.originalCardComponent == null` is a clean discriminator for "spell-copy → token" vs "Clone-style permanent copies" — reuses existing state, no new component.
- CR 707.10f citation verified correct against the current Comprehensive Rules.
- Layer-7b CDA regression test correctly excludes the opponent's higher-MV permanent.

## Fixes applied

**1. `stripSupertypes` → `removeLegendary` (with `TypeLine.withoutLegendary()`)**

The flag was wiping the full `supertypes` set (Snow, Basic, World, Legendary) when the oracle text only says "except the copy isn't legendary." The engine already has `TypeLine.withoutLegendary()` and the `removeLegendary: Boolean` convention on `CreateTokenCopyOfEquippedCreatureEffect` (Helm of the Host). Renamed for consistency and narrowed the mutation.

**2. Test the "isn't legendary" clause**

The only existing trigger test used Llanowar Elves, which isn't legendary — the flag could have been a no-op and the test would still pass. Added a scenario that casts a 1-MV legendary creature under Jackal and asserts:
- both the original and the token copy survive (legend rule CR 704.5j doesn't kill one),
- `projected.isLegendary(tokenId)` is `false`.

## Not yet addressed (minor, author's discretion)

- The targeted permanent-spell path returns `EffectResult.error(...)` (line 141-143). Could be promoted to a definition-time check, but Jackal uses `TriggeringEntity` so it doesn't hit this path in practice.
- `driveStormModalCopies` silently drops `removeLegendary` and `keywordsForCopy` for the modal-with-targets path. Docstring now describes this; threading the flag is left for the first card that needs it.
- The cross-reference to "CR 707.2" in `CopyOfComponent`'s docstring (pre-existing, not from this PR) is decorative — 707.2 is about copiable values; revert-on-leave is 400.7.

All four `JackalGeniusGeneticistScenarioTest` cases pass; `MockingbirdScenarioTest` (incl. revert-on-leave), `MischievousQuanarScenarioTest`, and the Storm/Harnesser scenarios remain green.

wingedsheep:
Addressed the three minor follow-ups in [`b11db23`](https://github.com/wingedsheep/argentum-engine/commit/b11db23fc) on the same [`review/pr-137-removeLegendary`](https://github.com/wingedsheep/argentum-engine/tree/review/pr-137-removeLegendary) branch:

1. **Targeted permanent-spell copy now works end-to-end.** Dropped the `"Cannot copy target permanent spell with targets"` runtime error in `CopyTargetSpellExecutor`. The retarget continuation (`StormCopyTargetContinuation`) already routes through `putSpellCopy`, which clones the source's `CardComponent` regardless of whether the spell is instant/sorcery or a permanent, and `StackResolver.enterPermanentOnBattlefield` already tags the result with `TokenComponent` per CR 707.10f. `StormCopyTargetContinuation.spellEffect` is now nullable (it was only forwarded, never read, on the permanent-spell path).

2. **`removeLegendary` + `keywordsForCopy` are now threaded through every path.** Both flags live on `StormCopyTargetContinuation` and `StormCopyModalTargetContinuation`; `driveStormModalCopies` takes them and propagates them into the continuation. Post-`putSpellCopy` mutation is consolidated in a single helper, `StormCopyEffectExecutor.applyCopyMutations`, used by all five copy paths (no-target, no-target modal, targeted, targeted-modal, no-legal-targets fizzle loop). Updated the `CopyTargetSpellEffect.removeLegendary` docstring to reflect this.

3. **Fixed the `CopyOfComponent` CR citation.** Dropped the spurious `CR 707.2` cross-reference — 707.2 covers copiable values, not zone-change object identity. Revert-on-leave (Clone, Mockingbird) is governed by CR 400.7 alone.

All previously-passing suites still green: `:rules-engine:test`, `:mtg-sdk:test`, `:game-server:test` (full), and the targeted copy-related scenarios (`Jackal`, `Mockingbird`, `MischievousQuanar`, `HarnesserOfStorms`, storm specs).

PR #154, opened 2026-05-14
Add Starfield Vocalist to Edge of Eternities

bolav:
Adds the EOE Panharmonicon-variant whose oracle text doesn't restrict the entering permanent to ones you control. Extends AdditionalETBTriggers with an `enteringMustBeYouControl` flag (default true, preserving Naban / Traveling Chocobo behavior) and threads it through TriggerDetector's ETB-doubling pass so opponent ETBs that tickle "whenever a creature enters" triggers on your permanents also get doubled.


wingedsheep:
Pushed two follow-up commits to address minor review feedback:

**`f8fc011`** — Merge `origin/main` into the branch. Resolves a conflict in `backlog/sets/edge-of-eternities/cards.md` (PR bumped 168→169, main went 169→173) by bumping to **174 / 261**.

**`54632b9`** — Two changes:

1. **Rename `AdditionalETBTriggers.creatureFilter` → `enteringFilter`.** The field gates any permanent (Starfield Vocalist passes `GameObjectFilter.Any`), not just creatures, so the old name was a slight lie now that the doubler can match non-creature ETBs. Updates the property, the three named-arg callsites (Naban, Traveling Chocobo, Starfield Vocalist), `mtg-sdk/reference.md`, and the `TriggerDetector` comment.

2. **New scenario test — "Vocalist sees its own ETB."** Covers the EOE ruling: *"If a permanent entering the battlefield at the same time as Starfield Vocalist (including Starfield Vocalist itself) causes a triggered ability of a permanent you control to trigger, that ability triggers an additional time."* Setup: Wretched Anurid in play, cast Vocalist; Vocalist's static ability is live by the time its own entrance triggers are detected, so Anurid's "another creature enters" trigger fires twice → 2 life lost.

All Vocalist + Naban scenario tests and the mtg-sdk / mtg-sets suites pass.

PR #155, opened 2026-05-14
Add Timeline Culler with additional-cost and graveyard warp support

bolav:
Extends KeywordAbility.Warp with optional additionalCost and a fromGraveyard opt-in, threads both through the cast flow, and adds auto-payment for AdditionalCost.PayLife so Timeline Culler's "Warp—{B}, Pay 2 life" cost and graveyard-cast clause work end to end.


wingedsheep:
## Review

Behaviorally correct and well-tested; the SDK shape is acceptable but the `fromGraveyard: Boolean` flag on `KeywordAbility.Warp` is the kind of single-card knob that tends to multiply across keywords — worth a conscious decision now whether to keep it on `Warp` or extract it.

I've merged `origin/main` into the branch (one trivial conflict in `backlog/sets/edge-of-eternities/cards.md`: `169` vs `173` implemented → `174` with Timeline Culler checked) and pushed two follow-up commits to `bolav/timeline-culler`. Feel free to revert any of them.

### What's good
- **`PayLife` auto-pay generalizes cleanly.** The pre-pass at `CastSpellHandler.kt:1359-1373` is keyword-agnostic — any future Flashback/Kicker/Adventure with a `PayLife` additional cost gets it for free. Marks `LifeLostThisTurn` and emits `LifeChangedEvent(PAYMENT)`, matching `CostHandler.payAdditionalCost` for `AbilityCost.PayLife`.
- **Excellent regression guard.** The fourth test (Sinister Cryologist from graveyard fails) directly defends the old hand-only contract.
- **Mirrors existing precedent.** `Warp.additionalCost` follows the same shape as `Flashback.additionalCost` / `Kicker.additionalCost`, and the wiring into `allAdditionalCosts` is identical to flashback's.
- **`hasWarpPermission` cleanup.** Replacing the hand-only check with a single function that consults both zones is a strict improvement over branching at every call site.

### Changes pushed
1. **Rename `enumerateWarpFromHand` → `enumerateWarp`** (`CastFromZoneEnumerator.kt:1159`). The method now enumerates both HAND and GRAVEYARD, so the `FromHand` suffix was misleading.
2. **Provisional-flag comment on `Warp.fromGraveyard`** (`KeywordAbility.kt`). Flags that a second card with the same "cast from graveyard using its X ability" wording on a different keyword should trigger extracting a generic cast-from-zone permission rather than duplicating the boolean.
3. **New test** `warp is rejected when player can't afford the life payment` in `TimelineCullerScenarioTest`. Exercises the CR 119.4 validation branch at `CastSpellHandler.kt:1141-1148` (P1 at 1 life cannot warp Timeline Culler; life total stays at 1). All 5 TimelineCuller tests + 9 existing `WarpMechanicTest` cases pass.

### Still open
- **`fromGraveyard` long-term.** With only Timeline Culler, the flag is fine. The right primitive for the next such card is something like `Effects.GrantCastFromGraveyard(filter, viaKeyword)` (cf. existing `MayCastFromZone`/`Forage`/flashback permissions). The new comment marks this.
- **`CR 119.4` citation** — couldn't fully verify online (yawgatog mirror returned truncated content). The number matches the standard rule for "can't pay more life than you have"; worth a quick double-check before merge.
- **Pre-existing**: `CostHandler.canPayAdditionalCost` at line 637 uses `life > cost.amount` (strict) rather than `>=`. Not in this PR's scope, but worth fixing in a follow-up — it would refuse exact-amount life payments through the ability-cost path.

PR #216, opened 2026-05-29
Add Sindbad with DrawRevealDiscardUnless effect

bolav:
Sindbad ({1}{U} 1/1): "{T}: Draw a card and reveal it. If it isn't a land card, discard it." No existing primitive covered draw-reveal-conditional-discard, so this adds a general, reusable effect.

Engine change:
- DrawRevealDiscardUnlessEffect (mtg-sdk) + Effects.DrawRevealDiscardUnless facade.
- DrawRevealDiscardUnlessExecutor (rules-engine): draws via the shared DrawCardPrimitive, reveals the drawn card to all players (emits CardsRevealedEvent so the opponent sees it whether kept or discarded), matches it against the filter (base state, hand zone), and on a non-match discards it via ZoneTransitionService.discardCard so the canonical CardsDiscardedEvent fires (discard triggers, madness, client log) instead of a bare zone change. Registered in DrawingExecutors.
- SindbadScenarioTest covers both branches (nonland discarded, land kept) plus the rule corners: the discard emits CardsDiscardedEvent, the drawn card is revealed in both branches, keeping a land fires no discard, and an empty-library activation discards nothing without crashing.
- Register ArabianNightsSet in ScenarioTestBase so ARN cards are testable.
- Document the effect in card-sdk-language-reference.md.


wingedsheep:
Pushed review fixes to `arn-sindbad` (now at `8597e12b0`):

- **`CardsDrawnEvent` now emitted.** The executor drew via `DrawCardPrimitive` directly, which by contract leaves the aggregate `CardsDrawnEvent` to its driver — so it was never emitted, and "whenever you draw a card" triggers (which match only on `CardsDrawnEvent`) never fired. Now emitted after the primitive draw, mirroring `DrawLoop`'s ordering. (Note: drawing via the primitive still bypasses `DrawLoop`'s draw-*replacement* dispatch — documented in a comment as a known, narrower gap.)
- **`revealToSelf = false`** on the reveal — the drawing player already saw the card; only the opponent needs the overlay.
- **Two new tests**: one asserting `CardsDrawnEvent` fires for the drawn card, and an end-to-end test with a "whenever you draw a card → gain 1 life" enchantment proving the trigger fires off Sindbad's draw.

Also merged `origin/main` into the branch. All 7 `SindbadScenarioTest` tests pass. 🤖 Pushed via Claude Code.

PR #222, opened 2026-05-29
﻿﻿Add Uthros Psionicist; generalize cost gating to NthOfTypePerTurn

bolav:
## Summary

Implements **Uthros Psionicist** (EOE) — `{2}{U}` 2/4 Jellyfish Scientist whose
static ability reads "The second spell you cast each turn costs {2} less to cast."

Supporting this card required generalizing the existing cost-gating primitive and
correcting cost-calculation ordering:

- **`CostGating.FirstOfTypePerTurn` → `CostGating.NthOfTypePerTurn(n)`.** The old
  boolean "first spell" gate becomes a 1-indexed Nth-spell gate. Eluge, the
  Shoreless Sea migrates to `NthOfTypePerTurn(1)` (no behavior change); Uthros uses
  `NthOfTypePerTurn(2)`. `gatingApplies` now counts prior matching spells this turn
  and fires when the count equals `n - 1`.
- **Cost ordering fixed per CR 601.2f.** `calculateEffectiveCost` now applies cost
  *increases* before *reductions*, with the generic component floored at `{0}`.
  Previously reductions ran first, which could leave a stale increase behind
  (`{U}` +{1} −{2} incorrectly yielded `{1}{U}` instead of `{U}`).
- **Description generation** handles the Nth-ordinal phrasing and singular/verb
  agreement ("The second ... spell you cast ... costs ... each turn").

Uthros itself counts toward the per-turn tally (per Scryfall ruling 2025-07-25):
if it's the first spell cast this turn, the next spell is the "second".

## Changes

- New card: `eoe/cards/UthrosPsionicist.kt` (+ manual scenario JSON)
- `CostStaticAbilities.kt` — `NthOfTypePerTurn(n)`, ordinal/agreement in descriptions
- `CostCalculator.kt` — CR 601.2f ordering, `filterForGating` helper
- `ElugeTheShoreslessSea.kt` — migrated to `NthOfTypePerTurn(1)`
- Docs: `card-sdk-language-reference.md` `ModifySpellCost` section
- Backlog: EOE marked 215/261

### DIFF
--- a/backlog/sdk-reusability-consolidation.md
+++ b/backlog/sdk-reusability-consolidation.md
@@ -79,12 +79,12 @@ one `ModifySpellCost(target, modification, gating)`:
   `ReduceGenericBy(CostReductionSource)` / `ReduceColored(symbols)` /
   `ReduceColoredPerUnit(symbols, countSource)` / `IncreaseGeneric(amount)` /
   `IncreaseGenericPerOtherSpellThisTurn(amountPerSpell)`.
-- `gating: CostGating` — `None` / `FirstOfTypePerTurn` (Eluge).
+- `gating: CostGating` — `None` / `NthOfTypePerTurn(n)` (Eluge uses `n = 1`; Uthros Psionicist `n = 2`).
 
 `CostCalculator` rewritten around the new shape: a single
 `scanBattlefieldModifySpellCost` walks every battlefield permanent once,
 `targetMatchesSpell` filters by target/scope (using *projected* controller, not
-base), `gatingApplies` checks first-of-type-per-turn, and `applyToSpellCast`
+base), `gatingApplies` checks nth-of-type-per-turn, and `applyToSpellCast`
 dispatches by modification kind. Face-down (`calculateFaceDownCost`) and morph
 (`calculateMorphCostIncrease`) walk the same scan with target-typed filters.
 `ReduceFaceDownCastingCost` was unused dead code and dropped entirely.
--- a/backlog/sets/edge-of-eternities/cards.md
+++ b/backlog/sets/edge-of-eternities/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 261 booster cards (excluding basic lands)
 **Release Date:** August 1, 2025
-**Implemented:** 213 / 261
+**Implemented:** 214 / 261
 ---
 
 - [x] Adagia, Windswept Bastion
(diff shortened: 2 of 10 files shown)


wingedsheep:
## Review — ✅ approved, one nit fixed and pushed

Reviewed with `origin/main` merged in (clean, no conflicts). Pushed two commits to `bolav:uthros-psionicist`:
- `b4c3e9f` — merge of `origin/main` so the PR reflects post-merge reality
- `4a028e1` — the fix below

### Verdict
Behavior is correct and the SDK shape is right. Generalizing the boolean `CostGating.FirstOfTypePerTurn` into a 1-indexed `NthOfTypePerTurn(n)` is a clean "parameterize for the next card" move — Eluge migrates to `n = 1` with zero behavior change, Uthros uses `n = 2`, no new card-specific type. The strict `priorMatching == n - 1` correctly scopes the discount to *only* the Nth spell.

### What's good
- **CR 601.2f ordering fix is correct and verified.** "The total cost is the mana cost or alternative cost … plus all additional costs and cost increases, and minus all cost reductions" — increases first, then reductions, mana floored at `{0}`. The new order in `CostCalculator` matches, and it's a general fix benefiting every cost-modified spell.
- **Tests hit the rule-corner that drove the change.** The Glowrider +{1} / Opt ({U}) case yields `{1}{U}` under the old order and `{U}` under the new — it actually fails if the ordering regresses. All 5 scenario tests pass locally.
- Clean migration (no stale `FirstOfTypePerTurn` refs), correct printing placement (`just check-card-printing` green), SDK reference doc updated, and the backlog count is now self-consistent (`main` had stated 213 / 214 actual boxes; this PR makes it 215 / 215).

### Fix applied (was: Minor)
`ModifySpellCost(YouCast(GameObjectFilter.Any), …)` generated **"The second _card_ spell you cast costs {2} less to cast each turn"** because `GameObjectFilter.Any.description` is `"card"`. Invisible here since `oracleText` is set explicitly, but the SDK reference now recommends this exact recipe, so any future card using it without an explicit `oracleText` would inherit the leak (the granted-ability path surfaces `ability.description`). Fixed by emitting no adjective when the filter narrows nothing → "The second spell you cast…". Added `ModifySpellCostDescriptionTest` covering the `Any` case, a typed filter (Eluge), and the ungated plural form.

Nothing blocking — good to merge.

PR #241, opened 2026-05-30
Add Command Bridge to Edge of Eternities with PayCost.Tap support

bolav:
Command Bridge's ETB trigger ("sacrifice it unless you tap an untapped permanent you control") is the first card to require a tap-based alternative cost in the PayOrSufferEffect punisher infrastructure.

Adds `PayCost.Tap(filter, count = 1)` to the SDK and wires it through PayOrSufferExecutor and SacrificeAndPayContinuationResumer alongside the existing discard / sacrifice / pay-life / mana / exile variants. Candidates are restricted to your untapped permanents (matching the filter) with the source auto-excluded; tapping each emits a TappedEvent so "becomes tapped" triggers fire normally.

Includes scenario tests covering the four flows (pay-and-survive, decline-and-die, auto-suffer-when-no-candidates, tapped-permanents- excluded), a manual DevScenarioController fixture, and the SDK language reference update with the new PayCost catalog.


wingedsheep:
Pushed a small follow-up fix (fcba5ea7b): the tap prompt rendered "**a** untapped card you control" because the a/an article was computed from the filter description's first letter, but "untapped" always immediately follows the article — so it's now hardcoded to "**an** untapped …" in both `PayCost.Tap.description` and the `PayOrSufferExecutor` decision prompt. All 4 `CommandBridgeScenarioTest` cases still pass.

Otherwise the PR looks great — `PayCost.Tap` is a well-justified new payment shape (not a relabel), projection is handled correctly via `BattlefieldFilterUtils`, `TappedEvent` is emitted per permanent, oracle text matches Scryfall (EOE #252), and the printing check passes.

PR #266, opened 2026-05-30
Add Invasion: Undermine, Absorb, Spite//Malice, Void

wingedsheep:
Implements four Invasion (INV) cards.

## Cards

- **Undermine** `{U}{U}{B}` — Instant. Counter target spell. Its controller loses 3 life. (composed from existing `CounterSpell` + `LoseLife(EffectTarget.TargetController)`)
- **Absorb** `{W}{U}{U}` — Instant. Counter target spell. You gain 3 life. (`CounterSpell` + `GainLife`)
- **Spite // Malice** `{3}{U}` // `{3}{B}` — Split instant. Spite counters a noncreature spell; Malice destroys a nonblack creature that can't be regenerated. (`CardLayout.SPLIT`, existing facades)
- **Void** `{3}{B}{R}` — Sorcery. Choose a number; destroy all artifacts and creatures with that mana value; then target player reveals their hand and discards all nonland cards with that mana value.

## New mechanic: choose-a-number

Void introduces a reusable combinator rather than a card-specific executor:

- `Effects.ChooseNumberThen(then, minValue, maxValue, prompt)` — pauses for a number, stamps it onto the effect context as **X**, then runs `then` once. Mirrors the existing `ChooseColorThen` shape (executor + continuation + resumer in `ColorChoiceContinuationResumer`).
- `CardPredicate.ManaValueEqualsX` / `GameObjectFilter.manaValueEqualsX()` — match objects whose mana value equals the chosen number. Resolution-time only, mirroring `ManaValueAtMostX`; stubbed `false` in the context-free matchers (`AffectsFilterResolver`, `CostCalculator`, `TriggerMatcher`).

Void then composes `DestroyAll(artifacts/creatures with MV=X)` + reveal-hand + gather-and-discard (nonland, MV=X) — the one chosen number drives both halves.

## Tests

- `VoidScenarioTest` — destroy + discard at a matching number, and the no-match case.
- Full `rules-engine` and `game-server` scenario suites pass; `mtg-sets` (image URIs) pass.

## Docs

- `docs/card-sdk-language-reference.md` updated for `ChooseNumberThen`, `ManaValueEqualsX`, and `manaValueEqualsX()`.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -429,6 +429,11 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 - `ChooseColorAndGrantProtectionToGroupEffect(filter)` — same, for a group.
 - `GrantProtectionFromColor(color, target, duration)` — grant protection from a **fixed** color to a target (no player choice); a thin recipe over `GrantKeyword("PROTECTION_FROM_<COLOR>")`. "{W}: Target creature gains protection from red until end of turn." (Crimson Acolyte).
 - `ChooseColorThenEffect(whenChosen)` — pick a color, then run a function of that color.
+- `Effects.ChooseNumberThen(then, minValue=0, maxValue=16, prompt)` — pick a number in `[minValue, maxValue]`,
+  then run `then` once with the chosen number exposed via the effect context as **X**. Atomic effects and filters
+  under `then` read it through `ManaValueEqualsX` (`.manaValueEqualsX()`). Compose with `CompositeEffect` for
+  multi-step cards (Void: destroy all artifacts/creatures with that mana value, then a target player reveals their
+  hand and discards all nonland cards with that mana value).
 - `GrantHexproofFromChosenColorEffect(target)` — hexproof from chosen color.
 - `ChooseCreatureTypeEffect(...)` — pause for creature-type pick.
 - `SelectTargetEffect(...)` — have a player pick from a valid set.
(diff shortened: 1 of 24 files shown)


wingedsheep:
## Review follow-up — applied fixes (pushed as `1331a473b`)

While writing the recommended scenario tests, the Undermine test surfaced a **real bug** that the original review missed: Undermine's *"its controller loses 3 life"* silently did nothing.

### Root cause (latent engine bug, not just this card)
`TargetResolutionUtils.resolvePlayerTargets` (plural) routed its `else` branch through the **stateless** 2-arg resolver, which has no case for state-dependent targets like `EffectTarget.TargetController`. So `LoseLife(3, TargetController)` resolved to *no player* and was skipped. Undermine is the first card to exercise `LoseLife(TargetController)` against a spell target.

**Fix:** route the `else` fallback through the **state-aware** 3-arg resolver (it tries the stateless path first, so this is a strict superset). The existing damage-to-controller cards (Agonizing Demise, Backlash) still pass, confirming no regression.

Also reordered Undermine to `LoseLife(TargetController) then CounterSpell()` so the controller is read while the spell is still on the stack (correct for control-changed spells, rather than relying on the owner fallback after the spell leaves). Both sub-effects happen in one resolution, so the order is imperceptible.

### Tests added (all green)
- **`UndermineScenarioTest`** — countered spell's controller loses 3 life; Undermine's caster is unaffected.
- **`AbsorbScenarioTest`** — counter + caster gains 3 life; opponent unchanged.
- **`SpiteMaliceScenarioTest`** — Spite counters a noncreature spell and *cannot* target a creature spell; Malice destroys a nonblack creature, *cannot* target a black creature, and kills through a regeneration shield (can't-be-regenerated).

### Minor doc note
`EffectContext.xValue` now documents that `ChooseNumberThen` reuses this slot for the chosen number, so a future card that both pays `{X}` *and* chooses a number would need the slot split.

Full `:rules-engine:test` and `:game-server:test` suites pass locally.

PR #267, opened 2026-05-30
Add Invasion enchantments: Aura Shards, Sterling Grove, Elfhame Sanctuary, Saproling Symbiosis

wingedsheep:
Implements four Invasion (INV) cards.

## Cards

- **Aura Shards** `{1}{G}{W}` — Enchantment. "Whenever a creature you control enters, you may destroy target artifact or enchantment." Optional triggered ability composed from existing primitives (`entersBattlefield` + `Effects.Destroy(Targets.ArtifactOrEnchantment)`). Also adds the Commander 2011 (`CMD`) reprint row, since that set is scaffolded.
- **Sterling Grove** `{G}{W}` — Enchantment. "Other enchantments you control have shroud" via `GrantKeyword(SHROUD, AllEnchantments.youControl().other())`, plus `{1}, Sacrifice: search your library for an enchantment, reveal, shuffle, put on top" (`searchLibrary(TOP_OF_LIBRARY, shuffleAfter = true)`).
- **Elfhame Sanctuary** `{1}{G}` — Enchantment. Upkeep search for a basic land into hand; if you do, skip your draw step this turn.
- **Saproling Symbiosis** `{3}{G}` — Sorcery. "Cast as though it had flash if you pay {2} more" reuses the shared FlashKicker plumbing (same as Ghitu Fire); creates a Saproling per creature you control.

## New engine mechanic: SkipNextDrawStep

Elfhame Sanctuary needed a general "skip your draw step" mechanic, which did not exist:

- SDK: `Effects.SkipNextDrawStep(target)` / `SkipNextDrawStepEffect`.
- Engine: `SkipNextDrawStepExecutor` adds a one-shot `SkipDrawStepComponent` to the player; `DrawPhaseManager.performDrawStep` consumes it and skips the draw. Registered for polymorphic serialization.
- The draw-step skip is gated with `IfYouDoEffect` (`SuccessCriterion.Auto`) so it only fires when the search actually puts a land into hand — declining the search draws normally.
- Documented in `docs/card-sdk-language-reference.md`.

## Tests

- `AuraShardsScenarioTest` — creature ETB destroys a target enchantment; declining leaves it intact.
- `SaprolingSymbiosisScenarioTest` — token count equals creatures controlled (tokens not self-counted).
- `ElfhameSanctuaryScenarioTest` — fetching a land skips the draw step; declining the search draws normally.

`./gradlew :mtg-sdk:test :rules-engine:test` and the new game-server scenario tests pass; full `./gradlew build` compiles. `just check-card-printing` is clean for all four cards.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -368,6 +368,7 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 - `Effects.Suspect(target)` — target becomes Suspected (MKM keyword). Composite: `SetSuspectedEffect` (named status, CR 701.60d dedup) + `GrantKeywordEffect(MENACE)` + `CantBlockEffect`.
 - `RemoveFromCombatEffect(target)` — yank target out of combat.
 - `SkipNextTurnEffect(target)` — target skips their next turn.
+- `Effects.SkipNextDrawStep(target = Controller)` (`SkipNextDrawStepEffect`) — target skips their next draw step. Adds a one-shot `SkipDrawStepComponent` marker consumed by `DrawPhaseManager.performDrawStep` (Elfhame Sanctuary's "you skip your draw step this turn").
 - `HijackNextTurnEffect(target)` — you control target's next turn.
 - `GrantCantBeBlockedByChosenColorEffect(target, duration)` — unblockable except by chosen color.
 - `CantCastSpellsEffect(target, until?)` — target can't cast spells.
--- a/game-server/src/test/kotlin/com/wingedsheep/gameserver/scenarios/AuraShardsScenarioTest.kt
+++ b/game-server/src/test/kotlin/com/wingedsheep/gameserver/scenarios/AuraShardsScenarioTest.kt
@@ -1,0 +1,74 @@
+package com.wingedsheep.gameserver.scenarios
+
+import com.wingedsheep.gameserver.ScenarioTestBase
+import com.wingedsheep.sdk.core.Phase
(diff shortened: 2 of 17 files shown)


wingedsheep:
Pushed review fixups (commit 093853065), and merged latest `main` into the branch (clean, no conflicts).

**Applied from review:**

- **CR-number fix** — `SaprolingSymbiosis.kt` cited *CR 608.2g* for the "for each creature you control" count. 608.2g covers mana abilities / casting spells during resolution; the rule for "information from the game is determined only once, when the effect is applied" is **CR 608.2h**. Corrected the comment. (Note: the engine systemically mislabels last-known-info as 608.2g in ~6 other spots — out of scope here, but worth a follow-up sweep.)

- **New `SterlingGroveScenarioTest`** — the only card in the PR previously without coverage. Two cases:
  - the shroud lord grants shroud to *another* enchantment you control while Sterling Grove itself stays targetable (`.other()` self-exclusion), verified via `StateProjector.hasKeyword`;
  - the `{1}, Sacrifice` ability offers only enchantment cards (filter), and the found card ends up on **top** of the library *after* the shuffle.

- **Hardened `SaprolingSymbiosisScenarioTest`** — added an opponent-controlled creature so the "creature *you* control" scoping is actually exercised (token count correctly stays at 2 rather than 3).

All affected scenario tests pass (`SterlingGroveScenarioTest`, `SaprolingSymbiosisScenarioTest`), and `:rules-engine:test` / `:mtg-sdk:test` remain green.

PR #269, opened 2026-05-30
Add Invasion cards: Liberate, Mourning, Obliterate, Rout, Winnow

wingedsheep:
Implements five Invasion (INV) cards, each canonical in its earliest real printing (INV).

## Cards

- **Liberate** `{1}{W}` Instant — Exile target creature you control; return it to the battlefield under its owner's control at the next end step. Uses `EffectPatterns.exileUntilEndStep`.
- **Mourning** `{1}{B}` Aura — Enchanted creature gets -2/-0; `{B}`: return this Aura to its owner's hand. `ModifyStats(-2, 0)` static + bounce activated ability (Shimmering Wings / Crown of Flames shape).
- **Obliterate** `{6}{R}{R}` Sorcery — Can't be countered; destroy all artifacts, creatures, and lands (no regeneration). Single `Effects.DestroyAll` over the combined `Artifact or Creature or Land` filter so everything is destroyed simultaneously; `cantBeCountered = true`.
- **Rout** `{3}{W}{W}` Sorcery — Optionally cast as though it had flash for `{2}` more; destroy all creatures (no regeneration). Uses `KeywordAbility.flashKicker("{2}")` (Breaking Wave pattern).
- **Winnow** `{1}{W}` Instant — Destroy target nonland permanent if another permanent with the same name is on the battlefield; draw a card. The destroy is gated by a new condition; the draw is unconditional.

## SDK addition

- `Conditions.AnotherPermanentWithSameNameAsTarget(targetIndex = 0)` — true when at least one *other* battlefield permanent shares the exact card name of the chosen target (the target itself is excluded). Added the SDK condition type, the `Conditions` facade entry, the `ConditionEvaluator` resolution branch, and a `card-sdk-language-reference.md` entry.

## Tests

`InvasionDestructionBatchScenarioTest` covers all five cards, including Liberate's exile-then-return-at-end-step, Mourning's -2/-0 and self-bounce, Obliterate/Rout board wipes, and both Winnow branches (destroys with a duplicate present, no-ops without one, and always draws). All scenario, discovery, validation, and condition tests pass.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1333,6 +1333,11 @@ keywordAbilities(KeywordAbility.Protection(Color.BLUE), KeywordAbility.Annihilat
   only — no targets/triggering/kicker — so it evaluates identically in resolution and in
   projection, which lets it gate a `ConditionalStaticAbility`. Used by the Invasion djinn cycle
   ("as long as [color] is the most common color among all permanents…" — Goham/Halam/Ruham/Sulam/Zanam).
+- `AnotherPermanentWithSameNameAsTarget(targetIndex = 0)` — true when at least one *other*
+  battlefield permanent shares the exact card name of the context target at `targetIndex`. The
+  target itself is excluded, so a lone copy never satisfies its own check; tokens compare by name
+  like any other permanent. Resolution-only (reads a chosen target). Used by Winnow ("Destroy
+  target nonland permanent if another permanent with the same name is on the battlefield").
 - `YouHaveCitysBlessing` — you have City's Blessing (10+ permanents).
 - `SourceIsRingBearer` — the source permanent is your Ring-bearer (CR 701.52e).
 
--- a/game-server/src/test/kotlin/com/wingedsheep/gameserver/scenarios/InvasionDestructionBatchScenarioTest.kt
+++ b/game-server/src/test/kotlin/com/wingedsheep/gameserver/scenarios/InvasionDestructionBatchScenarioTest.kt
@@ -1,0 +1,195 @@
+package com.wingedsheep.gameserver.scenarios
+
+import com.wingedsheep.engine.core.ActivateAbility
+import com.wingedsheep.engine.mechanics.layers.StateProjector
+import com.wingedsheep.gameserver.ScenarioTestBase
+import com.wingedsheep.sdk.core.Phase
+import com.wingedsheep.sdk.core.Step
+import io.kotest.assertions.withClue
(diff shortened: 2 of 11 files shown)


wingedsheep:
## Review fixups pushed (`045983f9a`)

I reviewed this PR and pushed the fixups below. I also merged `origin/main` into the branch — that merge is **required** here, not just housekeeping: the new "can't be countered" test relies on **Undermine**, which landed on `main` separately, so the branch now carries it.

Verdict going in was solid: oracle text / cost / type / collector number match Scryfall exactly for all five cards, four of them compose existing primitives with zero new types, and the one new SDK addition (`Conditions.AnotherPermanentWithSameNameAsTarget`) is genuinely novel, well-parameterized by `targetIndex`, resolution-gated, and documented. The changes below close the gaps.

### 1. C17 Rout printing row (was blocking the printing check)
`just check-card-printing "Rout"` exited non-zero: Rout is reprinted in **Commander 2017**, which is scaffolded in the repo, but had no `Printing(...)` row. Added `c17/cards/RoutReprint.kt` (following the `MosswortBridgeReprint` shape). Canonical stays in INV (earliest real printing); the check is now green.

### 2. Face-down guard on the new condition
`AnotherPermanentWithSameNameAsTarget` read raw `CardComponent.name`, but face-down permanents keep their hidden real name on the component (the "Face-down creature" label is computed at use sites, not stored). Per CR 708.2 a face-down permanent has no name, so two face-down creatures — or a face-up card and its face-down twin — must not count as same-named. Added a `FaceDownComponent` guard on both sides (target and candidate). Unreachable with the current card pool (no morph/manifest targetable in INV), so it's a defensive fix with no dedicated test, but it matters now that this is a shared SDK condition.

### 3. Tests for the clauses that distinguish these wipes
The original suite cast Rout only at sorcery speed and never exercised Obliterate's uncounterability — i.e. the two clauses that make these cards more than vanilla Wraths. Added:
- **Rout** — flash-kicker cast wiping the board on the opponent's turn, plus the sorcery-speed rejection without the kicker.
- **Obliterate** — Undermine *targets* it (uncounterable doesn't change targetability) but Obliterate still resolves and wipes the board, confirming it isn't countered.

### 4. Minor
Winnow's draw assertions changed from `handBefore - 1 + 1` to `shouldBe handBefore`.

All 9 scenario tests pass, plus `CardDiscoveryTest`, `PrintingRegistryTest`, `ConditionEvaluatorTest`, and `CardImageUriTest`. The `flashKicker` primitive itself remains covered by `GhituFireScenarioTest`.

PR #273, opened 2026-05-30
Add Kav Landseeker to Edge of Eternities with skipCurrentTurn delayed…

bolav:
…-trigger flag

Adds a `skipCurrentTurn` boolean to `CreateDelayedTriggerEffect` for "on your next turn"-style timing: the trigger is deferred past the current turn regardless of step. Combined with `fireOnlyOnControllersTurn`, it lands on the controller's upcoming own turn — distinct from `onControllerNextTurn`, which still lets the current turn's end step satisfy the trigger if it hasn't begun yet.

### DIFF
--- a/backlog/sets/edge-of-eternities/cards.md
+++ b/backlog/sets/edge-of-eternities/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 261 booster cards (excluding basic lands)
 **Release Date:** August 1, 2025
-**Implemented:** 237 / 261
+**Implemented:** 238 / 261
 ---
 
 - [x] Adagia, Windswept Bastion
@@ -120,7 +120,7 @@
 - [x] Kavaron, Memorial World
 - [x] Kavaron Skywarden
 - [x] Kavaron Turbodrone
-- [ ] Kav Landseeker
+- [x] Kav Landseeker
 - [x] Knight Luminary
 - [x] Larval Scoutlander
 - [x] Lashwhip Predator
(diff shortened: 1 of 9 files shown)


wingedsheep:
Pushed a follow-up commit (`2e2aa93`) applying the SDK-shape suggestion from review.

**What changed:** collapsed the two *timing* booleans on `CreateDelayedTriggerEffect` (`onControllerNextTurn`, `skipCurrentTurn`) into a single `DelayedTriggerTiming` enum:

- `CURRENT_TURN_OR_LATER` (default) — no turn floor (was: neither flag set)
- `NEXT_END_STEP` — "at the beginning of your next end step" (was: `onControllerNextTurn = true`)
- `NEXT_TURN` — "on your next turn" (was: `skipCurrentTurn = true`)

**Why:** those two booleans weren't independent — they're two of three points on one mutually-exclusive axis ("the earliest turn a step-based delayed trigger may fire"). As separate booleans they admitted a meaningless `both = true` state that was only disambiguated by `when`-branch ordering in `CreateDelayedTriggerExecutor`. An enum makes the axis total and illegal combinations unrepresentable, and the next "next turn"-style card just reuses the value instead of growing another flag.

`fireOnlyOnControllersTurn` is deliberately **left as a boolean** — it gates *whose* turn the trigger fires on (an orthogonal axis still used on its own by Meandering Towershell and Sekshaas, Early Sleeper), not *which* turn is the earliest eligible one.

**Migrated:** Kav Landseeker → `NEXT_TURN`, Dragonhawk, Fate's Tempest → `NEXT_END_STEP`, the executor `when`, the SDK reference doc, and the scenario/backlog references.

**Verified green locally:** `KavLandseekerScenarioTest` (2), `DragonhawkFatesTempestTest` (5), and `:mtg-sdk:test`.

PR #279, opened 2026-05-30
Add Invasion cards batch C

wingedsheep:
Implements four Invasion (INV) cards via the `add-card` workflow.

## Cards added

- **Exotic Curse** — `{2}{B}` Aura. Domain: enchanted creature gets -1/-1 for each basic land type among lands you control. Composed from `GrantDynamicStatsEffect` + `DynamicAmounts.domain()` negated with `Multiply(-1)` (mirrors Withering Hex). No engine changes.
- **Juntu Stakes** — `{2}` Artifact. Creatures with power 1 or less don't untap during their controllers' untap steps. `GrantKeyword(DOESNT_UNTAP)` over a battlefield `Creature.powerAtMost(1)` group, reusing the existing untap-prevention ability flag. No engine changes.
- **Kavu Lair** — `{2}{G}` Enchantment. Whenever a power-4+ creature enters, its controller draws. `ZoneChangeEvent(Creature.powerAtLeast(4))` trigger with `controlledByTriggeringEntityController` (mirrors Death Match). No engine changes.
- **Essence Leak** — `{U}` Aura. While the enchanted permanent is red or green, it has "At the beginning of your upkeep, sacrifice this permanent unless you pay its mana cost."

None skipped.

## Engine / SDK additions (for Essence Leak)

Both are general-purpose and reusable:

- **`PayCost.OwnManaCost`** — pay the affected permanent's *own* mana cost (read from `CardComponent.manaCost` at payment time), resolved into a concrete `PayCost.Mana` inside `PayOrSufferExecutor`. The two morph cost sites reject it explicitly.
- **`Conditions.EnchantedPermanentMatches(filter)`** — generalizes the narrow `EnchantedCreatureIsLegendary` / `EnchantedCreatureHasSubtype` conditions to any `GameObjectFilter` (color, type, …), evaluated against the Aura's `AttachedToComponent` in both resolution and projection.
- `TriggerAbilityResolver.getAttachedGrantedTriggeredAbilities` now unwraps `ConditionalStaticAbility`-wrapped grants (mirroring the existing `GrantWard` path), so a conditionally-granted triggered ability is detected while its gate holds.

## Tests

- `InvasionBatchCScenarioTest` (game-server): Juntu Stakes untap suppression (power 1 vs 2), Kavu Lair draw on power-5 / no-draw on power-3, and Essence Leak pay-its-mana-cost (survives) / can't-pay (sacrificed) / inactive-on-blue.
- Full `:rules-engine:test`, `:mtg-sets:test`, and related game-server scenario tests (VileConsumption, morph, Death Match) pass.

## Docs

- Updated `docs/card-sdk-language-reference.md` (PayCost and Conditions sections).

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -144,6 +144,11 @@ player to pay against an alternative consequence.
 
 - `PayCost.Mana(ManaCost)` — pay mana (auto-taps lands via the solver). "...unless you pay {U}{U}"
   (Vaporous Djinn).
+- `PayCost.OwnManaCost` — pay the mana cost of the permanent the cost applies to (its *own* mana
+  cost, read from `CardComponent.manaCost` at payment time). Use for granted abilities like
+  Essence Leak ("...sacrifice this permanent unless you pay its mana cost"), where the affected
+  permanent — not a fixed cost — owns the mana cost. The engine resolves it into a concrete
+  `PayCost.Mana` against that permanent before prompting.
 - `PayCost.PayLife(amount)` — pay N life; offered only when the player has more than N life.
   "...unless you pay 3 life."
 - `PayCost.Discard(filter = Any, count = 1, random = false)` — discard cards matching `filter`.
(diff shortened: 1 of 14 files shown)


wingedsheep:
Pushed two small follow-ups from a review pass (commit 26a1775af). Both are tidy-ups on the new Essence Leak engine pieces — no behavior change for the cards, all `InvasionBatchCScenarioTest` cases still green.

**1. `ConditionEvaluator.evaluateEnchantedPermanentMatchesCtx` — honest projection-mode controller.**
The projection branch fell back to `PredicateContext(controllerId = sourceId)` when no controller was in scope, i.e. it stuffed the Aura's *entity* id into a slot that expects a *player* id. Harmless for Essence Leak (a color filter never reads `controllerId`), but a latent bug the moment `EnchantedPermanentMatches` is gated by a filter carrying a controller predicate in projection mode. Now it falls back to the source's projected controller (`projected.getController(sourceId)`) and bails (`return false`) if even that is unknown, rather than miscasting an entity id as a player.

**2. `PayOrSufferExecutor` — document the `OwnManaCost` empty-cost case.**
`PayCost.OwnManaCost` reads `CardComponent.manaCost`, which is an *empty* `ManaCost` for a permanent with no mana cost (a land or token). That resolves to `{0}` and is trivially payable, so such a permanent is always kept. Added a comment on both the execute and `canPayCost` branches so the next reader doesn't mistake this for a bug, and clarified that the `ownCost != null` guard only covers a missing source entity (the cost itself is non-nullable).

PR #288, opened 2026-05-30
Add Invasion cards: Tsabo's Web, Orim's Touch, Turf Wound

wingedsheep:
Implements three Invasion (INV) cards via the add-card workflow.

## Cards

- **Tsabo's Web** ({2} Artifact, rare) — "When this artifact enters, draw a card. Each land with an activated ability that isn't a mana ability doesn't untap during its controller's untap step." ETB draw + a static `GrantKeyword(DOESNT_UNTAP)` over lands matching the new `CardPredicate.HasNonManaActivatedAbility`. The untap step already honors `DOESNT_UNTAP`.
- **Orim's Touch** ({W} Instant, common; Kicker {1}) — "Prevent the next 2 damage that would be dealt to any target this turn. If kicked, prevent 4 instead." Composes `Effects.PreventNextDamage` inside `ConditionalEffect(WasKicked, …)`.
- **Turf Wound** ({2}{R} Instant, common) — "Target player can't play lands this turn. Draw a card." Composite of the now-targeted `PreventLandPlaysThisTurnEffect` + draw.

## SDK / engine changes

- Generalized `PreventLandPlaysThisTurnEffect` from a `data object` to a `data class` with a `target` (default = controller). New facade `Effects.CantPlayLandsThisTurn(target)`. Migrated Rock Jockey to the facade; its executor now resolves the target.
- Added `CardPredicate.HasNonManaActivatedAbility`, backed by a precomputed `CardComponent.hasNonManaActivatedAbility` flag (computed via `CardDefinition.hasNonManaActivatedAbility` = has a printed activated ability that isn't a mana ability and isn't a loyalty ability). Set in `GameInitializer` and the scenario-test card factory. Wired through the exhaustive predicate matchers in `PredicateEvaluator`, `AffectsFilterResolver`, and `CostCalculator`.
  - Known limitation (documented in code + the SDK reference): reflects a land's *printed* abilities only — abilities granted by other continuous effects are not counted.
- Updated `docs/card-sdk-language-reference.md` for the new effect facade and predicate.

## Tests

- New `InvasionTsabosWebTurfWoundOrimsTouchScenarioTest` (game-server) covering all three cards, including the unkicked/kicked Orim's Touch branches and the Tsabo's Web positive (non-mana land) / negative (mana-only land) untap cases.
- Full `:rules-engine:test` and `:mtg-sdk:test` suites pass, including the polymorphic-registration and serialization round-trip hygiene tests; Rock Jockey regression test passes.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -381,6 +381,9 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 - `HijackNextTurnEffect(target)` — you control target's next turn.
 - `GrantCantBeBlockedByChosenColorEffect(target, duration)` — unblockable except by chosen color.
 - `CantCastSpellsEffect(target, until?)` — target can't cast spells.
+- `Effects.CantPlayLandsThisTurn(target = Controller)` (`PreventLandPlaysThisTurnEffect`) — the target player can't
+  play lands for the rest of this turn (sets remaining land drops to 0). Defaults to the controller (Rock Jockey);
+  pass `EffectTarget.ContextTarget(n)` for "target player can't play lands this turn" cards like Turf Wound.
 
 ### Forced sacrifice / discard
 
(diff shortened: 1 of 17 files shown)


wingedsheep:
Pushed a follow-up commit (a90bc8e0c) addressing review feedback — tests only + one cosmetic SDK tweak, no behavior change to the cards:

**Orim's Touch — actually exercise the kicked amount.** Both prevention tests now use a 5/5 attacker so the prevented amount is observable:
- unkicked → prevents 2 of 5, so **3 damage gets through** (`life - 3`)
- kicked → prevents 4 of 5, so **1 gets through** (`life - 1`)

Previously the kicked test only asserted the cast resolved without error, so a 4-vs-2 swap in the `ConditionalEffect(WasKicked, …)` branches would have passed silently. Now the 4-vs-2 distinction is the assertion.

**Tsabo's Web — end-to-end untap behavior.** Added a test that taps a land with a non-mana activated ability *and* a mana-only land, advances through the controller's untap step, and asserts the former stays tapped (`DOESNT_UNTAP`) while the latter untaps. The existing test only checked that the keyword was projected onto the land; this confirms the grant actually changes untap-step behavior.

**`PreventLandPlaysThisTurnEffect.description`** now renders `"Target player can't play lands this turn"` for a `ContextTarget` (Turf Wound) instead of the grammatically-incomplete `"Target can't play lands this turn"`. The controller default (Rock Jockey → `"You can't play lands this turn"`) is unchanged.

All five scenario cases pass, plus `RockJockeyTest` and the `:mtg-sdk:test` suite (serialization round-trips included).

PR #321, opened 2026-05-31
Add New Way Forward (prevent-and-react chain)

wingedsheep:
## What

Implements **New Way Forward** (Tarkir: Dragonstorm #211), Tier-3 engine-gap item 16 in `backlog/tdm-engine-gaps.md`:

> The next time a source of your choice would deal damage to you this turn, prevent that damage. When damage is prevented this way, New Way Forward deals that much damage to that source's controller and you draw that many cards.

## Approach — a reusable prevent-and-react chain

New Way Forward is **Deflecting Palm plus "draw that many cards."** Instead of a bespoke prevent-redirect-draw effect, the existing chosen-source prevention shield now carries a **composable chain of reactions** keyed to the prevented amount:

- **SDK** — new sealed `PreventionReaction` with two links:
  - `DealToSourceController` — deal the prevented amount to the source's controller (the old `reflect`)
  - `ControllerDrawsCards` — the protected player draws that many cards
  
  `PreventDamageEffect.reflect: Boolean` → `onPrevented: List<PreventionReaction>`. New facade `Effects.PreventNextDamageFromChosenSourceThen(vararg reactions)`; `DeflectNextDamageFromChosenSource()` is now just `[DealToSourceController]`.
  - Deflecting Palm: `[DealToSourceController]`
  - New Way Forward: `[DealToSourceController, ControllerDrawsCards]`

- **Engine** — reactions threaded through `DeflectDamageSourceChoiceContinuation` → resumer → the `DeflectNextDamageFromSource` shield. `DamageUtils.checkDeflectDamageShield` applies each reaction in order when the shield fires (reflect via `dealDamageToTarget`; draw via the shared `DrawCardPrimitive`, aggregated into one `CardsDrawnEvent`).

The reactions fire inline as the damage is prevented (matching how Deflecting Palm already reflects), so the draw bypasses draw-replacement prompts — acceptable for the current users and documented in the SDK reference.

## Notes

- The card does **not** target — the source is chosen on resolution (per Scryfall rulings, included in the card's `metadata`).
- Deflecting Palm is behaviorally unchanged; its scenario test still passes.

## Tests

- New `NewWayForwardScenarioTest` — prevents + reflects + draws for both a **combat** source (Alpine Grizzly, 4 damage → reflect 4, draw 4) and a **spell** source (Shock, 2 → reflect 2, draw 2).
- Regression pass over serialization-hygiene, damage, prevention, combat, and Deflecting Palm tests — all green.


wingedsheep:
**Revised the design after review feedback.** The first pass baked a closed `PreventionReaction` enum + a `PreventNextDamageFromChosenSourceThen` facade into the prevention effect — a mini-language that doesn't compose and pushes toward per-effect `*Then` variants.

The subtlety is that this "then" is **deferred**: it fires later, when the chosen source's damage is actually prevented, keyed to a runtime amount — so a sequential `CompositeEffect` can't express it. The reworked design instead has the chosen-source prevention shield carry an **arbitrary `onPrevented: Effect?`**, run through the normal effect pipeline when it fires, with:

- the prevented amount bound to `DynamicAmount.ContextProperty(PREVENTED_DAMAGE_AMOUNT)`
- the source's controller reachable as `EffectTarget.ControllerOfTriggeringEntity`
- the shield's own card as the effect source

So there's **no bespoke reaction type** — the payoff is plain effect composition using the same `ControllerOfTriggeringEntity` + damage-amount pair Tephraderm already uses:

- Deflecting Palm = `DealDamage(ControllerOfTriggeringEntity, preventedAmount)`
- New Way Forward = `Composite(DealDamage(...), DrawCards(preventedAmount))`

The shield runs the follow-up via an injected effect runner (wired in `EngineServices`), so any future "when prevented" payoff composes for free. See the second commit.

wingedsheep:
**Now fully rules-faithful** (3rd commit). The "When damage is prevented this way, …" payoff is a real triggered ability on the stack — opponents get priority and can respond (CR 603) — instead of running inline.

It turned out the engine already has an event-based delayed-trigger framework, so this is modeled as the actual MTG shape: two linked objects created on resolution —

1. the one-shot chosen-source prevention shield, and
2. a delayed triggered ability *"When damage is prevented this way, `<onPrevented>`"*.

When the shield prevents an instance it emits an internal `DamagePreventedEvent` carrying the link id; trigger detection fires **only that** delayed ability onto the stack (matched by id, so multiple shields on the same source don't cross-fire). Inside the trigger the prevented amount is `ContextProperty(PREVENTED_DAMAGE_AMOUNT)` and the source's controller is `ControllerOfTriggeringEntity` — so the payoff stays plain effect composition (`DealDamage` + `DrawCards`), and being a normal stack ability it can be interactive.

The SDK surface (`PreventDamageEffect.onPrevented`, the facades, the card) is unchanged from the previous commit. The earlier inline caveat in the docs is gone. All scenario + regression + serialization-hygiene tests pass.

PR #341, opened 2026-05-31
Rename LifeAuctionEffect → OpenLifeBidEffect, generalize over participants

wingedsheep:
Fixes audit item **#6 — `LifeAuctionEffect` — named/shaped for one card** from `backlog/sdk-quality-audit.md`.

## Problem
The open life-bid effect was named for its single user (Mages' Contest) and hardcoded the other bidder as "the controller of the targeted spell," violating audit standards #2 (name the mechanic, not the card) and #4 (don't shape a type around one card).

## Changes
- **Rename the SDK type** `LifeAuctionEffect` → `OpenLifeBidEffect` (`@SerialName` `LifeAuction` → `OpenLifeBid`, facade `Effects.LifeAuction` → `Effects.OpenLifeBid`, field `onCasterWins` → `onWin`).
- **Generalize over participants:** the other bidder is now a `participant: Player` reference (defaults to `Player.Opponent`), resolved against the effect context, rather than derived inline from the targeted spell. Mages' Contest passes `Player.ControllerOf("target spell")`. If `participant` resolves to the caster (or nobody), the caster is the sole bidder and wins at the opening bid of 1 — the same behavior the old "you control the targeted spell" branch produced.
- **Drop the vestigial `spellId`** from the continuation: it was only used to derive the spell controller at setup. The `StackResolver` already fizzles a spell whose only target leaves the stack before the effect runs, so the old defensive guard was redundant.
- Rename the engine plumbing to match: `OpenLifeBidExecutor` / `OpenLifeBidLogic` / `OpenLifeBidContinuation` / `OpenLifeBidStage`, plus the resumer and serialization registration.
- Update `docs/card-sdk-language-reference.md`.

The bespoke two-player alternating-bid machinery is kept (the audit explicitly allows it); only the card-shaped naming and the hardcoded participant are removed.

## Testing
- `MagesContestTest` (both scenarios) passes unchanged.
- `:rules-engine:test --tests MagesContestTest`, `:game-server:compileKotlin`, and `:mtg-sdk:test` all green.


wingedsheep:
**Correctness follow-up (commit fceb0af):** the first commit resolved the participant via `Player.ControllerOf`, which reads the stack object's `ControllerComponent`. But a spell on the stack keeps the `ControllerComponent` it had in hand (= the card's **owner** — `GameInitializer.kt:360`), and neither `moveToZone` nor `CastSpellHandler` updates it on cast. So `ControllerOf` gave the owner, which is wrong for theft-cast spells (casting a card you don't own: the spell's controller is the caster, not the owner).

Fixed by resolving a target's controller from `SpellOnStackComponent.casterId` first (the canonical spell controller), then `ControllerComponent`, then owner — applied to `TargetController`, `ControllerOfPipelineTarget`, and `Player.ControllerOf` via one helper. This restores the original `casterId` semantics while keeping the generalized `participant: Player` design, and makes "controller of target spell" correct for any future effect. Full `:rules-engine:test` suite is green.

PR #374, opened 2026-05-31
Add TDM cards (group C): Temur Devotee, Sultai Devotee, Mardu Monument, Nightblade Brigade

wingedsheep:
Implements four Tarkir: Dragonstorm (TDM) cards.

## Cards
- **Temur Devotee** ({1}{U} 3/3 Human Druid) — Defender; `{1}: Add {G}, {U}, or {R}. Activate only once each turn.` Uses the `keywords(DEFENDER)` helper plus the standard once-per-turn `AddManaOfChoice` mana ability (same shape as the other Devotees in the set).
- **Sultai Devotee** ({1}{G} 2/1 Zombie Snake Druid) — Deathtouch; `{1}: Add {B}, {G}, or {U}. Activate only once each turn.`
- **Mardu Monument** ({2} Artifact) — ETB searches the library for a basic Mountain/Plains/Swamp to hand then shuffles (atomic `EffectPatterns.searchLibrary`). `{2}{R}{W}{B}, {T}, Sacrifice this artifact:` create three 1/1 red Warrior tokens that gain menace and haste until end of turn (sorcery speed).
- **Nightblade Brigade** ({2}{B} 1/3 Goblin Soldier) — Deathtouch, Mobilize 1, and ETB surveil 1 (`EffectPatterns.surveil`).

## Engine change
Mardu Monument's "they gain menace and haste until end of turn" composes `CreateToken` with per-token `GrantKeyword` (duration EndOfTurn) addressed via `EffectTarget.PipelineTarget(CREATED_TOKENS, index)`. `CreateTokenExecutor` previously did not publish its created tokens to the `CREATED_TOKENS` pipeline (only `CreatePredefinedTokenExecutor` did), so this PR makes `CreateTokenExecutor` publish them too — a small, additive, reusable change letting any composite address freshly-created tokens. The SDK reference is updated accordingly.

## Tests
- Added `MarduMonumentScenarioTest` covering the ETB land search and verifying all three tokens are 1/1 with menace and haste.
- Full `:rules-engine` suite and the TDM Monument/Devotee scenario tests pass.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -357,7 +357,10 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 
 - `CreateToken(name, p, t, colors?, subtypes?, keywords?, count?, tapped?)` — make N tokens. `count` accepts an
   `Int` or a `DynamicAmount` (the latter for "create X tokens" wording — e.g. Verdeloth the Ancient passes
-  `count = DynamicAmount.XValue` to make X Saprolings when kicked).
+  `count = DynamicAmount.XValue` to make X Saprolings when kicked). Publishes the created token entity IDs to the
+  `CREATED_TOKENS` pipeline collection, so a sibling effect in a `CompositeEffect` can address each token via
+  `EffectTarget.PipelineTarget(CREATED_TOKENS, index)` — e.g. Mardu Monument grants menace and haste until end of
+  turn to each of its three freshly-created Warriors with one `GrantKeyword` per token.
 - `CreateDynamicToken(dynamicPower, dynamicToughness, colors?, creatureTypes, keywords?, count?, controller?, imageUri?)` —
   tokens whose P/T is computed at resolution (e.g. Pure Reflection's X/X Reflection where X = the cast spell's mana
   value, via `DynamicAmounts.triggeringManaValue()`). `controller` directs who gets the token (e.g.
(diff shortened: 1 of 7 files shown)


wingedsheep:
## Review — approved with one small cleanup applied

Behavior is correct and the SDK shape is right. Verified all four cards against Scryfall Oracle (mana cost, type line, P/T, and full text all match), confirmed printing placement (`just check-card-printing` passes for all four — TDM is each card's earliest/only printing), and ran the tests.

**The engine change is exemplary.** Publishing `CreateTokenEffect`'s output to the existing `CREATED_TOKENS` pipeline is the minimal, additive, reusable fix — it mirrors `CreatePredefinedTokenExecutor` line-for-line, introduces no new SDK type, and the gap was genuinely real (existing `EffectPatterns` only ever published from the *predefined* executor). Mardu Monument's "they gain menace and haste" then composes purely from existing primitives (`CreateToken` + `GrantKeyword` + `PipelineTarget`). The `CompositeEffectExecutor` already threads `updatedCollections` between sub-effects, so the wiring is end-to-end correct.

**Tests run green locally:**
- `MarduMonumentScenarioTest` (both cases) — verifies the ETB land search *and* that all three tokens are 1/1 with menace+haste **after the Monument is sacrificed** (the load-bearing assertion: EndOfTurn floating effects must outlive their source).
- `*Token*` / `*Mobilize*` in `rules-engine` — no regression from the executor change.

**Applied (pushed as a follow-up commit):** `MarduMonument.kt` used the raw `CreateTokenEffect(...)` constructor; CLAUDE.md prefers the `Effects.*` facade. Swapped to `Effects.CreateToken(...)` (drop-in — the facade takes every param incl. `imageUri`) and dropped the now-unused import. Scenario test still passes.

**Minor (not blocking, your call):** the two Devotees and Nightblade Brigade have no dedicated scenario tests — consistent with the existing three Devotees, and `mobilize`/`surveil` are covered by their own suites. The one genuinely untested axis anywhere is the once-per-turn `AddManaOfChoice` restriction, if you want a cheap test out of this.

LGTM. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #383, opened 2026-06-01
Add 5 EOE cards across three small engine extensions

bolav:
Implements Cosmogoyf, Roving Actuator (no engine work), Blade of the Swarm (target-enumeration for Zone.EXILE), Orbital Plunge (excess-damage condition), and Quantum Riddler (ModifyDrawAmount replacement). 242 → 247 / 261 EOE booster.

SDK additions:
- Conditions.IfTargetTookExcessDamage / TargetMarkedDamageExceedsToughness — post-damage check that the target creature's marked damage strictly exceeds toughness. The check is `marked > toughness` regardless of which preceding step dealt the damage; CompositeEffect doesn't interleave SBA or trigger processing between sub-effects, so for the canonical "deal N, then check" pipeline this is equivalent to "did the preceding step deal excess." The defensive guards (non-creature target, target left the battlefield) are unreachable under `Targets.Creature` + Composite and retained for future callers — documented in both KDocs and the SDK reference.
- ModifyDrawAmount(modifier, restrictions, appliesTo) replacement effect, shaped like ModifyLifeLoss with a Condition list for arbitrary gating.

Engine wiring:
- TargetEnumerationUtils gains Zone.EXILE support via findValidExileTargets, honoring ControlledByYou / OwnedByYou predicates the same way as graveyard.
- DrawReplacementDispatcher.applyDrawAmountModifier consulted once per draw instruction at every announcement site (CR 121.2a — modifications to the draw count happen before any individual card draws):
    * DrawCardsExecutor.execute for spell/ability draws
    * DrawPhaseManager.performDrawStep for the draw step
    * CycleCardHandler for "Discard this, Draw a card" (CR 702.32a)
    * EachPlayerDrawsForDamageDealtToSourceExecutor, per-player, since each player's instruction is independently announced (CR 121.2c) Exposed as DrawCardsExecutor.applyDrawAmountModifier so the cycle and per-player-draw call sites — which drive executeDraws directly instead of going through execute() — can wire the announcement without learning about the dispatcher abstraction. Resume paths (DrawReplacementContinuationResumer, CoreAutoResumerModule) intentionally skip — they're continuing an already-modified instruction.
- ModifyDrawAmount added to StaticAbilityHandler.isRuntimeReplacementEffect.

UI: generalize the graveyard targeting overlay to also handle Zone.EXILE so Blade of the Swarm's "target exiled card with warp" mode renders the same pile-selection UI with "Your Exile" / "Opponent's Exile" labels instead of falling back to BattlefieldTargetingUI.

Tests: per-card scenarios in game-server cover both branches of each new mechanic (excess vs equality vs sub-lethal damage; hand-size threshold met vs not; self-exile vs opponent-exile target). QuantumRiddlerScenarioTest additionally regresses the draw-step announcement site (turn 2 with hand = 0 draws 2) and the cycling announcement site (cycling Hundroog with the hand-size restriction holding draws 2). Manual dev scenarios under manual-scenarios/cards/ stage each card so the key behavior is reachable in the dev UI in a single cast or short play sequence.

### DIFF
--- a/backlog/sets/edge-of-eternities/bugs.md
+++ b/backlog/sets/edge-of-eternities/bugs.md
@@ -7,3 +7,5 @@
 - Rust Harvester, +1/+1 counter is not applied if target is invalid
 - Wedgelight Rammer did not turn to creature
 - Spacecraft can attack 1 turn after being played
+
+- Auxiliary Boosters - equipment not attached
--- a/backlog/sets/edge-of-eternities/cards.md
+++ b/backlog/sets/edge-of-eternities/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 261 booster cards (excluding basic lands)
 **Release Date:** August 1, 2025
-**Implemented:** 242 / 261
+**Implemented:** 247 / 261
 ---
 
 - [x] Adagia, Windswept Bastion
@@ -24,7 +24,7 @@
 - [x] Biomechan Engineer
 - [x] Biosynthic Burst
 - [x] Biotech Specialist
-- [ ] Blade of the Swarm
+- [x] Blade of the Swarm
 - [x] Blooming Stinger
 - [x] Bombard
 - [x] Breeding Pool
@@ -40,7 +40,7 @@
 - [x] Comet Crawler
 - [x] Command Bridge
 - [x] Consult the Star Charts
-- [ ] Cosmogoyf
+- [x] Cosmogoyf
 - [x] Cosmogrand Zenith
 - [x] Cryogen Relic
 - [x] Cryoshatter
@@ -155,7 +155,7 @@
 - [x] Nebula Dragon
 - [x] Nova Hellkite
 - [x] Nutrient Block
-- [ ] Orbital Plunge
+- [x] Orbital Plunge
 - [x] Oreplate Pangolin
 - [x] Ouroboroid
 - [x] Pain for All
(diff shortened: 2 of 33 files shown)


wingedsheep:
Pushed a fixup commit (fc2613002) addressing two review items:

**1. Cycling CR citation corrected → 702.29a.** The comments in `CycleCardHandler.kt` and the `QuantumRiddlerScenarioTest` cycling case cited **702.32a**, but 702.32 is *Fading*. Cycling is **702.29a** ("[Cost], Discard this card: Draw a card") — verified against the official Comprehensive Rules (20260417). The other three citations (121.2a, 121.2c, 707.12) check out.

**2. Added `RovingActuatorScenarioTest`.** Roving Actuator was the only new card without an automated test (just a manual scenario JSON). The card reuses the Shiko exile→copy→may-cast pipeline, so the new test focuses on the card's own wiring rather than the shared pipeline:
- Void satisfied → ETB exiles a graveyard Shock, copies it, casts the copy for free (opponent to 18; original stays exiled).
- Void off (no nonland permanent left / no spell warped this turn) → the ETB ability doesn't trigger.

All four existing scenario tests plus the two new ones pass. Nice work on the rest — both new SDK primitives (`ModifyDrawAmount` at the CR 121.2a announcement site, `TargetMarkedDamageExceedsToughness`) are well-justified and the other four cards are clean composition.

PR #390, opened 2026-06-01
Add Leyline of Hope to Duskmourn with leyline opening-hand mechanic

bolav:
Implements DSK #18 with three new SDK/engine primitives:

1. Leyline mechanic. `CardBuilder.leyline()` sets `CardScript.mayStartOnBattlefield`. After mulligans and bottoming complete (CR 103.6), `MulliganHandler.initiateLeylinePhase` populates each player's `pendingLeylineCardIds` and the engine walks turn order from the active player, pausing per leyline card with a YesNoDecision resumed by `LeylineDecisionContinuation`. "Yes" routes the card through `ZoneTransitionService` (hand -> battlefield as owner); "no" leaves it in hand. `SubmitDecisionHandler` then advances the UNTAP step into turn 1. The keep/bottom completion paths now funnel through a single `MulliganHandler.tryAdvancePastMulliganPhase` so the leyline phase has one point of truth.

2. `ModifyLifeGain` gains a flat `modifier` (mirrors `ModifyLifeLoss`) and is now wired into `GainLifeExecutor`, `OwnerGainsLifeExecutor`, the noncombat lifelink path in `DamageUtils`, and the combat lifelink path in `CombatDamageManager` via the shared `LifeGainModifiers.apply` helper. Stacks additively across multiple sources per the Scryfall rulings.

3. The conditional anthem composes existing primitives: `ConditionalStaticAbility(ModifyStats(+2/+2, creaturesYouControl), Compare(LifeTotal(You) GTE Add(StartingLifeTotal(You), 7)))`.

`ZoneTransitionService.moveToZone` now registers `ContinuousEffect` and `ReplacementEffect` source components on permanents entering the battlefield through the shared path (leyline starts, reanimation, returns from exile). Without this, a leyline-yes would land Leyline of Hope on the battlefield with no anthem or +1 life rider active. The cast pipeline keeps its own registration in `StackResolver`, so this is additive — no double-registration.

Tests: 8 scenario tests in `LeylineOfHopeScenarioTest` cover the +1 modifier (single + stacking + opponent isolation + combat lifelink via Healer's Hawk), the conditional anthem on both sides of the 7-life threshold, and the opening-hand yes/no flow built on top of `GameInitializer` (asserting the +1 rider survives the leyline-yes path via `ContinuousEffectSourceComponent` /
`ReplacementEffectSourceComponent` attachment).

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -57,6 +57,11 @@ section; do not let SDK additions land without a corresponding doc update.
 - `replacementEffect { ... }` — "instead/if … would" replacement.
 - `keywords(...)` / `keywordAbility(...)` / `keywordAbilities(...)` — add keyword abilities.
 - `spell { ... }` — define the spell payload for instants/sorceries and Adventure / Omen faces.
+- `leyline()` — Leyline mechanic ("If this card is in your opening hand, you may begin the game with it on the
+  battlefield"). Sets `CardScript.mayStartOnBattlefield = true`. After all mulligans and bottoming resolve, the
+  engine walks each player in turn order from the active player and presents a yes/no decision per leyline card
+  in their opening hand; a "yes" routes the card to the battlefield through the standard zone-change pipeline
+  before the first turn begins, a "no" leaves it in hand.
 
 ---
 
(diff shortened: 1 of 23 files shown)


wingedsheep:
Pushed a small fixup (`f02f14b`) addressing review feedback. Three changes:

1. **Projected controller in `LifeGainModifiers.apply`** — it was reading the base `ControllerComponent` to decide whose `Player.You` the `+1` matches. That's the base-vs-projected pitfall: if a Leyline of Hope changes controller (Confiscate, etc.), the base component still points at the owner, so the buff would stick to the wrong player. Now resolves via `state.projectedState.getController(entityId)` with a base fallback, matching the existing pattern in `DamageUtils`. Rare in practice, but it's the engine's load-bearing projection rule.

2. **CR citation fix (lifelink)** — the `DamageUtils` comment cited **CR 119.9**, which is actually about *triggered-ability* wording ("Whenever a player gains life…"). The relevant rules for lifelink causing a replaceable life-gain event are **CR 119.3 / 702.15b**. Updated.

3. **CR citation fix (leyline window)** — the `MulliganStateComponent` doc cited **CR 103.5** (the mulligan process). The opening-hand-actions window is **CR 103.6** (103.6a covers "begin the game with this on the battlefield"). Updated.

Also added a one-line note on the multiply-then-add ordering for mixed-shape `ModifyLifeGain` stacks (CR 616.1 lets the player choose order; currently unreachable since Alhammarret's Archive isn't implemented, and same-shape stacks are order-independent).

The SDK shape was already right — `ModifyLifeGain.modifier` mirroring `ModifyLifeLoss`, the anthem composing `ConditionalStaticAbility`/`ModifyStats`/`Compare`, and `leyline()` as a reusable flag for the whole cycle — so no structural changes. All 8 scenario tests plus the full `rules-engine` + `game-server` suites stay green.

Note: I rebased the fixup directly onto the PR head, so the branch stays linear (no `main` merge commit). `origin/main` merges in cleanly if you want to pull it.

PR #392, opened 2026-06-01
Add Weapons Manufacturing to Edge of Eternities

bolav:
Required two engine changes to support the card's Munitions artifact token:

- `CreateTokenEffect` gained a `nonCreature` flag so a single token effect can produce a non-creature artifact token with embedded triggered abilities, without falling back to the predefined-token registry for a one-off shape.
- `TriggerMatcher` was reading `ZoneChangeEvent.lastKnownWasToken` for the `IsNontoken`/`IsToken` predicates, which is only populated when a permanent leaves the battlefield. On enter-battlefield events the bit was always `false`, so token entries incorrectly matched `IsNontoken` filters — which would have caused Weapons Manufacturing to re-trigger on its own Munitions tokens. Read `TokenComponent` live when the event is not a leave.

### DIFF
--- a/backlog/sets/edge-of-eternities/cards.md
+++ b/backlog/sets/edge-of-eternities/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 261 booster cards (excluding basic lands)
 **Release Date:** August 1, 2025
-**Implemented:** 247 / 261
+**Implemented:** 248 / 261
 ---
 
 - [x] Adagia, Windswept Bastion
@@ -256,7 +256,7 @@
 - [x] Vote Out
 - [x] Warmaker Gunship
 - [x] Watery Grave
-- [ ] Weapons Manufacturing
+- [x] Weapons Manufacturing
 - [x] Wedgelight Rammer
 - [x] Weftblade Enhancer
 - [x] Weftstalker Ardent
--- a/backlog/sets/edge-of-eternities/missing-effects.md
+++ b/backlog/sets/edge-of-eternities/missing-effects.md
@@ -4,7 +4,7 @@ Engine features required to implement the remaining EOE cards. Each section list
 unblocks, the exact oracle clause that can't be expressed with current primitives, and a sketch of
 the engine/SDK work needed.
 
-As of the latest pass, **247 / 261** booster cards are implemented. The cards below remain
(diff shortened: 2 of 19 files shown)


wingedsheep:
Pushed two commits to `weapons-manufacturing`:

**1. Merged `origin/main`** into the branch (clean, no conflicts) so the review reflects post-merge reality. You can drop this merge commit and rebase if you'd rather keep history linear.

**2. Review fixups** (`31877c2`):

- **Removed the unused `CreateTokenEffect.nonCreature` flag** plus its `CreateTokenExecutor` branch and the `description` rewrite. No card sets `nonCreature = true` — Munitions is built through the predefined-token registry (`card("Munitions") { typeLine = "Artifact"; triggeredAbility {...} }`), which already supports noncreature type lines and embedded abilities. As written, the flag/executor branch/description refactor were dead, untested code, and the PR body's claim that the token is made "without falling back to the predefined-token registry" didn't match the actual implementation. `CreateTokenExecutor.kt` and `TokenEffects.kt` now match `main` exactly.
- **Reverted the "escape hatch" notes** in `missing-effects.md` §11 and `card-sdk-language-reference.md` — clarified that the predefined-token registry is the mechanism.
- **Fixed a stale CR citation** in `TriggerMatcher.kt`: the comment cited `704.5s` (that's the Saga-sacrifice SBA); the token-ceases-to-exist SBA is `704.5d` — matching the (correct) citation you added in `DamageEffects.kt`.

**What I kept as-is — this is the good part of the PR:**
- The `TriggerMatcher` fix is a real, correct bug fix. `IsNontoken`/`IsToken` were reading `lastKnownWasToken` (only populated on leave), so on enter-battlefield `IsNontoken` always passed (WM would re-trigger off its own Munitions) and `IsToken` never matched. Branching on `event.fromZone == Zone.BATTLEFIELD` (LKI for leaves, live `TokenComponent` for enters) is the right shape.
- The predefined-token implementation of Munitions is idiomatic (matches Cragflame/Sword/Treasure).
- The `DealDamageEffect` doc on null `damageSource` + LKI is accurate and useful (CR 704.5d verified).

Tests run green after the changes: `WeaponsManufacturingScenarioTest` (all 3, including the token-doesn't-re-trigger case), `TriggerMatcherIsNontokenTest`, `IncubateTest`, and the full `:mtg-sdk:test` suite.

PR #393, opened 2026-06-01
Add Erg Raiders with per-creature SourceAttackedThisTurn predicate

bolav:
Engine feature: a per-creature "attacked this turn" history predicate (CR 506.2-derived) — `StatePredicate.AttackedThisTurn`, paired with `ObjectFilter.attackedThisTurn()` and `Conditions.SourceAttackedThisTurn`.

The data already exists: `PlayerAttackersThisTurnComponent` is maintained on each player by `AttackPhaseManager.declareAttackers` (the same set behind raid / "you attacked with N creatures this turn" tribal triggers). The new predicate just reads it via the entity's controller, so no new component, lifecycle hook, or cleanup pass is needed.

Wires through:
 * `PredicateEvaluator` and `AffectsFilterResolver` evaluate it directly.
 * The exhaustive `when`s in `BeginningPhaseManager.matchesStatePredicate ForUntap` and `TriggerMatcher.matchesStatePredicateForTrigger` list it alongside the other history predicates with the same "no-op in this context" branch.

Card: Erg Raiders' end-step intervening-if composes as `AllConditions(Not(SourceAttackedThisTurn), Not(SourceEnteredThisTurn))`, the standard summoning-sickness proxy for "came under your control this turn". Effect: `DealDamage(2, EffectTarget.Controller, damageSource = EffectTarget.Self)`.

Scenario test covers all three branches:
 * settled, didn't attack → 2 damage to controller,
 * attacked this turn → trigger doesn't fire,
 * came under control this turn → trigger doesn't fire.

Existing `MarduSkullhunterTest` (the closest raid-style regression) still passes.

### DIFF
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 78 cards
 **Release Date:** December 17, 1993
-**Implemented:** 51 / 78
+**Implemented:** 53 / 78
 | Color | Count |
 |-------|-------|
 | White | 11 |
@@ -39,7 +39,7 @@
 - [x] Unstable Mutation
 - [ ] Cuombajj Witches
 - [x] El-Hajjâj
-- [ ] Erg Raiders
+- [x] Erg Raiders
 - [ ] Guardian Beast
 - [x] Hasran Ogress
 - [x] Junún Efreet
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1612,6 +1612,9 @@ contexts.
 - `SourceIsTapped` — source is tapped.
 - `SourceIsUntapped` — source is untapped.
 - `SourceEnteredThisTurn` — source entered the battlefield this turn.
+- `SourceAttackedThisTurn` — source was declared as an attacker at least once during the
+  current turn (per-creature, derived from the controller's `PlayerAttackersThisTurnComponent`).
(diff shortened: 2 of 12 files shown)


wingedsheep:
Reviewed and pushed a small fixup commit (`bf14562a`) with two changes from the review:

**1. Corrected the CR citation (508.1, was 506.2).** `StatePredicate.AttackedThisTurn`'s doc comment cited CR 506.2, but 506.2 designates which player is the *attacking/defending player* during combat — it says nothing about attack history. The attacker set is populated during the **declare-attackers step (CR 508.1)**, so the comment now reads "set during the declare-attackers step, CR 508.1". (Verified against the current Comprehensive Rules: 508.1 — "First, the active player declares attackers.")

**2. Tidied inline fully-qualified names.** `PredicateEvaluator` and `AffectsFilterResolver` referenced `PlayerAttackersThisTurnComponent` / `ControllerComponent` by full package path inline, while every neighboring branch uses the imported short name. Added the imports and switched to short names for consistency. Pure cosmetics — no behavior change.

The fixup is cherry-picked directly onto the PR head (no `main` merge in it). `ErgRaidersScenarioTest` (all 3 cases) and `:mtg-sdk:test` are green after the change.

Otherwise the PR is in great shape — the `AttackedThisTurn` primitive is exactly the right SDK shape: it reads per-creature attack history by reusing the existing `PlayerAttackersThisTurnComponent` (no new component, lifecycle hook, or cleanup pass), it's parameter-free and composable for future cards, and the canonical card def is correctly placed in `arn` (earliest real printing).

PR #437, opened 2026-06-02
Add Éowyn, Fearless Knight with source-relative power filter and mult…

bolav:
…i-color protection grant

Éowyn (LTR #201) needed two SDK additions:

- CardPredicate.PowerGreaterThanEntity(reference) — power comparison against a referenced entity (mirrors ManaValueAtMostEntity in shape). Powers the "exile target creature with greater power" target filter via GameObjectFilter.powerGreaterThanEntity(EntityReference.Source) and the matching TargetFilter passthrough.

- Effects.GrantProtectionFromColorsOfEntity(filter, source, duration) — reads the projected colors of the referenced entity and adds one PROTECTION_FROM_<COLOR> floating effect per color for the snapshot of matching permanents. One-shot at resolution. The LKI fallback to base CardComponent.colors fires only when the source has actually left the battlefield, so a legitimately-colorless on-battlefield source (Devoid, Layer-5 "becomes colorless") grants nothing per CR 105.2 — the projected empty set is authoritative. Composed with Exile in a CompositeEffect so the grant resolves while the target is still on the battlefield.

Engine plumbing: PredicateEvaluator resolves PowerGreaterThanEntity (Source → context.sourceId) against projected power; AffectsFilterResolver, CostCalculator, and TriggerMatcher get the inert false branches.

Scenario tests cover the happy path (4/4 UB target exiled, my legendary gains PROTECTION_FROM_BLUE/_BLACK, non-legendary unaffected), the strict greater-power filter (3/3 excluded), a colorless 4/4 target (no protection grants, CR 105.2), and the EndOfTurn duration cleanup.

A TODO on GrantProtectionFromColorsOfEntityEffect notes the decomposition path (ForEachColorOf combinator) for the next card that needs "for each color of [entity], do X".

SDK reference doc and LTR backlog updated.


wingedsheep:
Reviewed and pushed two follow-up changes to this branch (`ltr-eowyn-fearless`). Also merged current `origin/main` in (clean, no conflicts) so the diff reflects post-merge reality.

**1. Added a test that exercises the projected-vs-printed color distinction.**

The executor's documentation (and the PR description) centers on the grant reading the target's *projected* colors so Layer-5 "becomes colorless"/Devoid and color-*adding* effects are honored, with printed colors only as off-battlefield LKI. But none of the existing tests actually exercised projected ≠ base:
- the colorless test used a *printed*-colorless creature (base and projected both empty), and
- the happy path used a printed-UB creature (base and projected both UB).

Either passes whether you read projection or base, so the projection logic was effectively untested. The new test recolors the 4/4 exile target to green with a Layer-5 effect ("target creature becomes green until end of turn"), then asserts the legendary creatures gain `PROTECTION_FROM_GREEN` and **not** `PROTECTION_FROM_RED` — pinning the behavior to projected colors.

**2. Renamed `GrantProtectionFromColorsOfEntityExecutor.readColorsAsLastKnown` → `readSourceColors`.** On the battlefield it returns the source's *current projected* colors, not last-known information — "AsLastKnown" only described the off-battlefield fallback branch.

All 5 scenario tests green (`:rules-engine:test --tests "*EowynFearlessKnightScenarioTest"`).

Two notes left for your discretion, not changed:
- `GrantProtectionFromColorsOfEntityEffect` is a monolith (color-iteration + protection-grant fused). Landing it as-is is the right call — it's generically parameterized and building the `ForEachColorOf` combinator speculatively for one card would cut against "inline until a second user." Your TODO already records the decomposition path; just keep it honest when a second "for each color of [entity]" card lands.
- All cited CR numbers (105.2, 611.1, 113.7a, 608.2) check out against the current Comprehensive Rules. Printing placement is correct (`check-card-printing` passes; canonical in LTR, the earliest real printing).

wingedsheep:
Split the `GrantProtectionFromColorsOfEntityEffect` monolith into a reusable combinator (commit `643f7a155`).

**New primitive: `Effects.ForEachColorOf(source, effect)`** — the non-interactive sibling of `ChooseColorThen`. It runs the inner effect once per color of the referenced entity, placing that color in `EffectContext.chosenColor`, so the existing per-color atoms (`GrantProtectionFromChosenColor`, `GrantHexproofFromChosenColor`, `GrantCantBeBlockedByChosenColor`, …) compose inside it exactly as they do inside `ChooseColorThen`. Lives in the composite executor family (`ForEachColorOfExecutor`, mirroring `ForEachInGroupExecutor`) because it's colour-agnostic, not protection-specific.

Éowyn's grant is now the same recipe **Akroma's Blessing** already uses for "creatures you control gain protection from the chosen color" — just driven by an entity's colours instead of a player prompt:

```kotlin
ForEachColorOf(Target(0),
    ForEachInGroup(legendariesYouControl,
        GrantProtectionFromChosenColor(Self)))
```

The projected-over-base colour reading (Layer-5 / Devoid honored on the battlefield, printed colours as LKI off it, colorless → zero iterations) moves onto the iterator, so it's reusable rather than baked into a protection effect. Removed the old effect, its facade, and its executor.

Green: all Éowyn scenarios (multi-colour, colorless, Layer-5 recolor, end-of-turn), Akroma's Blessing, and the card/polymorphic serialization round-trips. SDK reference updated.

PR #441, opened 2026-06-02
Arabian Nights: state-triggered abilities + unblocked-attack trigger

bolav:
## Summary

Adds two engine primitives needed by a cluster of early-set cards, and lands three
ARN cards that compose on top of them. Branch `arn-state-trigger-no-lands` off `main`.

### Engine
- **State-triggered abilities (CR 603.8)** — new `StateTriggeredAbility` SDK type +
  `CardBuilder.stateTriggeredAbility { }` block. Engine polls each battlefield
  permanent's conditions at every priority check via a new `StateTriggerPoller` and
  emits `PendingTrigger`s on false→true transitions; a per-entity
  `StateTriggerLatchesComponent` prevents re-firing while the condition stays true
  (603.8d). `Conditions.YouControl(filter, negate = true)` for "you control no X".
- **`Triggers.AttacksAndIsntBlocked` (CR 509.7)** — new `BecomesUnblockedEvent`;
  detection piggybacks on `BlockersDeclaredEvent` (SELF attacker not in any
  blocker's blocked-attackers list). Mirrors `BecomesBlockedEvent`.
- `SacrificeExecutor` now captures permanent snapshots on the auto-sac path so
  `Conditions.SacrificedHadSubtype` works after `EffectPatterns.sacrifice(...)`.

### Cards (3 / 4 from the original batch)
- **Dandân** · **Island Fish Jasconius** · **Merchant Ship** — each with a
  scenario test and a manual `DevScenarioController` JSON for live testing.
- **Serendib Djinn deferred** — its "if you sacrifice an Island this way" clause
  still needs `SacrificeContinuation` to thread `updatedSacrificedPermanents` for
  the multi-land decision path. Tracked in `backlog/sets/arabian-nights/TODO.md`.

ARN backlog: 53 → 56 / 78.

### DIFF
--- a/backlog/sets/arabian-nights/TODO.md
+++ b/backlog/sets/arabian-nights/TODO.md
@@ -9,41 +9,27 @@
 Set scaffolding is done:
 
 - `mtg-sets/.../definitions/arn/ArabianNightsSet.kt` — registered in `MtgSetCatalog.all`.
-- `mtg-sets/.../definitions/arn/cards/` — empty, ready for one file per card.
+- `mtg-sets/.../definitions/arn/cards/` — one file per card.
 - `backlog/sets/arabian-nights/cards.md` — the 78-card checklist (mark `[x]` as you go).
 - `backlog/sets/arabian-nights/arn_set.json` — the **offline card-data source** for this set.
 
 Verify status anytime with: `scripts/card-status --set ARN` (and `--list --set ARN`).
 
-## Progress (as of 2026-05-26)
+## Progress (as of 2026-06-02)
 
-**Implemented: 46 / 78** (`scripts/card-status --set ARN`). Full per-card status lives in
+**Implemented: 56 / 78** (`scripts/card-status --set ARN`). Full per-card status lives in
 `cards.md` — that file's `Implemented:` count is the source of truth; keep it in sync as boxes
 are checked. The triage buckets below are pruned to **remaining work only**; a card is removed
 from its bucket once it lands on a branch.
 
-Note: **Banding is fully modelled in the engine** (CR 702.22 — `Keyword.BANDING`, handled in
-`CombatManager` / `CombatDamageManager` / `AttackPhaseManager`; `BandDeclarationTest`,
-`BandingTrampleDrainScenarioTest`, and `BandingCombatScenarioTest` all pass, and Mirage's Noble
-Elephant uses it with no per-card wiring). Banding cards therefore compose onto `arn-cards`.
+**In-flight engine-change branches** (each off `main`, pending its own PR):
 
-The original banding triage was inaccurate (verified against `arn_set.json`): Moorish Cavalry is
-Trample only and Hurr Jackal is regeneration-denial — neither has banding (both already done).
-Aladdin's control-change also composed onto `arn-cards`. The only true banding cards are **Camel**
-and **War Elephant**; Abu Ja'far is a death trigger, not banding (moved to Bucket B).
-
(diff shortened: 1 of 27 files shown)


wingedsheep (commented):
## Review fixes applied (commit `c4ca785`)

I reviewed this PR and pushed a follow-up commit addressing the findings, plus a merge of current `origin/main` (clean, no conflicts). The core behaviour was already correct and the SDK shape is right — `StateTriggeredAbility` is a genuinely novel primitive (condition-polling, not event-matching) and `BecomesUnblockedEvent` is a clean mirror of `BecomesBlockedEvent`. The fixes are accuracy, dead-API removal, and test coverage. All scenario tests + `:mtg-sdk:test` are green.

### 1. CR citation corrections (verified against the current Comprehensive Rules)
- **Unblocked-attack trigger: `509.7` → `509.3g`.** Rule 509 stops at 509.4b; the rule for *"Whenever [a creature] attacks and isn't blocked"* is **509.3g**. Fixed in `Triggers`, `GameEvent`, `TriggerMatcher`, the SDK reference, and the Merchant Ship test.
- **State triggers: `603.8a–d` / `603.8d` → `603.8`.** 603.8 is a single paragraph with no lettered subsections — the cited ones don't exist.
- **Sacrifice snapshot: `112.7a` → `113.7a`.** The last-known-information rule is 113.7a (608.2h literally says *"See rule 113.7a"*). Note: the rest of the codebase still uses the stale `112.7a` from an older CR edition (~20 references) — I only corrected the citation this PR introduced; a codebase-wide sweep is out of scope here.

### 2. Removed dead, speculative API
`Triggers.attacksAndIsntBlocked(filter)` (ANY-binding) could never fire — `TriggerMatcher` only handles the `SELF` binding for `BecomesUnblockedEvent` and never applied the filter — yet it was documented as working. Removed the factory and the now-dead `filter` field on `BecomesUnblockedEvent` (now a `data object`, like `StateConditionMetEvent`). The ANY-filtered variant should be added together with its matcher/detector branches when a card actually needs it.

### 3. Documented the deliberate latch divergence from CR 603.8
The printed rule resets the latch when the ability **leaves the stack** and re-triggers if the condition is still true; this engine resets when the condition next evaluates **false**. The two agree for every state trigger whose effect removes its source or clears the condition (all three shipped "sacrifice this creature" cards), and diverge only for a state trigger that leaves both source and condition intact — no such card exists yet. Documented in `StateTriggeredAbility`, `StateTriggerLatchesComponent`, and the SDK reference, with a note to wire reset-on-leaves-the-stack before authoring such a card.

### 4. Test coverage for the latch (the defining feature)
The three shipped cards all sacrifice their source on resolution, so none of them actually exercise 603.8 re-fire suppression. Added `StateTriggerLatchScenarioTest`: an inline card with a non-self-removing state trigger (*"control no Islands → gain 1 life"*) must gain **exactly 1 life** across a full turn of priority passes. A broken latch would re-fire every priority check and inflate the total.

### 5. Minor cleanups
- Dropped the dead `classLevel` computation + unused `filterByZone` param in `StateTriggerPoller`.
- Fixed the `stateTriggeredAbility { }` KDoc example (`Effects.Sacrifice` → `Effects.SacrificeTarget`).

### Left as-is
The `SacrificeExecutor` snapshot capture has no consumer in this PR (Serendib Djinn is deferred), but it's consistent with the existing `SacrificeTargetExecutor` and brings the auto-sac path to parity — kept it; it'll be exercised when Serendib Djinn lands.

PR #443, opened 2026-06-02
Add hope/verse/influence/burden counter types + Dawn of a New Age (LTR)

bolav:
## Summary
- Adds four passive counter types from LTR — `hope`, `verse`, `influence`, `burden` — wired through every counter layer: `CounterType` enum + `Counters` string constants (mtg-sdk), web-client `CounterType` + display names, mana-font icon mapping, badge getters/styles,
  and badge rendering.
- Documents the existing stun counter (CR 122.1d, already implemented via `untapOrConsumeStun`) in the SDK reference for completeness.
- Implements **Dawn of a New Age** composing the new `hope` counter: `EntersWithDynamicCounters` for the ETB-with-N-per-creature half, and a `Composite(Conditional, Conditional)` end-step chain modeling "remove a counter, if you do draw; then if zero counters, sac +
gain 4 life".
- Sourced oracle from `ltr_set.json` (per LTR backlog rule, not Scryfall).

## Scope
This is LTR backlog Gap 6. Dawn of a New Age is the only card that cleanly composes from the new counters. Lost Isle Calling needs a separate engine gap (read source counters at activation time after `ExileSelf` — CR 122.2 wipes them at zone change); Palantír of
Orthanc needs an opponent-makes-may-decision-on-your-trigger primitive plus a mana-value-sum-over-collection dynamic amount. Both documented inline in `TODO.md` for the next pass.


wingedsheep:
Pushed a review-fix commit (`87f108c`) addressing three points from review:

**1. Corrected CR citation in `DawnOfANewAge.kt`.** The "Then if … no hope counters" note cited `CR 605.3a / 700.1` — but 605.3a is about *activating mana abilities* mid-resolution and 700.1 is the generic "anything that happens is an event" definition; neither supports a resolution-time re-check. Replaced with **CR 608.2c** ("controller follows its instructions in the order written"), and clarified that this is *not* an intervening-"if" (CR 603.4 covers only the "if" immediately following a trigger condition). Behavior was already correct and tested — only the citation was wrong.

**2. Distinct icons for the new counters.** `influence` was mapped to `counter-charge`, byte-identical to `hope`, so the two would be visually indistinguishable on the board. Now each LTR passive counter has its own mana-font glyph: `hope → counter-charge`, `verse → counter-verse` (there's a dedicated one), `influence → counter-devotion`, `burden → counter-doom`.

**3. Deduped the badge rendering.** The four passive-counter badges were copy-pasted across three files (4 getters in `shared.ts`, 4 style objects in `styles.ts`, 4 JSX blocks in `GameCard.tsx`). Collapsed into a generic `getCounterCount(card, type)` + a `PASSIVE_COUNTER_TYPES` list + a palette-driven `passiveCounterBadgeStyle(type)`, rendered data-driven with a single `.map()`. Net −69 lines; adding the next passive counter is now a one-line palette + icon entry.

Verified: `DawnOfANewAgeScenarioTest` (all 4) green, web-client `tsc --noEmit` clean. The commit is on top of your original head `9e5b747` (no `main` merge folded in).

PR #451, opened 2026-06-03
Scaffold TMT set + implement 96/190 cards

bolav:
## Summary

Scaffolds the Teenage Mutant Ninja Turtles (TMT) set backlog and lands the first
implementation pass — 96 of 190 cards across 99 files (~6.2k insertions, 104 commits).

## What's in

- **Backlog scaffolding:** `cards.md` checklist, `mechanics.md`, and a `TODO.md`
  that tracks per-card progress, partitions cards by mechanic, and catalogs the
  remaining engine gaps (Sneak, Disappear, Alliance, Class, Mutagen, Sagas, Crew tracking, …).
- **96 implemented cards** — each as its own commit, all
  composing existing SDK primitives. No new effect types or executors were added;
  cards that would have required them are deferred behind labeled gaps in `TODO.md`.
- **Gap closures noted in TODO:** Gap K (Sagas) primitive-resolved; Gap J partly
  cleared via `Turtle Blimp`; Gaps MM and NN added for newly surfaced needs.
- **Final cleanup commit:** addresses review findings on the card batch.

## What's not in

- 94 remaining cards blocked on unresolved engine gaps (see `TODO.md` for the
  per-gap breakdown). Those will land after each gap is addressed via `add-feature`.
- No SDK / engine changes — this is content-only.

### DIFF
--- a/backlog/sets/teenage-mutant-ninja-turtles/TODO.md
+++ b/backlog/sets/teenage-mutant-ninja-turtles/TODO.md
@@ -27,8 +27,12 @@ The remaining 94 cards mostly cluster on a handful of unresolved gaps:
 - **Vehicles / Crew** — 1 card left (Gap J — partly cleared; `Turtle Blimp`
   shipped composably, `Turtle Van` still waits on Gap LL "creature that
   crewed this vehicle this turn" tracking)
-- **Sagas** — 2 cards (Gap K)
-- Plus various one-offs from Gaps M and N–LL
+- **Sagas** — 2 cards (Gap K — Saga primitive is wired, but each TMT Saga
+  has its own separate blocker: `The Cloning of Shredder` → Gap MM
+  `addedSubtypes` on `CreateTokenCopyOfTarget`; `The Last Ronin` → Gap AA
+  "Mill X. When you do, …" sub-trigger + Gap NN "attacks alone this turn"
+  trigger condition)
+- Plus various one-offs from Gaps M and N–NN
 
 Gaps **resolved across the runs so far**:
 - **Gap C — Alliance (partly cleared)**: 6 of the 10 Alliance cards (`East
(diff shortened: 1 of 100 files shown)


wingedsheep:
## Review + fixes pushed (`301afff`)

Reviewed the full diff (95 card files + backlog docs) against `main`. Primary lens: SDK elegance. **This is genuinely content-only** — zero new `Effect`/`Condition`/`Trigger`/`StaticAbility` types, every card composes existing primitives, and `:mtg-sets:compileKotlin` is green. Reprint plumbing (`Negate`, `Escape Tunnel`) correctly uses the `PrintingRegistry` pattern — a bare `Printing(...)` row pointing at the canonical def in FDN/MKM, no duplicated `CardDefinition`. `sealedSupported = true` on an `incomplete` set follows existing precedent (arn, eoe, ltr, tdm).

I pushed a commit directly to the branch with the fixes below.

### Blocking — fixed
- **`SplinterRadicalRat.kt:51` — inverted ability.** Oracle is `{1}{U}: Target Ninja **can't be blocked** this turn` (evasion), but it was implemented as `Effects.CantBlock(ninja)`, which makes the Ninja unable to **block** — the opposite effect. Changed to `Effects.GrantKeyword(AbilityFlag.CANT_BE_BLOCKED, ninja)`, matching the established unblockable pattern (Long River Lurker, Crafty Pathmage).

### Important — partially addressed
- **No tests for 95 cards.** The batch shipped with zero `scenarios/`/`*Test` coverage, which is exactly what let the Splinter inversion through — a one-line "declare blockers" scenario catches it. I added `SplinterRadicalRatTest` (rules-engine) that attacks with Splinter, activates the ability on himself, and asserts the projected `CANT_BE_BLOCKED` keyword + a failed block. **Remaining:** the other non-trivial cards (trigger-doublers like Mighty Mutanimals, the exile-return pairs, the reveal-until pipelines, Krang's affinity/draw-to-4) still want scenario coverage. Flagging as a tracked follow-up rather than blocking the scaffold — please confirm you're OK landing the rest of the cards without per-card tests, or I can add a focused batch.

### Minor — fixed
- **`TODO.md:276,280` — wrong CR number.** "you don't lose this mana as steps and phases end" was cited as **CR 106.4b**; that's the announce-mana-on-passing-priority rule. The mana-empties-each-step-and-phase rule the effect overrides is **CR 106.4** ("Each player's mana pool empties at the end of each step and phase, and the player is said to lose this mana"). Verified against the Comprehensive Rules; corrected both citations.

### Note (no action needed)
- `check-card-printing "Negate"` exits non-zero, but that drift is **pre-existing** — canonical lives in FDN while the earliest real print is MOR (unscaffolded), plus missing rows in m10/m12/m14/dtk/rix. None of it is introduced here; the TMT contribution is a correctly-placed `reprint` row. `check-card-printing "Escape Tunnel"` passes clean.

### What changed in `301afff`
- `SplinterRadicalRat.kt` — `CantBlock` → `GrantKeyword(CANT_BE_BLOCKED)` (+ `AbilityFlag` import)
- `SplinterRadicalRatTest.kt` — new regression test (passes: `:rules-engine:test --tests "*SplinterRadicalRatTest"`)
- `TODO.md` — CR 106.4b → 106.4

The other 94 cards match their header Oracle text, use projection-aware filters, and compose existing primitives — merge-ready once the test-coverage follow-up is agreed.

PR #452, opened 2026-06-03
Add Zero Point Ballad + ToughnessAtMostX SDK predicate

bolav:
## Summary
- New EOE card **Zero Point Ballad** ({X}{B} sorcery): destroy all creatures
  with toughness ≤ X, lose X life, and if X ≥ 6 reanimate one of them under
  your control. Composes `Effects.DestroyAll` + `ConditionalEffect` +
  `SelectFromCollectionEffect` + `MoveCollectionEffect` — no card-specific
  executor.
- New SDK predicate `CardPredicate.ToughnessAtMostX` and DSL builder
  `.toughnessAtMostX()`, mirroring the existing `ManaValueAtMostX` shape:
  resolves against `PredicateContext.xValue` at resolution time; trigger /
  layer-projection / cost calculation report `false` (no X context there).
- Legal-action layer plumbed end-to-end for client re-filtering: new
  `xConstrainsToughness` flag on `TargetInfo` / `LegalAction` /
  `LegalActionTargetInfo` and matching TS types, so the client targeting UI
  intersects `validTargets` with `card.toughness ≤ chosenX` after X is picked.
- Refactored `TargetEnumerationUtils` to a generic predicate-tree walker;
  future `*AtMostX` predicates need only a one-line `requirementUses…` helper
  plus a parallel flag.
- Updated `docs/card-sdk-language-reference.md` and added a manual-play
  scenario at `manual-scenarios/cards/z/zero-point-ballad.json`.

## Test plan
- [ ] `:game-server:test --tests "ZeroPointBalladScenarioTest"` — X=2, X<6,
      X=6 reanimation paths all green.
- [ ] `:rules-engine:test --tests "BlueSunsTwilightScenarioTest"` —
      `xConstrainsTargetManaValue` regression remains green.
- [ ] `web-client && npm run typecheck` — clean.
- [ ] `just check-card-printing "Zero Point Ballad"` — canonical lives in EOE.


wingedsheep:
## Review + applied fixups

Reviewed for SDK elegance, correctness, projection use, rules accuracy, and tests. The card behaves correctly and `CardPredicate.ToughnessAtMostX` is a legitimately novel primitive (reads `PredicateContext.xValue` at resolution, mirroring `ManaValueAtMostX`). I've pushed a fixup commit (`87081a9`) rather than just leaving notes.

### What's good (kept as-is)
- The predicate is the right shape, and the four "no X context → `false`" guards (`TriggerMatcher`, `AffectsFilterResolver`, `CostCalculator`, the cast-record branch) are correct and well-commented.
- The card composes existing primitives (`DestroyAll` + `LoseLife` + `ConditionalEffect(X≥6)` + `SelectFromCollection` + `MoveCollection`) — no card-specific executor.
- Tests exercise the real rule corners (toughness gate leaves 3/3 alone at X=2; X<6 no-reanimate; X=6 reanimate under caster's control). All green.
- Oracle text matches Scryfall exactly; `just check-card-printing` is green (canonical in EOE, the earliest real printing); web-client typecheck clean; the `BlueSunsTwilight` MV-X regression stays green.

### Changed — dropped the speculative target plumbing (Important)
Zero Point Ballad is a **non-targeted** `DestroyAll`, so the `xConstrainsTargetToughness` / `xConstrainsToughness` legal-action + client re-filter pipeline had **no consumer**: the flag was `false` for every card in the repo, and there was no test that could exercise it. That's untested dead surface — exactly the "parameterize for the next card, not this one" case the project guards against.

Reverted to `main`: `LegalAction.kt`, `LegalActionPresentation.kt`, `LegalActionEnricher.kt`, `CastSpellEnumerator.kt` (×7 call sites), `CastFromZoneEnumerator.kt` (×5), `TargetEnumerationUtils.kt` (the generic walker only existed to serve the toughness path), `messages.ts`, and `pipelinePhases.ts`. This should land in the PR that first ships a *"destroy target creature with toughness X or less"* card, where the client re-filter can actually be tested.

### Changed — two minor correctness tightenings
- **Reanimation filter → `.nontoken()`** (`ZeroPointBallad.kt`). Oracle says "a creature **card**"; tokens cease to exist on death. Harmless in practice but now precise.
- **`PredicateEvaluator.ToughnessAtMostX`: unbound X → match nothing, not everything.** With the targeted enumeration path removed, this predicate is only evaluated at resolution where X is bound. The old permissive `null → true` was a footgun for a non-targeted `DestroyAll` (a missing X would silently wipe the board); it now returns `false`.

### Verification
- `:game-server:test --tests "ZeroPointBalladScenarioTest"` — all 3 green after changes (X=6 confirms `.nontoken()` still selects correctly).
- `:rules-engine:test --tests "BlueSunsTwilightScenarioTest"` — green.
- `web-client` typecheck — clean.

PR #456, opened 2026-06-03
Add YouChoseOtherCreatureAsRingBearer condition + fix CR 701.54a cont…

bolav:
…rol change

Closes LTR Gap 3: ring-tempt triggers like Aragorn/Faramir/Gandalf/Galadriel gate their payoff on "if you chose a creature other than this as your Ring-bearer." The new `Conditions.YouChoseOtherCreatureAsRingBearer` is true when the controller currently has a Ring-bearer designated and that bearer isn't the source — so it's false both when the source was chosen and when no creature was available to choose. Lands Faramir, Field Commander as the demo card; the other three Gap-3 cards remain blocked on Gaps 4/7 for their second abilities.

Also fixes CR 701.54a: "until another player gains control of it." Today the designation hangs around as `RingBearerComponent` on the entity, suspended while another player controls it. When control reverts (Threaten end-of-turn) the projection re-activates the designation, contradicting the rule. Every control-change executor now calls `clearRingBearerOnControlChange` to strip the component eagerly the moment the new controller differs from the bearer's owner, so the revert keeps the bear bearer-less.

Drive-by: renames the Ring rules from CR 701.52 to 701.54 across the codebase. The CR shuffled these subrules (a–e map 1:1) between LTR's 2023 release and the 2026-04-17 rules; verified against the local MagicCompRules_20260417.txt.

### DIFF
--- a/backlog/sets/lord-of-the-rings/TODO.md
+++ b/backlog/sets/lord-of-the-rings/TODO.md
@@ -94,14 +94,15 @@ in its own PR, then attach all of its cards.
 - **Stalwarts of Osgiliath** — same trigger (ETB Ring tempts half is already composable).
 
 ### Gap 3 — "the Ring tempted you and you chose a creature other than this" condition
-**Engine change:** a condition usable with `Triggers.RingTemptsYou` meaning "you chose a
-creature other than {source} as your Ring-bearer" (CR 701.52d).
-- **Aragorn, Company Leader** — "…if you chose a creature other than Aragorn as your
-  Ring-bearer…" (also Gap 7 keyword counters).
-- **Faramir, Field Commander** — "…if you chose a creature other than Faramir…, create a token."
-- **Gandalf, Friend of the Shire** — "…if you chose a creature other than Gandalf…, draw a card."
-  (also Gap 4 cast-as-flash).
-- **Galadriel of Lothlórien** — chapter-one half.
+**Status:** LANDED as `Conditions.YouChoseOtherCreatureAsRingBearer` (PR
+`ltr-gap3-ring-bearer-other-than`). Same PR fixed CR 701.54a — control-change executors now strip
+`RingBearerComponent` eagerly via `clearRingBearerOnControlChange`, so a temporary Threaten-style
+steal no longer silently restores the designation when control reverts at end of turn.
+- **Faramir, Field Commander** — ✅ implemented.
+- **Aragorn, Company Leader** — still blocked on Gap 7 (keyword counters).
+- **Gandalf, Friend of the Shire** — still blocked on Gap 4 (cast-as-flash for sorcery spells).
+- **Galadriel of Lothlórien** — Ring-tempt half composes via this condition; second ability still
+  needs a "Whenever you scry" trigger + "reveal top; if land, put onto battlefield tapped".
 
 ### Gap 4 — "cast [filter] spells as though they had flash" permission
 **Engine change:** a static/one-shot permission granting flash-timing to a filtered set of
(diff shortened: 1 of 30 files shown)


wingedsheep:
## Review — SDK elegance + correctness

**Verdict:** Behavior is correct and the SDK shape is right. `YouChoseOtherCreatureAsRingBearer` is a genuinely novel, parameter-free state read (it inspects the Ring-bearer designation + identity-vs-source, which no existing primitive expresses) and it's built for the whole Aragorn/Faramir/Gandalf/Galadriel cycle rather than just Faramir. The CR 701.54a control-change fix is well-targeted. All rule citations, the card data, and the printing placement check out, and the full Ring + Faramir test suites pass locally.

### What's good
- **Right-sized condition.** No constants baked in; reusable across the four Gap-3 cards. Reads the same `controllerId`/projected-controller pair as `SourceIsRingBearer`, so a control-suspended designation correctly fails to count as "a bearer." Name matches semantics, and the false-when-no-creature case is documented.
- **Complete control-change coverage.** All five executors that emit `ControlChangedEvent` (`GainControl`, `GainControlByActivePlayer`, `GainControlByMost`, `GiveControlToTargetPlayer`, `ExchangeControl`) call `clearRingBearerOnControlChange`; I grepped for stragglers and found none. The shared helper only strips when control actually moves to a *different* player, so same-owner re-grants are left alone.
- **Rule numbers verified.** Downloaded `MagicCompRules 20260417.txt` and confirmed the Ring rules are now §701.54 (was 701.52 in the LTR-era CR; §701.52 is now "Roll to Visit Your Attractions"). 701.54a–e map 1:1 as the PR claims, and 701.54a's text ("…until another creature becomes your Ring-bearer or another player gains control of it") is exactly the basis for the eager strip. The blanket 701.52→701.54 rename across the codebase is correct.
- **Card data matches Scryfall** (`{3}{W}`, Legendary Creature — Human Soldier, 3/3, ltr #14 uncommon, color identity W, oracle text, artist, flavor, image URI).
- **Tests exercise the rule corners.** Both branches of the intervening-if (chose-other vs chose-self vs only-Faramir), and a dedicated CR 701.54a test that steals the bearer with a Threaten-style spell and asserts the designation stays gone *after control reverts* — the exact bug being fixed.
- **Docs updated in the same change** (`card-sdk-language-reference.md`, backlog TODO/cards count). Good discipline.

### Issues

**Minor (fixed in this push — commit `6e0c6e10c`):** `RingBearerComponent`'s kdoc still claimed losing control ends the designation "without the component having to be eagerly removed" — which this PR directly reverses. Updated it to describe the dual mechanism: projection *suspends* on transient/continuous control loss; the executors *strip* the component on an actual transfer to another player (701.54a), permanently.

**Minor (not fixed):** "Find a player's Ring-bearer" is now an ad-hoc battlefield scan in two places — `ConditionEvaluator` (new) and `ClientStateTransformer` (existing) — with slightly different control checks (the new one filters on projected controller, the old one doesn't). Not worth extracting a helper for two near-uses yet, but worth keeping in mind if a third appears.

**Minor (not fixed):** The eager `RingBearerComponent` removal is a state change with no dedicated `GameEvent`. This is consistent with how the tempt continuation already moves the designation (no event), and nothing triggers off "designation lost," so it's fine — flagging only because it brushes the "events, not silent mutations" principle.

### Note on this review branch
Reviewed in a worktree with `origin/main` merged in; the merge was clean (no conflicts). The fix above was cherry-picked onto the PR head and pushed as a single commit, so **no merge commit was introduced into this PR** — you can keep your branch as-is.

### Recommendation
Ship it. The doc fix is already pushed; nothing else is blocking. The remaining Gap-3 cards (Aragorn/Gandalf/Galadriel) are correctly left for Gaps 4/7 as the PR body states.

PR #458, opened 2026-06-03
Add Bioengineered Future with lands-entered-this-turn tracker

bolav:
Bioengineered Future ({1}{G}{G} green enchantment, EOE #172) needs to know "how many lands entered the battlefield under your control this turn" so its static ability can grant your other creatures an extra +1/+1 counter per land at ETB time. The existing `LANDS_PLAYED` tracker only counts from-hand land drops — not Lander-token search, Cultivate-style "put a land onto the battlefield" effects, or gift-a-land effects — so this commit adds a parallel `LANDS_ENTERED_UNDER_CONTROL` turn tracker that bumps on every land ETB regardless of source.

SDK: new `TurnTracker.LANDS_ENTERED_UNDER_CONTROL` enum value with oracle description and a `DynamicAmounts.landsEnteredUnderControlThisTurn(player)` facade. Engine: new per-player `LandsEnteredUnderControlThisTurnComponent`, incremented by `PermanentEntryTracker.record` whenever the entering permanent's projected types include `LAND` (covers both the standard zone-transition path and the ad-hoc `BattlefieldEntry.place` path), resolved by `DynamicAmountEvaluator`, cleared by `CleanupPhaseManager` at end of turn, and registered with the serialization module. The card itself composes the existing `EntersWithDynamicCounters(otherOnly = true)` replacement pattern (Grumgully, Dragonstorm Globe) with the new dynamic count.

Scenario tests cover the happy path, the multi-land scaling case, the pre-BF land ordering from the WotC ruling, per-player isolation (opponent's lands don't feed our count), and end-of-turn cleanup.


wingedsheep:
## Review — Bioengineered Future + `LANDS_ENTERED_UNDER_CONTROL` tracker

**Verdict:** Behavior is correct and the SDK shape is right. This is a genuinely novel tracker (reads state no existing primitive can — `LANDS_PLAYED` only counts from-hand drops), parameterized by `Player`, and the card itself composes the existing `EntersWithDynamicCounters(otherOnly=true)` + `DynamicAmount.TurnTracking` rather than growing a card-specific type. No over-specialization. Verified against Scryfall: oracle text, `{1}{G}{G}` Enchantment, EOE #172, and both 2025-07-25 rulings all match verbatim.

### What's good
- **Right SDK altitude.** New `TurnTracker` enum value + per-player component + facade — exactly the sanctioned "new tracker" shape, not a bespoke effect. The card is pure composition.
- **Idempotency reasoning is correct and documented.** The author noticed the land *count* tracker (unlike the set-based type tracker) must not be recorded twice, updated the `PermanentEntryTracker` doc comment to call this out, and confirmed both sanctioned paths (`ZoneTransitionService`, `BattlefieldEntry.place`) call `record` exactly once. Both paths read **projected** types, per the projection rule.
- **Full cross-layer wiring.** Cleanup at end of turn, serialization registration (`SerializationPolymorphicRegistrationTest` green), and `DynamicAmountEvaluator` resolution are all present. `Player.You` resolves against the BF source's controller in `EntersWithCountersHelper`, so the count reads the right player.
- `just check-card-printing "Bioengineered Future"` is clean (canonical in EOE, `peoe` promos correctly skipped).

### Issues

**Important (addressed in this push)** — *Rule-corner that drove the change was untested.* Every original scenario entered lands via `PlayLand`, which routes through `BattlefieldEntry.place` and would *also* bump the old `LANDS_PLAYED` — so none of them exercised the actual reason the new tracker exists: counting a land that enters **without** a land drop (the other code path, `ZoneTransitionService`). I added a case that resolves **Rampant Growth** (a real card already in the catalog — no fabricated test card) to fetch a basic land onto the battlefield, then asserts `landDropsUsed == 0` while the tracker reads `1` and a subsequent creature still gets its `+1/+1`. This is the divergence the whole feature is about. Pushed as `f26069c8`; all 6 scenarios green.

### Recommendation
Mergeable as-is with the added test. No code changes requested — the implementation is sound.

---
*Reviewed on a worktree with `origin/main` merged in (clean merge, no conflicts). The added test commit was cherry-picked onto the PR head and pushed to `bolav/bioengineered-future`, so the PR history stays free of a merge-main commit.*

PR #462, opened 2026-06-03
Add Old Man of the Sea to Arabian Nights; implement CR 506.4

bolav:
Adds the {1}{U}{U} 2/3 Djinn with its "for as long as this creature remains tapped and that creature's power remains less than or equal to this creature's power" steal duration, plus the engine work needed to make the second half of that duration actually fire.

SDK additions:
- `Duration.WhileSourceTappedAndAffectedPowerAtMostSource` — variant of WhileSourceTapped that also gates per-affected-entity on power ≤ source's power.
- `CardPredicate.PowerAtMostEntity(reference)` + `.powerAtMostEntity(ref)` DSL on ObjectFilter / TargetFilter — inverse of PowerGreaterThanEntity, used by the target requirement.

Engine wiring:
- StateProjector collects the duration's floating effects gated only on source-tapped; a post-Layer-7 fix-up (`applyAffectedPowerAtMostSourceGate`) compares each affected entity's final projected power to the source's final projected power and reverts the controller for any entity that's stronger. The post-Layer-7 timing is what makes the gate see every pump source — printed power, +1/+1 / -1/-1 counters, Layer-7 floating pumps (Giant Growth, Aggressive Urge), and lord-style anthems.
- CR 506.4 — "A permanent is removed from combat if its controller changes." New SBA `ControlChangedRemovesFromCombatCheck` walks every `AttackingComponent` / `BlockingComponent` entity and removes it from combat if its projected controller no longer matches the player whose role it's filling (attackers must be the active player; blockers must not). `RemoveFromCombatExecutor`'s cleanup is extracted into `CombatRemovalHelper` so the SBA and the explicit `RemoveFromCombatEffect` share the same removal logic.
- CleanupPhaseManager + SacrificeAndPayContinuationResumer drop the new duration's floating effect at untap-step cleanup the same way they handle WhileSourceTapped.
- PredicateEvaluator implements PowerAtMostEntity; the other entity-relative-predicate stubs (TriggerMatcher, AffectsFilterResolver, CostCalculator) mirror PowerGreaterThanEntity.

Eight scenario tests cover the activation flow, the predicate's target-validity gate, the untap-returns-control / stays-tapped fork, the per-entity power gate via +1/+1 counters AND Layer-7 floating pumps, and the CR 506.4 mid-attack removal — a stolen attacker pumped past Old Man's power loses control, leaves combat, and deals zero damage.

### DIFF
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 78 cards
 **Release Date:** December 17, 1993
-**Implemented:** 56 / 78
+**Implemented:** 57 / 78
 | Color | Count |
 |-------|-------|
 | White | 11 |
@@ -32,7 +32,7 @@
 - [x] Giant Tortoise
 - [x] Island Fish Jasconius
 - [x] Merchant Ship
-- [ ] Old Man of the Sea
+- [x] Old Man of the Sea
 - [ ] Serendib Djinn
 - [x] Serendib Efreet
 - [x] Sindbad
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -450,8 +450,10 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
   `Duration.WhileSourceTapped` (Callous Oppressor) or
   `Duration.WhileSourceTappedAndAffectedPowerAtMostSource` (Old Man of the Sea) for the classic
   "for as long as this creature remains tapped [and the stolen creature's power stays ≤ source's
-  power]" steal pattern; both durations are gated per-frame by `StateProjector` and cleaned up
-  at the next untap step.
+  power]" steal pattern. `StateProjector` gates these per-frame for the instantaneous view; the
+  one-way half of CR 611.2b ("for as long as" durations don't restart) is enforced by the
+  `EndedDurationExpiryCheck` state-based action, which physically removes the effect the moment
+  the condition fails — so a pump that wears off (or a re-tap) never re-grabs the creature.
 - `ExchangeControlEffect(target1, target2)` — swap control of two permanents.
 - `GainControlByMostEffect(metric, target?)` — the player with strictly the most of a `PlayerRankMetric` takes it (tie = no change). Metrics: `PlayerRankMetric.LifeTotal` (Ghazbán Ogre), `PlayerRankMetric.CreaturesOfSubtype(subtype)` (Thoughtbound Primoc). Facades: `Effects.GainControlByMostLife()`, `Effects.GainControlByMostOfSubtype(subtype)`.
 - `GiftGivenEffect(target)` — "gift" temporary control.
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -446,7 +446,12 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 
 ### Control & combat
 
-- `GainControlEffect(target, duration)` — gain control until end of turn (default).
+- `GainControlEffect(target, duration)` — gain control until end of turn (default). Pair with
+  `Duration.WhileSourceTapped` (Callous Oppressor) or
+  `Duration.WhileSourceTappedAndAffectedPowerAtMostSource` (Old Man of the Sea) for the classic
+  "for as long as this creature remains tapped [and the stolen creature's power stays ≤ source's
+  power]" steal pattern; both durations are gated per-frame by `StateProjector` and cleaned up
+  at the next untap step.
 - `ExchangeControlEffect(target1, target2)` — swap control of two permanents.
 - `GainControlByMostEffect(metric, target?)` — the player with strictly the most of a `PlayerRankMetric` takes it (tie = no change). Metrics: `PlayerRankMetric.LifeTotal` (Ghazbán Ogre), `PlayerRankMetric.CreaturesOfSubtype(subtype)` (Thoughtbound Primoc). Facades: `Effects.GainControlByMostLife()`, `Effects.GainControlByMostOfSubtype(subtype)`.
 - `GiftGivenEffect(target)` — "gift" temporary control.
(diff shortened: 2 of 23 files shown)


wingedsheep:
## Review — Old Man of the Sea + CR 506.4

**Verdict:** Behavior is right and the SDK shape is well-judged. The new `PowerAtMostEntity` predicate is a clean, reusable inverse of `PowerGreaterThanEntity`; the CR 506.4 SBA is a genuinely general engine fix (not card-specific), and `CombatRemovalHelper` is the right extraction. All 8 scenario tests pass, the full `:rules-engine` suite is green (the new global SBA regresses nothing), `:mtg-sdk` is green, and `just check-card-printing` reports the canonical correctly in `arn`. Verified: CR 506.4 text matches, the oracle text matches Scryfall, and `PowerAtMostEntity` is a faithful mirror of `PowerGreaterThanEntity` in `PredicateEvaluator`.

### What's good (keep it)
- **`CardPredicate.PowerAtMostEntity`** is parameterized by `EntityReference`, not hardcoded to Old Man — any "power ≤ referenced entity" card composes it. The stub additions in `TriggerMatcher` / `AffectsFilterResolver` / `CostCalculator` correctly mirror the sibling.
- **CR 506.4 modeled as a real SBA**, not a card hack. It fixes controller-change-during-combat for *every* control effect, and reuses `CombatRemovalHelper` so `BlockedComponent`/`BlockingComponent` back-references stay consistent. The two-player approximation is documented.
- **Projection discipline.** The post-Layer-7 gate reads the in-progress `projectedValues` map (not `state.projectedState`) to avoid the lazy-projection recursion, and the SBA reads `projected.getController` — so the gate's revert flows correctly into the SBA. The comment explaining *why* the power half can't run during Layer-2 collection is exactly the kind of non-obvious "why" worth writing.
- **Docs updated** for both the Duration and the predicate, per the SDK-change rule.

### Issues

**Minor**

1. **Reversible per-frame control gate vs. CR 611.2i "for as long as."** `applyAffectedPowerAtMostSourceGate` hides control while `affectedPower > sourcePower` and *restores* it once power drops back (the floating effect persists until untap-step cleanup). Strictly, a "for as long as" duration is one-way: once the condition fails the effect ends and does not restart, so a temporary pump (e.g. opponent's Giant Growth) that wears off while Old Man stays tapped should **not** re-steal the creature — but here it would. **However**, this is consistent with how the engine already models `WhileSourceTapped` (StateProjector.kt:456 also `continue`s when untapped and re-includes when tapped again), so it's a pre-existing engine-wide simplification, not a regression this PR introduces. Acceptable for current scope; worth a one-line note in the Duration KDoc and, someday, an SBA that physically drops the floating effect on first breach. The two power tests only cover the exceed→revert direction, never drop-back→stays-reverted.

2. **`ControlChangedRemovesFromCombatCheck` inline FQN** — `mutableListOf<com.wingedsheep.sdk.model.EntityId>()` used a fully-qualified name where the sibling `CombatRemovalHelper` imports it. **Fixed in this branch** (pushed as a follow-up commit).

3. **CR 506.4 test exercises the SBA via a manual `StateBasedActionChecker.checkAndApply` call** rather than proving it fires automatically in the priority loop. The registration in `CreatureSbaModule` means it runs for real, and the final life-total assertion after `passPriorityUntil(POSTCOMBAT_MAIN)` covers the damage outcome — but the automatic-firing-during-combat path isn't directly asserted. Author's discretion.

### SDK-elegance note (not blocking)
`Duration.WhileSourceTappedAndAffectedPowerAtMostSource` fuses two conditions and hardcodes the power comparison in `applyAffectedPowerAtMostSourceGate`, duplicating the comparison `PowerAtMostEntity` already encodes. The "parameterize for the next card" instinct would suggest `Duration.WhileSourceTappedAndAffected(predicate: CardPredicate)` reusing the predicate machinery. I looked at whether that's feasible and concluded **it isn't cheap here**: the gate runs mid-projection on the in-progress `projectedValues` map, while `PredicateEvaluator` reads `state.projectedState` — which would recurse. A generic predicate-carrying duration would need a projection-map-aware evaluator, a much larger lift. Given (a) Durations aren't composable in the SDK and (b) the recursion constraint, the fused variant is a defensible pragmatic choice, and its name honestly describes its specificity. No change requested.

### Recommendation
Merge-ready. Only the follow-up cleanup commit was applied. Consider (optional) a KDoc line on the Duration acknowledging the reversible-gate / CR 611.2i gap and a drop-back test to pin current behavior, but neither blocks.

wingedsheep:
## Follow-up: fixed the reversible-gate limitation (now one-way per CR 611.2b)

The earlier review flagged that the per-frame control gate was reversible — a temporary pump that pushed the stolen creature past Old Man's power and then wore off would *re-steal* the creature, contrary to CR 611.2b ("a 'for as long as' duration … doesn't last forever" and doesn't restart). I've now fixed this at the engine level (commit `c6aad2b7`), generalized across the whole "for as long as" family rather than patched for Old Man alone.

**What changed**
- New SBA **`EndedDurationExpiryCheck`** (`SbaOrder.DURATION_EXPIRY = 240`) physically removes a floating effect from `GameState.floatingEffects` the moment its "for as long as" condition fails, so no later projection can bring it back. It covers `WhileSourceTapped`, `WhileSourceTappedAndAffectedPowerAtMostSource`, `WhileSourceOnBattlefield`, and `WhileControlledByController`.
- It reads the **cached `state.projectedState`** (it's an SBA, not inside the projector — so no recursion, which is what made the in-projector approach awkward), shrinks the affected-entity set per condition, emits `ControlChangedEvent` on control reversions, and runs an internal fixpoint for effect-on-effect cascades.
- Ordered **before** `ControlChangedRemovesFromCombatCheck` (250) and the lethal/zero-toughness checks (300/400) so they see the reverted controller / ended toughness boosts in the same SBA pass.

**Design choice — additive, not a rewrite.** The existing `StateProjector` gates stay (they give the correct *instantaneous* view); the SBA supplies the *one-way* half. The untap-step cleanup is untouched (SBAs don't run during the untap step). This keeps the existing projection-only tests valid while making the end permanent.

**Tests / verification**
- New scenario: `CR 611.2b — control does NOT return after a temporary pump wears off (one-way latch)` — steal, pump past Old Man's power, run SBAs (control latches away + the `WhileSourceTapped…` floating effect is asserted *gone*, not merely hidden), remove the pump, re-run SBAs → control stays with the owner.
- Full **`:rules-engine`**, **`:game-server`**, and **`:mtg-sdk`** suites green — including the other duration users (`SuspendMechanicTest`, `EvergloveCourierTest`, `CallousOppressorTest`), so the global one-way behavior regresses nothing.
- KDocs (Duration, StateProjector, the card) and the SDK reference updated; CR 611.2b verified against the official rules text.

PR #463, opened 2026-06-04
Add Weftwalking with composable free-cast SDK primitive

bolav:
## Summary
- Adds Weftwalking (Edge of Eternities #86, {4}{U}{U} mythic enchantment): ETB-if-cast shuffles hand + graveyard into library and draws 7; static lets each player free-cast their first spell each of their own
turns.
- Introduces `MayCastWithoutPayingManaCost(controllerOnly, firstSpellOfTurnOnly)` — a composable battlefield "may cast without paying its mana cost" SDK type. Future cards in this family compose by flipping
flags instead of growing a sibling type.
- Wires the free-cast through a dedicated `CastWithoutPayingManaCost` LegalAction routing via `CastSpell.useWithoutPayingManaCost`, emitted **alongside** Jodah / flashback / harmonize / warp / evoke / impending
  / `selfAlternativeCost`. The player picks which alternative cost to apply (CR 118.9a — alt costs don't combine, and the choice is the player's, not handler priority). Mutually exclusive with
`useAlternativeCost`, validated at cast time.
- Cost calculator gate (`hasFreeCastPermission`) scans the full battlefield so an opponent's source still grants the active player their free cast unless `controllerOnly = true`. `firstSpellOfTurnOnly` enforces
  active-player + zero-spells-cast-this-turn.
- Mandatory additional costs (Embrace Oblivion sacrifice, etc.) still apply; kicker / blight / behold / runtime tax are skipped; X = 0 per CR 107.3b — matching the existing `PlayWithoutPayingCostComponent` flow
  used by Cascade / Omniscience.
- Updates the SDK reference doc, EOE backlog (254 → 255 implemented), and two manual scenarios.

## Test plan
- [x] `:rules-engine:test --tests WeftwalkingScenarioTest` — 11 scenarios green (shuffle+draw on cast, no shuffle on reanimate, free-cast happy path, gate close after first spell, opponent's-turn rejection,
opponent-controls-source, Jodah+Weftwalking dual LegalActions, mutex rejection, Embrace Oblivion mandatory sacrifice both with and without).
- [x] `:rules-engine:test`, `:mtg-sdk:test`, `:mtg-sets:test` — no regressions.
- [ ] Manual: load `manual-scenarios/cards/w/weftwalking-etb.json` and `weftwalking-free-cast.json` in the dev scenario UI; confirm both cast paths render and the cost-{0} button shows up alongside Jodah's alt
cost when both are legal.

### DIFF
--- a/backlog/sets/edge-of-eternities/cards.md
+++ b/backlog/sets/edge-of-eternities/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 261 booster cards (excluding basic lands)
 **Release Date:** August 1, 2025
-**Implemented:** 254 / 261
+**Implemented:** 255 / 261
 ---
 
 - [x] Adagia, Windswept Bastion
@@ -260,7 +260,7 @@
 - [x] Wedgelight Rammer
 - [x] Weftblade Enhancer
 - [x] Weftstalker Ardent
-- [ ] Weftwalking
+- [x] Weftwalking
 - [x] Wurmwall Sweeper
 - [x] Xu-Ifit, Osteoharmonist
 - [x] Zealous Display
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1527,18 +1527,23 @@ riders, matching how the engine already treats e.g. City of Brass's damage durin
   Yawgmoth's Agenda (`MayCastFromGraveyard(Nonland)`); `lifeCost = 1, duringYourTurnOnly = true` for
   Festival of Embers. Pair with `MayPlayLandsFromGraveyard` for "play lands and cast spells from
   your graveyard". Lands are *played*, not cast, so they need the lands permission separately.
-- `MayCastFirstSpellOfTurnWithoutPayingMana(controllerOnly = false)` — the first spell each
-  player casts during each of their **own** turns may be cast without paying its mana cost
-  (Weftwalking). Gated at cast time by `CostCalculator.hasFirstSpellOfTurnFreeCast`: requires the
-  active player to be the caster, requires the caster to have cast zero spells this turn, and
-  requires a battlefield source with this static (the source's controller may be anyone unless
-  `controllerOnly = true`). Surfaces as an alt-cost entry (`{0}`) in
-  `EnumerationContext.alternativeCastingCosts`; `CastSpellHandler.isFirstSpellOfTurnFreeCastChosen`
-  resolves the priority chain so the static only wins when no keyword alt (flashback / harmonize /
-  warp / evoke / impending), no `selfAlternativeCost`, and no Jodah-style `GrantAlternativeCastingCost`
-  applies — when it does win the cast is treated as `playForFree` (cost zeroed, kicker / blight /
-  behold / runtime tax skipped, matching the existing `PlayWithoutPayingCostComponent` flow used
-  by Cascade and Omniscience).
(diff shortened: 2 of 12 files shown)


wingedsheep:
## Review — Weftwalking + composable free-cast SDK primitive

**Verdict:** Behavior is correct and the SDK shape is right. `MayCastWithoutPayingManaCost` is a genuinely new battlefield-permission static (no existing primitive reads "may cast a spell without paying its mana cost" at cast-legality time — the existing free-cast paths stamp `PlayWithoutPayingCostComponent` on a *specific* card, which can't express "the first spell each player casts"). It's parameterized for the next card via two orthogonal gates rather than hardcoding Weftwalking. All 11 scenarios pass locally; the three cited CR numbers all check out against the 2026-04-17 Comprehensive Rules.

I pushed two small fixups to the branch (commit `7678abc`) — details under Minor below.

### What's good
- **Right SDK altitude.** Two boolean gates (`controllerOnly`, `firstSpellOfTurnOnly`) instead of a Weftwalking-specific type, with documented compositions for the obvious next cards. This is exactly the composition-over-monolith shape the repo wants.
- **CR citations are accurate** (verified against the official rules text): 118.9 ("without paying its mana cost" *is* an alternative cost), 118.9a (only one alternative cost per cast; the choice is the player's), 107.3b (X = 0 when cast without paying mana cost). The "emitted alongside, player picks" design correctly implements 118.9a rather than letting handler priority decide.
- **Projection-correct.** `hasFreeCastPermission` reads the controller from `state.projectedState.getController(...)` for the `controllerOnly` gate, and scans the whole battlefield so an opponent's source still grants the active player their free cast — matching "each player … during each of their turns."
- **Excellent test coverage of the rule-corners,** not just the happy path: intervening-if `WasCast` (reanimation does *not* shuffle/draw), gate-close after the first spell, opponent's-turn rejection, opponent-controls-the-source, Jodah + Weftwalking both surfacing as distinct legal actions, the alt-cost mutex, and Embrace Oblivion's mandatory additional sacrifice enforced both with and without nomination. The cited rules are each exercised.
- **Clean cross-layer wiring.** New `useWithoutPayingManaCost` flag → dedicated `CastWithoutPayingManaCost` legal action with `action.type == "CastSpell"`, so the dumb-terminal client picks it up through the existing multi-cast-options branch with no client change required. Mutual exclusion with `useAlternativeCost` validated in `CastSpellHandler`.

### Issues

**Minor (fixed in `7678abc`):**

1. `SpellStaticAbilities.kt` — the new `MayCastWithoutPayingManaCost` KDoc + `data class` were inserted *between* `PlayersCantCastSpells` and its existing KDoc block, orphaning the `@property affected/spellFilter/condition` doc and leaving `PlayersCantCastSpells` undocumented. Moved the new block above that doc block so each class keeps its own KDoc.

2. `CastSpellEnumerator.kt` — the free-cast legal action used a bare `"Cast {name}"` description, identical to the normal-cast button; every sibling alt-cost variant distinguishes itself in the label (`"Cast X (WUBRG)"`, `"Evoke X (…)"`, `"Cast X (Behold)"`). In the multi-button menu the free cast was distinguishable only by mana cost. Added a `(Free)` suffix at all three branches and updated the test assertion + name.

### Recommendation
Merge after the fixups. One thing worth a glance (left as-is): `firstSpellOfTurnOnly` also bakes in the active-player requirement, so it really means "first spell of *your own* turn." That's correct for Weftwalking and documented, but a hypothetical "first spell you cast each turn, regardless of whose turn" couldn't be expressed without splitting the gate. Fine to defer until such a card exists.

PR #466, opened 2026-06-04
Fix TMT bugs: Uneasy Alliance attached-creature filter, token art

bolav:
- Uneasy Alliance's CantAttack/CantBlock static abilities defaulted to GroupFilter.source(), which applied the restrictions to the Aura itself instead of the enchanted creature. Switch both to GroupFilter.attachedCreature() (Pacifism-style), so the enchanted creature actually can't attack or block. Add a scenario test covering both the static restrictions and the {5}, Sacrifice activated ability that exiles the enchanted creature and creates a 1/1 black Ninja token.
- Sally Pride, Lioness Leader and Slash, Reptile Rampager: stamp the shared TMT 2/2 red Mutant token art onto their CreateTokenEffect so the generated tokens render with the correct image.

### DIFF
--- a/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/tmt/cards/SallyPrideLionessLeader.kt
+++ b/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/tmt/cards/SallyPrideLionessLeader.kt
@@ -52,6 +52,7 @@ val SallyPrideLionessLeader = card("Sally Pride, Lioness Leader") {
                 toughness = 2,
                 colors = setOf(Color.RED),
                 creatureTypes = setOf("Mutant"),
+                imageUri = "https://cards.scryfall.io/normal/front/5/1/51e33613-7a24-461c-8d9f-12680af4b92a.jpg?1771590526"
             )
         )
     }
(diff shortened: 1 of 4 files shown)


wingedsheep:
## Review — TMT bug fixes (Uneasy Alliance filter + token art)

**Verdict:** Behavior is correct and the SDK shape is right. Both fixes compose existing primitives — no new types — and the change is well-tested. Approve. ✅

### What's good
- **Uneasy Alliance fix is the canonical pattern.** `CantAttack`/`CantBlock` already take a `GroupFilter`, and switching the default `GroupFilter.source()` → `GroupFilter.attachedCreature()` is exactly how `Pacifism` (ons) and `ManiacalRage` (exo) model "enchanted creature can't attack/block". No SDK growth, just using the primitive correctly. The original default genuinely was a bug — it pinned the restriction on the Aura permanent itself, which has no attack/block step.
- **Token art is pure data.** `CreateTokenEffect.imageUri` is an existing optional field; both cards just populate it. I cross-checked both Scryfall IDs:
  - `51e33613-…` → *Mutant*, 2/2 red (ttmt) — matches Sally Pride & Slash's "2/2 red Mutant" token. ✓
  - `a7b76498-…` → *Ninja*, 1/1 black (ttmt) — matches Uneasy Alliance's Ninja token. ✓
- **Test exercises the actual rule corner.** The scenario covers both the static restrictions (via `projectedState.cantAttack/cantBlock` — the projection source of truth) and the `{5}, Sacrifice` ability's exile + 1/1 black Ninja creation. The regression docstring explains *why* the filter matters. Both tests pass locally (`:rules-engine:test`).

### Issues
- **Minor (applied):** the test used inline fully-qualified `CardComponent`/`TokenComponent` references; every sibling scenario test imports them. I pushed a one-commit cleanup (`21bb902`) converting to imports — re-ran the test, still green. Nothing else to change.

### Recommendation
Merge as-is. Note: I reviewed with `origin/main` merged in (clean, no conflicts), so this is post-merge reality.

*Reviewed via the `review-changes` skill.*

PR #468, opened 2026-06-04
Fix summoning sickness for Vehicles and animated permanents

bolav:
## Summary

A Vehicle cast and crewed on the same turn was attacking the turn it entered
— a violation of CR 302.6 / 508.1a, which key off control-continuity of the
*permanent*, not whether it was a creature the whole turn. The same gap
affected animated lands and any "becomes a creature this turn" path.

## Fix

- `ZoneTransitionService` and `StackResolver` now attach `SummoningSicknessComponent`
  to every entering permanent, not just those whose printed type line is
  Creature. Downstream gates already short-circuit for lands and non-creature
  artifacts, so the marker is a no-op until the permanent becomes a creature.

- Downstream `{T}`-cost gates in `ActivateAbilityHandler`, `CostHandler`, and
  `CostEnumerationUtils` were reading `cardComponent.typeLine.isCreature`
  (base type) and `baseKeywords.contains(HASTE)` — both wrong for Vehicles
  post-crew and for creatures granted haste by a continuous effect. Switched
  to `state.projectedState.isCreature(...)` / `hasKeyword(..., HASTE)`,
  matching the CLAUDE.md invariant and the existing `ManaSolver` reads.

- `ObservationBuilder.summoningSick` now gates on `projected.isCreature(...)`
  so the gym observation no longer mislabels freshly-played Mountains and
  Equipment as sickness-bound.

## Tests

- New `VehicleEntersThisTurnAttackTest`: casts Turtle Blimp from hand, crews
  with Coral Eel, asserts attack is refused on the entry turn.
- Full `:rules-engine`, `:mtg-sdk`, `:gym`, and `:game-server` suites green.

### DIFF
--- a/gym/src/main/kotlin/com/wingedsheep/gym/contract/ObservationBuilder.kt
+++ b/gym/src/main/kotlin/com/wingedsheep/gym/contract/ObservationBuilder.kt
@@ -254,7 +254,14 @@ class ObservationBuilder(
             power = if (onBattlefield) projected.getPower(entityId) else null,
             toughness = if (onBattlefield) projected.getToughness(entityId) else null,
             tapped = onBattlefield && container.get<TappedComponent>() != null,
-            summoningSick = onBattlefield && container.get<SummoningSicknessComponent>() != null,
+            // Only creatures meaningfully suffer summoning sickness — the engine attaches the
+            // marker to every entering permanent so Vehicles / animated lands inherit the
+            // restriction when they become creatures, but for non-creatures the marker is a
+            // no-op (all {T}/attack gates are creature-conditional). Reporting it on a freshly
+            // played Mountain would mislead the agent into thinking it can't tap for mana.
+            summoningSick = onBattlefield
+                && container.get<SummoningSicknessComponent>() != null
+                && projected.isCreature(entityId),
             faceDown = container.get<FaceDownComponent>() != null,
             damageMarked = container.get<DamageComponent>()?.amount ?: 0,
             counters = container.get<CountersComponent>()?.counters
--- a/manual-scenarios/cards/t/turtle-blimp.json
+++ b/manual-scenarios/cards/t/turtle-blimp.json
@@ -1,0 +1,34 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Turtle Blimp"],
+    "battlefield": [
+      {"name": "Coral Eel"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"}
+    ],
+    "graveyard": [],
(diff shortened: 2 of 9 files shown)


wingedsheep:
## Review — Vehicle / animated-permanent summoning sickness

**Verdict:** Behavior is correct and the SDK shape is right. This is a pure correctness fix that adds **no new SDK types** — it routes existing battlefield reads through projected state instead of base type/keyword reads, which is exactly the CLAUDE.md invariant. The root-cause framing (control-continuity is a property of the *permanent*, per CR 302.6 / 508.1a, not of "was it a creature the whole turn") is the right model.

### What's good
- **Correct root cause, minimal surface.** Attaching `SummoningSicknessComponent` to *every* entering permanent (rather than special-casing "becomes a creature") is the principled fix — the marker is dormant for non-creatures because every `{T}`/attack gate is creature-conditional, and the start-of-turn removal in `BeginningPhaseManager`/`SacrificeAndPayContinuationResumer` clears it for all permanents regardless of type, so nothing gets stuck sick.
- **Projected-state reads are consistent.** I checked every reader of the component. The `{T}`-cost gates in `ActivateAbilityHandler`, `CostHandler`, `CostEnumerationUtils`, `ManaAbilityEnumerator`, `ActivatedAbilityEnumerator`, and all three `ManaSolver` spots gate on `!isLand && projected.isCreature(...)` and read haste via `projected.hasKeyword(...)`. Lands keep their carve-out, so freshly-played lands still tap for mana. `AttackRestrictionRules` needs no creature gate (only creatures are declared attackers) and now correctly refuses the just-entered Vehicle.
- **CR citations verified** against the 2025-04-04 Comprehensive Rules: 302.6 (summoning-sickness rule — both the `{T}`/`{Q}` restriction and the attack restriction), 508.1a (declare attackers: haste *or* controlled continuously since the turn began), and 602.5a in the test comment all match the actual rule text.
- **Test exercises the real path** — casts Turtle Blimp from hand through `ZoneTransitionService`, crews it, and asserts the attack is refused at declare-attackers. Good rule-corner coverage. Passes locally (`:rules-engine`).

### Issues

**Minor (addressed) — client DTO parity.** The PR fixed the gym `ObservationBuilder` to gate `summoningSick` on `projected.isCreature(...)`, but the parallel web-client path in `ClientStateTransformer.kt:694` was left ungated — so the server now reports `hasSummoningSickness = true` for freshly-played lands and Equipment. Today the only thing preventing a visible 💤 overlay on a new Mountain is a render-time `cardTypes.includes('CREATURE')` check in `GameCard.tsx:1226`; the DTO field itself was semantically wrong and any future consumer would inherit the bug. I pushed a one-line fix gating it at the source (`projectedState.isCreature(entityId)`), mirroring the `ObservationBuilder` change — commit `e1432525e`.

### Recommendation
Mergeable as-is with the pushed parity fix. No SDK changes needed. (Note: I validated against a local merge of `origin/main` — it merged cleanly with no conflicts, so no action needed there.)

PR #471, opened 2026-06-04
Add Lightstall Inquisitor to Edge of Eternities

bolav:
Implements Lightstall Inquisitor (EOE #24, {W} 2/1 Angel Wizard with vigilance):

▎ When this creature enters, each opponent exiles a card from their hand and may play that card for as long as it remains exiled. Each spell cast this way costs {1} more to cast. Each land played this way
▎ enters tapped.

SDK additions (collection-based, reusable):
- Effects.GrantPlayWithCostIncrease(from, amount) — stamps PlayWithCostIncreaseComponent on every card in a named pipeline collection, so CastSpellHandler's existing surcharge path taxes the cast at {amount}
more generic. Companion to the per-target ExileAndGrantOwnerPlayPermission (Soul Partition).
- Effects.GrantMayPlayFromExile(..., landEntersTapped) — new flag on the may-play permission. PlayLandHandler reads it off the active permission and stamps TappedComponent before the played land's own ETB
script runs. Per CR 616.1a both replacements end with the land tapped, so the shock-land "pay 2 life" prompt is intentionally elided under this grant.
- PlayLandHandler also now cleans up may-play permissions when a land plays from exile (lands skip the stack, so StackResolver's cleanup never ran for them — a pre-existing latent issue that benefits other
exile-grant cards too).

ETB pipeline: ForEachPlayer(EachOpponent) { GatherCards(hand) → SelectFromCollection → MoveCollection(exile) → GrantMayPlayFromExile(permanent, landEntersTapped) → GrantPlayWithCostIncrease(1) }. Each opponent
makes their own selection from their own hand.

Tests: Four scenario tests — markers stamped, land plays tapped, opponent's cast fails for "Not enough mana" when only base cost is available (proving the surcharge fires), and resolves onto the battlefield
when the surcharged cost is met.

Doc reference updated in the same change.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -324,8 +324,9 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 - `ReturnLinkedExileUnderOwnersControl()` — return under each card's owner.
 - `ReturnLinkedExileToHand()` — return all from linked exile to hand.
 - `ReturnOneFromLinkedExile()` — return one chosen card.
-- `GrantMayPlayFromExile(from, restriction?)` — owner may play matching cards from exile.
+- `GrantMayPlayFromExile(from, expiry?, withAnyManaType?, condition?, landEntersTapped?)` — controller may play matching cards from exile. `landEntersTapped=true` forces a played land tapped regardless of its own ETB script (Lightstall Inquisitor); PlayLandHandler reads the flag off the active `MayPlayPermission` at play time and stamps `TappedComponent` before the card's intrinsic `EntersTapped` branch runs.
 - `GrantPlayWithoutPayingCost(from)` — same, without paying mana costs.
+- `GrantPlayWithCostIncrease(from, amount)` — stamp `PlayWithCostIncreaseComponent(controllerId, amount)` on every card in the collection, so the next cast pays `{amount}` extra generic. Pair with `GrantMayPlayFromExile` for "each spell cast this way costs {N} more" clauses (Lightstall Inquisitor); for target-based "exile this permanent, owner may play it, opponents tax" effects use `Effects.ExileAndGrantOwnerPlayPermission` instead.
 - `GrantFreeCastTargetFromExile(target)` — cast specific exiled card for free.
 
 ### Stats & keywords
--- a/manual-scenarios/cards/l/lightstall-inquisitor.json
+++ b/manual-scenarios/cards/l/lightstall-inquisitor.json
@@ -1,0 +1,31 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Lightstall Inquisitor"],
+    "battlefield": [
+      {"name": "Plains", "tapped": false}
+    ],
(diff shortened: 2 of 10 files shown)


wingedsheep:
## Review — Lightstall Inquisitor (PR #471)

**Verdict:** Behavior is correct and the SDK shape is right. The card composes the existing exile-pipeline primitives, the two SDK additions are minimal and reusable (a collection-based companion to the per-target Soul Partition path, plus a persistent permission flag), and all four scenario tests pass against `main` merged in. One CR-number nit, fixed in-branch.

### What's good
- **Reuses the existing surcharge machinery instead of inventing it.** `GrantPlayWithCostIncreaseEffect` just stamps the pre-existing `PlayWithCostIncreaseComponent`, which `CastSpellHandler` already reads and keys on `controllerId == action.playerId`. No new cost path — the collection executor mirrors `GrantPlayWithoutPayingCostExecutor` exactly, pushing a cost up rather than waiving it. This is the right "collection variant of the target-based `ExileAndGrantOwnerPlayPermission`" framing.
- **`landEntersTapped` is modeled on the permission, not as a one-shot effect** — correct, because the land may be played many turns after the ETB resolves, so the tapped rider has to persist with the grant and be read at `PlayLandHandler` time. A `CompositeEffect` at ETB couldn't reach a future `PlayLand`.
- **The latent permission-cleanup fix is genuinely worth having.** Lands skip the stack, so `StackResolver` never ran `removeMayPlayPermissionsForCard` for them; routing it through `PlayLandHandler` on any exile→battlefield land play closes that for every exile-grant card. And `removeMayPlayPermissionsForCard` correctly drops only the single `cardId` from multi-card permissions (Etali/Narset-style), so it won't revoke "any number of spells" grants.
- **Tests exercise the rule corner, not just the happy path.** The {1}{G}→{2}{G} pair (2 Forests fail "Not enough mana", 3 Forests resolve) actually proves the surcharge fires rather than asserting a marker; the decoy second card forces `SelectFromCollection` to really prompt the opponent rather than auto-pick. Oracle text, set (EOE), collector # (24), and rarity all verified against Scryfall.
- The `permissionForcesTapped` read uses the pre-move `state` before `removeMayPlayPermissionsForCard` runs, and the empty branch falls through to the common land-drop/`ZoneChangeEvent`/trigger finish exactly like the normal tapland `else` — no double land-drop, no dropped landfall trigger.

### Issues

**Minor (fixed in this branch — commit `3aa3ffb`)** — `PlayLandHandler.kt:229` cited **CR 616.1a** for "both replacements apply and the controller picks the order." 616.1a is the self-replacement carve-out (CR 614.15), and neither Lightstall's "enters tapped" rider nor a shock land's enters-tapped is a self-replacement effect, so the governing rule is **616.1** (affected permanent's controller chooses one to apply) with the applicable step being the catch-all **616.1e**. Comment-only; behavior unchanged. The same `616.1a` reference appears in the PR description — worth correcting there too. The elision conclusion itself is sound: either ordering ends with the land tapped, so paying 2 life is strictly a no-op under this grant.

### Recommendation
Good to merge once the pushed comment fix is pulled. No code changes needed beyond the citation nit. (Reviewed with `origin/main` merged into the branch; the merge was clean.)

PR #472, opened 2026-06-04
Add Pyramids to Arabian Nights

bolav:
Implements Pyramids ({6} artifact, ARN). {2}: choose one — destroy target Aura attached to a land, or shield a target land from the next destruction this turn by clearing its damage instead.

New SDK primitives:

- StatePredicate.AttachedToCardType(cardType) — entity whose AttachedToComponent points to a permanent currently having the given top-level CardType in its projected types. Reusable for "Equipment attached to a creature", etc.
- RemoveDamageShieldEffect / SerializableModification.RemoveDamageShield — one-shot destruction shield similar in shape to regeneration, but without tapping the target or removing it from combat. ZoneMovementUtils checks it after the regen pass; not gated by "can't be regenerated" (the replacement isn't a regeneration ability).

ClientStateTransformer surfaces the shield as a "Shielded" effect badge so the UI shows when a land is protected.

Oracle text matches Scryfall verbatim; canonical hosted in ARN (the card's only printing); just check-card-printing "Pyramids" is happy.

Tests in PyramidsScenarioTest cover: shield replaces destroy + clears damage + does NOT tap, one-shot consumption (second destroy goes through), regen-preferred coexistence, AttachedToCardType match / non-match / missing-component, and a projection-flip test that animates a Forest into a creature via a Layer 4 AddType("CREATURE") floating effect and asserts
  the predicate matches both LAND and CREATURE — locking the rule that the predicate reads projected types, not base.

### DIFF
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -81,7 +81,7 @@
 - [ ] Jandor's Ring
 - [x] Jandor's Saddlebags
 - [ ] Jeweled Bird
-- [ ] Pyramids
+- [x] Pyramids
 - [ ] Ring of Ma'rûf
 - [x] Sandals of Abdallah
 - [x] Bazaar of Baghdad
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -878,8 +878,10 @@ work for abilities-on-stack (which carry no `CardComponent`).
 - `IsFaceDown` — currently face-down.
 - `HasCounter(type)` — has at least one counter of `type`.
 - `AttachedToCardType(cardType)` — Aura/Equipment whose `AttachedToComponent` points to a
-  permanent whose card includes the given top-level [`CardType`]. Used by filters like
-  "Aura attached to a land" (Pyramids) or "Equipment attached to a creature". False for
+  permanent that currently has the given top-level [`CardType`] in its **projected** type
+  set. Used by filters like "Aura attached to a land" (Pyramids) or "Equipment attached
+  to a creature". Reads the attached-to permanent's projected types, so a land animated
+  into a creature still matches `LAND` and additionally matches `CREATURE`. False for
   entities with no `AttachedToComponent`.
 - `IsWarpExiled` (filter builder `warpExiled()`) — card in exile via warp's
   end-of-turn delayed trigger (CR 702.185b).
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -265,6 +265,15 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 ### Destruction & exile
 
 - `Destroy(target)` — destroy target (respects indestructible).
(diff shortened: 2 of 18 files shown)


wingedsheep:
## Review — Pyramids (ARN)

**Verdict:** Behavior is right and the SDK shape is well-judged. The two new primitives are both genuinely novel (neither is expressible by composing existing ones), parameterized for the next card, and named honestly. Oracle text, cost, type, artist, and collector number all match Scryfall verbatim, and `check-card-printing` is green (ARN is the only printing, canonical hosted there). I pushed one correctness fix + test directly to the branch.

### What's good
- **`RemoveDamageShield` is correctly distinct from regeneration.** It mirrors `RegenerationShield`'s floating-effect/consume-on-destroy shape but skips the tap and the combat-removal, exactly matching "remove all damage marked on it instead." Reusing the regen pass structure (and slotting the executor into `RegenerationExecutors`) keeps it discoverable.
- **`AttachedToCardType` reads *projected* types.** `PredicateEvaluator` routes through `state.projectedState.hasType`, and the predicate is correctly placed in `StatePredicate` (attachment is state, not card identity). The projection-flip test (animate a Forest → also matches `CREATURE`) locks that in — nice.
- **All the exhaustive-`when` plumbing is covered:** `CardValidator`, `BeginningPhaseManager`, `TriggerMatcher`, `AffectsFilterResolver`, and the `toModification()` NoOp mapping. `ClientStateTransformer` surfaces a "Shielded" badge. Nothing left dangling.
- **Tests are clean and hit the right axes:** one-shot consumption, regen coexistence/preference, no-tap assertion, and the three `AttachedToCardType` match/non-match/missing-component cases.

### Issues

**Important (fixed in 010838c5)** — the shield was only consulted in `ZoneMovementUtils.destroyPermanent` (explicit "destroy" effects), **not** in `LethalDamageCheck` (the 704.5g/h lethal-damage SBA). But Pyramids' wording — "remove all damage marked on it" — is specifically the *damage* case: an animated land taking lethal combat damage is destroyed via the SBA, which bypassed the new shield entirely (it would die despite the shield). I mirrored the existing regen handling in `LethalDamageCheck` and added a test (`shield fires against lethal-damage destruction (SBA)`). This is the same intent the PR's own projection-flip test signals you care about (animated lands).

**Minor** — when both a regeneration shield and a remove-damage shield are on the same permanent, the engine resolves regen first (per the comment in `destroyPermanent`). Strictly, CR 616.1 lets the affected permanent's controller choose which replacement applies. This matches the engine's existing auto-ordering for shields and is an extreme corner (you'd need both Pyramids and a regen source on the same animated land), so it's noted, not a request.

**Minor** — neither `applyRemoveDamageReplacement` nor the existing `applyRegenerationReplacement` emit a `GameEvent` for the replacement firing. Consistent with current precedent, so no change requested — flagging only in case a "permanent was shielded/regenerated" event is ever wanted for triggers/animations.

### Recommendation
Good to merge with the pushed fix. No SDK changes needed — both primitives are the right shape. Reviewed in worktree `~/.claude/worktrees/argentum-engine/pr-472-review`; `origin/main` was already merged in (no merge commit needed). `PyramidsScenarioTest` is green at 8/8, and the regeneration/SBA suites pass.

PR #474, opened 2026-06-04
Add 6 TDM cards (lands, Narset, Rediscover the Way, Sarkhan Dragon Ascendant)

wingedsheep:
Implements six Tarkir: Dragonstorm (TDM) cards, plus three small, reusable SDK/engine features they need. All scenario tests pass and the full rules-engine + mtg-sdk + mtg-sets suites are green.

## Cards
- **Great Arashin City** (#257) — check-land; `{T}: Add {B}`; `{1}{B}, {T}, Exile a creature card from your graveyard: Create a 1/1 white Spirit`.
- **Mistrise Village** (#261) — check-land; `{T}: Add {U}`; `{U}, {T}: the next spell you cast this turn can't be countered` (existing `MakeNextSpellUncounterable` rider).
- **Maelstrom of the Spirit Dragon** (#260) — `{T}: Add {C}`; `{T}: any-color mana spendable only on a Dragon or Omen spell`; `{4}, {T}, Sacrifice: tutor a Dragon to hand`.
- **Narset, Jeskai Waymaster** (#209) — end step: may discard your hand, draw cards equal to spells cast this turn.
- **Rediscover the Way** (#215) — Saga; chapters I/II dig top three keep one; chapter III installs a per-cast targeting delayed trigger granting double strike to a chosen creature each noncreature spell you cast that turn.
- **Sarkhan, Dragon Ascendant** (#118) — ETB behold a Dragon → create a Treasure; whenever a Dragon you control enters, +1/+1 counter on Sarkhan and he becomes a Dragon with flying until end of turn.

## SDK / engine features (general-purpose, documented in `card-sdk-language-reference.md`)
- `ManaRestriction.SubtypeSpellsOnly(subtypes)` — mana spendable only on spells carrying any of N subtypes (Dragon **or** Omen).
- `CreateDelayedTriggerEffect.targetRequirement` — a target chosen each time a delayed trigger fires, threaded through `DelayedTriggeredAbility` → `TriggeredAbility.create`. Also wired `SpellCastEvent` into the filter-scoped delayed-trigger detector so "whenever you cast a [filtered] spell this turn, …" delayed triggers fire.
- `Effects.Behold(filter, ifBeheld)` — resolution-time behold effect (choose a controlled matching permanent or reveal a matching card from hand; then run the payoff), with executor + continuation, distinct from the existing cast-time `AdditionalCost.Behold`.

## Tests
New scenario tests cover restricted mana, the targeting delayed trigger, the behold→Treasure flow and Dragon-enters buff, Narset's discard/draw, and both check-lands. Full `:rules-engine:test`, `:mtg-sdk:test`, `:mtg-sets:test` pass.

## Not included
**Mardu Siegebreaker** is intentionally deferred — it needs a larger feature set than this batch (token copy of a *linked-exile* card + per-opponent attacking tokens with the correct defender + delayed sacrifice), which is multi-feature engine work better done on its own.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -306,6 +306,14 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
   express "if you don't, X" riders — e.g. SOI shadow lands wrap this in
   `OnEnterRunEffect(...)` with `otherwise = Effects.Tap(EffectTarget.Self)` for the
   "this land enters tapped" branch.
+- `Effects.Behold(filter, ifBeheld?)` — resolution-time **behold** (`BeholdEffect`): "you may
+  behold a `filter`. If you do, `ifBeheld`." The behold itself is optional — the controller may
+  choose a matching permanent they control **or** reveal a matching card from hand (revealing emits
+  `CardsRevealedEvent`; battlefield permanents are merely chosen). If they decline, or control no
+  matching permanent and hold no matching card, `ifBeheld` does not run. Distinct from the cast-time
+  `AdditionalCost.Behold` / `AdditionalCost.BeholdOrPay` (which behold as a casting cost). Sarkhan,
+  Dragon Ascendant ETB: `Effects.Behold(GameObjectFilter.Any.withSubtype(Subtype.DRAGON),
+  ifBeheld = Effects.CreateTreasure())`.
 
 ### Library reveal & free cast
 
@@ -1286,6 +1294,14 @@ Triggers.youCastSpell(
     control on your next attack). With `fireOnce = false` (default) it fires on every matching event
     until expiry (double-strike combat damage). One-shot consumption happens when the trigger goes
     on the stack (`TriggerProcessor`), so a second matching event the same turn won't re-fire it.
+  - `targetRequirement = <TargetRequirement>` — a target chosen **each time** the delayed trigger
+    fires, exposed to `effect` as `EffectTarget.ContextTarget`. Use for delayed triggers whose payoff
+    targets: Rediscover the Way chapter III installs
+    `CreateDelayedTriggerEffect(trigger = Triggers.YouCastNoncreature, fireOnce = false,
+    expiry = EndOfTurn, targetRequirement = Targets.CreatureYouControl,
+    effect = Effects.GrantKeyword(Keyword.DOUBLE_STRIKE))` — "whenever you cast a noncreature spell
+    this turn, target creature you control gains double strike". Works on both event-based and
+    step-based delayed triggers; null (default) for non-targeting delayed triggers.
 
 ---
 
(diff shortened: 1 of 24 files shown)


wingedsheep:
## Review — 6 TDM cards + 3 SDK features

**Verdict:** Behavior is correct and the SDK shape is right. All six cards match Oracle text, all cited mechanics are exercised, and the three new SDK primitives are genuinely novel and parameterized for reuse rather than baked to one card. Merged `origin/main` cleanly (no conflicts) before reviewing. I added one missing test and pushed it to the branch.

### What's good
- **`BeholdEffect` is the right call, not over-generalized.** Resolution-time behold is a distinct interaction shape (choose-a-permanent-you-control **or** reveal-from-hand, with conditional reveal) that no existing effect covered — the cast-time `AdditionalCost.Behold` is a cost, not an effect. The executor mirrors the cast-time behold convention exactly (`CostHandler:751` — battlefield via projected state, hand via base), so the two stay consistent. It correctly stays behold-specific (a `filter` + `ifBeheld` payoff) instead of trying to parameterize reveal/exile/mill.
- **`CreateDelayedTriggerEffect.targetRequirement`** is clean plumbing: threaded `Effect → DelayedTriggeredAbility → TriggeredAbility.create(targetRequirement=)` (which already accepted it), with the matching `SpellCastEvent` branch added to the filter-scoped delayed-trigger detector so the noncreature filter + casting-player predicate are honored. Per-firing targeting on a delayed trigger genuinely wasn't expressible before.
- **`ManaRestriction.SubtypeSpellsOnly(Set<String>)`** is parameterized over any subtype set (OR-joined), not hardcoded to Dragon/Omen — the next multi-subtype land composes it directly.
- **Projection is correct everywhere** — the behold executor uses `matches(state, projected, …)` and `getBattlefieldControlledBy` (the projection-aware path; note `matchesWithProjection` no longer exists, CLAUDE.md naming is stale), so a creature that became a Dragon via a continuous effect is beholdable.
- **Printings verified:** `just check-card-printing` is clean for all six — canonical lives in TDM (earliest real printing); #209/#118 are correctly `RARE` (only the boosterfun variants #407/#417 are mythic).
- Tests are well-targeted (restricted-mana pays Dragon but not Ogre; per-cast targeting delayed trigger; behold→Treasure; Narset discard/draw=spell count; both check-lands). I ran all of them green locally.

### Issues

**Minor**
- `EffectAndTriggerContinuationResumer.kt:371` — `resumeBehold` emits the hand-reveal event into `events`, then runs `ifBeheld`; if `ifBeheld` *pauses* it does `return result`, dropping the buffered reveal event from the emitted stream (the state change survives, only the `GameEvent` is lost). No current payoff pauses (`CreateTreasure` doesn't), so it's latent — but worth a comment or a fix before a behold whose payoff prompts. Left as-is rather than risk a half-baked continuation-threading fix.
- `BeholdEffect` is hardcoded optional ("you may behold"). A mandatory "behold a Dragon" effect (no *may*) would need a flag. Fine for now — every effect-side behold printed so far is "you may".
- `SubtypeSpellsOnly` overlaps the single-subtype `SubtypeSpellsOrAbilitiesOnly`; eventually the latter could be subsumed, but keeping a focused spell-only variant avoids reshaping an existing serializable type. Acceptable.
- No test covers the Omen branch of `SubtypeSpellsOnly` (Omen may not be modeled yet) — harmless and forward-compatible.

### Change I pushed
- **Added `SarkhanDragonAscendantScenarioTest` "behold a Dragon you control (battlefield)" case.** The executor's *battlefield* branch (chosen, not revealed — `handOptionIds` exclusion + no reveal) was untested; only behold-from-hand and decline were covered. New test puts the only beholdable Dragon on the battlefield and confirms a Treasure is created. Green locally alongside the rest of the suite.

### Recommendation
Ship as-is. The pushed test closes the one coverage gap; the Minor items are author's discretion (the `resumeBehold` event-loss is the only one I'd consider following up before the next behold card with a pausing payoff).

PR #475, opened 2026-06-04
Add 7 Tarkir: Dragonstorm cards (Surrak, Tersa, Teval, Ugin, Ureni x2, Zurgo)

wingedsheep:
Implements 7 missing Tarkir: Dragonstorm (TDM) cards, each with a Kotest scenario test.

## Cards

| Card | Notes |
|------|-------|
| **Ureni's Rebuff** | Bounce sorcery + harmonize — existing primitives only. |
| **Surrak, Elusive Hunter** | Uncounterable, trample; draw whenever a creature **or creature spell** you control is targeted by an opponent. |
| **Tersa Lightshatter** | Haste; ETB loot (discard up to two, draw that many); attack trigger (7+ graveyard) exiles a random card you may play this turn. |
| **Zurgo, Thunder's Decree** | Mobilize 2; during your end step, Warrior tokens you control can't be sacrificed (so the mobilize tokens survive). |
| **Ureni, the Song Unending** | Flying, protection from white/black; ETB divides X damage (X = lands you control) among any number of opponent creatures/planeswalkers. |
| **Teval, Arbiter of Virtue** | Flying, lifelink; spells you cast have delve; lose life equal to each cast spell's mana value. |
| **Ugin, Eye of the Storms** | Colorless planeswalker; cast + colorless-spell triggers exile a colored permanent; +2 / 0 / −11 (search colorless nonland cards → exile → free-cast until EOT). |

## Engine / SDK additions (minimal, reusable)

- **Spell targets emit `BecomesTargetEvent`** (StackResolver) so "a creature spell you control becomes the target" triggers fire; `TriggerMatcher` resolves a spell target's controller from its caster. Ward stays battlefield-only.
- **`AbilityFlag.CANT_BE_SACRIFICED` + `CantBeSacrificed` static** + **`IsInStep` condition** (dual-mode, projection-safe). `SacrificeTargetExecutor` honors the projected flag.
- **`DividedDamageEffect.dynamicTotal`** — a `DynamicAmount` total computed at resolution / stack placement (Ureni). Zero chosen targets ⇒ no-op.
- **`CardPredicate.IsColored`** ("one or more colors"), wired through every color-predicate site.
- `docs/card-sdk-language-reference.md` updated for each new building block.

## Tests

Per-card scenario tests under `rules-engine/.../scenarios/`. Full `:rules-engine:test` and `:mtg-sdk:test` suites pass; `:mtg-sets` builds.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1214,13 +1214,15 @@ Triggers.youCastSpell(
 - `CreatureTurnedFaceUp(player?)` — when a creature you control turns face up.
 - `GainControlOfSelf` — you gain control of source.
 - `BecomesTarget(filter?)` — source becomes target of spell/ability. The engine emits the
-  underlying `BecomesTargetEvent` for both permanent targets and spell targets on the stack, and
-  the `filter` is matched against the targeted object's card data — so a `Creature` filter also
-  matches a creature spell on the stack (Surrak, Elusive Hunter: "a creature you control or a
-  creature spell you control becomes the target"). Ward never sees spell targets because it is
-  generated only from battlefield permanents.
-- `CreatureYouControlBecomesTargetByOpponent(filter?)` — your creature (permanent or creature spell)
-  gets targeted by an opponent's spell or ability.
+  underlying `BecomesTargetEvent` for both permanent targets and spell targets on the stack, but the
+  trigger matches **permanent targets only** by default — "a creature you control" is a battlefield
+  creature, not a creature spell. Set `includeSpellTargets = true` on the event for the "... or a
+  creature spell you control" wording (Surrak, Elusive Hunter); the `filter` is then also matched
+  against the spell's card data, so a `Creature` filter matches a creature spell on the stack. Ward
+  never sees spell targets because it is generated only from battlefield permanents.
+- `CreatureYouControlBecomesTargetByOpponent(filter?, includeSpellTargets = false)` — your creature
+  gets targeted by an opponent's spell or ability. Permanent-only unless `includeSpellTargets = true`
+  (Surrak), which also fires when an opponent targets a matching creature spell you control.
 - `Transforms` — source transforms (either direction).
 - `TransformsToFront` — to front face.
 - `TransformsToBack` — to back face.
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -238,6 +238,13 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 - `DealDamage(amount, target)` — deal fixed/dynamic damage.
 - `DealXDamage(target)` — deal X damage (spell's X).
 - `Fight(target1, target2)` — two creatures each deal damage equal to their power to each other (CR 701.12).
+- `DividedDamageEffect(totalDamage, minTargets, maxTargets, dynamicTotal?)` — "N damage divided as you
+  choose among target ..." The targets come from the ability's target requirement; pair with
+  `TargetCreature(count, minCount)` (Forked Lightning, Skirk Volcanist) or, for "any number of target"
+  + a dynamic total, a `TargetObject(optional = true, dynamicMaxCount = ..., filter = ...)`. Set
+  `dynamicTotal` (a `DynamicAmount`) for totals computed when the ability resolves/goes on the stack —
+  Ureni, the Song Unending: `dynamicTotal = DynamicAmounts.landsYouControl()`. Works for creatures and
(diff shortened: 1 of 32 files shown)


wingedsheep:
## Review — 7 TDM cards + becomes-target / divided-damage / can't-be-sacrificed primitives

**Verdict:** Behavior is correct and the new primitives are well-shaped and reusable — with one real regression I've fixed and pushed: the spell-target change made *every* permanent-only "a creature you control becomes the target" trigger also fire on creature **spells**. Oracle text, costs, color identities, and collector numbers all verified against Scryfall and match.

### What's good (keep it)
- **`DividedDamageEffect.dynamicTotal`** is a clean, minimal extension — the total is a `DynamicAmount` evaluated at resolution, the target set comes from the ability's `TargetObject(optional, dynamicMaxCount=…)`, and zero targets cleanly no-op. Both the resumer and the executor compute the total the same way. Good reuse, not over-specialized.
- **`IsInStep(steps, yoursOnly)`** is board-derived (reads `state.step` + active player) so it evaluates identically at resolution and under projection — exactly what a `ConditionalStaticAbility` gate needs. Nicely generic.
- **`CardPredicate.IsColored`** wired through *all* color-predicate sites (`PredicateEvaluator` ×2, `AffectsFilterResolver`, `CostCalculator`) — no fall-through gaps.
- The `emitBecomesTarget` refactor collapses four near-identical loops in `StackResolver` into one helper — good cleanup.
- Tests are thorough: Ureni covers the zero-target path, Zurgo asserts the projected flag *and* token survival, Surrak covers spell + permanent + player targets.
- `card-sdk-language-reference.md` updated for every new building block.

### Blocking (fixed in 5c13eaa)
- **Spell-target emission regressed permanent-only becomes-target triggers.** `StackResolver` now emits `BecomesTargetEvent` for spell targets, but `matchesBecomesTargetTrigger` never checked `targetIsSpell`. Every ANY-bound trigger with a matching card filter — **Pawpatch Recruit** (`CreatureYouControlBecomesTargetByOpponent()`), **Daru Spiritualist** (`BecomesTarget(Cleric.youControl())`) — would now fire when an opponent targets a creature *spell* you control (e.g. counters it). Per the comprehensive rules, "a creature you control" is a battlefield creature, not a creature spell; only Surrak's oracle says "… **or a creature spell** you control". (Elrond escaped only because its `withCounter` filter excludes spells; Skulking Ghost because it's SELF-bound.)

  **Fix:** added `includeSpellTargets` (default `false`) to `EventPattern.BecomesTargetEvent`; `TriggerMatcher` skips spell-target events unless opted in; `CreatureYouControlBecomesTargetByOpponent` takes the flag and Surrak passes `true`. Added a regression test (Pawpatch Recruit stays quiet when an opponent counters your creature spell) — it fails without the gate, passes with it.

- **Valiant first-time tracking could be corrupted by spell targets.** `emitBecomesTarget` was calling `markTargetedByController` on the spell's stack entity. A permanent spell's stack entity can be reused as the resolved permanent, leaking a stale "targeted this turn" flag onto the permanent and suppressing a later Valiant trigger. **Fix:** spell targets no longer participate in the per-controller targeted-this-turn tracking (they carry `firstTime = true` and leave the tracking untouched).

### Minor (author's discretion)
- **`CantBeSacrificed` only honored in `SacrificeTargetExecutor`.** There are three sacrifice paths (`SacrificeTargetExecutor`, `SacrificeSelfExecutor`, `SacrificeExecutor`); the projected `CANT_BE_SACRIFICED` flag is checked only in the first. That's sufficient for Zurgo (mobilize's delayed sacrifice routes through `SacrificeTargetExecutor`, and the test proves it), but the primitive's name promises more than it delivers — a future "Goblins can't be sacrificed" meant to stop a sac-outlet or edict would silently fail. Consider centralizing the flag check in the shared sacrifice/zone-move helper so all three paths honor it, or note the scope limit in the doc.
- **`DividedDamageEffect` now carries both `totalDamage: Int` and `dynamicTotal: DynamicAmount?`.** Slightly redundant — a single `total: DynamicAmount` (fixed cards pass `DynamicAmount.Fixed(n)`) matches the engine's "amounts are `DynamicAmount`" convention and avoids the override dance. Not worth blocking; the dual fields are clearly documented. Flagging only for the convention.

### Tests run
`:mtg-sdk:test` and `:rules-engine:test` both green after the fix (including the new Surrak regression test and the existing Pawpatch / Valiant / Manifold Mouse cases). `mtg-sets` compiles.

### Recommendation
The fix is pushed to this branch (also merges current `origin/main`, cleanly). Ready to merge once you've eyeballed the gate change. Consider the two Minor items as follow-ups.

*Reviewed in worktree `~/.claude/worktrees/argentum-engine/pr-475-review`.*

PR #476, opened 2026-06-04
LTR: cast-as-flash permission + ETB/LTB doubler, +3 cards

bolav:
## Summary
Closes Gap 4 and Gap 4b in LTR's TODO and implements the three cards that
needed them.

**Gap 4 — "cast [filter] spells as though they had flash"**
- New `Effects.GrantFlashToSpells(target, spellFilter, duration)` →
  `FlashGrantsThisTurnComponent` on the target player. Turn-scoped,
  player-scoped sibling of the existing permanent-static
  `GrantFlashToSpellType`; both are consulted by
  `CastPermissionUtils.hasGrantedFlash` / `CastZoneResolver.hasGrantedFlash`
  (CR 601.3b, 702.8a). Cleared at cleanup (CR 514.2).

**Gap 4b — "triggers an additional time" for ETB *or* LTB**
- Existing `AdditionalETBTriggers` (Panharmonicon family) reshaped into
  `AdditionalETBOrLTBTriggers` with a `Set<BattlefieldDirection>`. Default
  `{ENTERING}` keeps Panharmonicon / Naban / Starfield Vocalist / Traveling
  Chocobo behaviour byte-identical; add `LEAVING` for Gandalf-the-White-
  style "entering or leaving the battlefield" doublers. LTB path reads
  `ZoneChangeEvent.lastKnownController` (CR 603.10a). CR 603.2d.

**Cards (LTR)**
- Borne Upon a Wind — Effect-only.
- Gandalf, Friend of the Shire — Flash + sorcery-flash static + existing
  Ring-tempts-you condition.
- Gandalf the White — Flash + legendary/artifact flash static + reshaped
  ETB-or-LTB doubler with `mustBeYouControl = false`.

Implemented count: 194 → 197 / 291.

## Test plan
- [x] `:rules-engine:test` — full suite green, incl. 22 new tests across
  `BorneUponAWindTest`, `GandalfFriendOfTheShireTest`,
  `GandalfTheWhiteTest`, `GrantFlashToSpellsEffectTest`.
- [x] `just check-card-printing` clean for all three cards (canonical in
  earliest real LTR printing).

### DIFF
--- a/backlog/sdk-reusability-consolidation.md
+++ b/backlog/sdk-reusability-consolidation.md
@@ -323,7 +323,7 @@ ZoneChange/SpellCast/Attack/DealsDamage variants:
 - **`PlayFromTopOfLibrary`** / **`CastSpellTypesFromTopOfLibrary`** /
   **`LookAtTopOfLibrary`** / **`PlayLandsAndCastFilteredFromTopOfLibrary`** →
   `LibraryTopAccess(view, play)`. 4 → 1.
-- **`AdditionalETBTriggers`** vs **`AdditionalSourceTriggers`** →
+- **`AdditionalETBOrLTBTriggers`** vs **`AdditionalSourceTriggers`** →
   `AdditionalTriggers(triggerScope, sourceFilter)`.
 
 ### Targeting (`scripting/targets/`)
--- a/backlog/sets/lord-of-the-rings/TODO.md
+++ b/backlog/sets/lord-of-the-rings/TODO.md
@@ -107,12 +107,25 @@ steal no longer silently restores the designation when control reverts at end of
   needs a "Whenever you scry" trigger + "reveal top; if land, put onto battlefield tapped".
 
 ### Gap 4 — "cast [filter] spells as though they had flash" permission
-**Engine change:** a static/one-shot permission granting flash-timing to a filtered set of
-spells.
-- **Borne Upon a Wind** — "You may cast spells this turn as though they had flash."
-- **Gandalf, Friend of the Shire** — "You may cast sorcery spells as though they had flash."
-- **Gandalf the White** — "You may cast legendary spells and artifact spells as though they
-  had flash." (also needs the extra-trigger replacement below).
+**Status:** LANDED as `Effects.GrantFlashToSpells(target, spellFilter, duration)` plus
+`FlashGrantsThisTurnComponent` (PR `ltr-gap4-cast-as-flash`). The Effect is the turn-scoped,
+player-scoped sibling of the existing permanent-static `GrantFlashToSpellType`; both are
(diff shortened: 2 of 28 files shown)


wingedsheep:
## Review + fix pushed

Reviewed the full diff (with `origin/main` merged in — clean, no conflicts). The SDK shape is good: reshaping `AdditionalETBTriggers` → `AdditionalETBOrLTBTriggers` with a `Set<BattlefieldDirection>` (instead of a new type) is the right compose-don't-proliferate move, default `{ENTERING}` keeps Naban/Starfield/Chocobo byte-identical, and `GrantFlashToSpellsEffect` has complete cross-layer plumbing (component → serialization → cleanup → `ClientStateTransformer`). All five cited CR rules verified correct (702.8a, 601.3b, 603.2d, 603.10a, 514.2). Printing placement clean for all three cards.

### One correctness gap — fixed in 483644a9d

`TriggerDetector.duplicateETBOrLTBTriggers` re-checked its filter against the cause with `predicateEvaluator.matches(...)` on **live state**. For the `LEAVING` direction that diverges from the engine's canonical leave-matching (`TriggerMatcher.matchesZoneChangeTrigger`), which prefers the `ZoneChangeEvent`'s `lastKnownTypeLine` per **CR 603.10a** (leaves-the-battlefield checks look back in time).

Effect: a permanent that was an artifact/legendary **only via a continuous effect** reverts to its printed type line once it dies, so Gandalf the White's `Artifact ∨ Legendary` filter silently failed and the trigger it caused wasn't doubled. (A token swept by 704.5d hits the same gap.)

Fix extracts the cause-filter check into `causeMatchesDoublerFilter`: `ENTERING` keeps the projected-state match; `LEAVING` overlays `lastKnownTypeLine` onto the cause entity (synthesizing a minimal one if it was swept) and reuses the **full filter engine**, so `or`/color/subtype predicates all resolve. Added two `GandalfTheWhiteTest` cases — a continuous-effect artifact dying (which **fails without the fix**) and a legendary token dying. Full `:rules-engine:test` suite green.

### Minor (left for author's discretion)

- `GrantFlashToSpellsEffect`'s KDoc lists `UntilYourNextTurn` as a typical duration, but the executor collapses every non-`Permanent` duration to `EndOfTurn` (binary `PlayerEffectRemoval`, same as the `GrantSpellsCantBeCountered` sibling). Behavior is consistent — just drop the `UntilYourNextTurn` mention or note the limitation.
- The turn-grant lookup is duplicated in both `CastZoneResolver.hasGrantedFlash` and `CastPermissionUtils.hasGrantedFlash` (mirrors the pre-existing static-check duplication). A `// keep in sync` note would help.

PR #480, opened 2026-06-05
Add Tannuk, Steadfast Second to Edge of Eternities with hand-warp gra…

bolav:
…nt support

Tannuk's haste anthem composes existing `GrantKeyword(HASTE, Creature.youControl().other())`, but its second clause — "Artifact cards and red creature cards in your hand have warp {2}{R}" — needed a new SDK seam: a battlefield static that grants warp to matching cards in the controller's hand. Added `GrantWarpToCardsInHand(filter, cost)` plus a `WarpGrants` engine helper (mirroring `HarmonizeGrants`) that funnels every printed-or-granted warp lookup through one chokepoint, so the cast-from-zone enumerator, `CastSpellHandler` (validate + execute, cost + bundled additional cost, and the `wasWarped` resolution flag), and `CastZoneResolver.hasWarpPermission` all treat a granted warp identically to a printed one — same alternative-cast legal action, same end-step exile trigger, same `WarpExiledComponent` recast-from-exile path.

The granter scan iterates `state.controlledBattlefield(playerId)` rather than the ownership-keyed `getBattlefield(playerId)`, so the grant follows whoever currently controls the source permanent (CR 109.4) — a stolen Tannuk grants warp to its new controller, not its owner. Recasts from exile go through the standard `MayPlayPermission` added by `WarpExileExecutor` (regular mana cost per CR 702.185a), so a granted-warp card with no printed warp on its definition is still recastable from exile; the scenario test asserts the permission is wired up after the end-step exile.

### DIFF
--- a/backlog/sets/edge-of-eternities/cards.md
+++ b/backlog/sets/edge-of-eternities/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 261 booster cards (excluding basic lands)
 **Release Date:** August 1, 2025
-**Implemented:** 255 / 261
+**Implemented:** 257 / 261
 ---
 
 - [x] Adagia, Windswept Bastion
@@ -125,7 +125,7 @@
 - [x] Larval Scoutlander
 - [x] Lashwhip Predator
 - [x] Lightless Evangel
-- [ ] Lightstall Inquisitor
+- [x] Lightstall Inquisitor
 - [x] Lithobraking
 - [x] Loading Zone
 - [x] Lost in Space
@@ -226,7 +226,7 @@
 - [x] Syr Vondam, the Lucent
 - [x] Systems Override
 - [x] Tannuk, Memorial Ensign
-- [ ] Tannuk, Steadfast Second
+- [x] Tannuk, Steadfast Second
 - [x] Tapestry Warden
 - [x] Temporal Intervention
 - [ ] Terminal Velocity
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1619,7 +1619,10 @@ riders, matching how the engine already treats e.g. City of Brass's damage durin
   by CR 702.185a and the granters' "in your hand" wording — the grant doesn't extend warp to other
   zones. Routed through `WarpGrants.effectiveWarp` alongside printed warp; the granter's controller
   is the only beneficiary. (Tannuk, Steadfast Second = `GrantWarpToCardsInHand(filter = artifact OR
-  red creature, cost = {2}{R})`.)
+  red creature, cost = {2}{R})`.) When the granted warp lands on a card that *also* has another
+  alternative cost (e.g. a red evoke creature), both casts are offered and disambiguated by
+  `CastSpell.alternativeCostType` (see `engine-server-interface.md`) — picking "Evoke" charges the
+  evoke cost, not warp, even though warp would win a naive priority order.
 - `MayCastWithoutPayingManaCost(controllerOnly = false, firstSpellOfTurnOnly = false, spellFilter = Any)` — a
   battlefield permission to cast a spell without paying its mana cost (CR 118.9). Composable
   gates: `controllerOnly = true` restricts the benefit to the source's controller ("you" wording);
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1612,6 +1612,14 @@ riders, matching how the engine already treats e.g. City of Brass's damage durin
   Yawgmoth's Agenda (`MayCastFromGraveyard(Nonland)`); `lifeCost = 1, duringYourTurnOnly = true` for
   Festival of Embers. Pair with `MayPlayLandsFromGraveyard` for "play lands and cast spells from
   your graveyard". Lands are *played*, not cast, so they need the lands permission separately.
+- `GrantWarpToCardsInHand(filter, cost)` — cards in the controller's hand matching `filter` gain
+  warp (CR 702.185) with mana cost `cost`. Behaves identically to a printed warp keyword: surfaces a
+  "Cast (Warp)" legal action, marks `wasWarped` on resolution, and the post-resolution permanent is
+  exiled at the next end step and can be cast again from exile for its regular mana cost. Hand-only
+  by CR 702.185a and the granters' "in your hand" wording — the grant doesn't extend warp to other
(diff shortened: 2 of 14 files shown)


wingedsheep:
## Review — Tannuk, Steadfast Second (`GrantWarpToCardsInHand`)

**Verdict:** Behavior is correct and the SDK shape is right. The new static is a genuinely
novel seam (a battlefield permanent granting warp to matching cards in its controller's
hand — no existing primitive reads that), it's parameterized for the next card (`filter` +
`cost`, no baked-in constants), and the name matches its semantics. Mergeable as-is; I have
no blocking or important findings.

### What's good
- **Single chokepoint.** `WarpGrants.effectiveWarp` mirrors `HarmonizeGrants` and funnels
  every warp read site (enumerator, cast-handler validate/cost/additional-cost, `wasWarped`,
  `CastZoneResolver.hasWarpPermission`) through one resolver, so a granted warp behaves
  identically to a printed one — same legal action, same end-step exile, same recast path.
- **Control-correct.** `findGrantedWarp` scans `state.controlledBattlefield(playerId)`
  (projected control, CR 109.4) rather than the ownership-keyed battlefield, so a stolen
  Tannuk grants warp to its new controller. Nice detail, and tested implicitly by the
  leave-the-battlefield case.
- **Printed-vs-granted precedence** is handled cleanly: the enumerator and `CastZoneResolver`
  read printed warp first (preserving the `fromGraveyard` opt-in), then fall back to the
  hand-only grant — granted warp correctly never satisfies a graveyard pass.
- **Tests cover the rule corners,** not just the happy path: self-excluded haste anthem,
  opponent exclusion, both filter branches (red creature / artifact), a negative
  (green non-artifact gets nothing), the full warp lifecycle (cast → `WarpedComponent` →
  end-step exile → `WarpExiledComponent` → `MayPlayPermission` recast for a card with **no
  printed warp**), and grant-revocation on leave. All 7 pass locally; `WarpMechanicTest` and
  `SongcrafterMageTest` still green.
- Doc updated in the same change (`card-sdk-language-reference.md`), and
  `just check-card-printing` is clean (canonical in EOE, `peoe` promos correctly not scaffolded).
- CR citations (702.185 / .185a / .185b, 109.4) match the warp implementation already on main —
  no rule-number drift.

### Minor (author's discretion)
- **Enumeration hot-path cost.** `enumerateWarpFromZone(HAND)` now calls
  `effectiveWarp` → `findGrantedWarp` for every non-warp hand card, and each call scans the
  whole controlled battlefield (`CastFromZoneEnumerator.kt:1328`). In the common case (no
  granter out) that's `hand × battlefield` map-lookups per `getLegalActions`, which matters
  for MCTS/self-play. It mirrors the existing `CastPermissionUtils` scan style and the
  absolute cost is small (~hand×battlefield instanceof checks), so I left it rather than
  fracture the single-chokepoint design — but a one-time "does this player control any
  `GrantWarpToCardsInHand`?" guard before the loop would erase the multiplier if enumeration
  perf shows up in a profile.
- **`wasWarped` disambiguation (pre-existing).** `wasWarped = useAlternativeCost &&
  effectiveWarp != null` (`CastSpellHandler.kt:2073`) would also read true if a card both
  matched the grant and were cast via a *different* alternative cost (e.g. evoke). This
  ambiguity already exists for printed warp + evoke and isn't introduced here; flagging only
  so it's on record. No real card hits it today.
- **Backlog `+2`.** `cards.md` checks off both Tannuk and Lightstall Inquisitor (255→257).
  Lightstall was already implemented on `main` with a stale unchecked box — the extra check
  is a correct drive-by, not an orphaned claim.

**Recommendation:** Ship it. No changes requested.

🤖 Reviewed against `origin/main` (already merged in; no merge commit needed) via the
review-changes skill.

wingedsheep:
## Follow-up: fixed a real bug this PR made reachable — committed `87486e1`

While verifying the earlier "minor footnote" about `wasWarped` disambiguation, I found it's an actual, reachable bug (my first review under-rated it — sorry). Reproduced it empirically before fixing.

### The bug
`CastSpell` carried only a single `useAlternativeCost: Boolean`. The handler resolved the actual cost through a **fixed priority chain** (`flashback > harmonize > warp > evoke > impending > self-alt > granted`). When two alternative costs are legal for the same card+zone, the two legal actions were **byte-identical** `CastSpell(useAlternativeCost = true)`, so the handler always picked the higher-priority one.

Tannuk's granted warp makes this reachable for a whole class of cards. With Tannuk out and a **red evoke creature in hand** (e.g. **Vibrance** `{3}{R/G}{R/G}` or **Catharsis** `{4}{R/W}{R/W}` — both already implemented; hybrid mana makes them red), clicking **Evoke** charged the **warp** cost and marked the spell warped:

```
Evoke action  → cost {R},    useAlt=true   ┐ identical CastSpell payload
Warp  action  → cost {2}{R}, useAlt=true   ┘
clicking Evoke → 3 lands tapped ({2}{R}), spellWarpedThisTurn=true   ← wrong
```

### The fix
Added `CastSpell.alternativeCostType` (`WARP/EVOKE/IMPENDING/FLASHBACK/HARMONIZE/SELF_ALTERNATIVE/GRANTED`). The legal-action enumerators stamp it on every alternative-cost action (all 24 emit sites across `CastSpellEnumerator` + `CastFromZoneEnumerator`); `CastSpellHandler` gates each branch of cost resolution, additional-cost validation, and the `wasWarped/wasEvoked/wasImpending` flags by the chosen type. `null` falls back to the original priority chain, so legacy/AI/hand-built actions are unaffected. Same shape `useWithoutPayingManaCost` already used for the same reason — CR 118.9a (only one alternative cost per cast) + CR 601.2b (the player announces which).

The client is unaffected (dumb terminal — it echoes the server's `action` JSON verbatim; no client field references these).

### Verification
- New `AlternativeCostChoiceScenarioTest`: distinct discriminators on the two actions; **Evoke charges `{R}`** (1 land), not warped; **Warp charges `{2}{R}`** (3 lands) and marks warped; legacy null-discriminator path still resolves.
- Full `rules-engine` suite green: **2906 tests, 0 failures** (incl. the existing warp/harmonize/impending/flashback/Jodah/Tannuk tests). All modules compile.

Files: `GameAction.kt` (+enum/field), `CastSpellHandler.kt` (gating), the two enumerators (stamping), docs (`engine-server-interface.md`, `card-sdk-language-reference.md`), + the new test.

🤖 via the review-changes → add-feature skills. Review worktree: `~/.claude/worktrees/argentum-engine/pr-480-review`.

PR #481, opened 2026-06-05
Add Jandor's Ring to Arabian Nights with last-drawn discard cost

bolav:
Adds the Arabian Nights artifact "{6} — {2}, {T}, Discard the last card you drew this turn: Draw a card." Composing with existing primitives wasn't sufficient: the discard cost picks a specific entity (the *one* card you drew most recently this turn), which the existing `AbilityCost.Discard(filter, count)` can't express — its filter-and-count shape would force a player selection rather than auto-resolving on the tracked entity. New SDK seam:

- `AbilityCost.DiscardLastDrawnThisTurn` — payable when the per-player tracker points at a card still in the controller's hand; unpayable otherwise (matches the Scryfall ruling "If you do not have the card still in your hand, you can't pay the cost"). Engine picks the entity automatically — no `discardChoices` round-trip — so the activation doesn't need a pre-payment selection UI flow.

- `GameState.lastCardDrawnThisTurnByPlayer: Map<EntityId, EntityId>` — updated at every `CardsDrawnEvent` emit site during a turn (the *last* id in a multi-card batch wins, per the second Jandor's Ring ruling "If you draw more than one card due to a spell or ability, you must discard the last one of those drawn") and cleared in `TurnManager.startNewTurn`. `DrawLoop` (both success and pause paths) and `DrawRevealDiscardUnlessExecutor` write the field next to where they emit the event; mulligan and opening-hand draws deliberately do not touch it.

`CostHandler` (canPay + pay) and `ActivatedAbilityEnumerator` (both the top-level cost branch and the composite sub-cost branch — Jandor's Ring wraps the new cost in `Composite(Mana, Tap, DiscardLastDrawnThisTurn)`) consult `WarpGrants`-style: the gate marks the ability unaffordable when the tracker is empty or the entity has left the hand, so the UI still surfaces the action as a greyed-out entry rather than dropping it. DSL reference (§Costs) updated in the same change per the project rule.

Scenario tests cover all four behaviours the rulings hinge on: happy path (cost discards the tracked card and the replacement draw retargets the tracker at the freshly drawn card), both unpayable branches (no draw this turn; tracked card has left the hand), the draw step populating the tracker via its `CardsDrawnEvent`, the multi-card-draw rule (`Effects.DrawCards(2)` spell → tracker holds the second-drawn id → ring discards that one, not the first), and the turn-boundary reset clearing the per-player entry.

### DIFF
--- a/backlog/sets/arabian-nights/TODO.md
+++ b/backlog/sets/arabian-nights/TODO.md
@@ -15,9 +15,9 @@ Set scaffolding is done:
 
 Verify status anytime with: `scripts/card-status --set ARN` (and `--list --set ARN`).
 
-## Progress (as of 2026-06-02)
+## Progress (as of 2026-06-05)
 
-**Implemented: 56 / 78** (`scripts/card-status --set ARN`). Full per-card status lives in
+**Implemented: 59 / 78** (`scripts/card-status --set ARN`). Full per-card status lives in
 `cards.md` — that file's `Implemented:` count is the source of truth; keep it in sync as boxes
 are checked. The triage buckets below are pruned to **remaining work only**; a card is removed
 from its bucket once it lands on a branch.
@@ -30,6 +30,9 @@ from its bucket once it lands on a branch.
   this way" clause needs `SacrificeContinuation` to thread `updatedSacrificedPermanents`
   back through the resumer (the auto-sac path now captures snapshots, but the multi-land
   decision path does not).
+- `arn-jandors-ring` — `AbilityCost.DiscardLastDrawnThisTurn` + per-player
+  `GameState.lastCardDrawnThisTurnByPlayer` tracker (updated at every `CardsDrawnEvent`
+  emit site during a turn, cleared in `TurnManager.startNewTurn`) + Jandor's Ring.
 
 ## Data sources — do NOT hit the network
 
@@ -111,7 +114,6 @@ Remaining:
 - **Untap-cost lock**: Magnetic Mountain (blue creatures don't untap unless {4} each).
 - **Cumulative counters dealing damage**: Cyclone.
 - **Card-from-outside-the-game**: Ring of Ma'rûf (wish-style).
-- **Last-drawn-card tracking**: Jandor's Ring.
 - **Ante**: Jeweled Bird — needs the ante mechanic (often unsupported / house-ruled).
 - **Subgame**: Shahrazad — a full Magic subgame; the largest single feature here.
 - **Jihad** — multi-condition enchantment that references a chosen color/player. Implement
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 78 cards
 **Release Date:** December 17, 1993
-**Implemented:** 57 / 78
+**Implemented:** 59 / 78
 | Color | Count |
 |-------|-------|
 | White | 11 |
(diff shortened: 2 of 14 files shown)


wingedsheep:
## Review — Jandor's Ring / `DiscardLastDrawnThisTurn`

**Verdict:** Behavior is correct and the SDK shape is right. This is a genuinely novel cost (not over-specialization), the tracker plumbing is complete, and the rules/Oracle citations all check out. Mergeable — I pushed one small test-strengthening commit (`184b2d58f`, see below).

### What's good (keep it)
- **The new cost is a justified seam, not a one-card type.** `AbilityCost.DiscardLastDrawnThisTurn` reads state no existing primitive can (`lastCardDrawnThisTurnByPlayer`) and auto-resolves a *specific* entity rather than forcing the player-selection shape that `AbilityCost.Discard(filter, count)` would impose. It's parameterless, so the next card with the same notion reuses it as-is. Correct call to add a new `AbilityCost` here.
- **Tracker coverage is actually complete.** I grepped every `CardsDrawnEvent` emit site: the only in-turn ones are `DrawLoop` (success + pause) and `DrawRevealDiscardUnlessExecutor` — all three updated. `GameInitializer`/`MulliganHandler` are correctly skipped (pre-turn-1, and the turn-boundary clear neutralizes them anyway). Replacement/cycle draws route through `DrawLoop`, so they're covered too.
- **New `GameState` field is clean.** Engine-internal (not referenced in the view/client layer — no masking leak), mirrors `lastCastSpellColors`, defaulted to `emptyMap()` so save/undo serialization stays backward-compatible.
- **Rules + card data verified.** CR 121 *is* "Drawing a Card" (confirmed against the current Comprehensive Rules — 120 is Damage). Oracle text, `{6}` Artifact, collector #64, Dan Frazier, and both 2004-10-04 rulings all match Scryfall exactly. `just check-card-printing` is green (canonical correctly placed in ARN, the earliest real printing).
- **Test coverage is excellent** — happy path, both unpayable branches, draw-step population, the multi-card "last one wins" ruling, and the turn-boundary reset. All six pass locally.

### Issues

**Minor (already applied — `184b2d58f`).** The two "unpayable" scenarios assert only that no *affordable* activation is offered, but their comments claim the ability is still emitted as a greyed-out unaffordable entry. Since Jandor's Ring's cost is `Composite`, it does take the `costAffordable = false` path (not a `continue` drop), so that claim is true and worth locking in. I strengthened both tests to assert exactly one Jandor's Ring activation, unaffordable — matching the comment and the intended "visible but disabled" UX. This also dropped a redundant `as? ActivateAbility` cast (smart-cast covers it; was a compiler warning).

**Minor (consideration, not a request).** The tracker write — `state.copy(lastCardDrawnThisTurnByPlayer = ... + (player to drawnCards.last()))` — is duplicated at three driver sites. `DrawCardPrimitive.drawOne` is the single per-card chokepoint every real draw routes through, and it already maintains the sibling `CardsDrawnThisTurnComponent`, so the tracker would arguably be more robust there (future draw paths get it for free). The tradeoff is that it pulls a global `GameState` concern into an otherwise pure single-card mover and would rely on the turn-boundary clear to neutralize opening-hand/mulligan draws. The current explicit driver-level approach is defensible and well-documented; flagging only as a judgment call.

**Minor (forward-looking, not a regression).** `GraveyardAbilityEnumerator` has `else -> {}` catch-alls, so a hypothetical *graveyard-activated* ability carrying this cost would be treated as always-payable. No current card hits it, and sibling costs (`DiscardSelf`, `SacrificeSelf`, …) are equally unhandled there — so this is consistent with the existing pattern, just noting it for whoever adds the next graveyard cost.

### Recommendation
Merge as-is (with the pushed test commit). The centralization note is optional.

---
*Reviewed in a worktree with `origin/main` merged in (merged cleanly, no conflicts). Only the single test commit `184b2d58f` was cherry-picked onto the PR head and pushed — the review-side merge commit was not pushed.*

PR #506, opened 2026-06-06
Add Goad mechanic for LTR Gap 5 with Glóin, Dwarf Emissary

bolav:
Implements CR 701.15 (goad) as a reusable engine primitive plus the LTR card that exercises it. `GoadEffect` records the effect's controller in a `GoadedComponent(goaderIds: Set<EntityId>)`; the set semantics give CR 701.15d (same-goader re-goad = no-op) and CR 701.15c (multi-goader stacking) for free. `AttackPhaseManager` adds both goad attack requirements inline alongside the existing must-attack pass; `TurnManager.startTurn` drops the active player from every goader set (clearing the component when empty) for the "until your next turn" duration (CR 701.15a). Goaded creatures surface to the client as a `Goaded` badge listing goader names.


wingedsheep:
## Review (with `origin/main` merged in) — behavior ✅, SDK shape ✅

Goad is modeled exactly as CR 701.15 describes — a runtime *designation* (`GoadedComponent`), not a fake static ability — and the per-goader `Set<EntityId>` gives 701.15c stacking and 701.15d dedup for free. All four sub-rules (701.15a–d) were verified against the current Comprehensive Rules; the backlog's stale `701.41` was correctly fixed to `701.15`. Full cross-layer wiring is present (GameEvent → Serialization → ClientEvent → ClientStateTransformer badge). Nice work.

I pushed two follow-up commits to the branch (maintainer edit):

**`Fix goad review findings`**
- **Blocking (fixed):** the LTR card snapshot golden was never re-blessed, so `CardDefinitionSnapshotTest > LTR` was failing on the unblessed Glóin. Re-blessed (`LTR.json`, +1 card only).
- **Important (fixed):** `validateGoadedRequirements` counted non-goader *planeswalkers* as a satisfying alternative, but CR 701.15b's second requirement is about attacking a *player*. Narrowed the "if able" pool to opponent players only — this also removed an untested, structurally-unreachable branch. The chosen-defender check still maps a chosen planeswalker to its controller, so attacking a goader's planeswalker is caught when an unaffected player was available.

**`mtgish: make Glóin auto-generatable`**
- Since the corpus already contains Glóin's full IR, wired the bridge + emitter so it renders as a complete **AUTO** draft. Each mapping is general-purpose (helps other cards too): `GoadCreature → Effects.Goad`, `TriggerOnceEachTurn → oncePerTurn = true`, `WhenAPlayerCastsASpell{You,IsHistoric} → Triggers.YouCastHistoric`, `CreateTokens[TreasureToken] → Effects.CreateTreasure`, and the `IsArtifactType` sacrifice-cost filter. The emitted draft is essentially identical to the hand-authored card.

**Verification:** `GoadEffectTest` (8/8) green after the planeswalker change; the re-blessed snapshot suite passes; the emitted Glóin draft compiles and scores AUTO with **0 capability mismatch** vs the golden; POR calibration stays 100% and cross-set fidelity is unchanged.

**Heads-up (not from this PR):** the whole-set `just coverage-verify LTR` gate is red on an *unrelated, pre-existing* emitter bug — `Ioreth of the Healing House` emits `Effects.UntapEachTarget()`, which doesn't exist in the SDK (only `TapEachTarget`). Worth a separate fix to the `TapEachPermanent` handler; it does not affect Glóin.

PR #524, opened 2026-06-07
Add Ydwen Efreet to Arabian Nights

bolav:
Whenever it blocks, flip a coin; on a loss it leaves combat and can't block this turn, and any attacker it was sole blocker of becomes unblocked — the pre-modern remove-from-combat behavior the oracle text preserves over modern CR 509.1h. Adds an `unblockSoleBlockedAttackers` flag to RemoveFromCombatEffect and threads it through CombatRemovalHelper so the override is opt-in.

### DIFF
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 78 cards
 **Release Date:** December 17, 1993
-**Implemented:** 59 / 78
+**Implemented:** 60 / 78
 | Color | Count |
 |-------|-------|
 | White | 11 |
@@ -58,7 +58,7 @@
 - [ ] Magnetic Mountain
 - [x] Mijae Djinn
 - [x] Rukh Egg
-- [ ] Ydwen Efreet
+- [x] Ydwen Efreet
 - [ ] Cyclone
 - [x] Desert Twister
 - [ ] Drop of Honey
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -518,7 +518,9 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 - `CantAttackGroupEffect(filter, condition?)` — group-scoped can't-attack.
 - `CantBlockGroupEffect(filter, condition?)` — group-scoped can't-block.
 - `Effects.Suspect(target)` — target becomes Suspected (MKM keyword). Composite: `SetSuspectedEffect` (named status, CR 701.60d dedup) + `GrantKeywordEffect(MENACE)` + `CantBlockEffect`.
-- `RemoveFromCombatEffect(target)` — yank target out of combat.
+- `RemoveFromCombatEffect(target, unblockSoleBlockedAttackers = false)` — yank target out of combat.
+  Set `unblockSoleBlockedAttackers = true` for the old-rules behavior (Ydwen Efreet): attackers the
+  target was sole blocker of become unblocked (CR 509.1h normally keeps them blocked).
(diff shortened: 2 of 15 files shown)


wingedsheep:
## Review

**Verdict.** Behavior is correct and well-tested; the rules override is the right call and the CR citation checks out. Mergeable essentially as-is.

### What's good
- **Tight, opt-in scope.** `unblockSoleBlockedAttackers` defaults to `false`, so every existing `RemoveFromCombat` site (Gustcloak, the CR 506.4 control-change SBA) is untouched. Staying-blocked stays the default, as CR 509.1h requires.
- **The rules corner is actually tested.** The fourth test ("with a second blocker… keeps the attacker blocked by the other blocker") exercises the precise *sole blocker* distinction that makes the `updatedBlockerIds.isEmpty()` check load-bearing. Win/lose are both forced and the lose case verifies real downstream damage. All four pass locally.
- **CR citation verified.** `509.1h` is exactly right — confirmed against the 2026-04-17 Comprehensive Rules: *"A creature remains blocked even if all the creatures blocking it are removed from combat."* Code, KDoc, and PR body all cite it consistently.
- **Printing placement correct.** `check-card-printing` is clean: canonical in ARN (earliest real printing), the ME1 reprint is unscaffolded so needs no row.
- **Complete cross-layer plumbing.** SDK effect → `Effects` facade → executor → helper all thread the flag; the docs reference and the ARN snapshot are updated in the same PR.

### Notes (non-blocking)
- **Boolean param vs. composition.** Adding `unblockSoleBlockedAttackers` to `RemoveFromCombatEffect` is the right shape here, even though we usually prefer composing atomic effects: the effect *already* rewrites the blocked attackers' `BlockedComponent` in its `isBlocking` branch, so this toggles a variation of behavior it already owns. The alternative (a new `UnblockAttackersSoleBlockedBy` effect + executor) would be *more* SDK surface duplicating logic the helper already runs.
- **"this combat" simplification.** The impl keys off the *current* blocker set (`updatedBlockerIds.isEmpty()`) rather than the set at the moment of blocking. These diverge only in a contrived case (Ydwen + X block, X leaves earlier, then Ydwen's flip is lost). Vanishingly rare and there's no per-combat "became blocked by" history to key off — fine to leave; an optional one-line comment at the check would help the next reader.

---

## Changes pushed (commit `ea22f94`)

Added **mtgish-tooling auto-generator coverage** for this card. Before, the generator classified Ydwen Efreet as **BLOCKED** (three unmapped IR capabilities). Now that the engine carries the `unblockSoleBlockedAttackers` flag, I wired them through both dictionaries:

**Bridge (capability):** `WhenACreatureBlocks` trigger · `RemoveCreatureFromCombatAndUnblockBlockers` → `RemoveFromCombat` · `FlipACoin_OnLose` → `FlipCoin`

**Emitter (rendering):** `WhenACreatureBlocks` → `Triggers.Blocks` · `RemoveCreatureFromCombatAndUnblockBlockers` → `RemoveFromCombat(..., unblockSoleBlockedAttackers = true)` · `CreatePermanentRuleEffectUntil` → `Effects.CantBlock`/`CantAttack` (CantBlock/CantAttack + EOT only) · `FlipACoin_OnLose` → `FlipCoinEffect(lostEffect = ...)`

Each is a reusable per-action mapping, not a one-off recognizer.

**Verification:** `coverage --card "Ydwen Efreet"` → **COVERABLE-NOW**; `coverage-fidelity --emit` → tier **AUTO** (the emitted tree matches the hand-authored card — `Effects.Composite(...)` serializes the same as `.then(...)`); `:mtgish-tooling:test` green incl. `EmitterGoldenTest`; **POR calibration 200/200 (100%)** — no calibrated-set output changed.

> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #539, opened 2026-06-07
Add Moonlit Meditation; generalize Mirrormind Crown's token-copy replacement to Auras

bolav:
Adds Moonlit Meditation (EOE #69, {2}{U} Enchantment — Aura): the first time you'd create one or more
tokens each turn, you may instead create that many copies of the enchanted permanent.

Rather than introducing a parallel SDK type, this generalizes the existing Mirrormind Crown primitive:

- SDK rename: ReplaceTokenCreationWithEquippedCopy → ReplaceTokenCreationWithAttachedCopy, with a new
attachmentVerb ("equipped" / "enchanted" / …) for display. The handler reads AttachedToComponent, so
both Equipment and Auras (and any future Fortification, etc.) share one code path. Attachment-type
validation stays at cast/attach time on equipmentTarget / auraTarget.
- Engine fix: token copies only receive SummoningSicknessComponent when the copy itself is a creature
(CR 302.6). Previously harmless for Mirrormind Crown (always equipping a creature), but matters now
that an Aura can copy an artifact.

Tests

- MoonlitMeditationScenarioTest — yes-replace, no-decline, once-per-turn-after-decline, and the
non-creature (Sol Ring) copy path.
- MirrormindCrownScenarioTest — regression coverage for the equip path of the renamed primitive.
- CardDefinitionSnapshotTest reblessed for ECL and EOE.

Docs

- card-sdk-language-reference.md updated alongside the SDK rename per repo convention.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -2703,6 +2703,16 @@ replacementEffect {
   "Devour land N" wording. **Scope today:** only the stack-spell entry path is wired; reanimation and
   token entries skip Devour (which is fine for printed cards — Devour creatures all cost real mana to
   cast).
+- `ReplaceTokenCreationWithAttachedCopy(optional, oncePerTurn, attachmentVerb, appliesTo)` —
+  "the first time you would create one or more tokens each turn, you may instead create that
+  many tokens that are copies of [attached] permanent." Works for both Equipment and Auras —
+  the engine reads the source's `AttachedToComponent` to find the permanent to copy.
+  `optional = true` surfaces a yes/no during resolution; `oncePerTurn = true` adds
+  `TokenReplacementOfferedThisTurnComponent` after the first offer (cleared at end of turn).
+  `attachmentVerb` is a display-only label ("equipped", "enchanted", "fortified") — the
+  attachment-type validation already happens at cast/attach time via `equipmentTarget` /
+  `auraTarget`. Token copies are summoning-sick only when the copy is a creature (CR 302.1).
+  Mirrormind Crown: `attachmentVerb = "equipped"`; Moonlit Meditation: `attachmentVerb = "enchanted"`.
 - `EntersAsCopy(optional, copyFilter, copyFromZone, filterByTotalManaSpent, additionalSubtypes, additionalKeywords, nameOverride, powerOverride, toughnessOverride, exileCopiedCard)` —
   "enter as a copy of …". As the permanent resolves, the controller picks an object matching
   `copyFilter` and the permanent enters as a copy (Rule 707 copiable values), with any overrides
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -2711,7 +2711,7 @@ replacementEffect {
   `TokenReplacementOfferedThisTurnComponent` after the first offer (cleared at end of turn).
   `attachmentVerb` is a display-only label ("equipped", "enchanted", "fortified") — the
   attachment-type validation already happens at cast/attach time via `equipmentTarget` /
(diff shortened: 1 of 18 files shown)


wingedsheep:
## Review — Add Moonlit Meditation; generalize the token-copy replacement to Auras

**Verdict:** Behavior is correct and the SDK shape is right. This is the model way to add a card — instead of a parallel `ReplaceTokenCreationWithEnchantedCopy`, it generalizes the existing `ReplaceTokenCreationWithEquippedCopy` into `ReplaceTokenCreationWithAttachedCopy`, driven entirely by `AttachedToComponent`, with a display-only `attachmentVerb`. One primitive now covers Equipment, Auras, and any future Fortification. I pushed two small fixups (commit `c5efb42a4`) and made Moonlit Meditation auto-generatable from the mtgish IR.

### What's good (keep it)
- **Generalization over duplication.** The rename + `attachmentVerb` param is exactly right; the handler reads `AttachedToComponent` and attachment-type validation stays at cast/attach time via `auraTarget`/`equipmentTarget`. No behavioral re-check duplicated in the replacement path.
- **The summoning-sickness fix is a genuine catch.** Only adding `SummoningSicknessComponent` when the token copy is a creature (CR 302.6) is correct and was latent-wrong before (harmless only because Mirrormind Crown always copied a creature). The Sol Ring test pins the non-creature axis.
- **Test coverage is thorough and on the right axes:** yes/no/once-per-turn-after-decline + non-creature copy, plus a dedicated `MirrormindCrownScenarioTest` regression for the renamed primitive's `equipped` path. All 6 pass locally.
- **Printing placement** verified — `just check-card-printing "Moonlit Meditation"` is green (EOE canonical, `peoe` promo correctly skipped). Oracle text matches Scryfall exactly. Card auto-registers via `CardDiscovery` package scan.

### Issues

**Minor (fixed in `c5efb42a4`)** — `TokenCreationReplacementHelper.createAttachedPermanentCopies` KDoc cited summoning sickness as *CR 302.1*; the rule is *CR 302.6* (302.1 is just "casting a creature card"). The inline comment in the loop already had 302.6 right — verified against the 2026-04-17 Comprehensive Rules.

No blocking or important issues. Projection isn't a concern here: the token copies the *printed* (copiable) characteristics of the attached permanent, so reading the base `typeLine.isCreature` of the freshly-copied card at creation time is correct per Rule 707.

### Bonus: mtgish auto-generation (also in `c5efb42a4`)
Per the request, I wired this card family into `:mtgish-tooling`:
- **Bridge:** mapped the mtgish `ReplaceAPlayerWouldCreateTokens` rule to `supported` (the engine's `ReplaceTokenCreationWithAttachedCopy`). Moonlit Meditation flips **BLOCKED → COVERABLE-NOW**.
- **Emitter:** renders the *first-time-each-turn / you-may / copy-the-host* shape to `replacementEffect(ReplaceTokenCreationWithAttachedCopy(attachmentVerb = …))` (verb from the attachment subtype), declining any other token-replacement shape to a scaffold. Also widened `auraTargetBlock` to render "enchant artifact or creature you control" as `TargetPermanent(TargetFilter.CreatureOrArtifact.youControl())`.
- **Result:** `just coverage-fidelity --emit "Moonlit Meditation"` now renders the card **whole (fidelity AUTO)** — the generated DSL matches the hand-authored definition. POR `EmitterGoldenTest` still matches card-for-card, and `:mtgish-tooling:test` is green.
- **Note:** *Mirrormind Crown* now shows the token-copy capability as `[supp]`, but stays BLOCKED on the unmapped `Equip` keyword and an `If` "as long as attached" wrapper — both generic tooling gaps unrelated to this feature, left out of scope.

**Recommendation:** Merge as-is. The fixups are already on the branch.

PR #553, opened 2026-06-07
Add Ali from Cairo with LifeLossFloor replacement effect

bolav:
Adds the new `LifeLossFloor` replacement (mtg-sdk) for "damage that would reduce your life total to less than 1 reduces it to 1 instead" — a damage-only life-loss cap parameterized by floor amount, restrictions, and the protected player. Wired into the damage→life-loss pipeline (CR 120.3a) in DamageUtils and CombatDamageManager; LoseLifeExecutor deliberately skips it so non-damage life loss bypasses the floor (per the printed ruling). The damage event still fires at full amount, so lifelink and damage-dealt triggers see the unchanged amount. Same primitive will later power Worship with a `YouControlACreature` restriction.

DamageUtils' `applyStaticLifeLossModification` (ModifyLifeLoss / Bloodletter) and the new `applyLifeLossFloors` share a private inline reified `forEachLifeLossReplacement<T>` scaffold for the battlefield scan, player-match, and restriction-gate boilerplate.

Drive-by: corrected CR 119.3 → CR 120.3a in pre-existing damage→life-loss comments in DamageUtils.kt, CombatDamageManager.kt, and BloodletterOfAclazotz.kt; updated the lifelink comment in DamageUtils.kt to cite CR 120.3f / 702.15b.

Scenario tests cover: lethal damage floored to 1, non-lethal unchanged, opponent unprotected, direct life-loss (LoseLifeEffect) bypasses the floor, post-leaves-battlefield no-op, lifelink sees the unfloored damage amount, and Ali still protecting his controller when both take lethal damage in the same event.

### DIFF
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 78 cards
 **Release Date:** December 17, 1993
-**Implemented:** 60 / 78
+**Implemented:** 61 / 78
 | Color | Count |
 |-------|-------|
 | White | 11 |
@@ -50,7 +50,7 @@
 - [x] Stone-Throwing Devils
 - [x] Aladdin
 - [x] Ali Baba
-- [ ] Ali from Cairo
+- [x] Ali from Cairo
 - [x] Bird Maiden
 - [x] Desert Nomads
 - [x] Hurr Jackal
--- a/backlog/sets/lord-of-the-rings/cards.md
+++ b/backlog/sets/lord-of-the-rings/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 291 cards (261 Draft / 30 Extra)
 **Release Date:** June 23, 2023
-**Implemented:** 197 / 291
+**Implemented:** 198 / 291
 Run `scripts/card-status --set LTR` (and `--list --set LTR`) to verify status at any time.
 The split below mirrors that script: **Draft** = Scryfall `booster: true`; **Extra** =
 starter-deck/special cards and basic lands.
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -2744,6 +2744,16 @@ replacementEffect {
   instances stack (×s multiply, +s sum) — two Leylines of Hope add 2 to every life-gain event.
 - `ModifyLifeLoss(multiplier, modifier, restrictions, appliesTo)` — same shape as `ModifyLifeGain` for life loss
   events (`LifeLossEvent`), plus a `restrictions: List<Condition>` list that further gates the replacement.
+- `LifeLossFloor(floor, restrictions, appliesTo)` — cap damage-induced life loss so the resulting life total
+  is ≥ `floor`. `appliesTo` is a `LifeLossEvent` whose `player` filter gates who is protected (default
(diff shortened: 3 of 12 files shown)


wingedsheep:
## Review — `LifeLossFloor` / Ali from Cairo

**Verdict:** Behavior is right and the SDK shape is right. `LifeLossFloor` is a genuinely novel replacement (it floors the *resulting life total*, which needs `currentLife` — not expressible as the affine `multiplier/modifier` of `ModifyLifeLoss`), parameterized for the next card rather than hardcoded to Ali, and correctly scoped to the damage→life-loss path. No code changes needed from my side — verified locally and recommending merge.

### What's good
- **The shared `forEachLifeLossReplacement<T>` reified-inline scaffold** is a clean DRY win — it unifies the `ModifyLifeLoss` and `LifeLossFloor` battlefield scans (player-match + restriction gate) while each function keeps its own arithmetic in the callback. Right altitude.
- **Genuinely novel, and parameterized for the *next* card.** `floor`, `restrictions`, and `appliesTo` are all open, so Worship drops in as `LifeLossFloor(floor = 1, restrictions = listOf(YouControlACreature), appliesTo = LifeLossEvent(Player.You))` with no new type. Multiple floors pick the strictest order-independently.
- **Scope matches the printed ruling exactly.** Both damage call sites (`DamageUtils.dealDamageToTarget`, `CombatDamageManager`) apply the floor; the direct-loss path (`loseLife`/`LoseLifeExecutor`) deliberately skips it — and there's a test (`Drain Five`) proving direct life loss bypasses the floor.
- **Rule precision is excellent.** I downloaded the current CR and verified: `120.3a` (damage by a non-infect source → lose that much life) ✓, `120.3f`/`702.15b` (lifelink) ✓. The drive-by `119.3 → 120.3a` fixes in the damage comments are correct, and the *remaining* `119.3` in the `loseLife` doc is correctly left as-is (119.3 governs effect-based life change, the right citation for that path).
- **Tests map 1:1 to the three rulings** — lethal floored to 1, non-lethal unchanged, opponent unprotected, direct-loss bypass, leaves-battlefield no-op, lifelink/`DamageDealtEvent` seeing the *unfloored* 5, and the same-event Earthquake case (Ali still protects before SBAs bury him). All 7 pass locally; `CardDefinitionSnapshotTest` green (serialization auto-registers the new sealed subtype).
- **Printing placement verified** — `just check-card-printing "Ali from Cairo"` is clean: canonical lives in ARN (earliest real printing); the ME4 reprint is correctly left unscaffolded.

### Minor (no action required for this PR)
- **Pre-existing projection gap, not introduced here:** the shared scan resolves the replacement source's controller via base `ControllerComponent` (`DamageUtils.kt:1337`) rather than `state.projectedState.getController(...)`. Under a control-changing effect (a stolen Ali / Bloodletter), `Player.You`/`Player.Opponent` would track the printed controller instead of the current one. This idiom appears at ~5 sites in `DamageUtils`, so it's a codebase-wide pattern that deserves its own pass — out of scope for a card PR, flagging only so it's on the radar.
- **Multi-replacement ordering:** when both Bloodletter (`ModifyLifeLoss`) and Ali (`LifeLossFloor`) apply, the engine uses a fixed modify-then-floor order rather than letting the affected player choose per CR 616.1. The fixed order happens to be the survival-favorable one, and player-ordered replacements are a broader engine limitation — noting, not blocking.

`origin/main` merged cleanly into the review branch (no conflicts), so this reflects post-merge reality.

PR #565, opened 2026-06-08
EOE: add Tezzeret + Terminal Velocity, bug-fix sweep

bolav:
## Summary
- Implement the last two unimplemented EOE cards: Tezzeret, Cruel Captain
  (planeswalker, with a -7 emblem that puts +1/+1 counters on a target
  artifact and turns it into a 0/0 Robot artifact creature) and
  Terminal Velocity (cheat-into-play sorcery granting haste + LTB ping +
  end-step sacrifice). EOE is now 261/261.
- Engine fixes surfaced while playing the set:
  - CR 608.2b alignment in StackResolver / EffectContext / DealDamageExecutor —
    a sub-effect that references a target dropped at resolution now fizzles
    instead of shifting onto the next surviving target (Diplomatic Relations).
  - CR 704.5p enforcement in UnattachedAurasCheck — an Equipment that becomes
    a creature (Tezzeret's emblem on Atomic Microsizer) auto-unattaches.
  - Two-step UI pause for ActivateAbility costs that need a player choice:
    TapXPermanents (Secluded Starforge "Tap X artifacts") raises
    ChooseNumber → SelectCards instead of silently activating with X=0;
    ExileFromGraveyard (Rust Harvester) prompts the player to pick which
    graveyard card to exile instead of auto-picking the first match.
- Per-card bug fixes: Auxiliary Boosters attaches to its freshly-created
  Robot token (PipelineTarget instead of ContextTarget), Kavaron Harrier
  token is sacrificed rather than exiled at end of combat, Glacier Godmaw
  Landfall uses EffectTarget.Self inside ForEach, Larval Scoutlander gets an ETB regression test (its sacrifice gates the library search).
- Card metadata backfill (artist / flavorText / imageUri) on 14 EOE cards
  flagged by the set audit, plus Thrumming Hivepool's Sliver token image.

## Test plan
- [x] `just test` — full suite green
- [x] New scenario tests pass: DiplomaticRelationsPartialIllegalTargetTest,
      EquipmentAsCreatureUnattachTest, TezzeretCruelCaptainScenarioTest (5),
      TerminalVelocityScenarioTest (2), AuxiliaryBoostersAttachOnEtbTest,
      KavaronHarrierTest, LarvalScoutlanderScenarioTest,
      RustHarvesterExileCostChoiceTest, SecludedStarforgeTest UI-flow case
- [x] `just check-card-printing` clean for both new cards (canonical EOE)

### DIFF
--- a/backlog/sets/edge-of-eternities/bugs.md
+++ b/backlog/sets/edge-of-eternities/bugs.md
@@ -1,14 +1,18 @@
 - Kavaron Harrier (exile, not sacrifice)
 - Diplomatic Relations gives +1/+0 to TO creature if FROM creature is dead
 - Secluded Starforge: 2,T: can not tap artifacts.
-- Chrome Companion: should keep the bottom of the library visible.
-- Intrepid Tenderfoot, stacks pumps, but it's sorcery.
 - Rust Harvester, cannot choose from graveyard, card gets picked automatically
 - Rust Harvester, +1/+1 counter is not applied if target is invalid
-- Wedgelight Rammer did not turn to creature
-- Spacecraft can attack 1 turn after being played
 - Auxiliary Boosters - equipment not auto-attached
-- Glacier Godmaw - Landfall not working
 - Larval Scoutlander - Lands did not come to the battlefield, deck was not shuffled
 - Artifact Equipment attacks, gives separate damage
   - Atomic Microsizer 3/3 attached to Chrome Companion - Tezzeret, Cruel Captain
+
+Not bugs:
+- Chrome Companion: should keep the bottom of the library visible.
+- Intrepid Tenderfoot, stacks pumps, but it's sorcery.
+
+Fixed:
+- Wedgelight Rammer did not turn to creature
+- Spacecraft can attack 1 turn after being played
+- Glacier Godmaw - Landfall not working
--- a/backlog/sets/edge-of-eternities/bugs.md
+++ b/backlog/sets/edge-of-eternities/bugs.md
@@ -1,6 +1,4 @@
-- Secluded Starforge: 2,T: can not tap artifacts.
 - Rust Harvester, cannot choose from graveyard, card gets picked automatically
-- Rust Harvester, +1/+1 counter is not applied if target is invalid
 - Auxiliary Boosters - equipment not auto-attached
 - Larval Scoutlander - Lands did not come to the battlefield, deck was not shuffled
 - Artifact Equipment attacks, gives separate damage
--- a/backlog/sets/edge-of-eternities/bugs.md
+++ b/backlog/sets/edge-of-eternities/bugs.md
@@ -1,4 +1,3 @@
-- Kavaron Harrier (exile, not sacrifice)
 - Diplomatic Relations gives +1/+0 to TO creature if FROM creature is dead
 - Secluded Starforge: 2,T: can not tap artifacts.
 - Rust Harvester, cannot choose from graveyard, card gets picked automatically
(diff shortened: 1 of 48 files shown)


wingedsheep:
## Review — SDK elegance + correctness pass

**Verdict:** Behavior is correct and the SDK shape is right. Both new cards compose existing primitives with no card-specific types; the three engine fixes (CR 608.2b partial-illegal-target fizzle, CR 704.5p equipment-becomes-creature unattach, two-step cost-choice pauses) are well-targeted and fully plumbed (serialization + resumer registration). All 13 new scenario tests pass, the card snapshot is green, `check-card-printing` is clean for both new cards, and `damageSource` regressions still pass. CR citations verified against the current Comprehensive Rules (2026-04-17).

### What's good
- **608.2b fix is elegant.** `buildAlignedValidated` keeps the compacted drop-illegal list every executor has always seen, but threads a *positionally-aligned, null-padded* parallel list into `buildNamedTargets` so a `BoundVariable` reference to a dropped target resolves to `null` and fizzles instead of sliding onto the next survivor. Identity (`===`) alignment is safe because `validateTargets` is a `filter` that preserves instances and ordering.
- **704.5p fix** asks creatureness of the *projected* state (layer-4 type change), checks `RECONFIGURE`, cleans up the reverse attachment link, and the CR renumber (704.5m Aura / 704.5n Equipment / 704.5p creature-attached) is correct.
- The three cost-choice continuations are genuinely novel SDK surface (a player-interaction pause no executor produced), all `@Serializable`, registered in `engineSerializersModule` + `ContinuationHandler`, with subset/count validation in the resumers.
- `GameObjectFilter.ArtifactCreature` is a clean two-predicate composition, not a bespoke type.
- Tests exercise the real rule corners: FROM dying in response, equipment-becomes-creature with full board reprojection, X=0 vs X=3 tap-selection branches, and the `may`-decline path.

### Nits — fixed in follow-up commit (pushed to this branch)
I pushed `bcd4c55` addressing the minor issues:

1. **`SbaOrder.kt` CR citation collision** — `COUNTER_ANNIHILATION` is CR **704.5q**, not 704.5m; the (correct) Unattached-Auras relabel to 704.5m had created a duplicate. Fixed the comment.
2. **Duplicated candidate-finders.** The handler's `findUntappedTapXCandidates` / `findGraveyardExileCandidates` mirrored payment-side logic, so the pause could in principle offer a different set than payment accepts. Removed both; the handler now calls `CostHandler.findUntappedMatchingPermanentsUnified` and `findMatchingCardsUnified` (now `internal`) — one source of truth, no drift.
3. **`DealDamageExecutor`** returned `EffectResult.error` for a legal 608.2b fizzle; changed to `success(state)` (no-op skip, surrounding `Composite` resolves) with a comment.
4. **Test comment** dropped the stale `†` on the Diplomatic Relations collector-number reference (the card's `collectorNumber` was corrected to `177` in this PR).

(Also touched the PR description to reflect that the Larval Scoutlander change is a regression test, not a code fix.)

Re-ran the affected suites after the de-dup + fizzle change — all green (`SecludedStarforgeTest` incl. the X=3 tap flow, `RustHarvesterExileCostChoiceTest`, `DiplomaticRelationsPartialIllegalTargetTest`, plus `LongstalkBrawlTest`/`UncontrolledInfestationTest` for the normal `damageSource` path).

Heads-up: I also merged `origin/main` into the branch (clean, no conflicts) so the review reflected post-merge reality — that merge commit is included in the push.

**Recommendation:** Mergeable.

PR #580, opened 2026-06-08
LTR Cluster 1: six cards + per-turn history SDK primitives

bolav:
Adds six LTR cards plus three small engine primitives the cluster needed.

**Cards (all canonical in LTR per `check-card-printing`):**
- **Nasty End** — additional-cost sacrifice; draw 2, or draw 3 if it was legendary.
- **Gríma Wormtongue** — "Your opponents can't gain life"; tap+sac edict that amasses Orcs 2 on a legendary sac.
- **Rise of the Witch-king** — symmetric edict; reanimate a permanent card from your graveyard if you sacrificed.
- **Shortcut to Mushrooms** — Ring-tempt ETB + per-turn-LTB-gated end-step +1/+1.
- **Samwise the Stouthearted** — flash; optional graveyard return + Ring tempts (fires on decline).
- **Lobelia Sackville-Baggins** — flash + menace; exile opponent's recent-grave creature for X Treasures.

**New SDK / engine plumbing:**
- `PermanentSnapshot.supertypes` + `SacrificedWasLegendary` / `YouSacrificedThisWay` conditions; `ForceSacrificeExecutor` snapshots flow through `SacrificeContinuation` into the
  surrounding `EffectContext`.
- `PermanentLeftBattlefieldThisTurnComponent` per-player tracker + `Conditions.YouHadPermanentLeaveBattlefieldThisTurn` (broader than `ControlledCreatureDiedThisTurn` — counts
lands, tokens, any destination).
- `PutIntoGraveyardFromBattlefieldThisTurnMarker` per-entity marker + `StatePredicate.PutIntoGraveyardFromBattlefieldThisTurn` (wiped at each untap step).
- `CreatePredefinedTokenEffect.dynamicCount: DynamicAmount?` for Lobelia's "X = exiled power".

**Engine fixes surfaced by the new cards:**
- `EffectAndTriggerContinuationResumer` — "up to N target" abilities with no `elseEffect` now still put the ability on the stack with `targets=[]` so sibling effects (Samwise's
`TheRingTemptsYou()`) execute (CR 115.6).
- `DamageUtils.isLifeGainPrevented` — `Player.EachOpponent` was silently falling through the `when`, so Gríma's static never fired. Same pairing as the sibling code at line
1353.

**Tests:** 23 scenarios across 8 test classes; both branches of every conditional covered (legendary vs nonlegendary sac, choose vs decline, intervening-if true vs false, marker
  stripped/wiped, regression on Sunspine Lynx). All green on a clean build. `docs/card-sdk-language-reference.md` updated.


wingedsheep:
## Review — LTR Cluster 1 (six cards + per-turn history SDK primitives)

Reviewed with `origin/main` merged in (clean, no conflicts). All six Oracle texts verified exact against Scryfall; CR citations 115.6 / 603.4 / 400.7 verified against the 2026-04-17 Comprehensive Rules; printing placement correct for all six (LTR canonical). I pushed one commit (`880acd`) addressing the findings below — full module suite green.

### Verdict
Behavior correct for five of six cards and the SDK shape is genuinely good (the new conditions/predicate/component read state no existing primitive can, and they're parameterized rather than card-specific). One real rules bug on Rise of the Witch-king, now fixed.

### What's good (kept as-is)
- **SDK primitives are correctly novel, not over-fitted** — `PermanentSnapshot.supertypes` (generic), `PermanentLeftBattlefieldThisTurn(player)` (takes a `Player`, not baked to "you"), and the graveyard marker all read state no `Modification`/`DynamicAmount` could express.
- **Thorough defensive predicate handling** — `PutIntoGraveyardFromBattlefieldThisTurn` returns `false` in all four non-graveyard contexts (battlefield projection, untap filter, trigger gating, marker-stripped-on-entry), each with a why-comment.
- **The CR 115.6 resumer fix** (zero-target ability still resolves siblings so Samwise's Ring temptation fires on decline) is correct and well-scoped.
- **Tests exercise the rule corners**: both legendary/nonlegendary sac branches, marker stripped-on-leave / wiped-next-turn / mill-doesn't-match, decline-target-still-tempts.

### Fixed in `880acd`

**Blocking — Rise of the Witch-king reanimated the creature you just sacrificed.** The gather pulled `FromZone(GRAVEYARD, You, Permanent)` with no exclusion, so the sacrificed creature (now in your graveyard) was a legal reanimation choice. CR ruling is explicit: *"You cannot return the same permanent card that you sacrificed"* (oracle: "return **another** permanent card"). Fixed by adding `CardSource.FromZone.excludeSacrificedThisWay` — drops entities whose IDs are in `EffectContext.sacrificedPermanents` (the snapshots already carry the entity id), wired in `GatherCardsExecutor`. Reusable for any future edict-then-reanimate card.

**Important — missing `YouSacrificedThisWay` false-branch test.** Both existing Rise tests gave the active player a creature, so the condition was always true. Added two cases: (1) the sacrificed creature is **not** among the `SelectCardsDecision.options` while a separate graveyard permanent **is**, and (2) the rider does not fire when you control no creature.

**Minor — docs/citations.**
- Documented `CreatePredefinedTokenEffect.dynamicCount` and the new `excludeSacrificedThisWay` flag in `card-sdk-language-reference.md`.
- Lobelia: now cites **CR 400.7j** (an effect can find the object it became in the public zone it moved to) instead of 400.7, whose main clause says the opposite.

Re-blessed the LTR card snapshot golden for the new gather field.

### Not changed (intentional)
- The `Player.Any` arm added to `DamageUtils.isLifeGainPrevented` is unused by this PR's cards but is correct ("any player can't gain life" → prevent for all); left in place.
- The "sibling code at line 1353" reference is in the PR description prose (not code), so nothing to fix in-tree — worth dropping the line number from the description though.

PR #589, opened 2026-06-09
Add three LTR Sagas (Scroll of Isildur, There and Back Again, Long List of the Ents)

bolav:
Implements three LTR Sagas, plus the two engine primitives they need and one delayed-trigger extension. Also fixes a latent Threaten bug on Aladdin (ARN).

Cards
- Scroll of Isildur ({2}{U}) — I: gain control of an artifact for as long as you control this Saga + Ring tempts; II: tap up to two creatures, stun-counter each; III: draw a
card per tapped creature target opponent controls.
- There and Back Again ({3}{R}{R}) — I: target creature can't block for as long as you control this Saga + Ring tempts; II: tutor a Mountain to the battlefield; III: create
Smaug (legendary 6/6 Dragon, flying/haste, "When Smaug dies, create fourteen Treasures").
- Long List of the Ents ({G}) — six chapters that each note a new creature type and install a one-shot delayed trigger adding a +1/+1 counter to the next creature spell of that
type cast this turn.

SDK / engine additions
- Duration.WhileYouControlSource(sourceDescription) — source-side mirror of WhileControlledByController. StateProjector gates per-frame after Layer 2 via a new
sourceControllerGate; EndedDurationExpiryCheck physically removes the effect when the source's projected controller drifts or it leaves play (CR 611.2b one-way). Also fixes
Aladdin, which previously used WhileSourceOnBattlefield and would not release a stolen artifact if Aladdin himself was Threatened.
- Effects.NoteCreatureType(storeAs) + per-permanent NotedCreatureTypesComponent — "note a creature type that hasn't been noted for this ". Dedup is dynamic against the source's
current noted set; persistence lives on the source's container so it vanishes with it per CR 400.7.
- CreateDelayedTriggerExecutor.bakeChosenValuesIntoTrigger — substitutes CardPredicate.HasSubtypeFromVariable inside a SpellCastEvent filter at trigger-creation time, so the
delayed trigger spec is self-contained once the originating EffectContext has evaporated.

Tests — 11 new scenario tests covering the one-way latch (CR 611.2b), Threaten-the-source case, per-source independence of noted sets, leaves-play cleanup, type-mismatch trigger
  skip, and fireOnce semantics. LTR + ARN snapshot goldens re-blessed.

Backlog: LTR bumped to 207 / 291.


wingedsheep:
Reviewed and pushed a follow-up commit (d394bf3) addressing the three minor nits from review. No blocking or important issues were found — behavior is correct, all cited CR rules (611.2b, 400.7, 122.1d) verify, oracle text matches Scryfall exactly, printing placement is correct, and all original tests pass green.

**Fixes in d394bf3:**

1. **CR citation corrected (400.7, not 122.2).** The doc (`card-sdk-language-reference.md`) and the `NoteCreatureTypeEffectTest` header cited *CR 122.2* for `NotedCreatureTypesComponent` vanishing on zone change — but 122.2 is specifically about *counters* ceasing to exist. The component is not a counter; the correct rule (already cited in the component/`Duration` KDoc) is **CR 400.7** (a permanent that changes zones becomes a new object with no memory). Now consistent everywhere.

2. **Strengthened the "leaves play" test.** It previously only asserted the source had left the battlefield. It now re-enters the permanent and asserts the noted set starts fresh (Elf selectable again), actually exercising the 400.7 guarantee.

3. **Added the multi-chapter test the header described but didn't implement.** `DelayedTriggerFromNotedTypeTest` now has the two-activations case: two types noted, each installs its own one-shot delayed trigger, and casting one creature of each adds a counter to each — the heart of Long List of the Ents. (Writing it also confirmed the engine handles two independent same-source delayed triggers correctly; the only fix needed was draining the full stack between the two sorcery-speed casts in the test.)

**Praise worth keeping on a rewrite:** `WhileYouControlSource` is the right new primitive (it fills a genuine gap between `WhileSourceOnBattlefield` and `WhileControlledByController`), is wired through every duration-switch site, and the Aladdin latent-bug fix is a nice catch with its own Threaten-the-source test. `NoteCreatureType` + `NotedCreatureTypesComponent` is a clean, reusable, generically-parameterized primitive.

All 12 affected scenario tests pass; `CardDefinitionSnapshotTest` (LTR + ARN goldens) and `just check-card-printing` for all four cards are green.

PR #609, opened 2026-06-10
Add "opponent draws except first each draw step" trigger + Orcish Bow…

bolav:
…masters

Implements the Orcish Bowmasters draw-punisher clause as a reusable trigger primitive, then the two LTR cards that need it.

Engine/SDK:
- EventPattern.DrawEvent gains `exceptFirstInDrawStep`; new Triggers facades `OpponentDraws` and `OpponentDrawsExceptFirstEachDrawStep`.
- DrawEvent already fires once per individual card drawn (CR 121.2); the new flag subtracts the one card a player draws first in their own draw step (CR 504.1's turn-based draw). The exempt card is identified via a per-player draw-step-start snapshot (GameState.drawStepStartDrawCountByPlayer), set in DrawPhaseManager.performDrawStep and cleared each turn by TurnManager.
- TriggerMatcher.drawTriggerFiringCount computes the firing count; the TriggerDetector DrawEvent branch uses it instead of a raw event.count.

Cards:
- Orcish Bowmasters (the printed card folds enters + opponent-draws into one ability, modeled as two siblings sharing "deal 1 to any target, then amass Orcs 1").
- A-Orcish Bowmasters (Alchemy rebalance — drops the enters trigger).

Tests: OrcishBowmastersTest proves every clause — the for-turn draw is exempt, second+ draw-step draws fire, N draws outside the step fire N times, opponent draws on your turn fire, the controller's own draws never fire, the enters trigger (and the A- card's lack of one). Snapshot golden re-blessed.

Docs: card-sdk-language-reference.md updated; LTR TODO/cards/missing-effects refreshed (209/291).


wingedsheep:
## Review

**Verdict:** Behavior is correct and exceptionally well-tested, and the SDK shape is right — a defaulted boolean on the existing `DrawEvent` pattern plus thin `Triggers` facades, not a new card-specific event type. I found one blocking issue and one corner-case engine flaw; **both are fixed in the two commits I pushed to this branch** (`02998c5` merge of main + `ffb21d2`).

### What's good

- **SDK shape.** `DrawEvent(exceptFirstInDrawStep = true)` is the minimal-surface design: the clause is named for what it generically means, the snapshot state (`drawStepStartDrawCountByPlayer`) is genuinely novel (nothing else can identify "first card drawn in this draw step"), and it composes with the existing `Player` scoping.
- **Rules fidelity.** The snapshot-of-counts approach correctly implements the ruling that the exemption applies to the first card *actually drawn* in the step even when the turn-based draw is replaced or skipped — deliberately captured on the turn-1 skip path too, with a comment explaining why. Multiple draw steps per turn each get their own exemption.
- **Verification.** All cited CR numbers check out against the official rules text (121.2 individual draws, 504.1 turn-based draw, 603.2 trigger matching). Both cards' Oracle text matches Scryfall exactly. `check-card-printing` is clean for both (LTR is the earliest real printing).
- **Tests.** Every clause has a paired assertion. The two-sibling modeling of the printed card's single "enters and whenever…" ability is reasonable (the conditions can never fire from one event) and documented in the card file.

### Issues (all fixed in `ffb21d2`)

1. **Blocking — `AOrcishBowmasters.kt`: missing `inBooster = false`.** Scryfall has `booster: false` for the Alchemy rebalance, but `CardDiscovery` auto-registers everything in the `ltr.cards` package, so it would have entered the LTR draft/sealed booster pool as an extra rare. Fixed with `inBooster = false` (same pattern as the BLB extras); LTR snapshot golden re-blessed.

2. **Important — batch-boundary miscount in `drawTriggerFiringCount`.** It derived `countBefore = countAfter - event.count` from the *post-execution* state, but `detectTriggers` evaluates every event of an execution against that same final state. If a single resolution drew via two separate effects (e.g. `Composite(DrawCards(1), DrawCards(1))`) while the exempt slot was still open — turn-1 first player's draw step, or a replaced turn-based draw — the snapshot window missed every event and the trigger fired 2× instead of 1×. Fixed: `TriggerDetector` now passes each `CardsDrawnEvent` the number of same-player draws later in the batch, so the matcher reconstructs that event's own count boundary. Added a regression test (verified it fails without the fix).

3. **Minor — `Triggers.OpponentDraws` shipped with no user and no direct test.** Kept it (future Underworld Dreams-style cards want it) and added a test pinning its contrast with the exceptFirst variant: the for-turn draw *does* fire the plain version.

### Also note

- I merged `origin/main` into the branch (`02998c5`). `TriggerMatcher.kt` had an additive conflict — main's new `matchesTokenCreationTrigger` (Mirkwood Bats) and this PR's `drawTriggerFiringCount` landed in the same spot; resolved by keeping both.

### Test results (post-merge, post-fix)

- `OrcishBowmastersTest`: 9/9 (7 original + 2 new)
- Full `:rules-engine:test`: 3,418 tests, 0 failures
- `:mtg-sets:test` incl. `CardDefinitionSnapshotTest`: green

PR #610, opened 2026-06-10
Add Magnetic Mountain (ARN) + dynamic per-player mana payment

bolav:
Magnetic Mountain (ARN, {1}{R}{R} Enchantment):
- "Blue creatures don't untap during their controllers' untap steps." — the DOESNT_UNTAP continuous restriction (same shape as Choke / Juntu Stakes).
- "At the beginning of each player's upkeep, that player may choose any number of tapped blue creatures they control and pay {4} for each creature chosen this way. If the player does, untap those creatures." — composed from the existing GatherCards -> SelectFromCollection(ChooseAnyNumber, on-battlefield) -> ConditionalOnCollection -> Gate.MayPay -> ForEachInCollection(Untap) pipeline, with the upkeep player as both chooser and payer.

New engine capability: PayDynamicManaCostEffect(amount, payer) — the dynamic, payer-parametric twin of PayManaCostEffect. It pays a DynamicAmount of generic mana at resolution from any Player reference (defaulting to the controller), the building block for "pay {N} for each X" and "that player pays" each-player triggers. Gate.MayPay's affordability pre-pass now recognizes it, resolving the cost's own payer so an unaffordable selection skips the prompt instead of erroring at payment — correct even when payer differs from the gate's decisionMaker. The fixed and dynamic mana-payment executors share their auto-tap/deduct core via the new payGenericFromPool helper.

Tests: MagneticMountainTest covers the untap restriction (blue held / non-blue untaps / cross-controller), the upkeep player paying {4} to untap (charging the triggering player, not the enchantment's controller), the per-creature scaling (choosing two costs {8} and untaps both; a {7} pool gates the {8} away), declining, and the can't-afford skip. Docs + ARN snapshot updated; manual scenario added.


wingedsheep (approved):
## Review

**Verdict:** Correct behavior and the right SDK shape. `PayDynamicManaCostEffect` is a justified new primitive — a fixed `ManaCost` can't express a resolution-computed amount, and nothing existing could charge a non-controller — and it's parameterized for the next card (any `DynamicAmount`, any `Player`), not just this one. The card itself is pure composition of existing pipeline primitives.

Verified during review:
- **Oracle text, rarity, collector number, artist, image URI** all match Scryfall exactly; `just check-card-printing` confirms ARN is the correct canonical home (3ED/4ED/FBB/SUM/4BB aren't scaffolded, so no `Printing` rows due).
- **CR 502.3** as cited is correct ("effects can keep one or more of a player's permanents from untapping").
- The gather path uses projected state; registration is automatic via `CardDiscovery`; the empty-pool upkeep resolves silently.
- Green: `MagneticMountainTest`, full `:rules-engine:test` (warranted by the `PayManaCostExecutor` refactor), `:mtg-sets:test`, `:mtg-sdk:test`, `:game-server:test`.

**What's good**
- `canAfford` resolving the cost's *own* `payer` rather than trusting the gate's `decisionMaker` is the right correctness call for cross-player gates.
- Extracting the shared auto-tap/deduct core instead of copy-pasting it.
- Test coverage maps 1:1 to every behavior claimed in the PR body, including charging the upkeep player rather than the enchantment's controller, per-creature scaling with exact tapped-land counts, and the decline path.

**Follow-up pushed to this branch** (`af3615a53`, after merging `main` cleanly):

1. **Affordability-capped selection** (was the one Important finding): with {7} available and two creatures chosen, the gate silently skipped and the player untapped *nothing* — a UX trap on every future "pay {N} for each chosen X" card. New generic `SelectionRestriction.MaxAffordablePayment(manaPerSelected, payer)` caps the selection at `floor(available mana / per-card cost)`, folded into the decision's `maxSelections` (already enforced by `DecisionValidators`); a zero cap under `ChooseAnyNumber` skips the prompt entirely. This is also rules-faithful: "choose … and pay" is one action, so an unpayable choice was never legal in paper either.
2. **Computed pay prompt**: the `Gate.MayPay` confirm button now shows the evaluated total ("Pay {8}"), not the per-creature formula.
3. **Honest helper name**: `payGenericFromPool` → `payManaCostFromPool` (it pays colored fixed costs for `PayManaCostExecutor` too).

Tests updated accordingly ({7}-pool asserts `maxSelections == 1` then pays for one; {3}-pool asserts no prompt at all; the {8} confirm asserts the button label), docs extended, ARN snapshot re-blessed. All suites re-run green after the changes.

One pre-existing note, no action needed: because `decisionMaker` is set, this gate intentionally takes the generic auto-tap path rather than the `asOptionalManaPayment` manual-source-selection UX — fine here (generic-only cost), worth remembering if Treasures/sub-cost sources ever matter for a card in this family.

PR #611, opened 2026-06-11
Add Sneak alternative-cost keyword (CR 702.190) + 17 TMT cards

bolav:
Implements the **Sneak** keyword from Teenage Mutant Ninja Turtles and the
17 cards that use it.

## Sneak (CR 702.190)
"Any time you could cast an instant during your declare blockers step, you may
cast this spell by paying [cost] and returning an unblocked creature you control
to its owner's hand rather than its mana cost." A permanent spell so cast enters
tapped and attacking the same defender the returned creature was attacking
(702.190b / 506.3a).

Modelled as a real alternative cost (`AlternativeCostType.SNEAK`) reusing the
existing alt-cost pipeline, not a one-off:
- `SneakWindow` centralizes the declare-blockers timing + unblocked-attacker reads
  (projected controller).
- A dedicated `SneakCastEnumerator` surfaces the instant-speed cast only inside the
  window, with a `BouncePermanent` additional cost.
- `StackResolver` enters the permanent tapped/attacking; falls back to "not
  attacking" when the carried defender is no longer legal (506.3c).
- `Conditions.SneakCostWasPaid` exposes the durable flag for riders (Leonardo,
  The Last Ronin's Technique, Turncoat Kunoichi).

## Cards
17 TMT cards (Foot Ninjas, the Technique cycle, Shredder/Leonardo/Donatello, …),
all composing existing SDK primitives — no card-specific types.

## Tests
`SneakTest` (window legality, bounce cost, combat damage same combat, 506.3c
defender-left, normal-cast unaffected) + one scenario test per card. DSL reference
and TMT backlog (96→113) updated. All green.


wingedsheep (commented):
## Review — Sneak (CR 702.190) + 17 TMT cards

**Verdict:** Behavior is right and the SDK shape is exactly right: zero new `Effect`/`StaticAbility`/card-specific types — Sneak rides the existing alternative-cost pipeline plus the pre-existing `BouncePermanent` additional-cost plumbing, and all 17 cards compose existing primitives. One real correctness gap was found (CR 509.1h blocked-status stickiness); it and three smaller findings are **fixed in 7f3287b68**, pushed to this branch.

### Verified
- **All 17 cards char-exact vs Scryfall** (name, mana cost, type line, P/T, full oracle text, sneak cost, rarity, collector number) — zero mismatches, including Leonardo, Leader in Blue's sneak cost being *more* than its mana cost and Turncoat Kunoichi's off-color `{2}{W}{B}` sneak with `colorIdentity = "WB"`.
- **`just check-card-printing` passes 17/17** — TMT is the earliest real printing for every card.
- **Every cited CR number checked against the official rules text**: 702.190a/b, 506.3c, 509.1h, 118.9a, 601.2c all correct (702.190b itself cross-references 506.3a, so those mentions are fine).
- **Full `mtg-sdk`, `mtg-sets`, `rules-engine`, `game-server` suites green**, before and after the fixes.
- The `AutoPassManager` refactor is a real fix, not a drive-by: the cast-action-type allowlist was duplicated 7× and missing every `CastWith*` variant, so the client would speed past the only window where Sneak is legal. The new named sets match the actual enumerator-emitted action types (diffed them).

### What was fixed in 7f3287b68

1. **CR 509.1h sticky blocked status (the one real bug).** `SneakWindow.unblockedAttackers` decided "unblocked" by scanning the battlefield for creatures whose `BlockingComponent` currently lists the attacker. But blocked status is sticky — *"A creature remains blocked even if all the creatures blocking it are removed from combat"* — and the engine already encodes this deliberately: `CombatRemovalHelper` keeps `BlockedComponent` on the attacker when its blockers leave combat (Ydwen Efreet is the documented sole override). So a blocked attacker whose blocker died or bounced during the declare blockers step became returnable for a sneak cost — an illegal cast, reachable through both the enumerator and `CastSpellHandler` validation. Now reads `BlockedComponent == null` (also drops the O(n²) scan). `isWindowOpen` additionally requires `BlockersDeclaredThisCombatComponent` (same gate `CombatEnumerator` uses), since CR 509.1h only assigns blocked/unblocked status once the declare-blockers turn-based action happens.
2. **Two new tests** pinning the corner: a blocked attacker whose blocker left combat is not returnable (cast rejected, pool empty), and the defending player cannot sneak during the attacker's combat.
3. **Stale KDoc** on `SpellOnStackComponent.sneakAttackDefenderId` claimed the resolver "falls back to the caster's opponent" when the carried defender is gone — it actually enters the creature not attacking (CR 506.3c, and the existing test pins that). Comment corrected.
4. **Wrong CR citation** in `SneakCastEnumerator`: 601.3e (alternative characteristics) → 601.3 (cast permissions/prohibitions).

### Kept as-is (deliberate, looks right)
- `SneakWindow` as the single seam for window + pool reads, projected-state controllers.
- Durable `ChoiceSlot.SNEAK` flag mirroring `KICKED`; `Conditions.SneakCostWasPaid` mirroring `WasKicked` with the resolution-context fallback for non-permanent spells.
- Card compositions follow precedent: Jennika = the Fiery Cannonade `ForEachInGroup` shape, Karai = `modal(chooseCount = 2, minChooseCount = 1)`, Turncoat Kunoichi = the established linked-exile pair; Shredder's Technique documents its accepted indestructible-enchantment edge.

PR #616, opened 2026-06-11
LTR equipment-aware cards: Shagrat, Forge Anew, Frodo + equipment-aware engine primitives

bolav:
Implements the equipment-aware subset of the LTR backlog, plus the engine primitives it needs.
(Audit found `missing-effects.md` stale — most "equipment-aware" cards fork into unrelated gaps;
this PR covers the genuinely equipment-aware ones. The other 5 remain blocked on distinct features.)

## Engine (one commit)
- **Equipment-filtered attachment count** — `AttachmentCount(AttachmentKind)` +
  `DynamicAmounts.equipmentAttachedToSelf()` (counts only Equipment, not Auras).
- **Equip timing/cost permissions** — `EquipAbilitiesAtInstantSpeed` and `FreeFirstEquipEachTurn`
  statics (keyed off a new `ActivatedAbility.isEquipAbility`), tracked per-turn via
  `EquipActivationsThisTurnComponent`. Wired into the enumerator + `ActivateAbilityHandler`,
  mirroring the existing flash-permission scan. No new client surface.

## Cards
- **Shagrat, Loot Bearer** — attack-trigger attach + amass Orcs X = Equipment attached.
- **Forge Anew** — ETB reanimate Equipment from GY + instant-speed equip + free first equip.
- **Frodo, Determined Hero** — MV-2-or-3 attach on enter/attack + during-your-turn damage
  prevention (composed, no engine change).

## Fixes / tests
- Free-equip cost display: a cost reduced to `{0}` rendered blank, so the client showed the
  Equipment's printed cost — now surfaced as `{0}`.
- Engine scenario tests for each card + the permissions (incl. enumeration, not just validate);
  re-blessed card snapshots (only `isEquipAbility: true` added); SDK reference + backlog updated.
- Dev scenarios under `manual-scenarios/`.

LTR: 209 → 212 / 291.


wingedsheep:
## Review + fixes pushed (`c6b5baab`)

Reviewed through the SDK-elegance lens. The two new equip permissions are genuinely novel (timing/cost semantics no existing primitive expresses), `AttachmentCount(kind: AttachmentKind)` is parameterized for the next card rather than hardcoded for Shagrat, and all three cards compose existing primitives. Merged `origin/main` in cleanly (no conflicts) and verified before reviewing.

I pushed two fixes directly to this branch:

**Blocking (caused the failing test) — `Serialization.kt`.** `EquipActivationsThisTurnComponent` was never added to the `polymorphic(Component::class)` block, so a paused `GameState` carrying the per-turn equip tracker failed round-trip. This is exactly what `SerializationPolymorphicRegistrationTest > every Component implementation is polymorphically registered` guards — that's why it was red. Added the `subclass(...)` line; the full test (all 6 hierarchies) now passes.

**Important — projection (`DynamicAmountEvaluator.kt`).** The new `AttachmentCount(EQUIPMENT/AURA)` branches read the attachment's type from the printed `CardComponent.typeLine`. Attachments live on the battlefield, where a continuous effect can change what is/isn't an Equipment or Aura, so per the projected-state rule the count must read projection (`getSubtypes`, with a base-state fallback when projection is empty). Switched to projection and imported `AttachmentKind`/`Subtype` so the branches read unqualified rather than fully-qualified inline.

**Minor — left as-is (heads-up).** `ActivateAbilityHandler` inlines its own copy of the equip-permission scan + the `applyFreeFirstEquip` discount, mirroring the versions in `CastPermissionUtils`/`ActivatedAbilityEnumerator`. This follows the file's existing precedent (it already mirrors `CastPermissionUtils.isActivationPrevented`, since the handler isn't constructed with `CastPermissionUtils`). Consolidating would mean injecting `CastPermissionUtils` into the handler for just this one path — inconsistent with `isActivationPrevented` and broader than warranted here, so I left it. Worth a follow-up if the handler ever gains `CastPermissionUtils` as a dependency.

### Verification
All re-run green after the fixes:
- `SerializationPolymorphicRegistrationTest` (was failing) ✅
- `ShagratLootBearerScenarioTest`, `FrodoDeterminedHeroScenarioTest`, `ForgeAnewScenarioTest`, `EquipPermissionsScenarioTest` ✅ (incl. the Shagrat test that exercises the projection path)
- `CardDefinitionSnapshotTest` ✅
- `just check-card-printing` ✅ for all three (canonical in LTR, the cards' earliest real printing)

Tests cover the corners well — enumeration (not just `validate`), the `{0}` free-equip display, equipment-only counting vs. an Aura, instant-speed timing in combat, and first-free-then-full.

wingedsheep:
## Follow-up: equip-permission duplication consolidated (`7aefb6e0`)

Did the cleanup I'd flagged as a heads-up rather than leaving it. `ActivateAbilityHandler` no longer carries its own copies of the equip-permission scan or the free-first-equip discount — and the pre-existing `isActivationPreventedByStatic` mirror is gone too, so the "handler doesn't carry `CastPermissionUtils`" precedent that justified the duplication is removed.

- **`EngineServices`** now holds a single shared `CastPermissionUtils`, threaded into `ActivateAbilityHandler` via `create()`. (No init-order issue — it's constructed after `predicateEvaluator`/`conditionEvaluator`.)
- **`CastPermissionUtils.applyFreeFirstEquipDiscount(...)`** is now the one place the {0} first-equip discount lives; both the enumerator (offered/displayed cost) and the handler (paid cost) call it, so they can't drift.
- Deleted from the handler: `isActivationPreventedByStatic`, `applyFreeFirstEquip`, `hasFreeFirstEquipGrant`, `canEquipAtInstantSpeedGrant`, `hasActiveEquipPermissionGrant`; from the enumerator: the private `applyFreeFirstEquip`. Dropped the now-unused `PreventActivatedAbilities` import.

Net **−91 lines** (42 insertions / 133 deletions). `create()` is the only construction site, so the new constructor param is fully wired.

### Verification
- Full `:rules-engine:test` suite green (incl. the equip + serialization tests and the `PreventActivatedAbilities`/activation-prevention coverage that exercises the now-shared `isActivationPrevented`).
- `:game-server` main + test compile clean.

PR #617, opened 2026-06-11
Close the fail-open defaults (sdk-analysis §1.3)

wingedsheep:
Implements [§1.3 of the revised SDK plan](../blob/main/backlog/sdk-analysis-2026-06-revised.md) — three small fail-open paths become explicit failures or composable semantics.

## 1. `GameObjectFilter.and` no longer silently discards predicates

AND-ing two filters used to keep only the right side's controller predicate on conflict (`ObjectFilter.kt`), and silently dropped a right-hand `anyOf` union. Both are now `require` failures at filter construction time — since filters are built when card definitions load, the corpus-wide tests surface any violation at build time. The error message points at the explicit alternative (`withControllerPredicate(ControllerPredicate.And(...))`).

## 2. `ControllerPredicate` combinators (`And` / `Or` / `Not`)

Mirrors `StatePredicate`'s existing combinators (`ControllerAnd`/`ControllerOr`/`ControllerNot` serial names). Heterogeneous controller/owner filters — "creatures you own but don't control" — now compose in one predicate instead of needing the `anyOf` escape hatch.

Evaluator support is a shared structural fold, `ControllerPredicate.evaluateWith(leaf)`, next to the predicate type: each engine site keeps only its own leaf semantics (live projected controller, zone-change last-known-controller, grant-provider fast paths) and shares the combinator recursion. Wired through:

- `PredicateEvaluator.matchesControllerPredicate` (explicit recursion, exhaustive `when` kept)
- `TriggerMatcher` (zone-change LKI + targeted-trigger sites)
- `TriggerAbilityResolver` (three grant-resolution sites)
- `ZoneMovementUtils`, `AffectsFilterResolver`
- `StaticAbilityHandler`: the controller fast path now maps unknown predicates to `AffectsFilter.Generic` (full evaluation) instead of silently widening them to `AllCreatures` — the same fail-open class, closed in passing

DSL entry point: `GameObjectFilter.withControllerPredicate(p)`.

## 3. `SuccessCriterion.Auto` rejects shapes it can't infer, at card load

`Gate.DoAction` with the default `Auto` criterion used to score *any* unrecognized action shape as "it happened" (fail open). Now:

- The shape probe is the SDK's single source of truth: `SuccessCriterion.Auto.canInfer` + the terminal-move walkers, which `GatedEffectExecutor` delegates to — the validator and the runtime snapshot recognize exactly the same shapes.
- `CardValidator` walks the card's **serialized JSON tree** to find every `Gate.DoAction` (spell effects, triggered/activated abilities, modes, faces, granted abilities — stale-proof against new effect containers) and errors with "specify an explicit criterion" when `Auto` can't infer.
- `SuccessCriterionValidationTest` (mtg-sets) enforces this corpus-wide at build time. The current corpus is already clean — every existing `IfYouDo` either has an inferable terminal zone move or an explicit criterion, so no card migrations were needed.
- Runtime keeps a graceful fallback only for the genuinely runtime-unresolvable residue (e.g. a destination player that doesn't resolve), which validation can't see statically.

## Tests

- `GameObjectFilterCompositionTest` (mtg-sdk): `and` conflict/anyOf rejection, combinator fold semantics, JSON round-trip, descriptions.
- `CardValidatorTest` additions: uninferable `Auto` flagged (incl. nested inside a mode inside a triggered ability), inferable shapes and explicit criteria accepted.
- `ControllerPredicateCombinatorScenarioTest` (rules-engine): steals a creature with Threaten (control change visible only via projection) and proves `And(OwnedByOpponent, ControlledByYou)` and `Not(ControlledByYou)` evaluate against the *projected* controller — a base-state read fails both tests.
- `SuccessCriterionValidationTest` (mtg-sets): the corpus-wide gate.

Full suite green: 4,218 tests across mtg-sdk / mtg-sets / rules-engine / game-server / mtgish-tooling, 0 failures. Card snapshot goldens unchanged (no serialized forms changed). No mtgish bridge/emitter updates needed: nothing was renamed, and the emitter never renders composed controller predicates.

Docs: `card-sdk-language-reference.md` updated (chained-predicates section + both `SuccessCriterion.Auto` entries).

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -678,9 +678,13 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
     success → `then`, failure → `otherwise`. This is "[action]. If you do, [then]" (a declined or
     no-op action runs `otherwise`, not `then`), distinct from `MayDecide` (gates on the yes/no) and
     `MayPay` (gates on paying a cost). `successCriterion` defaults to `SuccessCriterion.Auto`, which
-    infers success from the action's terminal zone-move (a pipeline `MoveCollection`, or a single
-    `MoveToZone` of the source itself) growing its destination zone; non-zone-move actions fall
-    through to fail-open. The executor pre-pushes a `GatedActionContinuation` so a paused action
+    infers success from the action's terminal zone-move (a pipeline `MoveCollection` to a zone, or a
+    single `MoveToZone` of the source itself) growing its destination zone. **Auto is only legal on
+    shapes it can infer** (`SuccessCriterion.Auto.canInfer`): card-load validation (`CardValidator`,
+    enforced corpus-wide by `SuccessCriterionValidationTest`) rejects an Auto criterion on any other
+    action — non-zone-move actions (deal damage, gain life, …) must state `SuccessCriterion.Always`
+    or `CollectionNonEmpty(name, min)` explicitly instead of silently inheriting a fail-open
+    "it happened". The executor pre-pushes a `GatedActionContinuation` so a paused action
     auto-resumes and evaluates after its own continuations drain. Replaces `IfYouDoEffect` (see
     "Sequencing & conditional" below).
   - `Gate.MayPayX` — **not a yes/no, a number chooser.** "You may pay {X}. If you do, [then]." The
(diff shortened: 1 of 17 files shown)


wingedsheep:
**Follow-up commit** (6858c7b) closing the two residual fail-opens noted in self-review:

1. **`SuccessCriterion.CollectionNonEmpty` now actually gates.** It was a runtime no-op (fell back to the Auto zone-probe with an empty snapshot → always "happened") — a problem this PR made worse by recommending it in the validation error message. The existing inject-collections-into-next-frame pattern (duplicated in `LibraryAndZoneContinuationResumer` ×3 and `AmassContinuationResumer`) is now one shared `exposeCollectionsToNextFrame` seam that also recognizes `GatedActionContinuation`; both `EffectContinuation` drain paths and the executor's synchronous path route through it. New scenario test pins both the chosen and the declined path — the declined path wrongly drew a card before this fix.

2. **Three-valued `evaluateWith`.** Site leaf lambdas returned `true` for leaf kinds they can't evaluate, which `Not(...)` would have flipped into rejecting everything. Leaves now return `Boolean?` (`null` = unknown), unknowns propagate with Kleene logic through `And`/`Or`/`Not`, and resolve to "don't constrain" only at the root.

Full suite green: 4,221 tests, 0 failures.

wingedsheep:
The follow-up commit above landed after this PR merged, so it moved to its own PR: #619.

PR #629, opened 2026-06-11
Add Abu Ja'far to Arabian Nights

bolav:
Abu Ja'far ({W} 0/1 Human): "When this creature dies, destroy all creatures blocking or blocked by it. They can't be regenerated."

A dies trigger can't read live combat state — by the time it resolves the engine has already torn down the combat cross-references (and the dying creature's own combat components). So this captures the CR 509 combat pairing as last-known information when the creature leaves the battlefield and threads it through to the trigger's effect:

- ZoneChangeEvent.lastKnownBlockingOrBlockedByIds — captured in ZoneTransitionService from the leaver's BlockedComponent (its blockers) + BlockingComponent (the attackers it blocked).
- Carried through TriggerContext -> EffectContext (triggerLastKnownBlockingOrBlockedByIds), including the on-stack (TriggeredAbilityOnStackComponent) and targeted-trigger continuation paths.
- New CardSource.LastKnownCombatPairedWithSource resolves it (restricted to creatures still on the battlefield) so the standard gather+destroy pipeline can consume it.
- Facade: Effects.DestroyCreaturesBlockingOrBlockedBySource(noRegenerate).

Scenario test covers Abu Ja'far dying as a blocker (kills the attacker it blocked, spares a bystander) and as an attacker (kills its blocker). SDK reference updated.

### DIFF
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 78 cards
 **Release Date:** December 17, 1993
-**Implemented:** 64 / 78
+**Implemented:** 65 / 78
 | Color | Count |
 |-------|-------|
 | White | 11 |
@@ -15,7 +15,7 @@
 
 ---
 
-- [ ] Abu Ja'far
+- [x] Abu Ja'far
 - [x] Army of Allah
 - [x] Camel
 - [ ] Eye for an Eye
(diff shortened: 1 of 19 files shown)


wingedsheep:
Reviewed — solid PR. The combat-pairing LKI is genuinely novel (no existing `CardSource`/filter could reconstruct "was in combat with the source when it left"), and it composes the standard gather→destroy pipeline rather than adding a bespoke executor. Threading mirrors the established `lastKnownAttachedTo` / `lastKnownDamageDealtByPlayers` pattern, the combat-component semantics check out (`BlockedComponent.blockerIds` ∪ `BlockingComponent.blockedAttackerIds`, restricted to survivors), printing placement is correct (canonical in ARN, the earliest real printing), and the three scenario tests cover blocker/attacker death plus the `noRegenerate` corner with a real regen shield up.

One minor follow-up pushed as `dc8ad25`: the divided-damage resume path (`EffectAndTriggerContinuationResumer.resumeTriggerDamageDistribution`) was the one LKI carrier that dropped the new field — `TriggerDamageDistributionContinuation` didn't have `triggerLastKnownBlockingOrBlockedByIds`, so the original comment noted it stays null there. I added the field to that continuation and populated it at its creation site, so the threading is now complete and a future "divide damage among creatures blocking or blocked by it" trigger keeps the pairing across a damage-distribution pause. `AbuJafarTest` still green.

LGTM.

PR #630, opened 2026-06-11
Add Drop of Honey to Arabian Nights

bolav:


wingedsheep:
## Review

*(Reviewed with `origin/main` merged into the branch, so this reflects post-merge reality — the PR is based on an older main, so either pull main or accept a merge commit before merging.)*

### Verdict
Behavior is correct and the SDK shape is right. The card faithfully implements Drop of Honey, the one new primitive (`HasLeastPowerAmongAllCreatures`) is genuinely novel and correctly named, and everything else composes existing primitives. All tests pass locally (Drop of Honey scenario suite, `mtg-sdk` tests, snapshot round-trip, CardLint). **Ready to merge** modulo two optional minor notes.

### What's good
- **Clean composition.** The card mirrors Serendib Djinn's structure exactly (upkeep trigger + `stateTriggeredAbility` self-sacrifice). `Conditions.NoCreaturesOnBattlefield` is just a named `Exists(Player.Each, …, negate = true)` recipe — no new condition type. The destroy is a standard Gather→Select→Move pipeline in `GroupPatterns`.
- **The naming distinction is correct, not a name-lie.** `HasGreatestPower` is *controller-scoped* (`PredicateEvaluator` filters `ctrl == entityController`), whereas Drop of Honey needs *global* least power across both players. The verbose `HasLeastPowerAmongAllCreatures` name and its docs make that scope difference explicit rather than papering over it. Good call.
- **Projection-aware throughout.** Both the resolution path (`PredicateEvaluator`, via `state.projectedState` / `projected.isCreature` / `projected.getPower`) and the layer path (`AffectsFilterResolver.hasLeastPowerAmongAllCreaturesInProjection`, with `baseStats.basePower` fallback) read projected state.
- **Exhaustive-allowlist discipline.** The new predicate was correctly added to the no-`else` `when` blocks in `BeginningPhaseManager` and `TriggerMatcher` (both classify it as "no constraint").
- **Tests cover the right axes**: global reach incl. destroying an *opponent's* creature, controller-chooses-on-tie (and auto-select when the minimum is unique — test 1 resolves with no prompt), and empty-board self-sacrifice. Oracle text is an exact match to Scryfall; `{G}` Enchantment, ARN #47.
- **Canonical placement correct.** ARN (1993) is the earliest real printing; the only reprint is ME4 (2011 masters set), so the canonical `card(...)` belongs in ARN.
- **Docs updated in the same change** (`card-sdk-language-reference.md` for both the effect and the condition), and the ARN snapshot + backlog count (64→65) regenerated.

### Issues

**Minor (author's discretion)**
- **"It can't be regenerated" has no paired test.** `noRegenerate = true` is wired through `MoveCollectionEffect`, but no scenario gives the least-power creature a regeneration shield to prove the shield is ignored. The flag is well-tested elsewhere, so this is optional — add a fourth test only if you want the rule-citation covered.
- **`destroyLeastPowerCreature` is currently single-use** — but `FacadeBoundaryTest` forbids raw pipeline construction in card files, every sibling pattern follows the same style, and Porphyry Nodes (PLC) has word-for-word the same ability, so the named pattern has a known second user. No change needed.

### Recommendation
Merge as-is. Optionally add a regeneration-shield assertion to back the "can't be regenerated" clause.

wingedsheep:
Pushed two follow-up commits addressing the review's remaining minor point and the conflict with current `main`:

**`09b72f79` — Merge `origin/main` into the branch (conflict resolution).** Main gained Abu Ja'far (#629) since this PR was opened, and both branches added a destroy pipeline at the same insertion point. Resolved by keeping both: `destroyLeastPowerCreature` and `destroyCombatPairedWithSourcePipeline` in `GroupPatterns`, both `Effects.*` facades, and both bullets in the DSL reference. One subtle catch the auto-merge would have silently gotten wrong: both branches recorded their own card as ARN's 65th implemented card, and the identical `64 → 65` line merged without conflict — the count is now correctly **66 / 78** (matches the 66 checked boxes).

**`fb3b8165` — Test for the "It can't be regenerated" clause.** With the upkeep destroy trigger pending, the least-power creature (Grizzly Bears) gets a regeneration shield — installed directly into `floatingEffects` after the turn change so cleanup can't expire it, the same idiom as `AbuJafarTest` — and the test asserts the shield does not save it. With this, every sentence of the card's oracle text has a paired test.

Verified on the merged result: `DropOfHoneyScenarioTest` 4/4, `AbuJafarTest` 3/3, `CardDefinitionSnapshotTest` 236/236 (ARN golden contains both new cards), `FacadeBoundaryTest` and `CardLintTest` green.

PR #631, opened 2026-06-11
Fix AI loop: "target spell" must not match abilities on the stack

bolav:
A counterspell's TargetSpell() requirement uses TargetFilter.SpellOnStack (GameObjectFilter.Any, zone = STACK). Both stack-target finders filtered state.stack by the base filter alone, and Any matches every entity — so triggered/activated abilities on the stack were counted as valid "target spell" targets. The enumerator then offered an illegal counterspell cast that TargetValidator rejected ("Target must be a spell on the stack"), looping the AI forever when an opponent ability sat on the stack.

Restrict findSpellTargets / findValidSpellTargets to entities that are actually spells (carry SpellOnStackComponent); ability-on-stack entities have no such component. A spell is a card on the stack (CR 112.1), whereas an activated/triggered ability put on the stack is an ability, not a spell (CR 113.3b/c), and once on the stack it exists as its own object independent of its source (CR 113.7a) — so "target spell" can never select it.

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/TargetFinder.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/TargetFinder.kt
@@ -10,6 +10,7 @@ import com.wingedsheep.engine.state.components.player.PlayerHexproofComponent
 import com.wingedsheep.engine.state.components.player.PlayerShroudComponent
 import com.wingedsheep.engine.state.components.identity.CardComponent
 import com.wingedsheep.engine.state.components.identity.ControllerComponent
+import com.wingedsheep.engine.state.components.stack.SpellOnStackComponent
 import com.wingedsheep.engine.mechanics.layers.ProjectedState
 import com.wingedsheep.engine.mechanics.targeting.PlayerTargetRestriction
 import com.wingedsheep.sdk.core.Keyword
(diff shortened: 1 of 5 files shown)


wingedsheep:
Reviewed and pushed two minor fixups (`63bba0d`):

- **PR description CR citation corrected.** The body previously cited CR 115.4 for "a spell targeting a spell can't target an ability" — but 115.4 is actually about damage spells requiring "any target". Updated it to the rules the fix actually rests on: CR 112.1 (a spell is a card on the stack), CR 113.3b/c (activated/triggered abilities are put on the stack as abilities), and CR 113.7a (an ability on the stack exists independently of its source). The in-code comments were already correct.
- **Centralized the dispatch check.** `CounterEffectExecutor`'s `SpellOrAbility` branch had the same inline `has<SpellOnStackComponent>()` test this PR factored into `GameState.isSpellOnStack` — routed it through the helper so the spell-vs-ability distinction lives in one place.

Verified green: `TargetSpellExcludesAbilitiesTest` (both cases), plus `WillbenderTest`, `StackWarTest`, and `MysticDenialScenarioTest` to confirm the counter-dispatch refactor is behavior-preserving.

Nice fix — note it also closes a latent bug: the old `findValidSpellTargets` filtered on `CardComponent != null`, which would have wrongly excluded **spell copies** (a copy of a spell is a spell but not a card); `isSpellOnStack` includes them. LGTM.

PR #632, opened 2026-06-12
Add Aladdin's Lamp to Arabian Nights

bolav:
Aladdin's Lamp ({10} Artifact): "{X}, {T}: The next time you would draw a
card this turn, instead look at the top X cards of your library, put all
but one of them on the bottom of your library in a random order, then
draw a card. X can't be 0."

The {X},{T} ability installs a one-shot ReplaceNextDraw shield whose
replacement is a lookAtTopAndKeep dig (keep one on top, rest to the
bottom in random order) followed by a real DrawCards(1) that draws the
kept card. The shield is consumed before that inner draw, so it doesn't
re-trigger itself.

Engine fix (reusable): the draw-replacement shield now captures the
activation-time {X} (ReplaceDrawWithEffect.xValue, set from the
activation context in ReplaceNextDrawWithExecutor and restored into the
EffectContext by DrawReplacementShieldConsumer), so a replacement effect
set up with an {X} cost can read DynamicAmount.XValue when it fires at
draw time — previously the X was lost between activation and the draw.

Scenario test verifies X-capture (X=2), the dig keeping the chosen card
(drawn), the rest going to the bottom, and that cards beyond the top X
are untouched. SDK reference updated.

### DIFF
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 78 cards
 **Release Date:** December 17, 1993
-**Implemented:** 64 / 78
+**Implemented:** 65 / 78
 | Color | Count |
 |-------|-------|
 | White | 11 |
@@ -70,7 +70,7 @@
 - [x] Sandstorm
 - [x] Singing Tree
 - [x] Wyluli Wolf
-- [ ] Aladdin's Lamp
+- [x] Aladdin's Lamp
 - [x] Aladdin's Ring
 - [x] Bottle of Suleiman
 - [x] Brass Man
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -305,7 +305,10 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 - `EachPlayerReturnPermanentToHand()` — each player bounces a permanent.
 - `EachPlayerDrawsForDamageDealtToSource()` — each player draws equal to damage source took this turn.
 - `ReadTheRunes()` — draw N, then discard N (or sacrifice permanents).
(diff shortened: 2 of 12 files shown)


wingedsheep:
Reviewed and pushed fixes for the two minor items from review (commit `616df97`):

**1. Consolidated the X-mana coverage duplication.** Extracted `ManaPool.xCoveragePlan(xAmount, xManaRestriction)` — the single "colorless first (unless X is color-restricted), then allowed colors" rule for how much of an `{X}` cost floating mana covers. It now backs three call sites that previously hand-rolled the same loop:
- `ActivateAbilityHandler.autoTapForManaCost` reduces its tap target by the plan's `.size`;
- `CastPaymentProcessor.payFromPool` and `.autoPay` both iterate the plan to spend each unit and tally per-color X spend.

This also removed a *second* copy of the loop I hadn't called out (it lived in both `payFromPool` and `autoPay`), so the consolidation nets three duplicates down to one rule.

**2. Covered the X=0 edge.** Confirmed the "X can't be 0 is harmless" claim is actually correct — with X=0 the dig gathers zero cards, `SelectFromCollection` short-circuits on the empty collection (no prompt, empty kept/rest), and the inner `DrawCards(1)` runs as a plain top-of-library draw. Replaced the unverified comment with a precise one and added a scenario test (`X=0 degenerates to an ordinary draw`) so the claim isn't decorative.

**Verification:** `AladdinsLampScenarioTest` (3 tests incl. the new X=0 one) green; re-ran the X-cost / X-mana-restriction suites that exercise both payment paths — `XCostSpellTest`, `ProsperityXCostTest`, `SoulBurnAndAtalyaXManaTest` (color-restricted X), `CastXDurableValueTest`, `CommanderTaxTest`, `CabalInterrogatorTest` — all pass.

The fix commit is rebased directly on the PR head (no `origin/main` merge folded in), so the PR diff stays clean.

PR #634, opened 2026-06-12
Add Cuombajj Witches with opponent-chosen target support

bolav:
Cuombajj Witches (ARN): "{T}: This creature deals 1 damage to any target
  and 1 damage to any target of an opponent's choice." The canonical card
  lives in Arabian Nights (earliest printing); CMR reprint rows (#116, #646)
  added.

  Implements a faithful, announcement-time "… of an opponent's choice"
  target mechanic rather than a resolution-time approximation:

  - SDK: TargetChooser { Controller, Opponent } enum + `chooser` on
    TargetRequirement / AnyTarget, and Targets.AnyChosenByOpponent. The
    chosen target is a real target of the controller's ability — announced
    together with the controller's own targets and equally respondable —
    but an opponent selects which legal object/player it is.
  - Engine: the activated-ability enumerator surfaces only controller-chosen
    requirements (gating activation on opponent reqs being satisfiable);
    ActivateAbilityHandler validates only controller targets, then pauses at
    announcement before paying any cost (CR 601.2c precedes 601.2g–h) to
    route a ChooseTargetsDecision to an opponent. Legal targets are computed
    relative to the controller, so hexproof/protection are measured against
    the ability's controller (matching the printed ruling). A new
    continuation + resumer merges the opponent's pick into the aligned
    target list and re-enters to pay the cost and put the ability on the
    stack. Reuses existing decision/event plumbing — no server/client change.

  Multiplayer "controller picks which opponent" defaults to the sole
  opponent for now (documented); honored for activated abilities, the only
  printed use.

  Tests: CuombajjWitchesScenarioTest covers controller+opponent pings, the
  hexproof-relative-to-controller ruling, the opponent aiming at the
  controller's own creature, and the controller's ping killing a creature
  while the opponent retaliates.


wingedsheep (commented):
## Review — opponent-chosen target support

**Verdict:** behavior is correct and the SDK shape is right. `TargetChooser` is the minimal, orthogonal generalization (defaulted on the base interface, read only at the announcement layer), and the opponent's pick reuses the existing `ChooseTargetsDecision` plumbing — no new `GameEvent`, no client/server change. All 5 scenario tests, the card snapshot net, the lint net, and `just check-card-printing "Cuombajj Witches"` pass. CR citations verified against the Nov-2024 CR (601.2c precedes 601.2g–h; pause is correctly placed before cost payment). The `opponentTargetsChosen` resume-flag guard is real and tested. Canonical lives in ARN (earliest printing); CMR reprint rows are correct.

I pushed two commits (maintainer edit):

1. **Merge `origin/main`** — resolved a conflict in `backlog/sets/arabian-nights/cards.md` (your base counted 65; main advanced to 66 without this card → resolved to **67 / 78**, Cuombajj checked). Pull this back or accept the merge commit.
2. **Address the review findings:**
   - **`CardLinter` rule (`UnsupportedOpponentChooser`)** — closes the latent footgun where a `TargetChooser.Opponent` target on a *spell* or *triggered ability* would silently let the **controller** choose it (only activated abilities route to an opponent today). Anchored on the `AnyTarget` type discriminator so it can't collide with the unrelated `chooser` field on pipeline selection steps (`Chooser` also has an `Opponent` value). Backed by `CardLinterTest` cases + the corpus `CardLintTest` gate; documented in the SDK reference.
   - **Fixed-count guard moved earlier** — `pauseForOpponentChosenTargets` now rejects non-fixed-count requirement shapes *before* raising the opponent's decision, not only after the pick on the resume path (the resumer keeps its check as defense-in-depth).
   - **CR citation fix** — the "which opponent" choice is 601.6a/602.3a; the controller-goes-first ordering is 601.6b/602.3b.
   - **Prompt polish** — stripped the redundant "of an opponent's choice" suffix from the prompt shown to the choosing opponent.

Nothing blocking. Good to merge once you've reconciled the merge commit on your end.

PR #635, opened 2026-06-12
Unify the ForEach effect family onto ForEachEffect + IterationSpace

wingedsheep:
## What

Collapses the five sibling iteration effects — `ForEachTargetEffect`, `ForEachPlayerEffect`, `ForEachInCollectionEffect`, `ForEachInGroupEffect`, `ForEachColorOfEffect` — into a single compiled `ForEachEffect(space: IterationSpace, body: Effect)` with a sealed iteration axis:

| `IterationSpace` | Iterates | Per-iteration binding |
|---|---|---|
| `Targets` | chosen targets | `ContextTarget(0)` = current target, collections wiped |
| `Players(selector)` | matching players | `controllerId`/`opponentId` rebound, collections wiped |
| `Collection(name)` | a pipeline collection | `iterationTarget` (→ `EffectTarget.Self`), collections preserved |
| `Group(filter, noRegenerate)` | battlefield filter snapshot | same as Collection |
| `ColorsOf(source)` | entity's colors, WUBRG | chosen-color channel |

Same unification move as `Gate`/`GatedEffect` and `CostAtom`: the five names survive as lowering factory functions (the `IfYouDoEffect → GatedEffect` precedent), so **card source and the mtgish emitter's rendered DSL are unchanged** — only the compiled/serialized shape moved to `"ForEach"` (43 snapshot goldens re-blessed; the diff is purely structural, e.g. `ForEachInGroup{filter, effect}` → `ForEach{space: IterationSpace.Group{filter}, body}`).

## Why

- `ForEachTargetExecutor` and `ForEachPlayerExecutor` were ~175-line near-identical copies, each embedding a third copy of the composite sub-effect loop. One `ForEachExecutor` + one `ForEachContinuation` now replace five executors and two continuation frames (**net −924/+121 in touched files**, plus the new ~490-line SDK type + executor).
- **Latent correctness bug fixed:** the group / collection / colors-of executors propagated a mid-iteration pause with no continuation for the remaining entities — a body pausing for a decision silently dropped every iteration after it. The unified executor pre-pushes the continuation for *every* space. `ForEachPauseResumeScenarioTest` pins the group and collection cases (one decision per iteration, decline-one-continue-rest); these tests fail on main.
- Multi-effect bodies now lower to `CompositeEffect`, whose executor is a strict superset of the duplicated loops (it also threads `storedNumbers`/`chosenValues` between sub-effects).

## Notes

- `ForEachCapturedControllerEffect` stays separate (its three-collection tally shape doesn't fit the axis cleanly) but resumes through the shared player-space frame, preserving its existing resume semantics.
- The never-used `ForEachInGroupEffect.simultaneous` flag is dropped — snapshot-before-iterating is now the documented contract of every space (only `SeasonOfGathering` passed it, always `true`).
- `CardLinter`'s collection-read dataflow entry moves to the `IterationSpace.Collection` node; the linter's generic JSON-tree walk covers the new shape otherwise.
- Purely internal compiled-shape refactor: no new decisions, events, or client-visible state — decisions raised by iteration bodies are unchanged, so no web-client work.
- SDK reference updated with the unified entry (`docs/card-sdk-language-reference.md`).

## Testing

- Full `./gradlew test` green across all modules (rules-engine, mtg-sdk, mtg-sets incl. lint/facade/snapshot, game-server, ai, mtgish-tooling, mtg-search, gym).
- New `ForEachPauseResumeScenarioTest`: 4 scenarios proving pause/resume for the previously-broken spaces.
- `CardSerializationRoundTripTest` updated to the new shape; snapshot goldens re-blessed via `-DupdateSnapshots=true`.


wingedsheep:
Merged latest `main` and fixed the `EmitterGoldenTest` POR failure that surfaced with it. Root cause was unrelated to the ForEach refactor: `PipelineBuilder.kt` (PR #636) added the extension `fun GameObjectFilter.namedFromVariable(...)`, and the mtgish `Registry` declaration scan indexed the *receiver type* as a top-level `dsl` declaration — so `importsFor()` started emitting the nonexistent `com.wingedsheep.sdk.dsl.GameObjectFilter` import (test fails on current main too). Fixed the scanner to skip extension receivers (`fun`/`val` names followed by `.` or `<`); the committed golden was already correct and stays untouched. Full `./gradlew test` green on the merged tree.

PR #649, opened 2026-06-12
Collapse four condition near-clones into EntityMatches (§2.3)

wingedsheep:
Implements **§2.3 "Condition hierarchy: stop the one-off accumulation"** from `backlog/sdk-analysis-2026-06-revised.md`.

## What

The four `Condition` near-clones that asked "does *this entity* match a `GameObjectFilter`" — differing only in *which* entity — collapse into one primitive:

```kotlin
data class EntityMatches(val entity: EffectTarget, val filter: GameObjectFilter) : Condition
```

| Deleted clone | Now a facade over |
|---|---|
| `SourceMatches(f)` | `EntityMatches(EffectTarget.Self, f)` |
| `EnchantedPermanentMatches(f)` | `EntityMatches(EffectTarget.EnchantedPermanent, f)` |
| `TargetMatchesFilter(f, i)` | `EntityMatches(EffectTarget.ContextTarget(i), f)` |
| `TriggeringSpellMatchesFilter(f)` | `EntityMatches(EffectTarget.TriggeringEntity, f)` |

The entity is named via the existing `EffectTarget` vocabulary, so the SDK can express new roles (e.g. the equipped creature) for free instead of minting another condition class.

## Behavior preservation

`ConditionEvaluator` now has **one** `EntityMatches` branch that dispatches on the entity role to the already-correct matching strategy, so every prior behavior is preserved:

- **`Self`** / enchanted-or-equipped attachment → live predicate match; **dual-mode** (resolution + static-ability projection).
- **`ContextTarget`** → resolution-only; resolve the chosen target (a player target ⇒ `false`) and match live.
- **`TriggeringEntity`** → resolution-only; match by the spell's **static cast characteristics** so the answer survives the spell leaving the stack (CR 603.4 intervening-'if' re-check).

The `Conditions.*` facades are unchanged, so card sources don't move — only the five cards that imported the raw types switched to the facade. The **card snapshot golden is re-blessed**; the diff is purely `type: "SourceMatches"` → `type: "EntityMatches"` + a nested `entity`, and `TargetMatchesFilter`'s raw `targetIndex` folding into `ContextTarget(index)`. As a bonus, the `CardLinter` now validates that `ContextTarget` index (previously hidden in a raw int field).

Parts 2 & 3 of the item land as forward-looking rules in `docs/card-sdk-language-reference.md` and the `add-feature` checklist: tracker-shaped checks route through `Compare` + a tracked `DynamicAmount`; set-mechanic conditions stay quarantined in mechanic-named files.

## Tests

- `:mtg-sdk:test` + `:mtg-sets:test` green (lint, snapshot, polymorphic-registration).
- Full `:rules-engine:test` green (ConditionEvaluator is widely used).
- `:mtgish-tooling:test` green (emitter golden unaffected — it emits facades).
- `:ai` / `:game-server` / `:gym` / `:gym-trainer` compile.


wingedsheep (commented):
## Review

### Verdict

The right refactor, done the right way: one "entity matches filter" primitive that names the role via the existing `EffectTarget` vocabulary instead of minting a condition class per role — exactly what §2.3 asked for. Behavior is preserved: full `:rules-engine:test`, `:mtg-sdk:test`, `:mtg-sets:test`, `:mtgish-tooling:test` verified green **with `origin/main` merged in** (clean merge, no conflicts). Two Important findings were identified and have been **fixed in b0e38af** (details below).

### What's good

- **The collapse itself.** The `Conditions.*` facades stay stable, so the ~40 corpus call sites didn't move; only the five files importing raw types changed, and those changes are themselves a boundary improvement.
- **Dispatch preserves the per-role strategies verbatim**, with the dual-mode vs resolution-only split intact (`(ctx as? Resolution) ?: false` matches the old `ifResolution` semantics exactly).
- **CR 603.4 citation verified correct** against the official CR text (intervening-'if', re-checked on resolution).
- **The linter bonus is real**: `CardLinter.visitNode` walks `ContextTarget` nodes generically, so the index `TargetMatchesFilter` hid in a raw int field is now validated against the requirement scope.
- **Snapshot re-bless is clean**: purely `type` rename + nested `entity`; descriptions aren't serialized.

### Issues (all addressed in b0e38af)

**Important**

1. **Silent `else -> false` for unsupported entity roles** (`ConditionEvaluator.evaluateEntityMatches`). `EffectTarget` has ~15 other variants; `EntityMatches(EffectTarget.PipelineTarget("stored"), f)` — a plausible role — silently evaluated to a constant `false`. Same fail-open shape the project closed in §1.3 / PR #606, and the Atmospheric Greenhouse no-op class the `CardLinter` exists to catch. **Fixed:** new lint check (`CardValidationError.UnsupportedEntityMatchesRole`) rejects any entity outside the six dispatched roles at card load; the evaluator `else` stays as defense in depth. Doc'd in §21 + unit tests.

2. **The `EnchantedPermanent` dispatch arm had zero test coverage.** Essence Leak (its only user) has no test, and this refactor rewired that arm through both the projection gate and the `TriggerAbilityResolver` conditional-grant path. The advertised `EquippedCreature`/`EnchantedCreature` roles were also untested capability claims. **Fixed:** `EntityMatchesAttachmentScenarioTest` (6 tests) pins the projection gate on/off, the trigger-resolver granted-upkeep-trigger on/off (Essence Leak's shape), and the `EquippedCreature` role on/off.

**Minor (also addressed)**

3. The `SourceIs*` vals spelled out `EntityMatches(EffectTarget.Self, ...)` ten times despite the `SourceMatches` helper in the same object — now desugar through the helper.
4. `FacadeBoundaryTest` didn't cover condition constructors (the gap that let five cards import the raw types in the first place) — raw `EntityMatches(` construction in card files is now banned (lookbehind excludes the `Conditions.EntityMatches` facade).

### Notes

- The branch now includes a clean merge of `origin/main` (`29594f8`) — no conflicts; it just brings the PR up to date.
- Post-fix verification: `:mtg-sdk:test`, `:mtg-sets:test` (snapshot + corpus lint + facade boundary), full `:rules-engine:test`, `:game-server:compileKotlin` all green.

PR #651, opened 2026-06-12
Use vertical space on crowded battlefields; hide replay share buttons on mobile

wingedsheep:
## Summary

The battlefield slot-sizing hook previously shrank cards until **every permanent in a row fit on one horizontal line**. On crowded boards that made cards tiny while vertical space sat unused, and on portrait phones (little width, lots of height) it was the worst possible trade. Three changes:

1. **Wrap crowded rows into multiple lines.** `useSlotSizedResponsive` now searches wrap-line combinations (1–3 per row) and picks the layout that maximizes card size while all lines still fit the slot height. Single lines win ties, so roomy boards keep today's flat layout. Each battlefield row reserves `minHeight` matching its budgeted lines, so a wrapped row can never collapse or push into its neighbour / the center HUD.
2. **Balanced wrapping.** When a row of single cards wraps, the group gets a `maxWidth` sized for `ceil(n / lines)` cards, so flex-wrap fills lines evenly (11 + 11) instead of greedily orphaning the remainder (21 + 1). The width model accounts for tapped cards' rotated container (1.4 × width + 8) and, after rebasing on #650's permanent grouping, for stack peek offsets — the fit constraints count **rendered stacks**, not raw cards (grouping moved up to the outer `Battlefield` so it isn't computed twice).
3. **Replay on mobile:** the *Share as scenario* / *Save snapshot* / *Share replay* buttons are hidden below the 640px mobile breakpoint in both the public `ReplayPage` and the admin `ReplayViewer` — they didn't fit the header there.

### Crowded desktop board (22 distinct creatures, 1600×1200)

| Before | After |
|---|---|
| before | after |

One cramped line of floor-size cards with an orphan spilling into the divider → two balanced lines of much larger cards.

### Mobile portrait (14 creatures, 390×844)

| Before | After |
|---|---|
| before | after |

On phones the old code already overflowed into ad-hoc wrap lines (the rows only reserved one card height, so they leaked toward the HUD/hand); now the lines are budgeted, spaced and balanced.

### Replay header

| Desktop (buttons kept) | Mobile (buttons hidden) |
|---|---|
| replay desktop | replay mobile |

## Notes

- When even the 60px readability floor can't fit (e.g. 20+ permanents on a phone), overflow is geometrically unavoidable; the hook now re-derives line counts from what actually fits at the floor so the reservations track reality. That degenerate case rendered as uncontrolled overlap before and degrades comparably now.
- Verified with `npm run typecheck`, `npm run build`, and live screenshots against a local stack (dev scenario endpoint). The screenshots branch `battlefield-vspace-screenshots` is throwaway.

### DIFF
--- a/web-client/src/components/admin/ReplayViewer.tsx
+++ b/web-client/src/components/admin/ReplayViewer.tsx
@@ -6,6 +6,7 @@ import { CombatArrows } from '../combat/CombatArrows'
 import type { SpectatingState } from '@/store/slices'
 import { reconstructSnapshots, type ReplayData } from '@/replay/reconstructSnapshots.ts'
 import { buildReplayScenarioUrl } from '../scenario/shareScenario'
+import { useViewportSize } from '@/hooks/useResponsive.ts'
 
 // ============================================================================
 // Types
@@ -379,6 +380,11 @@ function ReplayView({
     return () => ro.disconnect()
   }, [])
 
+  // Same breakpoint as useResponsive's isMobile. The scenario/snapshot/share
+  // buttons don't fit the header on phones — hide them there (they're
+  // desktop-tooling features anyway).
+  const isMobile = useViewportSize().width < 640
+
   const [scenarioCopied, setScenarioCopied] = useState(false)
   const handleShareAsScenario = async () => {
     const url = buildReplayScenarioUrl(window.location.origin, snapshot.gameSessionId, currentStep)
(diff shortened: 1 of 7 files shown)


wingedsheep:
Follow-up from mobile-portrait feedback (commit 11274b6):

- **Cards no longer overlap the center turn HUD.** The 60px readability floor could produce a layout taller than the slot on phones; the floor is now adaptive — when 60px can't fit, the search reruns with the breathing margin sacrificed (10px kept to clear the StepStrip chevron) and shrinks cards (down to a 40px hard minimum) until the plan truly fits the slot. Verified geometrically: every row's box stays inside its slot with ~13px clearance to the HUD.
- **The floating pass button is compact on mobile** (label-sized, 32px tall); desktop keeps 170×42.

Mobile portrait with a board like the reported one (no HUD overlap, small button):



The "after" screenshots in the PR body were re-captured with the final code.

wingedsheep:
Second round of mobile-portrait fixes (commit e955789):

- **Player names** no longer fall off-screen — on phones they render small *under* the life orbs instead of beside them.
- **Battlefield can no longer paint over the center HUD.** The wrap search now goes up to 4 lines per row with a 32px minimum card width so crowded boards genuinely fit; the tight HUD gap grew to 16px to clear the StepStrip glow; and the battlefield slot clips its content (`overflow: clip`) as a hard guarantee for pathological boards. Verified numerically: every row's box stays inside its slot on the crowded scenario.
- **Pass button** shrinks further on phones: 28px tall and the "Pass to" prefix becomes an arrow ("→ Attackers").
- **Game log toggle is hidden on phones** and the hand row reclaims the side reservations, so hand cards use nearly the full viewport width (Vincent's suggestion).

Crowded board / hand-space result on 390×844:

 

Desktop is unchanged (side names, log button, full-size pass button).

wingedsheep:
Commit 8709db8: on phones the player's hand now shifts left into the space the (hidden) log toggle used to occupy — the row reserves 8px on the left and 110px on the right for the pass/auto cluster, and the fan centers between those reserves. A full hand spans from the left edge to just clear the buttons instead of sliding underneath them. Desktop reserves stay symmetric, so nothing moves there.

PR #736, opened 2026-06-14
Add Cyclone to Arabian Nights

bolav:
Cyclone ({2}{G}{G} Enchantment): "At the beginning of your upkeep, put a
wind counter on this enchantment, then sacrifice this enchantment unless
you pay {G} for each wind counter on it. If you pay, this enchantment
deals damage equal to the number of wind counters on it to each creature
and each player."

New/extended primitives:
- WIND counter type (passive marker) wired across SDK (CounterType enum +
  Counters constant) and the web client's passive-counter rendering
  (enums, passive-counter list, icon, badge palette).
- PayDynamicManaCostEffect / Effects.PayDynamicMana gain an optional
  `color`: pay the evaluated amount as that many colored symbols
  ({G}{G}…) instead of generic mana — the building block for "pay {G}
  for each X" (cumulative-upkeep-shaped costs). Honored by the executor,
  and by Gate.MayPay's affordability check and prompt label.

The card composes these: each upkeep adds a wind counter, then an
OptionalCostEffect (Gate.MayPay) pays {G}×counters → damage to each
creature (ForEachInGroup) and each player, or sacrifices Cyclone on
decline / when unaffordable. Both cost and damage scale off
DynamicAmounts.countersOnSelf(WIND).

Scenario test covers pay→damage, decline→sacrifice, and
unaffordable→sacrifice. SDK reference updated.

### DIFF
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 78 cards
 **Release Date:** December 17, 1993
-**Implemented:** 67 / 78
+**Implemented:** 69 / 78
 | Color | Count |
 |-------|-------|
 | White | 11 |
@@ -59,7 +59,7 @@
 - [x] Mijae Djinn
 - [x] Rukh Egg
 - [x] Ydwen Efreet
-- [ ] Cyclone
+- [x] Cyclone
 - [x] Desert Twister
 - [x] Drop of Honey
 - [x] Erhnam Djinn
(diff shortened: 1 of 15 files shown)


bolav:
Code Review — arn-cyclone (Cyclone, ARN #45)

A clean, well-composed card. The SDK change (a color parameter on PayDynamicMana so "pay {G} for each X" templates as {G}{G}… rather than generic mana) is a genuine reusable
generalization wired consistently through the gate executor, snapshot golden, docs, and client. Scenario tests cover pay / decline / can't-afford. Verified deep against engine
conventions:

- ForEach Self binding — inside ForEachInGroup, EffectTarget.Self rebinds to the current iterated creature (TargetResolutionUtils.kt:32, iterationTarget ?: sourceId), so damage
correctly hits each creature. Matches the canonical idiom used by Inferno / Bloodfire Colossus / Sandstorm. ✔
- YourUpkeep controller restriction — carried by StepEvent(player = Player.You) (omitted from the snapshot only because it's the serialization default) and enforced at
TriggerMatcher.kt:1082 (controllerId == activePlayerId). Does not fire on opponents' upkeeps. ✔
- ManaCost.parse edge cases — parse("") returns ZERO; Color.GREEN.symbol == 'G'; parse("{G}{G}") yields two green pips. ✔
- Web-client counter-vortex — confirmed a real glyph in mana-font v1.18.0. ✔

Findings

1. ~~`computedCostLabellacked theamount <= 0` guard its sibling call-sites have~~ — FIXED
rules-engine/.../composite/GatedEffectExecutor.kt:353. Was a latent landmine: a colored PayDynamicMana behind a MayPay gate that resolved to 0 rendered a cosmetic "Pay " label,
and a negative amount would have thrown IllegalArgumentException from "{G}".repeat(...) during label rendering. Not reachable from Cyclone (a wind counter is added first, so
amount ≥ 1) but now hardened for future colored dynamic costs. Resolved this turn.

No correctness bugs found in the card behavior. The sole finding has been fixed and the module recompiles.

PR #737, opened 2026-06-14
Fix AI infinite loop casting a 'target spell' instant (Reprieve)

bolav:
A single-target spell that targets "a spell on the stack" (Reprieve, a
counterspell, ...) is surfaced by the legal-action enumerator with
`targetRequirements = null` and only a flat `validTargets` list — i.e. with
no `targetZone`, since the enumerator only attaches per-requirement
`TargetInfo` (which carries the zone) for multi-requirement spells.

The AI's Strategist.toChosenTarget then fell through to its `else` branch and
wrapped every chosen entity as a `ChosenTarget.Permanent`. When the chosen
target was actually a spell on the stack, the engine's target validator
rejected the cast ("Target must be a spell on the stack"), the AI's safe
fallback logged the failure, and — because the same CastSpell remained a
legal action — the AI re-picked it forever.

Fix: in the `else` branch, consult authoritative game state via
`state.isSpellOnStack(entityId)` and wrap a stack entity as a
`ChosenTarget.Spell`. This is general (any "target spell" effect with a single
requirement), not a Reprieve special-case, and leaves player/permanent
targets unchanged. The engine enumerator already correctly gates the offer on
having a legal target (it does NOT offer the cast with an empty stack), so the
remaining gap was purely the AI mis-typing the chosen target.

Adds two regression tests in AIPlayerTest:
- the AI casts a "target spell" instant with a ChosenTarget.Spell and the
  resulting action is accepted by the engine (anti-loop guarantee);
- a no-target spell and an "up to one target" spell remain castable with no
  legal targets (guards against over-correcting into a false gate).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/Strategist.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/Strategist.kt
@@ -318,9 +318,20 @@ class Strategist(
         }
         "STACK" -> ChosenTarget.Spell(entityId)
         else -> {
+            // `targetZone` is only populated for multi-requirement spells; a single-target
+            // spell (Reprieve, a counterspell, …) surfaces `validTargets` with `targetZone = null`.
+            // So fall back to authoritative game state: a target that is a spell on the stack must
+            // become a `ChosenTarget.Spell`, not a `Permanent`. Wrapping it as a `Permanent` here
+            // made the engine reject the cast ("Target must be a spell on the stack"), and the AI
+            // re-picked the same failing action forever.
+            val isSpell = state.isSpellOnStack(entityId)
             val isPlayer = state.getEntity(entityId)
                 ?.get<com.wingedsheep.engine.state.components.identity.PlayerComponent>() != null
-            if (isPlayer) ChosenTarget.Player(entityId) else ChosenTarget.Permanent(entityId)
+            when {
+                isSpell -> ChosenTarget.Spell(entityId)
+                isPlayer -> ChosenTarget.Player(entityId)
+                else -> ChosenTarget.Permanent(entityId)
+            }
         }
     }
 
--- a/ai/src/test/kotlin/com/wingedsheep/ai/engine/AIPlayerTest.kt
+++ b/ai/src/test/kotlin/com/wingedsheep/ai/engine/AIPlayerTest.kt
@@ -23,6 +23,7 @@ import io.kotest.matchers.comparables.shouldBeGreaterThan
 import io.kotest.matchers.ints.shouldBeGreaterThan
 import io.kotest.matchers.nulls.shouldNotBeNull
 import io.kotest.matchers.shouldBe
+import io.kotest.matchers.types.shouldBeInstanceOf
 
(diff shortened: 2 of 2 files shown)


bolav:
Verdict

Behavior is correct and the fix is the right shape. A single-target "target spell" cast (Reprieve, counterspells) was being submitted by the AI with a ChosenTarget.Permanent
wrapper, rejected by the validator, and re-picked forever. The fix consults authoritative game state (state.isSpellOnStack) in the one code path that lacked target-zone
metadata. No new SDK types, no over-fitting to Reprieve — this is a generic AI-layer correction. Ship it.

What's good

- Root-caused, not patched at the symptom. The commit message correctly identifies why the single-target path has targetZone = null: the enumerator only attaches per-requirement
TargetInfo (with the zone) for multi-requirement spells, and targetInfosFor (Strategist.kt:269) synthesizes a targetZone = null TargetInfo for the single-target case. The fix
lands exactly in that gap.
- General, not a special-case. isSpellOnStack works for any "target spell" effect, and SpellOnStackComponent (the marker isSpellOnStack checks) correctly excludes abilities on
the stack, so abilities won't be mis-wrapped as Spell.
- Consistent with the existing "STACK" branch. Multi-requirement stack targets already became ChosenTarget.Spell via info.targetZone == "STACK"; the else branch now reaches the
same conclusion for single-target. Same semantics, two paths.
- The decisive test assertion is the right one. driver.submit(chosen).isSuccess shouldBe true (AIPlayerTest) proves the engine actually accepts the chosen action — that's the
real anti-loop guarantee, not just "the wrapper type is Spell". And the second test guards against over-correction (no-target / "up to one" casts stay legal), which is the
natural way this kind of fix regresses.
- No anti-corruption gap. ChosenTarget.Spell already has its ClientChosenTarget.Spell counterpart (ClientDTO.kt:733) and validator support — no new wiring required.

I ran the full AIPlayerTest suite (isolated Gradle home, since another agent's build held the journal lock): 16/16 pass, including both new regression tests.

Issues

Minor

- Strategist.kt:321-326 — the comment is good but slightly redundant with the (excellent) commit message. Fine to keep; the why here is genuinely non-obvious, so it clears the
CLAUDE.md bar.
- Test comment cites CR 601.2c for "blocks a cast whose target requirement has no legal target and requires at least one." I verified against MagicCompRules_20260417.txt: 601.2c
is the target-announcement step ("announces their choice of an appropriate object … for each target the spell requires"), which is the correct anchor for why min-targets-0
stays castable. Accurate enough — no change needed.
- Pre-existing, out of scope: a spell targeting "target activated/triggered ability on the stack" (Stifle-style) would fall through isSpellOnStack (false for abilities) to
Permanent and could loop the same way. There's no ChosenTarget.Ability variant, so this can't be fixed here, and such cards may not be implemented. Worth a mental note, not a
blocker for this PR.

Recommendation

Merge as-is. The fix is minimal, correctly scoped to the AI layer, general across all single-target "target spell" effects, and backed by a test that exercises the actual
rejection-loop path end-to-end. No changes requested.

PR #746, opened 2026-06-14
Add Eye for an Eye to Arabian Nights

bolav:
Eye for an Eye ({W}{W} Instant): "The next time a source of your choice would deal damage to you this turn, instead that source deals that much damage to you and Eye for an Eye deals that much damage to that source's controller."

This reuses the chosen-source damage-reaction machinery (Deflecting Palm) with a new `preventDamage` flag on PreventDamageEffect / PreventNextDamageFromChosenSourceShield (default true). When false (Eye for an Eye), the chosen source's damage is NOT prevented — it still hits the caster in full — but the shield is still consumed and its linked "deal that much to the source's controller" reaction still fires with the captured amount.

checkDeflectDamageShield now returns a sealed DeflectOutcome (Prevented short-circuits as before; Reflected merges the reaction event and lets the damage proceed), handled by both the DamageUtils and CombatDamageManager damage paths. Facade:
Effects.ReflectNextDamageFromChosenSourceToController().

Scenario tests cover both reflection paths: a chosen attacker's combat damage (CombatDamageManager) and a chosen direct-damage spell (DamageUtils.applyDamage) each still reduce the caster's life AND deal an equal amount to the source's controller. Also documents that the internal DamagePreventedEvent fires even when the damage is not prevented. SDK reference updated and a dev scenario added.


bolav:
Verdict

Behavior is correct and the SDK shape is right. This is a model composition: the whole feature is one generic preventDamage flag threaded through the existing chosen-source
shield machinery, plus a DeflectOutcome sealed type to fork prevent-vs-reflect at the two damage call sites. No card-specific engine code, no new Effect type. Ready to merge.

What's good

- Composition over a new type. Reuses the onPrevented → linked delayed trigger, DynamicAmounts.preventedDamage(), and EffectTarget.ControllerOfTriggeringEntity trio (same as
Deflecting Palm). The only genuinely new thing — "consume the shield but let damage through" — is a timing semantic the AST couldn't otherwise express, so it earns its place as
a primitive.
- DeflectOutcome sealed interface (DamageUtils.kt) cleanly models the fork and forces both call sites (DamageUtils.applyDamage, CombatDamageManager) into an exhaustive when.
- Both damage paths now tested. EyeForAnEyeScenarioTest covers combat (CombatDamageManager, Grizzly Bears) and non-combat (DamageUtils.applyDamage, Lightning Bolt) reflection —
each asserts the caster still takes the damage and the source's controller takes an equal amount. This was the one real gap from the prior round and it's closed.
- No spurious wiring. ClientEvent.kt maps DamagePreventedEvent -> null (no false "prevented" animation) and TriggerMatcher returns false for generic matching, so only the
linkId-matched delayed trigger fires. The event's KDoc now documents that it fires even when damage isn't prevented.
- Oracle text matches Scryfall exactly; canonical correctly placed in ARN (earliest of 8 printings); doc reference and ARN snapshot updated in the same commit.

Issues

None blocking or important.

Minor (author's discretion, unchanged from last round):
- PreventDamageEffect(preventDamage = false) reads as self-contradictory ("a prevent-damage effect that doesn't prevent"). Thoroughly documented via KDoc; renaming would churn
the SDK + serialization + ARN snapshot for cosmetic gain. Leaving it is reasonable.
- The reflect facade (Effects.kt) duplicates the two-line onPrevented = DealDamageEffect(...) block from DeflectNextDamageFromChosenSource. Trivial; the explicit form is more
readable than a .copy(preventDamage = false).

Tests

Both scenario cases run green (verified this session). The squash changed only commit topology, not file content, and main's additive delayed-trigger change doesn't touch this
path, so the result holds post-merge.

Recommendation

Merge as-is. The two minor naming nits are cosmetic and safe to leave.

PR #753, opened 2026-06-14
Add Guardian Beast to Arabian Nights

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/746

{3}{B} 2/4 Beast. As long as it's untapped, noncreature artifacts you
control can't be enchanted, have indestructible, and can't be gained
control of by other players.

Modeled as three SourceIsUntapped-gated GrantKeyword statics over a
noncreature-artifacts-you-control filter, composing existing primitives
rather than adding a bespoke ability. Introduces two non-displayed
restriction flags on AbilityFlag and a notCreature() filter helper:

- CANT_BE_ENCHANTED — enforced at Aura-cast target legality in
  TargetValidator (CR 303.4); guards targeting only, not non-targeted
  Aura attach/move.
- CANT_GAIN_CONTROL — enforced in the gain / by-most / exchange control
  executors (an exchange fails entirely if either side can't be gained).

Scenario test covers all three properties, the tap-state toggle, the
exchange and by-most branches, and that already-attached Auras are not
removed. Updates the SDK reference and the ARN snapshot.

### DIFF
--- a/backlog/sets/arabian-nights/cards.md
+++ b/backlog/sets/arabian-nights/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 78 cards
 **Release Date:** December 17, 1993
-**Implemented:** 70 / 78
+**Implemented:** 73 / 78
 | Color | Count |
 |-------|-------|
 | White | 11 |
@@ -40,7 +40,7 @@
 - [x] Cuombajj Witches
 - [x] El-Hajjâj
 - [x] Erg Raiders
-- [ ] Guardian Beast
+- [x] Guardian Beast
 - [x] Hasran Ogress
 - [x] Junún Efreet
 - [x] Juzám Djinn
@@ -55,7 +55,7 @@
 - [x] Desert Nomads
 - [x] Hurr Jackal
 - [x] Kird Ape
-- [ ] Magnetic Mountain
+- [x] Magnetic Mountain
 - [x] Mijae Djinn
 - [x] Rukh Egg
 - [x] Ydwen Efreet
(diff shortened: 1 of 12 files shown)


bolav:
Verdict: Ship it. Behavior is correct, the SDK shape is right, all prior findings are resolved, and the branch is a single clean commit (acc8f3514) sitting on the current main
tip with no drift. No new issues.

The review ran in the worktree ~/.claude/worktrees/argentum-engine/arn-guardian-beast-review; the branch already contains main (verified git merge-base --is-ancestor main HEAD),
so there's nothing to merge and no conflicts.

What's good

- Composes existing primitives — three SourceIsUntapped-gated GrantKeyword statics over Artifact.notCreature().youControl(), with the two novel restrictions as generic
AbilityFlags reusable by future cards. No card-specific Effect/StaticAbility/Modification type.
- Projection-correct enforcement at every site (projectedState.hasKeyword/getController).
- Rules-accurate: exchange full-no-op matches CR 701.12a; the targeting block matches CR 303.4/303.4a; the scope-note now correctly cites CR 303.4i/303.4j for the uncovered
non-targeted-attach cases.
- Thorough tests — all three properties, tap-toggle, both control branches, and the already-attached-Aura corner; snapshot + lint green; printing canonical correct.

Issues

- Blocking / Important: none.
- Minor (carry-over, discretionary): the only remaining unasserted corner is an artifact animated into a creature dropping out of the notCreature() grant via projection. The
filter is projection-evaluated (the indestructible test proves selection), so this is a confidence test, not a suspected bug. Optional. The earlier optional DRY suggestion (a
shared cantBeGainedControlOf extension across the three executors) also remains author's discretion — the duplication is one line per site.

Recommendation

Merge as-is. The two discretionary minors don't justify another revision. Both prior rounds' findings (Keyword→AbilityFlag, missing exchange/by-most tests, inline FQNs, CR
citation, doc sync, filter placement, stale snapshot) are fixed and verified.

PR #758, opened 2026-06-14
Add Oubliette to Arabian Nights

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/753

Implements Oubliette ({1}{B}{B} Enchantment) using its modern phasing
Oracle text. Adds a reusable linked-phasing primitive — the phasing
analogue of ExileUntilLeaves:

- PhaseOutUntilLeavesEffect / PhaseInLinkedToSourceEffect (+ Effects
  facades), executors, and PhasedOutComponent.phaseInOnSourceLeaves /
  tapOnPhaseIn fields. The target phases out linked to the source,
  skips its untap-step phase-in, and phases back in (tapped) on the
  source's leaves-battlefield trigger. Indirect phasing carries Auras/
  Equipment out and back with it.

Scenario tests cover phase-out/in tapped, the skipped untap step,
controller preservation, and indirect Aura phasing. CR phasing
citations verified against the Comprehensive Rules (702.26a untap-step
phase-in, 702.26d no control change, 702.26g indirect phasing).


bolav:
Verdict

Behavior is correct and matches the current phasing Oracle exactly; the SDK shape is right. All issues from the prior round are resolved. Merge-ready.

What's good

- Correct mechanic, correct primitive. Phasing (not exile) per current Oracle, implemented as a reusable PhaseOutUntilLeaves / PhaseInLinkedToSource pair — the phasing analogue
of ExileUntilLeaves. It deliberately does not shoehorn phasing into the linked-exile return machinery, which would be wrong: exile-return is a zone change with an ETB, whereas
phasing-in is a status flip (702.26c) that preserves counters, tapped state, and attachments for free. Keeping them as parallel-but-separate subsystems is the right call.
- Link stored on the phased-out permanent, not the source — makes the untap-step skip in BeginningPhaseManager (phaseInOnSourceLeaves == null) a local check instead of a reverse
lookup, and is idempotent (if (...has<PhasedOutComponent>()) continue guards double phase-out).
- Projection used correctly. PhaseOutUntilLeavesExecutor:43 reads controller via state.projectedState.getController; collectWithAttachments walks base AttachmentsComponent,
which is correct since attachment links are structural, not projected.
- No new GameEvent — reuses PhasedOutEvent/PhasedInEvent, so no ClientEvent.kt gap; new effects are @Serializable with @SerialName and round-trip (snapshot + lint green).
- Tests now cover every cited rule. Three scenarios: phase-out/in tapped, the skipped untap step (702.26a), controller preservation (702.26d), and indirect Aura phasing
(702.26g) — the last asserting the Aura vanishes, returns still attached, and does not tap while the creature does. Every CR number in the code is paired with a test.
- CR citations verified against MagicCompRules_20260417.txt: 702.26a (untap-step phase-in), 702.26d (no control change), 702.26g (indirect phasing of Auras/Equipment). The
earlier 702.26e/j mistakes — including the two pre-existing ones in PhaseOutExecutor/BeginningPhaseManager — are all corrected.
- Printing canonical correctly in ARN (earliest of 5 printings; check-card-printing clean).

Issues

None blocking, important, or minor. The two pre-existing phasing files (PhaseOutExecutor.kt, BeginningPhaseManager.kt) carry one-line citation corrections that are in-scope for
this change and verified; flagging only so you're aware the commit touches files outside the new card.

Recommendation

Ship it. Optional follow-up (not for this PR): the dev-only manual-scenarios/cards/o/oubliette.json stocks P2 with Holy Strength/Naturalize for manual play — now that the Aura
case has an automated test, that fixture is redundant but harmless. Leave the worktree in place until merged.

PR #795, opened 2026-06-16
Fix AI infinite loop on sacrifice-cost activated abilities

bolav:
The engine AI repeatedly activated an ability with a 'Sacrifice an artifact/creature' additional cost (Sage of Lat-Nam, Atog, Orcish Mechanics, Ashnod's Altar) but submitted a bare ActivateAbility with no sacrifice chosen. CostHandler.paySacrificeList then hard-failed 'Not enough sacrifice targets chosen (need 1, got 0)', the server's safeFallbackActions did not break the loop, and it spun forever.

Mirrors the existing exile-from-graveyard cost-choice pattern:
- ActivateAbilityHandler pauses for a SelectCards decision when a Sacrifice cost has more matching permanents than its count (a real choice), pushing a new ActivateAbilitySacrificeContinuation. The AI's existing respondToDecision path and the client's existing decision UI handle it (same as exile-from-graveyard).
- CostHandler.paySacrificeList auto-picks when the choice is forced (candidates <= required count) instead of failing, so the common case (e.g. exactly one artifact) resolves with no decision.
- ActivateAbilitySacrificeContinuation + resumer fill costPayment.sacrificedPermanents and re-dispatch; registered for serialization.
- GamePlayHandler.safeFallbackActions passes priority FIRST in a declare-step priority window (where PassPriority is legal), so a failed AI action can't loop on a do-nothing DeclareBlockers/Attackers no-op.

Adds SacrificeCostActivationScenarioTest (forced auto-pick, real choice pause+respond, AI bare-action regression).

### DIFF
--- a/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/GamePlayHandler.kt
+++ b/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/GamePlayHandler.kt
@@ -1159,11 +1159,24 @@ class GamePlayHandler(
         aiPlayerId: EntityId
     ): List<com.wingedsheep.engine.core.GameAction> {
         val pass = com.wingedsheep.engine.core.PassPriority(aiPlayerId)
+        // If PassPriority is currently a legal action, we're in a priority window (e.g. after
+        // blockers/attackers were already declared), not the actual declare-step decision. In that
+        // case a do-nothing DeclareBlockers/DeclareAttackers can succeed as a no-op and hand
+        // priority straight back to the AI → infinite loop. Pass FIRST so the loop breaks; keep the
+        // do-nothing declaration only as a fallback for the genuine declare-step decision (where
+        // PassPriority is NOT yet legal).
+        val passIsLegal = gameSession.getLegalActions(aiPlayerId).any {
+            it.action is com.wingedsheep.engine.core.PassPriority
+        }
         return when (gameSession.getStateForTesting()?.step) {
-            com.wingedsheep.sdk.core.Step.DECLARE_BLOCKERS ->
-                listOf(com.wingedsheep.engine.core.DeclareBlockers(aiPlayerId, emptyMap()), pass)
-            com.wingedsheep.sdk.core.Step.DECLARE_ATTACKERS ->
-                listOf(com.wingedsheep.engine.core.DeclareAttackers(aiPlayerId, emptyMap()), pass)
+            com.wingedsheep.sdk.core.Step.DECLARE_BLOCKERS -> {
+                val declare = com.wingedsheep.engine.core.DeclareBlockers(aiPlayerId, emptyMap())
+                if (passIsLegal) listOf(pass, declare) else listOf(declare, pass)
+            }
+            com.wingedsheep.sdk.core.Step.DECLARE_ATTACKERS -> {
+                val declare = com.wingedsheep.engine.core.DeclareAttackers(aiPlayerId, emptyMap())
+                if (passIsLegal) listOf(pass, declare) else listOf(declare, pass)
+            }
             else -> listOf(pass)
         }
     }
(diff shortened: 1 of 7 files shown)


bolav:
Verdict

Behavior is correct and the SDK shape is right — the fix follows the established per-cost-type continuation idiom (ActivateAbilityExileFromGraveyardContinuation →
ActivateAbilitySacrificeContinuation) rather than inventing a one-off, and the candidate computation is shared between the pause and the payment path so they can't drift. All
sacrifice-cost tests (mine + the 5 newly-merged ATQ cards = 12 tests) pass on the merged tree. Mergeable.

What's good

- Consistency between pause and pay. Both ActivateAbilityHandler (pause decision) and CostHandler.paySacrificeList (payment) compute candidates via the same
costHandler.findMatchingCardsUnified(...) with identical excludeSelf filtering — the options offered are exactly the options that will be accepted. This is the right way to
avoid the classic "offered X but payment rejects X" bug.
- Projection-correct. findMatchingCardsUnified passes state.projectedState into the evaluator, so an artifact-that's-currently-a-creature etc. is matched correctly.
- The forced/auto-pick narrowing is genuinely safe. Old paySacrificeList failed on empty choices regardless; the new code only adds success for the forced case (candidates ==
requiredCount) and still fails when candidates > requiredCount. It cannot mis-sacrifice — when count equals candidates there is exactly one legal payment.
- Multi-candidate AI loop is actually closed, not just the single-candidate one. The pause prompt is "Select N permanent(s) to sacrifice for <card>", and the AI's
DecisionResponder.respondSelectCards has an isSacrifice branch (keyed on the prompt containing "sacrifice") that ranks by BoardPresence.permanentValue and returns a valid
response. So the AI answers the decision instead of re-submitting a bare action.
- The new ATQ cards are untouched — they all pre-fill sacrificedPermanents, so alreadySacrificing short-circuits the pause. Verified by their passing tests.
- Tests cover the three real axes (forced auto-pick, multi-candidate pause+respond, bare-payment regression) and the doc comments explain why (the loop), not the code.

Issues

Important

- Headline "AI infinite loop" claim has no test for the multi-candidate AI path. SacrificeCostActivationScenarioTest:122 (the "AI / loop regression" test) exercises only the
single-candidate case, which is fixed by the auto-pick — not by the AI responding to a decision. The genuinely dangerous original case (AI controls several artifacts, must
answer a SelectCardsDecision) is verified only by reading respondSelectCards. The "Choice case" test submits the selection manually, not via the AI. Since the fix lives partly
in the ai module and GamePlayHandler, consider a game-server scenario that drives the AI to activate Sage/Atog with 2+ candidates and asserts the game advances. Otherwise the
loop fix for the case that motivated the commit is asserted-by-reasoning.

Minor

- AI routing is coupled to the prompt string. The multi-candidate fix only stays optimal because the prompt contains the literal word "sacrifice" (ActivateAbilityHandler builds
it, respondSelectCards greps for it). A future prompt-wording change would silently degrade the AI's pick (it'd fall through to a generic branch). No loop risk (it still returns
a response), but worth a one-line comment at the prompt site noting the AI keys on "sacrifice".
- CostHandler.paySacrificeList > requiredCount branch is dead in the activated-ability flow and its message lies. It returns "...need $requiredCount, got 0" (CostHandler.kt:676
area) — the count isn't 0, it's candidates.size. It's only reachable if some non-activated caller hits paySacrificeList with empty choices and surplus candidates (the pause
never ran). Fine as a defensive net, but the "got 0" wording will mislead anyone who hits it; say "no selection supplied for a multi-candidate sacrifice" instead.
- extractSacrificeCost doesn't recurse into nested Composites (ActivateAbilityHandler helper) and returns only the first Sacrifice atom. No current card needs more (SDK
composites are flat: Tap + Sacrifice), and Sage confirms the flat case works — just flagging that a hypothetical double-sacrifice or nested-composite cost would only pause for
the first.

Recommendation

Merge as-is is acceptable. If you want to harden the commit to match its own title, add one game-server AI scenario for the multi-candidate sacrifice activation (the case the
loop-fix reasoning rests on), and fix the misleading "got 0" message. Everything else is optional polish.

PR #842, opened 2026-06-18
LTR: engine & SDK foundation for the rest of the cards

bolav:
Description:
Foundation PR for the Lord of the Rings set — engine + SDK only,
no card definitions. First in a stacked series of 8 (PR2–8 add cards on top);
merge in order, no squash.

What's here:
- New SDK primitives the set needs: guess-top-card, move-counters, player
  protection, copy-each-target-spell, damage-can't-be-prevented, plus
  supporting conditions/triggers/filters. Card-decoupled — the engine compiles
  and passes with zero LTR cards registered.
- Anti-corruption wiring for the two new events (CardPlayedFromPermission,
  SagaChapterResolved) and six new continuations, all serialized + dispatched.
- docs/card-sdk-language-reference.md updated for the new types.

Fixes:
- CreateTokenExecutor: an additional-token replacement (Peregrin Took's Food)
  was added per token instead of once per creation event (CR 614); a custom
  multi-token batch double-added. Now fires once per event.
- MoveCollectionExecutor: a Food sacrificed via MoveType.Sacrifice didn't tag
  the controller for "you sacrificed a Food this turn" payoffs. Now carried.
- CR citation corrections (113.2→113.7, 603.10e→603.10a), verified against the
  2026-04-17 Comprehensive Rules.

Tests: engine-level regression tests (inline cards) for the additional-token
batch, player-protection damage prevention, and the MoveCollection Food tag.
just build is green (engine + suite).


bolav:
Verdict

Correct and merge-ready. No new SDK types are introduced, so the elegance lens is a non-issue — these are a localized correctness fix, four CR-number corrections, and three
regression tests. All CR corrections were re-verified against MagicCompRules 20260417.txt, and every test was run (the MoveCollectionExecutor one was proven to fail without the
fix).

What's good

- The MoveCollectionExecutor fix reads the single source of truth. It carries over whatever trackPermanentSacrifice decided
(tracked.getEntity(controllerId).has<SacrificedFoodThisTurnComponent>()) rather than re-deriving Food-ness from the moved cards — so it can't drift from the canonical tagging
logic, and the marker being a data object makes the re-with idempotent.
- All three tests exercise the real corner, with inline cards (correct ScenarioTestBase scope for a card-less PR): the additional-token test hits a custom multi-token
CreateTokenEffect batch (CR 614 once-per-event, the only shape that surfaced the double-add); the player-protection test uses a non-targeted Player.Each source to hit the
DamageUtils prevention branch specifically, not the already-covered targeting path; the MoveCollection test drives MoveCollectionEffect(MoveType.Sacrifice) and asserts the tag
persists. All pass; the last fails without the fix (verified by revert).
- CR corrections are accurate: 113.7 ("The source of an ability is the object that generated it") and 603.10a (leaves-the-battlefield look-back) both match the official text;
the CastPermissionUtils edit adds a one-line gloss that earns its keep.

Issues

Blocking / Important — none.

Minor
- MoveCollectionExecutor.kt:730-735 — the explanatory comment is four lines. The why (the tag lives on a different entity and is otherwise dropped) is genuinely non-obvious so a
comment is warranted, but it could tighten to two lines. Author's discretion.
- 112.7a / 112.3b stale citations remain — these don't exist in this CR edition (LKI is 608.2h; copy-ceases-to-exist is 707.10a), and this branch adds 7 new uses, but they
mirror a ~16-use pre-existing convention on main. I deliberately left them to avoid diverging from main; the honest follow-up is a separate repo-wide sweep, not a partial fix
here. This is the one "should eventually fix" item.
- Test doc-comments reference the pre-fix behavior ("double-added per batch"). For regression tests, naming what they guard is appropriate, but it brushes against CLAUDE.md's
"no 'added for X' notes" — fine to keep.

Recommendation

Ready to commit to ltr-pr-1. Two operational notes carried over from the last round: (1) committing here means rebasing the stacked ltr-pr-2…8 onto the new tip; (2) the
112.7a/112.3b sweep is worth a standalone cleanup PR against main. Want me to commit these (fix + 3 tests) to ltr-pr-1 and cascade the rebase, or leave them staged for you?

PR #845, opened 2026-06-18
LTR: add 13 Lord of the Rings cards (The One Ring, Sméagol, Glorfindel, …)

bolav:
Adds 13 cards to The Lord of the Rings: Tales of Middle-earth (LTR). Stacked on
`ltr-pr-3` — review the 14 commits since that base. Pure card content: no new SDK
types, every card composes existing primitives.

## Cards
- **The One Ring** + **A-The One Ring** (Alchemy) — indestructible, player-level
  protection from everything (cast-gated), escalating burden draw/upkeep loss
- **Sméagol, Helpful Guide** — end-step Ring tempt + steal-a-land reveal
- **Galadriel of Lothlórien** — Ring-tempt scry; reveal-top-land-on-scry
- **The Ring Goes South** — Ring tempt + reveal X lands (X = legendary creatures)
- **Breaking of the Fellowship** — opponent's creature fights another, then tempt
- **Legolas, Master Archer** — cast-targeting triggers (counter / power-damage)
- **Glorfindel, Dauntless Rescuer** — on-scry modal pump + block restrictions
- **Gandalf's Sanction** — graveyard-scaled damage, excess to controller
- **Display of Power** — copy any number of instant/sorcery spells; can't be copied
- **Phial of Galadriel** — draw/life-gain replacements + any-color mana
- **Barrow-Blade** — equip; blocked creature loses all abilities
- **Troll of Khazad-dûm** — can't be blocked by fewer than three; swampcycling

## Testing
- A scenario test per card; rule corners pinned (protection blocks targeting,
  burden growth/608.2c, intervening-if/603.4, mode-2 block cap, excess damage,
  multi-spell copy, non-land reveal stays on top).
- All 13 scenario tests pass; CardDefinitionSnapshotTest + CardLintTest green.

All cited CR numbers (702.16, 603.4, 608.2c, 120.4a, 707.10) verified against the
Comprehensive Rules; oracle text and earliest-printing placement checked on Scryfall.

### DIFF
--- a/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/ltr/cards/AOneRing.kt
+++ b/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/ltr/cards/AOneRing.kt
@@ -1,0 +1,72 @@
+package com.wingedsheep.mtg.sets.definitions.ltr.cards
+
+import com.wingedsheep.sdk.core.Counters
+import com.wingedsheep.sdk.core.Keyword
+import com.wingedsheep.sdk.dsl.Conditions
+import com.wingedsheep.sdk.dsl.Costs
+import com.wingedsheep.sdk.dsl.DynamicAmounts
+import com.wingedsheep.sdk.dsl.Effects
+import com.wingedsheep.sdk.dsl.Triggers
+import com.wingedsheep.sdk.dsl.card
+import com.wingedsheep.sdk.model.Rarity
+import com.wingedsheep.sdk.scripting.events.CounterTypeFilter
+import com.wingedsheep.sdk.scripting.targets.EffectTarget
+
+/**
+ * A-The One Ring (Alchemy rebalance of [TheOneRing])
+ * {4}
+ * Legendary Artifact
+ *
+ * Indestructible
+ * When The One Ring enters, if you cast it, you gain protection from everything until your next turn.
+ * At the beginning of your upkeep, you lose 1 life for each burden counter on The One Ring.
+ * {1}, {T}: Put a burden counter on The One Ring, then draw a card for each burden counter on The One Ring.
+ *
+ * Identical to the printed [TheOneRing] except the draw ability gains a {1} mana cost
+ * (the Alchemy nerf). Reuses Gap 8 player-level protection — see [TheOneRing] for the full
+ * engine notes.
+ */
+val AOneRing = card("A-The One Ring") {
+    manaCost = "{4}"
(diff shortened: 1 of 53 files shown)


bolav:
Verdict

Ship it. Round 1 already established the 13 cards are behaviorally correct (oracle text, CR citations, and printing placement all verified against Scryfall and the CR). This
round only needed to vet the fix commit — all five minor issues are resolved correctly, and no new issues were introduced.

Fix-commit assessment

- Legolas filter (the only functional change). Creature.opponentControls() → withControllerPredicate(ControllerPredicate.Not(ControlledByYou)) correctly models "a creature you
don't control." I traced the trigger-match path: SpellCastPredicate.TargetsMatching (TriggerMatcher.kt:1340) routes through PredicateEvaluator.matches with the ability's
controller as "you", and ControllerPredicate.Not is evaluated via projected state (PredicateEvaluator.kt:616) — so it's both semantically right and projection-correct. Both
Legolas tests still pass (self-target → counter; opponent-creature target → damage), and the golden change is exactly and only that predicate.
- The One Ring — both rule corners now have paired tests. The growth test activates across two turns (untapping via the player-2-active trick) and asserts the second activation
draws 2, exercising the CR 608.2c counter-before-draw ordering. The cast-gate test uses a minimal inline Test Reanimate (Effects.Move(..., BATTLEFIELD, fromZone = GRAVEYARD),
targeted with the correct ChosenTarget.Card(..., GRAVEYARD)) to make the Ring enter without being cast, then asserts it's on the battlefield (ETB evaluated) but grants no
PlayerProtectionComponent — a genuine test of the 603.4 intervening-if negative branch, not just of the builder.
- Galadriel non-land test. Reveals a non-land, asserts it does not enter and remains in the library — exactly the gather/reveal-displacement risk I flagged.
- Breaking of the Fellowship. Now asserts TheRingComponent.temptCount == 1, so the trailing "The Ring tempts you" is exercised.
- Comment hygiene. Glorfindel and The One Ring kdoc trimmed of the drift-prone cross-file details (BlockPhaseManager wiring; validator/enumerator/executor list) while keeping
the modeling "why."

What's good (unchanged from round 1)

Pure composition — no new SDK type anywhere in this diff. Every cited CR rule now has a paired test (702.16, 120.4a, 707.10 from round 1; 608.2c and 603.4 added this round).
Golden is additive; existing cards byte-identical. All LTR scenario tests + snapshot + lint green.

Issues

Blocking / Important / Minor: none. All round-1 findings are resolved; no regressions.

Recommendation

Merge-ready.

PR #919, opened 2026-06-21
LTR: card corrections, field-verification net, and engine bug fixes

bolav:
Completes the Lord of the Rings set work: corrections to existing card definitions, the remaining scenario tests, the LtrCardFieldVerificationTest net (every registered LTR card checked against Scryfall), and manual scenarios. Along the way, four engine/SDK bugs surfaced by these cards are fixed:

- Equipment/Aura stays attached after its host blinks (Meneldor, Swift Savior). moveToZone reuses the host's EntityId across an exile->battlefield round-trip, so the unattached-permanents SBA (keyed on the host id still being on the battlefield) left the attachment in place. Mark the leaving host's attachments with a transient AttachmentHostLeftComponent and detach on that marker after triggers are detected (so Forebears Blade's ATTACHED-binding dies/leaves trigger still fires). CR 400.7 / 704.5m-n.

- The One Ring's "protection from everything" did not prevent combat damage. Combat damage is applied by CombatDamageManager (not DamageUtils), which never consulted PlayerProtectionRules. Add a PlayerProtectionModifier to the combat-damage Modify phase; honors global "damage can't be prevented".

- Elrond, Master of Healing's scry trigger offered 0 targets. The cap dynamicMaxCount = ContextProperty(TRIGGER_SCRY_COUNT) resolved to 0 because snapshotDynamicCount built the EffectContext without triggerScryCount; pass it through. Also stop TargetObject.description from doubling "up to " when the author-supplied id already begins with the quantifier (21 cards).

- Earlier ltr-pr fixes folded in: Wizard's Rockets X choice, Spiteful Banditry batched Treasure, Pippin's Bravery unpayable-sacrifice gate, Ring-bearer designation not surviving a blink, Bilbo's Ring "attacks alone" requires, Elrond becomes-target statePredicates, Friendly Rivalry TargetOther.

Covered by the LTR scenario tests, PlayerProtectionDamageScenarioTest, MeneldorBilbosRingDetachScenarioTest, and ElrondMasterOfHealingScryTargetCountTest.


bolav:
Verdict

Solid, ship-quality PR. The behavior is right and the SDK shape is right: it adds zero new SDK types, reuses existing primitives for both behavioral card changes, fixes four
genuine engine bugs with tight tests, and corrects a previously-wrong CR citation. I ran the full scenario set (11 tests), the snapshot net, and the lint net — all green — and
verified every changed oracle text against Scryfall and every CR number against the local rules file.

What's good (keep all of it)

- Pippin's Bravery rewrite (PippinsBravery.kt:41): converting "you may sacrifice a Food" from OptionalCostEffect to IfYouDoEffect over a Gather → ChooseUpTo(1) → Sacrifice
pipeline is more rules-correct — it's a resolution-time optional action, not a cost — and it composes existing primitives rather than adding a type. Gating +4/+4 on
SuccessCriterion.CollectionNonEmpty("sacrificedFood") cleanly distinguishes "sacrificed" from "declined / none controlled." Three tests cover all three branches, including the
no-prompt-when-no-Food path.
- Spiteful Banditry (SpitefulBanditry.kt:39): swapping the per-creature leavesBattlefield(binding = ANY) for the existing batched
Triggers.OneOrMoreCreaturesAnOpponentControlsDie() is the correct fix — a per-creature trigger over-fires on a board wipe and fights oncePerTurn. The two tests prove
single-Treasure batching and opponent-scoping. (Confirmed against Scryfall the Treasure is untapped, so CreateTreasure(1) is right.)
- Attachment fall-off fix (AttachmentHostLeftComponent, ZoneMovementUtils.markAttachmentsHostLeft, UnattachedAurasCheck): the leave-time marker is the right mechanism for the
genuinely hard case — moveToZone reuses the host's EntityId across a blink, so the SBA's id-absence check can't see "host left and returned as a new object." Deliberately
leaving AttachedToComponent intact so aurasByTarget still surfaces ATTACHED-binding dies/leaves triggers (Forebears Blade), then consuming the marker in the SBA after trigger
detection, is correctly ordered. Marker is registered in Serialization.kt and stripped on re-entry in stripBattlefieldComponents. The Meneldor test drives the real SBA pipeline
(not direct moveToZone) and stamps a RingBearerComponent to prove the blink actually happened — nice touch.
- The old comment's CR 704.5q was wrong (704.5q is +1/+1 vs −1/−1 counter annihilation); the new code correctly cites 704.5m (Aura → graveyard) and 704.5n (Equipment unattaches,
stays). Both verified in MagicCompRules_20260417.txt.
- PlayerProtectionModifier (CombatDamagePipeline.kt:117): plugs the real bug — CombatDamageManager.applyDamageToPlayer reduces life directly and never consults
PlayerProtectionRules, so combat damage to a "protection from everything" player (The One Ring) wasn't prevented. Uses projected controller for the source, mirrors the keyword
ProtectionModifier, and respects global prevention-disable (Fear, Fire, Foes!). The test trio (non-targeted damage, combat damage, and a no-protection control) is exactly right.
- Elrond scry-count fix (TriggerProcessor.kt:884): threading triggerScryCount into the dynamic-cap context so "up to X target creatures" resolves to the cards scried instead of
0; paired with the TargetObject.description de-dup that prevents "up to up to X." Both exercised by the Elrond test (maxTargets == 2, description asserts).
- Oracle/flavor/ruling corrections: every one of the six oracle-text edits (Orcish Bowmasters, Fall of Cair Andros, Surrounded by Orcs, Foray of Orcs, Treason of Isengard,
Rohirrim Lancer) matches Scryfall's oracle_text verbatim — including the subtle asymmetry that Orcish Bowmasters carries no amass reminder while the amass spells do. The One
Ring rulings match Gatherer.

Issues

Blocking: none.

Important: none.

Minor (author's discretion):
- PippinsBravery.kt:41 — the inline Gather→Select→Move composition is correct but verbose. If "you may sacrifice a [type]; if you do…" recurs (it will — it's a common LTR/MTG
shape), consider lifting it into a Patterns.Group/Patterns.Mechanic helper so the next card is a one-liner. Not needed for this PR.
- The BilbosRingAttacksAlone, WizardsRocketsXChoice, FriendlyRivalry, and RingBearer… tests guard behavior whose engine code is not in this diff (e.g. AttachmentTriggerDetector
requires, the X-choice pause). They're pure regression nets — valuable, but worth knowing they don't cover code changed in this commit, so a green run here doesn't validate
those paths beyond their current state.
- No docs/card-sdk-language-reference.md update is required: the diff adds no SDK-surface type (the batched trigger and IfYouDoEffect/Gather/Select/Move are all pre-existing;
the new pieces are engine-internal). Confirming this matches the CLAUDE.md hard rule.

Recommendation

Merge as-is. The two behavioral card rewrites move toward more correct rules modeling via existing primitives (exactly the SDK-elegance direction), the four engine fixes are
real and well-tested, the wrong CR citation is corrected, and all nets are green. The only optional follow-up is extracting the Pippin's Bravery "you may sacrifice → if you do"
composition into a Patterns helper if it recurs.

PR #930, opened 2026-06-23
Antiquities batch: artifact tap/activate triggers + untap-control family

bolav:
Implements 8 Antiquities cards (+ a 10E reprint row) across two related mechanics,
built by composing/extending existing SDK primitives rather than adding card-specific types.

### Cards
- **Haunting Wind, Powerleech, Artifact Possession** — "whenever an artifact becomes
  tapped or its ability is activated without {T} in its cost" punisher template.
- **Damping Field** — "players can't untap more than one artifact during their untap steps."
- **Phyrexian Gremlins** — taps a target artifact and locks it while Gremlins stays tapped.
- **Ashnod's Battle Gear (+2/-2), Tawnos's Weaponry (+1/+1)** — tap-locked stat buffs.
- **Colossus of Sardia** — mandatory "doesn't untap" + pay-{9}-during-upkeep to untap;
  canonical lives here (earliest printing), with a 10E `Printing` row.

### SDK / engine
- `ModifyStats` gains a `duration` param threading into the existing `ModifyStatsEffect`
  (reuses the `WhileSourceTapped` one-way latch).
- `Triggers.activatesAbilityWithoutTap` + `requireNoTapInCost`/`sourceFilter` flags on
  `AbilityActivatedEvent`, distinguishing "without {T} in cost" from "isn't a mana ability".
- New `UntapLimitPerStep(filter, max)` static ability (global untap-count cap).
- `GrantKeywordExecutor` generalized from creature-only to any battlefield permanent.
- The {T} TappedEvent of a mana ability now runs trigger detection (fixes "becomes tapped"
  triggers not firing when tapping for mana).

### Tests
Engine-level `ArtifactTapActivateTriggerScenarioTest` pins every axis (tap vs ability half,
mana vs non-mana, {T} vs non-{T}, source filter, opponent scoping) plus per-card scenario
tests. SDK reference doc, snapshot golden, and backlog updated. All cited CR numbers
(502.3, 605.1, 605.3) verified.

### DIFF
--- a/backlog/sets/antiquities/TODO.md
+++ b/backlog/sets/antiquities/TODO.md
@@ -114,13 +114,17 @@ confirm the noted clause during `add-card`; if it surprises you, move it to List
 
 Each group is a candidate single feature PR. Confirm with `add-card` before building.
 
-### Artifact-tapped / artifact-ability-activated trigger
+### Artifact-tapped / artifact-ability-activated trigger ✅ DONE
 *Trigger: "whenever an artifact becomes tapped, or a player activates an artifact's ability
-without `{T}` in its cost."* No such trigger today (`becomesTapped` exists but not typed to
-artifacts, and there's no ability-activated-without-tap event).
-- **Haunting Wind** — global; 1 damage to the artifact's controller.
-- **Powerleech** — opponents' artifacts only; you gain 1 life.
-- **Artifact Possession** — single enchanted artifact; 2 damage to its controller.
+without `{T}` in its cost."* Implemented: the tap half reuses `Triggers.becomesTapped(filter)`;
+the ability half is `Triggers.activatesAbilityWithoutTap(player, sourceFilter, binding)`, backed by
+`EventPattern.AbilityActivatedEvent(sourceFilter, requireNoTapInCost)`. The engine now emits
+`AbilityActivatedEvent` (carrying `costsTap`/`isManaAbility`) for every activated ability whose cost
+lacks `{T}` — mana abilities included — so the literal "{T}-in-cost" wording is honored, distinct
+from the existing "isn't a mana ability" wording.
+- [x] **Haunting Wind** — global; 1 damage to the artifact's controller.
+- [x] **Powerleech** — opponents' artifacts only; you gain 1 life.
+- [x] **Artifact Possession** — single enchanted artifact (ATTACHED); 2 damage to its controller.
 
 ### "Doesn't untap" control family (untap restrictions / pay-to-untap / tap-locked buffs)
 No "you may choose not to untap", "doesn't untap during your untap step", per-permanent untap
--- a/backlog/sets/antiquities/cards.md
+++ b/backlog/sets/antiquities/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 85 cards
 **Release Date:** March 4, 1994
-**Implemented:** 48 / 85
+**Implemented:** 51 / 85
 | Color       | Count |
 |-------------|-------|
 | White       | 7     |
@@ -39,9 +39,9 @@
 - [ ] Transmute Artifact
 
 ### Black
-- [ ] Artifact Possession
+- [x] Artifact Possession
 - [x] Gate to Phyrexia
-- [ ] Haunting Wind
+- [x] Haunting Wind
 - [ ] Phyrexian Gremlins
 - [ ] Priest of Yawgmoth
 - [ ] Xenic Poltergeist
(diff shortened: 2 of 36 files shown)


bolav:
1. Verdict

  Behavior is correct and the SDK shape is right. This is a careful, well-tested batch: eight ATQ cards + one 10E reprint row, built almost entirely by composing or minimally extending existing primitives rather than minting card-specific types. Mergeable; only minor
  notes below.

  2. What's good (keep this)

  - ModifyStats gained a duration param that threads into the pre-existing ModifyStatsEffect.duration — no new effect type for the tap-locked buffs (Ashnod's Battle Gear, Tawnos's Weaponry). The WhileSourceTapped one-way latch
  (CleanupPhaseManager.cleanupWhileSourceTappedEffects) is reused, not reinvented.
  - The "tap / activate without {T}" trigger family is two flags on the existing AbilityActivatedEvent / EventPattern.AbilityActivatedEvent (requireNoTapInCost, sourceFilter) plus one DSL sugar (activatesAbilityWithoutTap) — no new event or trigger type. Crucially it
  correctly disambiguates the two distinct MTG wordings ("isn't a mana ability" vs "without {T} in its activation cost"), and the doc/comments spell out exactly why the global form uses PlayerRef(TriggeringPlayer) (the ability entity is absent for mana abilities) while
  the ATTACHED form uses ControllerOfTriggeringEntity.
  - UntapLimitPerStep(filter, max) is genuinely novel and properly parameterized — a global untap-count cap with a keep-tapped player decision that no existing primitive expresses, and it's generic (filter + max) rather than hardcoded to "1 artifact". activeUntapLimits
  gathers it from the registry exactly like the existing Seedborn/Ivorytusk path, so it's consistent with convention.
  - GrantKeywordExecutor generalized from creature-only to any battlefield permanent is the right minimal change (combat keywords are inert no-ops on noncreatures; DOESNT_UNTAP is meaningful), enabling Phyrexian Gremlins to compose Tap then GrantKeyword(DOESNT_UNTAP,
  WhileSourceTapped).
  - A latent bug got fixed in passing: tapping an artifact for a {T} mana ability now runs trigger detection over the {T} TappedEvent (activationCostEvents), so "whenever an artifact becomes tapped" finally fires on mana-tapping — pinned by
  ArtifactTapActivateTriggerScenarioTest.
  - Test coverage is excellent and corner-targeted. ArtifactTapActivateTriggerScenarioTest is engine-level (not card-level) and pins every axis: tap-half vs ability-half, mana vs non-mana, {T} vs non-{T} cost, source filter rejecting a creature, opponent-scoping.
  Per-card tests cover the latch drop on untap, no-stacking across turns, the inert-cap case, and restriction gating. FlamescrollCelebrant/AshnodsAltar regressions stay green.
  - SDK reference doc and backlog updated in the same change; printing placement correct (canonical in ATQ, the earliest real set; 10E carries only a Printing row).

  3. Issues

  Blocking — none.

  Important — none.

  Minor
  1. Client-log behavior change for existing non-{T} mana abilities. ActivateAbilityHandler now emits AbilityActivatedEvent for any mana ability whose cost lacks {T} (sac-for-mana: PhyrexianAltar, BogInitiate, PhyrexianLens, …). ClientEvent.kt:905 maps that to a
  client-visible AbilityActivated with abilityDescription = "". Previously these produced no ability-activated event. It's arguably more correct, but it's a UI change beyond the ATQ scope with no test asserting the client-facing result, and the empty description reads
  oddly in a log. Worth a conscious confirm (and maybe populating a description) — rules-engine/.../ActivateAbilityHandler.kt:1240 and ClientEvent.kt:905.
  2. forcedKeepCount sums shortfalls across distinct filters independently (BeginningPhaseManager.kt:~127). For Damping Field (single filter) this is exact. With a second UntapLimitPerStep of a different filter that overlaps the option pool, minSelections could
  over-count; and when the keep-tapped pool also contains a non-matching MAY_NOT_UNTAP permanent, minSelections can be satisfied while still leaving a filter over its cap. The resumer catches the latter fail-closed with a clear re-prompt
  (SacrificeAndPayContinuationResumer.kt:784), and it's documented — so correctness holds, but the single-filter assumption deserves a one-line comment at the forcedKeepCount site so the next author doesn't extend it blindly.

  4. Recommendation

  Ship it. Optionally address the two minor notes inline: drop a one-line comment at forcedKeepCount flagging the single-filter assumption, and decide whether the new AbilityActivated client event for non-{T} mana abilities should carry a description (or be suppressed)
  — neither blocks merge.

PR #932, opened 2026-06-23
ATQ batch 2: choose-a-number CDA + counter↔token provenance (6 cards)

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/930

Adds 6 Antiquities cards, introducing a few reusable SDK primitives rather
than card-specific types.

### Cards
- **Shapeshifter** — as-enters + upkeep "choose a number 0–7"; P/T = n / 7−n
- **Primal Clay** — as-enters mode choice (3/3 · 2/2 flying · 1/6 defender)
- **Tetravus** — +1/+1 counters ↔ Tetravite tokens, both directions
- **Cursed Rack** — chosen opponent's max hand size is 4
- **Priest of Yawgmoth** — pure composition, no engine work
- **Hurkyl's Recall** — return all artifacts target player *owns* to hand
- Reprints: Hurkyl's Recall (10E), Primal Clay (M13)

### New SDK primitives
- `EntersWithChoice` gains `ChoiceType.NUMBER` (reuses the entry-choice path)
- `ChooseNumberForSourceEffect` — durable, re-settable number on a permanent,
  read by a CDA via `DynamicAmount.CastChoice` (the upkeep re-choice)
- `ConvertCountersToTokensEffect` — remove N counters, mint N stamped tokens
- `StatePredicate.CreatedBySource` + `CreatedByComponent` — token provenance
  ("tokens created with this creature"); the reverse direction composes from atoms
- `ControllerPredicate.OwnedByTargetPlayer` — owner axis for "target player owns"

### Notes
- SDK reference doc updated in the same change
- New scenario tests per card + a provenance feature test (two same-named
  sources can't reabsorb each other's tokens); snapshot + lint nets re-blessed
- Stacked on `atq-batch-1-no-tap-untap`; rebase onto main before merge


bolav:
Verdict

Behavior is correct and the SDK shape is right. This batch adds five new primitives across six cards, and each new type earns its place — most notably the work reuses existing
infrastructure (EntersWithChoice, CastChoicesComponent, DynamicAmount.CastChoice, the decision helper) instead of growing parallel machinery. I reviewed in place; the branch is
clean and I did not merge origin/main (you asked for the diff against the stacked base atq-batch-1-no-tap-untap, not main — worth a separate rebase check before this lands on
main). All 8 new scenario classes, the corpus snapshot, and the lint net pass under an isolated Gradle invocation. Issue 1 below has been fixed and committed (e938e3e1c).

What's good (keep this)

- EntersWithChoice gained ChoiceType.NUMBER rather than a new replacement type. Folding the as-enters number choice into the existing parameterized replacement (with
minValue/maxValue) is exactly the right call — it reuses the entire spell/battlefield pause+resume path (StackResolver, PermanentEntryReplacements, both resumers in
ModalAndCloneContinuationResumer).
- The forward/reverse asymmetry on Tetravus is principled. Tokens→counters composes from atoms (gather → chooseAnyNumber → exile → AddDynamicCounters), and only the
counters→tokens direction got a dedicated ConvertCountersToTokensEffect — justified because the prompt needs a dynamic max (current counter count) and a dynamic removal, neither
of which ChooseNumberThenEffect/RemoveCountersEffect can express today. The KDoc even explains why the reverse needs no primitive.
- CounterTokenProvenanceScenarioTest proves the exact gap. The third case (two same-named sources; B can't reabsorb A's tokens) is the precise scenario "{filter} tokens you
control" cannot express — it's the whole reason StatePredicate.CreatedBySource/CreatedByComponent exist, and it's tested directly. Good adversarial framing.
- Hurkyl's owner-vs-controller corners are covered four ways (HurkylsRecallScenarioTest), including the post-control-change case where owner ≠ controller and the "target the
controller, not the owner" negative — which is the only thing that distinguishes OwnedByTargetPlayer from ControlledByTargetPlayer.
- Shapeshifter's CDA timing is handled as a true as-enters replacement (CR 614.1c) — the test asserts the prompt precedes the permanent existing, so it never flickers at default
P/T. Source-relative predicate plumbing (AffectsFilterResolver, TriggerMatcher, BeginningPhaseManager exhaustive lists) was all updated for the new predicate.

Issues

Minor

1. ~~**Tetravus over-constrains the reabsorb gather with .youControl()**~~ — FIXED (e938e3e1c). The gather now uses GameObjectFilter.Any.createdBySource() over Player.Each,
matching the oracle ("exile any number of tokens created with this creature" — no control clause, confirmed against Scryfall). Prompt and doc comment updated; ATQ snapshot
golden re-blessed; Tetravus + snapshot + lint nets re-run green.
2. Stale guidance in the Effects.ChooseNumberForSource facade KDoc — Effects.kt:1968-1975 says "Use at entry (wrapped in OnEnterRunEffect) and/or from an upkeep trigger." But
the effect's own KDoc (PlayerEffects.kt:679-685) and the SDK reference now explicitly steer entry choices to EntersWithChoice(ChoiceType.NUMBER) and warn that OnEnterRunEffect
runs after placement (flickers a P/T-defining CDA). The facade comment contradicts the chosen design — update it to point at EntersWithChoice for the entry case.
3. Shapeshifter's "you may choose a number" upkeep is modeled as non-declinable — Shapeshifter.kt:53-57. Not a bug: re-choosing the current number is equivalent to declining
(the CDA reads the last choice either way), and the comment + two tests document this. Flagging only so it's a conscious choice; if a future card's "may choose" has an else
clause, this shortcut won't generalize.

Recommendation

Ship it. Issue 1 is resolved; the remaining items (2–3) are one-line doc/modeling notes, neither blocking. Before this stack reaches main, rebase/merge main into the batch-1
base and re-run the snapshot net, since this review was scoped to the stacked diff.

PR #933, opened 2026-06-23
ATQ: Power Artifact, Reverse Polarity, Golgothian Sylex (+ 3 SDK primitives)

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/932

Implements three Antiquities cards, each adding one reusable engine/SDK primitive.

## Cards
- **Power Artifact** — {U}{U} Aura: enchanted artifact's activated abilities cost {2}
  less, floored at one mana.
- **Reverse Polarity** — {W}{W} Instant: gain twice the artifact-source damage dealt to
  you this turn.
- **Golgothian Sylex** — {4} Artifact: sacrifice each nontoken permanent originally
  printed in ATQ (sparing itself).

## SDK additions
- `ReduceActivatedAbilityCost(filter, amount, manaFloor)` — generic-only activated-ability
  cost reduction over a `GroupFilter`, shared by the enumerator and activation handler so
  displayed and paid costs agree. Backed by `ManaCost.reduceGenericWithManaFloor`.
- `OriginallyPrintedInSet(setCode)` filter — matches a card's canonical set
  (`CardComponent.originalSetCode`); reprints still match their original set.
- `SacrificeAll(filter, excludeTriggering)` — "is sacrificed by its controller"
  (CR 701.21), the sibling of `DestroyAll`.
- `TurnTracker.DAMAGE_RECEIVED_FROM_ARTIFACTS` — per-turn artifact-source damage
  accumulator; source typing reads projected types on the battlefield, base types as LKI.

## Tests
Scenario tests for all three cards + a `ManaCost` floor-math unit test (colored-pip
branches). Card-definition snapshot re-blessed; SDK reference doc updated.


bolav:
Verdict

Behavior is correct and the SDK shape is right. All three new primitives are genuinely novel and parameterized for the next card. Oracle text matches Scryfall verbatim; CR
citations (701.21a, 118.7a) verified accurate. The two issues you asked me to fix are done and committed.

What's good

- ReduceActivatedAbilityCost(GroupFilter, amount, manaFloor) generalizes over a GroupFilter with a parameterized floor — a future "your creatures' abilities cost {1} less" lord
drops in for free. Correctly registered as a non-continuous cost static.
- One reduction function, two callers — enumerator (displayed) and ActivateAbilityHandler (paid) share applyActivatedAbilityCostReduction, so they can't drift.
- OriginallyPrintedInSet wired into all four matcher sites (PredicateEvaluator main + record-fallback, AffectsFilterResolver projection, CostCalculator).
- Set-code self-stamp in AntiquitiesSet so the rules-engine test path gets a reliable signal.
- Docs, snapshot re-bless, backlog count all updated in-change.

Issues

Fixed (committed in cb7c290c5)
1. Untested floor-with-colored-pip math → added ManaCostReduceFloorTest covering {3}{U}→{1}{U}, {2}{U}→{U}, floor-met-by-colored, and the floor=0 ≡ reduceGeneric equivalence.
2. Base-vs-projected fallback in DamageUtils.trackDamageReceivedByPlayer → projected types now authoritative on the battlefield; base typeLine.isArtifact used only as LKI once
the source has left.

Minor (open, author's discretion)
3. Static-description name-lie — Power Artifact's GroupFilter.attachedCreature() (Scope.AttachedTo.description = "enchanted/equipped creature") makes the generated static
description say "creature" for a card that enchants an artifact. Player-facing oracleText is correct; this is internal text. Consider a type-neutral scope description or an
attached() alias.
4. originalSetCode rides on mutable setCode — withPrinting overwrites setCode, so a later-set printing of an ATQ card (limited/booster path only) would report the wrong original
set. Not triggerable by anything in this PR; the set-stamp contract is also enforced in two places now — worth centralizing in CardDiscovery/MtgSet before the next
OriginallyPrintedInSet user (City in a Bottle → ARN).
5. Sacrifice regen-immunity (701.21a) cited but not tested — low value since it's inherent to MoveType.Sacrifice.

Recommendation

Ship-able. The one gap that left real logic unverified (#1) and the projection trap (#2) are fixed and committed. #3–#5 are discretionary.

PR #934, opened 2026-06-23
ATQ batch 4: Artifact Ward, Armageddon Clock, Energy Flux, Transmute Artifact, Urza's Avenger

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/933

Implements five Antiquities cards (58 → 61 implemented).

### Cards
- **Artifact Ward** {W} Aura — can't be blocked by artifact creatures, prevent
  all artifact-source damage, and can't be targeted by abilities from artifact sources.
- **Armageddon Clock** {6} — accrues a doom counter each upkeep, deals that many
  damage to each player in the draw step; {4} (any player, upkeep only) removes one.
- **Energy Flux** {2}{U} — all artifacts gain a pay-{2}-or-sacrifice upkeep trigger.
- **Transmute Artifact** {U}{U} — sacrifice→search→pay-the-difference, composed from
  existing pipeline atoms (no new engine type).
- **Urza's Avenger** {6} 4/4 — {0}: −1/−1 + your choice of banding/flying/first strike/trample.

### Engine / SDK changes
- New `CantBeTargetedBySourceTypeAbilities(sourceType, filter)` static — hexproof keyed
  to the *ability source's card type* (CR 113.7), not controller or color. Enforced via
  the shared `SourceTypeTargeting` helper from both `TargetFinder` and `StackResolver`.
- New passive `DOOM` counter type.
- New `withCardAttachedTo` scenario-test fixture for pre-attached Auras.

### Tests
13 new scenario tests (all five cards), plus re-blessed ATQ snapshot golden. Snapshot
and card-lint nets pass; Oracle text verified against Scryfall for all five.

Docs: `card-sdk-language-reference.md` updated for the new static + doom counter.


bolav:
Verdict

Behavior is right and the SDK shape is right. Five Antiquities cards land with one genuinely new SDK type (CantBeTargetedBySourceTypeAbilities), one new passive counter (DOOM),
and a reusable test-fixture helper — everything else composes existing primitives. All 13 new scenario tests, the snapshot net, and the lint net pass. Both review findings are
now fixed and verified in commit b1f295080.

What's good

- CantBeTargetedBySourceTypeAbilities is the right abstraction, not a one-off. Parameterized by sourceType: CardType + filter, so the next "can't be targeted by abilities from
<type> sources" card is free. It deliberately resists collapsing into GrantProtectionFromCardType (which would also block equip/enchant/block/damage), and the doc comment spells
out why.
- SourceTypeTargeting as a shared object keeps the two enforcement legs (TargetFinder enumeration + StackResolver resolution-time fizzle) from drifting, with the keyword name
and source-type resolution in one place. Mirrors the existing PROTECTION_FROM_CARDTYPE_<TYPE> idiom; uppercase normalization keeps grant and check sides consistent.
- Not-controller-gated check is correctly ordered (before the entityController == controllerId short-circuit), and the test proves P1's own Triskelion can't target P1's warded
creature while P1's Prodigal Sorcerer can.
- Transmute Artifact composed purely from pipeline atoms, with all three mana-value branches plus a youControl()-scoping test. ChooseUpTo(1) correctly models "search … you may
fail to find."
- Energy Flux / Armageddon Clock / Urza's Avenger reuse existing statics cleanly. withCardAttachedTo is a tidy deterministic fixture helper that correctly registers
static/replacement components via withCardOnBattlefield.
- Oracle text verified verbatim against Scryfall for all five — including that Artifact Ward says "abilities" (not "activated abilities"), so blocking triggered abilities too is
correct.

Issues

Blocking — ✅ FIXED

1. Energy Flux was missing its MMQ Printing(...) row. Resolved by adding mtg-sets/.../definitions/mmq/cards/EnergyFluxReprint.kt (collector #78, artist Qiao Dafu, 1999-10-04),
auto-discovered via CardDiscovery.findPrintingsIn. just check-card-printing "Energy Flux" now reports ok: canonical is in the card's earliest real printing and all scaffolded
reprints have rows; :mtg-sets:compileKotlin is green. Canonical remains correctly in ATQ. The other four cards already passed.

Minor — ✅ FIXED

2. CR 109.5 was an off citation for source typing. Replaced with CR 113.7 ("The source of an ability is the object that generated it") across the SDK doc
(ProtectionStaticAbilities.kt), SourceTypeTargeting.kt, and StackResolver.kt. The pre-existing CR 109.5 / 702.16 reference in PlayerComponents.kt was left untouched (not in this
branch's diff, and about damage-source typing).

3. (Optional, not changed) targetingSourceType == ANY applies the source-type check. Same conservative stance as the existing opponent-ability branch; every real ability path
threads ABILITY explicitly, so this is fine as-is.

Recommendation

Merge-ready. Both findings are fixed and verified; SDK shape, enforcement plumbing, projection use, and tests are all in good shape.

PR #935, opened 2026-06-23
Antiquities batch 5: Titania's Song, Tawnos's Coffin, The Rack, Urza's Miter

bolav:
Implements four Antiquities artifacts, each adding a small reusable SDK primitive
rather than card-specific logic.

## Cards
- **Titania's Song** — noncreature artifacts become MV/MV creatures and lose
  abilities; the "continues until end of turn" linger replays as floating effects.
- **Tawnos's Coffin** — state-preserving blink: exile a creature + its Auras, note
  its counters, return tapped with counters restored and Auras re-attached on
  leave/untap.
- **The Rack** — choose-an-opponent-as-it-enters; deals 3 − cards-in-hand to that
  player on their upkeep.
- **Urza's Miter** — "if it wasn't sacrificed" pay-{3}-to-draw on artifact death.

## SDK / engine additions
- `Effects.MassAnimate(filter, power, toughness, …)` — one-shot, locked-set
  (CR 611.2c) group animate with per-entity dynamic P/T; companion to the
  continuous group statics.
- `ContinuousEffect.controllerId` fallback so dynamic-P/T floating effects still
  resolve a controller after their source leaves the battlefield.
- `excludeSacrifice` on `leavesBattlefield(...)`, backed by a central
  `ZoneChangeEvent.wasSacrificed` signal (CR 701.21) stamped by the sacrifice hook.
- `ChosenOpponentUpkeep` step trigger resolving `Player.ChosenOpponent` against the
  source's stored choice.
- `ExileWithAurasNotingCounters` / `ReturnNotedExileTappedWithAuras` effect pair +
  `NotedExileComponent` (preserved across the source's zone change as LKI).

Scenario tests cover both Coffin return paths, the sacrifice vs. non-sacrifice
distinction, the chosen-opponent trigger (incl. negative cases), and the
until-EOT linger. SDK reference docs updated in sync.


bolav:
Verdict

Behavior is correct and the SDK shapes are well-judged. All four cards (Titania's Song, Tawnos's Coffin, The Rack, Urza's Miter) are registered, their scenario tests pass, every cited CR number checks out, and canonical printings sit in the earliest real set. One real
bookkeeping bug (backlog count/checkboxes) and a few coverage gaps; nothing blocking. The branch already contains all of origin/main (0 commits behind), so no merge was needed.

What's good

- wasSacrificed signal is the right shape. Rather than threading a flag through the dozen sacrifice call sites, trackPermanentSacrifice stamps pendingSacrificeIds and moveToZone consumes it once to set ZoneChangeEvent.wasSacrificed, exposed to cards as
leavesBattlefield(..., excludeSacrifice = true). Central hook, event-carried, consumed-and-cleared so a later non-sacrifice move can't be mis-tagged. Reusable for every "if it wasn't sacrificed" card. (ZoneTransitionService.kt:419-457, GameState.kt:154)
- MassAnimateEffect is parameterized for the mechanic, not the card. filter + per-entity DynamicAmount P/T + loseAllAbilities + duration, composing existing AddType/RemoveAllAbilities/SetPowerToughnessDynamic modifications. It's the explicit one-shot/locked-set (CR
611.2c) companion to the continuous group statics, and the per-entity AffectedEntity→ManaValue resolution is implicitly proven by the test (Tinker Engine MV 3 → 3/3, not Song's MV 4).
- Controller-fallback for source-left floating effects is a clean, minimal fix. ContinuousEffect.controllerId captured at creation and used by EffectApplicator only when the source's ControllerComponent is gone — exactly what Titania's Song's until-EOT linger needs,
with no impact on static-ability effects. (EffectApplicator.kt:60-65, ProjectionModels.kt, StateProjector.kt)
- Last-known-info preservation for NotedExileComponent mirrors the existing LinkedExileComponent handling so the LTB return can read it after the Coffin is in the graveyard, and it's stripped on battlefield re-entry per Rule 400.7. Correctly handled in both
ZoneTransitionService and ZoneMovementUtils.
- Tests exercise the rule corners that drove the change: Urza's Miter has both a non-sacrifice death (fires) and a sacrifice (doesn't), plus an end-to-end cast of The Rack through the real EntersWithChoice replacement (not just the injected-component shortcut) and a
negative case for a non-chosen player's upkeep. Both Coffin return paths (leaves + untap) are covered.
- All CR citations verified against the local rules: 701.21 (Sacrifice), 704.5m (unattached-Aura SBA), 611.2c (set locked when continuous effect begins), 604.3 (CDA), 400.7 (new object). Docs in card-sdk-language-reference.md updated in sync.

Issues

Minor

1. Backlog count and checkboxes are out of sync — backlog/sets/antiquities/cards.md:5,105,109. The Rack and Urza's Miter are fully implemented and registered (both appear in the ATQ.json snapshot golden), but their boxes are still - [ ] and Implemented: reads 63 / 85.
This branch adds four cards on top of base's 61, so it should be 65 / 85 with all four checked. The file's own note says "count is authoritative — keep this file's Implemented: line in sync as boxes are checked." Flip both boxes and bump to 65.
2. ReturnNotedExileTappedWithAuras bakes Tawnos-specific semantics into its name/shape. It's a parameterless data object with "Tapped" and "WithAuras" and "owner's control" hardcoded. Genuinely novel state (NotedExileComponent + re-attach) makes the pair defensible,
and decomposing further would just spawn more primitives — but if a second "return exiled with noted counters" card ever appears without the tapped/aura specifics, this won't stretch. Acceptable as-is; worth a comment that it's intentionally Tawnos-shaped.
3. Coverage gaps for documented corners (no test):
  - Titania's Song's locked-set behavior — an artifact that is already a creature should be unaffected, and multiple artifacts with different mana values should each get their own MV. Only a single noncreature artifact is tested.
  - The Coffin's "Auras that can't legally re-attach go to the graveyard via the 704.5m SBA" fallback is documented in the executor and reference doc but not exercised by a test.

These are edge cases; the core behavior is well covered. Add if you want the citations to be non-decorative.

Recommendation

Ship after fixing the backlog bookkeeping (Issue 1 — flip The Rack and Urza's Miter to [x], set count to 65 / 85). Issues 2–3 are author's discretion. The engine work is clean and the new SDK primitives are appropriately generic; no rewrites needed.

PR #938, opened 2026-06-24
Antiquities batch 6: stat counters, end-of-combat triggers, animate addTypes

bolav:
Adds four Antiquities cards and the minimal SDK/engine primitives they need.

## Cards
- **Battering Ram** — begin-combat banding + destroy a blocking Wall at end of combat.
- **Clockwork Avian** — enters with four +1/+0 counters; sheds one at end of combat
  if it attacked or blocked; `{X},{T}` upkeep refill capped at four total.
- **Mishra's Factory** — animates into a 2/2 Assembly-Worker **artifact** creature
  that's still a land; `{T}` pump for Assembly-Workers.
- **Mishra's War Machine** — upkeep "3 damage to you unless you discard; if it
  dealt damage, tap it."

## SDK / engine
- Asymmetric P/T stat counters `+1/+0` `+0/+1` `-1/-0` `-0/-1` (layer 7c, CR 613.4c),
  with filters, dynamic-amount/trigger/replacement wiring, and client display.
- `SuccessCriterion.DamageDealt(recipient)` — gates on damage *actually* dealt
  (prevented/replaced damage doesn't count).
- `AttackedThisCombat` / `BlockedThisCombat` predicates (per-combat, survive removal
  from combat) + `EachEndOfCombat` / `YourEndOfCombat` triggers.
- `BecomeCreature.addTypes` — animate effects can grant extra card types (e.g. artifact)
  while keeping existing types.

## Tests
16 scenario tests (all branches incl. prevented-damage and block-only shed) pass;
corpus snapshot + lint nets green. All cited CR numbers verified.


bolav:
Verdict: Behavior is correct and the SDK shape is right. This is a clean, well-composed batch — four Antiquities cards plus minimal, genuinely-novel primitives. All 16 new
scenario tests pass, the corpus snapshot + lint nets stay green, oracle text matches Scryfall exactly for all four cards, and every cited CR number checks out. The two minor
nits from the first pass are now fixed in ef47c9a71. No outstanding blocking or important issues. (Branch already has origin/main merged — fast-forward, no merge commit.)

What's good (keep all of it)

- The new SDK types are all justified, not over-specialized.
  - SuccessCriterion.DamageDealt(recipient) reads DamageDealtEvent history — something no existing criterion (Always/Auto/CollectionNonEmpty) can express — and correctly
distinguishes dealt from prevented/replaced. Parameterizing DamageRecipient.{Any,Controller} instead of hardcoding "you" sets up the next card cleanly. The event-flow plumbing
(evaluationEvents defaulting to priorEvents, sync path passing result.events, paused path routing accumulated events without double-prepending) is genuinely careful — and the
test proves all three branches (no-discard→tap, discard→no-tap, prevented→no-tap).
  - AttackedThisCombat/BlockedThisCombat components+predicates are novel: they survive removal-from-combat (unlike live AttackingComponent/BlockingComponent) and reset
per-combat (unlike AttackedThisTurn). Wired through every predicate site (PredicateEvaluator, AffectsFilterResolver, TriggerMatcher, BeginningPhaseManager, Serialization) and
cleared in CombatManager.endCombat.
  - The four asymmetric counters compose into the existing layer system; EffectApplicator splits the power/toughness math correctly and the if (powerMod != 0) / if (toughnessMod
!= 0) guards preserve the "don't materialize a null base into 0" semantics.
- BecomeCreature.addTypes is the right minimal fix for "becomes an artifact creature, still a land" — additive Layer-4 type grant, not a new effect type. Now exposed on both
overloads (the DynamicAmount parity fix).
- Tests hit the real corners: power-only modification (Clockwork Avian → 3/4 not 3/3), the "or blocked" half of the shed via a creature that only blocked, prevented-damage
gating, the cap formula min(X, 4-current), and the upkeep timing restriction.
- CR citations verified against local rules: 122.1a (+X/+Y counter math), 122.3 (only ±1/±1 annihilate), 613.4c (layer 7c). All accurate.

Issues — Resolved

1. ~~Wrong layer label in a test comment. ClockworkAvianScenarioTest.kt:19 said "layer 7d"; counters are 7c (7d is P/T switching).~~ ✅ Fixed in ef47c9a71.
2. ~~BecomeCreature addTypes only on the Int overload. The DynamicAmount overload silently couldn't grant extra types.~~ ✅ Fixed in ef47c9a71 — addTypes added and forwarded,
purely additive default, no callers break; snapshot net still green.

Issues — Minor (author's discretion, not addressed)

3. Client folds the asymmetric counters into passive badges, not the stat-modifier path. web-client/src/components/game/board/shared.ts getCounterStatModifier/hasStatCounters
(lines 423–436) still only sum PLUS_ONE_PLUS_ONE/MINUS_ONE_MINUS_ONE, and the new counters were added to PASSIVE_COUNTER_TYPES. Displayed P/T stays correct (server projects it;
client is a dumb terminal), so this is cosmetic — but verify the badge treatment is what you want for a +1/+0 (renders as a generic green badge rather than participating in any
+N stat overlay the way +1/+1 does).

Recommendation

Mergeable. The two actionable nits are committed; only the cosmetic client item (3) remains, and it's genuinely optional given the server is authoritative on P/T. Nothing else
to change — the primitives are the right shape and the test coverage is exactly where it should be.

PR #939, opened 2026-06-24
Antiquities: Circle of Protection: Artifacts, Martyrs of Korlis, Rocket Launcher

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/938

Implements three Antiquities cards and the SDK primitives they need.

### Cards
- **Circle of Protection: Artifacts** — `{2}`: prevent the next instance of damage
  from a chosen artifact source.
- **Martyrs of Korlis** — while untapped, redirect all artifact damage dealt to you
  onto itself.
- **Rocket Launcher** — `{2}`: 1 damage to any target, self-destructs next end step;
  activate only if controlled since your most recent turn began.

### SDK / engine additions
- `PreventionSourceFilter.ChosenSourceMatching(filter)` — parameterized "a [quality]
  source of your choice", plus an explicit `nextInstanceOnly` flag on
  `PreventDamageEffect` (single-instance Circle-of-Protection shield vs all-damage).
- `RedirectDamage.condition` — optional gate on a static redirect (Martyrs while untapped).
- `ActivationRestriction.ControlledSinceYourMostRecentTurn` — artifact "summoning
  sickness", reusing the engine's existing summoning-sickness tracking across all permanents.
- `SerializableModification.PreventNextDamageInstanceFromSource` for the consumed shield.

### Tests
Scenario coverage for each card, including the rule corners: untapped/tapped/non-artifact
redirect, summoning-sick vs controlled-since-turn activation, and single-instance vs
all-damage prevention. `:mtg-sets:test` (snapshot + lint) and `:rules-engine:test` green.

Branches off `atq-batch-6-addtypes-stat-counters`.

PR #940, opened 2026-06-24
ATQ batch 8: dynamic-P/T animate, until-your-next-upkeep, target-player graveyard

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/939

Implements 3 Antiquities cards, taking ATQ to 84/85 (only Bronze Tablet — ante — postponed).

### Cards
- **Xenic Poltergeist** — `{T}`: target noncreature artifact becomes an artifact creature
  with P/T each equal to its mana value, until your next upkeep.
- **Drafna's Restoration** — put any number of target artifact cards from target player's
  graveyard on top of their library in any order.
- **Goblin Artisans** — `{T}`: flip a coin; win → draw, lose → counter target artifact spell
  you control not already targeted by another Goblin Artisans.

### Engine / SDK additions
- **Dynamic-P/T animate** — `BecomeCreatureEffect` gains optional `dynamicPower`/`dynamicToughness`;
  facade `Effects.BecomeCreatureWithManaValueStats` wires P/T = the permanent's own mana value via
  the existing `SetPowerToughnessDynamic` (no new effect type).
- **`Duration.UntilYourNextUpkeep` expiry** — mirrors the until-your-next-turn path; expires at the
  start of the controller's upkeep, before re-granting upkeep triggers (Erhnam Djinn).
- **`StatePredicate.NotTargetedByAbilityFromSameNamedSource`** — generic self-referential targeting
  restriction keyed off the source's own name.
- **Target-player graveyard** — `TargetValidator` threads the chosen player target into graveyard-card
  validation so `OwnedByTargetPlayer` resolves; also fixes an `Int.MAX_VALUE` overflow in the
  per-cast target-count cap for `unlimited` requirements.

### Tests / docs
- Scenario tests for all three cards (both coin-flip branches, self-ref lock, animate→revert-at-upkeep,
  target-player ordering + negative-ownership rejection).
- ATQ snapshot re-blessed; PZ2 reprint row for Drafna's Restoration; SDK reference updated.


bolav:
Verdict. Ready to merge. Behavior is correct, the SDK additions are well-shaped (generically named, parameterized for the next card, composing existing machinery rather than new
monoliths), and all tests pass. The two findings from the prior round — the missing PZ2 printing row and the missing negative-ownership test — are now fixed and verified in
commit 3654fe4c4.

(Reviewed in place on a clean tree, diffing against the atq-batch-7… base as requested. origin/main was not merged, since the base is a sibling feature branch in a stack rather
than main — worth confirming the stack rebases cleanly onto main before the eventual merge.)

What's good

- Dynamic P/T as an extension, not a new effect. BecomeCreatureEffect gains optional dynamicPower/dynamicToughness, and BecomeCreatureWithManaValueStats is a thin facade wiring
DynamicAmount.EntityProperty(AffectedEntity, ManaValue) into the existing SetPowerToughnessDynamic path — the single-target companion to MassAnimate, reusable by any future
"becomes a creature with P/T equal to <property>" card.
- NotTargetedByAbilityFromSameNamedSource is correctly generic — it reads the source's own name from PredicateContext.sourceId rather than hardcoding "Goblin Artisans", and is
properly inert in all three non-target contexts (AffectsFilterResolver, TriggerMatcher, BeginningPhaseManager), matching the CreatedBySource precedent. The doc comment honestly
flags the copied-name edge.
- UntilYourNextUpkeep expiry mirrors expireUntilYourNextTurnEffects and is sequenced correctly: it runs at the start of Step.UPKEEP before priority, so a re-granting upkeep
trigger (Erhnam Djinn) re-applies afterward with a fresh timestamp.
- The TargetValidator overflow fix is a real catch — summing Int.MAX_VALUE for an unlimited requirement would wrap negative; the Long coercion + "skip the cap when any
requirement is unlimited" both handle it.
- CR numbers corrected and verified — 613.4c → 613.4b (layer 7b set P/T) + 608.2h (info determined once), both confirmed against MagicCompRules_20260417.txt.
- Tests exercise the rule corners — both coin-flip branches, the self-referential lock between two Goblin Artisans, animate→persist-through-opponent-turn→revert-at-your-upkeep,
player-chosen ordering, and (new) rejection of a card in a non-targeted player's graveyard.

Issues

Resolved this round
1. ~~Drafna's Restoration missing pz2 reprint row~~ — fixed: DrafnasRestorationReprint.kt added (auto-discovered via CardDiscovery.findPrintingsIn); check-card-printing
"Drafna's Restoration" now reports ok, canonical in ATQ.
2. ~~Drafna's test didn't isolate the negative ownership axis~~ — fixed: new case targets P2 as the player but an artifact in P1's own graveyard, asserting the cast is rejected.
Passes.

Minor (open, author's discretion)

3. chosenPlayerTarget takes the first player target only. TargetValidator.kt:197 (allTargets.firstNotNullOfOrNull { (it as? ChosenTarget.Player)?.playerId }) is correct for
these single-player-target relational spells but silently assumes one. A one-line comment noting the assumption (or keying off the requirement index) would save a future
multi-player-target card from a confusing bug.

Verification run

- :rules-engine:test — all five new scenario tests pass (3× Goblin Artisans, Xenic Poltergeist, 2× Drafna's incl. new negative case).
- :mtg-sets:test — CardDefinitionSnapshotTest + CardLintTest green; no re-bless needed (Printing rows aren't part of the card snapshot).
- check-card-printing — ok for all three new cards.
- Oracle text verified verbatim against Scryfall for all three.

Recommendation

Merge as-is. Optionally address the one-line chosenPlayerTarget comment (#3) if convenient. Before the stack lands on main, confirm atq-batch-7… → main rebases cleanly since
this review didn't merge main.

PR #1014, opened 2026-06-27
Engine bug fixes from TMT self-play: cleanup damage, trigger slots, cast-from-library, enters-with-counters

bolav:
Five tightly-scoped engine fixes surfaced during TMT self-play.

- **Cleanup damage removal.** Extract the CR 514.2 turn-based actions (remove marked
  damage; end per-turn combat/miracle markers) into
  `CleanupPhaseManager.applyCleanupTurnBasedActions` and run it on both cleanup paths.
  The CR 514.1 hand-size discard early-returns *before* 514.2, so its continuation now
  finishes the step — fixes marked damage (notably deathtouch suppressed by an expiring
  until-end-of-turn indestructible) surviving into the next turn.
- **Trigger target-slot alignment.** Multi-"up to N" triggers: drop declined slots and
  narrow each kept slot's requirement to the chosen count via new `TargetRequirement.withCount`,
  keeping the flat target↔requirement index walks aligned. Fixes Don & Leo-style fizzles
  where a partially-filled slot absorbed a later slot's target.
- **Cast-from-library tracking.** Add `CastFromLibraryComponent` (full parity with the
  hand/graveyard markers) so `WasCastFromZone(LIBRARY)` works for permanents — e.g. Mikey
  & Don's "+1/+1 if cast from the top of your library."
- **Non-self enters-with-counters condition** now evaluates against the *entering*
  creature, mirroring the dynamic-counters path (Leonardo, Sewer Samurai).

New tests cover each fix; `docs/card-sdk-language-reference.md` updated.
Cited CR numbers (514.1/514.2/608.2b/702.94) verified. `:rules-engine` compiles green.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -4170,9 +4170,13 @@ answer it and would silently return `false`.
 - `WasCastFromHand` — cast specifically from hand.
 - `WasCastFromZone(zone)` — cast from a specific zone. For resolving spells it reads the spell's
   cast-origin; for a permanent already on the battlefield it falls back to the cast-origin marker
-  stamped as it entered (`HAND` → `CastFromHandComponent`, `GRAVEYARD` → `CastFromGraveyardComponent`),
-  so it can gate an entering permanent's own replacement effect (Hundred-Battle Veteran: "enters with
-  a finality counter if cast from your graveyard").
+  stamped as it entered (`HAND` → `CastFromHandComponent`, `GRAVEYARD` → `CastFromGraveyardComponent`,
+  `LIBRARY` → `CastFromLibraryComponent`), so it can gate an entering permanent's own replacement
+  effect (Hundred-Battle Veteran: "enters with a finality counter if cast from your graveyard") or a
+  *non-self* `EntersWithCounters` (`selfOnly = false`), whose condition is evaluated against the
+  entering creature — Leonardo, Sewer Samurai ("creatures you cast from your graveyard enter with a
+  finality counter") and Mikey & Don ("creatures you cast from the top of your library enter with an
+  extra +1/+1 counter", `WasCastFromZone(Zone.LIBRARY)`).
 - `WasKicked` — cast with kicker / multikicker / offspring (i.e. an `OptionalAdditionalCost` with `branchesEffect = true` whose extra cost was paid). FlashKicker payments are intentionally invisible to this condition.
 - `SneakCostWasPaid` — the source was cast for its `Sneak` cost (CR 702.190 — mana + returning an unblocked attacker). Reads the durable `ChoiceSlot.SNEAK` flag on a resolved permanent, falling back to the resolution context for a non-permanent spell's own effect. Backs riders like Leonardo, Leader in Blue and The Last Ronin's Technique.
 - `NoManaSpentToCast` — "it wasn't cast or no mana was spent to cast it": the standard free-cast
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/targets/TargetRequirement.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/targets/TargetRequirement.kt
@@ -452,6 +452,35 @@ data class TargetOther(
 }
 
 /**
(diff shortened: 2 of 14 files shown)


bolav:
Review: engine-bugfixes-from-tmt

Verdict

Five tightly-scoped, correct bug fixes surfaced during TMT self-play. Behavior is right, the SDK additions are minimal and reusable (no card-specific types), and each fix is paired with a test that exercises the actual rules corner. Mergeable. The branch is a clean
descendant of main (main == origin/main == merge-base — no merge commit needed).

What's good

- TargetRequirement.withCount() is the right SDK shape. It's a small generic primitive over the sealed hierarchy — not a card-specific patch. The when(this) is exhaustive across all 11 subtypes with no else, so a future TargetRequirement forces a compile error rather
than silent misalignment. minCount is clamped only for the two types (AnyTarget, TargetObject) that store it independently; the rest derive minCount from count via the interface default, so copy(count=…) carries it correctly. The unit test pins exactly this (clamp +
effectiveMinCount staying 0 for optional, no-op identity, TargetOther recursion).
- The slot-alignment fix is the correct mechanism. Both index walks — EffectContext.buildNamedTargets (EffectContext.kt:386) and StackResolver.getRequirementForTargetIndex (StackResolver.kt:2649) — advance by req.count. Dropping declined slots and narrowing kept slots
keeps the flat targets list parallel to alignedRequirements, which is what putTriggeredAbility stores together in TargetsComponent for resolution-time re-validation. The "documents the bug" test (un-narrowed wide slot swallows the next target) is a nice touch.
- CastFromLibraryComponent has full parity. It appears in exactly the same 5 sites as CastFromHandComponent (serializer registration, ConditionEvaluator, StackResolver stamping, ZoneMovementUtils strip-on-leave, component def). It reuses the existing
WasCastFromZone(LIBRARY) condition instead of inventing a new condition type. The stamp sits in the permanent-resolution path beside its siblings.
- The cleanup extraction is clean and not double-applied. performCleanupStep early-returns inside if (cardsToDiscard > 0) before the 514.2 block, so applyCleanupTurnBasedActions runs exactly once on either path. The scenario test reproduces the real failure
(deathtouch damage suppressed by an expiring until-end-of-turn indestructible, creature dying on the next turn's SBA check with no destroy event), plus a simpler ordinary-damage regression.
- EntersWithCounters fix correctly mirrors the dynamic path. evaluateWasCastFromZone reads context.sourceId, and the fix sets sourceId = enteringEntityId (keeping controllerId as the source's, so "you control" still resolves correctly) — identical to the adjacent
EntersWithDynamicCounters branch.

All cited CR numbers verify against the local rules: 514.1 (hand-size discard), 514.2 (remove damage + end until-end-of-turn), 608.2b (all-targets-illegal fizzle), 702.94 (Miracle).

Issues

Minor

- ~~EffectAndTriggerContinuationResumer.kt:84-91 — inline FQNs vs. import.~~ Fixed in this branch: ChosenTarget and TargetRequirement are now imported and the two mutableListOf declarations use the short names. (Pre-existing inline ChosenTarget FQNs in the
entityIdToChosenTarget helper at ~201–214 were left as-is — outside this branch's scope.)
- EffectAndTriggerContinuationResumer.kt:91 — defensive getOrNull can desync the two lists. If a non-empty slot's index had no matching requirement, the target is added to selectedTargets but nothing to alignedRequirements, re-introducing the misalignment this fix
removes. In practice slot keys always map 1:1 to requirements so this can't fire — but a non-null assertion / error would fail loud rather than silent. Not blocking.
- Comment density. The new comments are long, but every one explains a non-obvious why (rules interaction, index-walk consequence) rather than restating code — consistent with house style. Flagging only so it's a conscious call.

Recommendation

Ship it. The one actionable minor nit (inline FQNs) is now addressed; a compile check of :rules-engine is running in the background to confirm the import change is clean. No behavioral changes requested.

Tests: TargetRequirementWithCountTest confirmed green (4/4); the combined Gradle run exited 0, so TriggerTargetSlotAlignmentTest and CleanupHandSizeDiscardDamageRemovalTest ran green too (their result XMLs were wiped post-run by a sibling container — the documented
shared-/workspace hazard, not a failure).

PR #1046, opened 2026-06-27
TMT merge — PR 4/5: Mana abilities & ETB triggers

bolav:
Part 4 of a 5-PR stack landing the Teenage Mutant Ninja Turtles (TMT) set.
**Base this PR on `tmt-3-...`; PR 5 bases on this branch.** Merge the stack in order.

### Engine / SDK changes in this PR
- Add `ManaRestriction.AbilityActivationOnly` — "spend only to activate an ability"
  (any ability of any source); composes under `AnyOf` for "cast X or activate an ability".

### Cards added (10)
- Purple Dragon Punks · Leonardo, Sewer Samurai · Krang & Shredder
- Groundchuck & Dirtbag · Leatherhead, Swamp Stalker · Donatello, Mutant Mechanic
- Lita, Little Orphan Amphibian · Mondo Gecko · Raphael, Ninja Destroyer
- Tokka & Rahzar, Terrible Twos · Turtles in Time

### Notes
- Every card composes existing primitives; the only new SDK type is the mana restriction
  above (wired in `ManaPool`, documented in the SDK reference).
- Scenario tests cover Groundchuck (off-stack mana), Krang Disappear (casts one chosen
  card), Leonardo (finality on graveyard-cast), Don & Leo (optional-target no-fizzle),
  and the any-ability mana restriction. `TMT.json` snapshot re-blessed; lint green.
- Follow-up: behavioral scenario tests still owed for Tokka & Rahzar, Donatello,
  Mondo Gecko, and Turtles in Time (snapshot-only at present).

### DIFF
--- a/backlog/sets/teenage-mutant-ninja-turtles/TODO.md
+++ b/backlog/sets/teenage-mutant-ninja-turtles/TODO.md
@@ -29,8 +29,14 @@ commits all carry `flavorText` in metadata.
 >   reflexive "when you do" (HEXPROOF counter + entersWith already exist).
 > - **Purple Dragon Punks** — *any*-ability mana ("artifact spell or activate an ability";
 >   `CardTypeSpellsOrAbilitiesOnly` ties abilities to the card type, too narrow).
-> - **The Last Ronin** — a Saga chapter that installs a *this-turn* triggered ability
->   ("whenever a creature attacks alone this turn …"; the attacks-alone trigger exists).
+> - **The Last Ronin** — chapters I (`DestroyAll`) and II (`mill` + `ReflexiveTrigger` return)
+>   compose cleanly; chapter III installs a turn-scoped delayed `attacks(filter=youControl,
+>   requires=Alone, binding=ANY)` trigger via `CreateDelayedTriggerEffect` (Rediscover the Way's
+>   tested chapter-III idiom + Beastmaster Ascension's filtered-attack spec). Built it, but a
+>   combat scenario test left the lone attacker with 0 counters — the delayed attacks-alone
+>   trigger didn't fire/apply. Reverted pending a correct chapter-III combat verification (likely
+>   a test-flow or binding nuance, not a missing capability — attack delayed triggers are
+>   supported per TriggerDetector.detectEventBasedDelayedTriggers).
 > - **Tokka & Rahzar** — "mana spent to cast it was less than its mana value" trigger condition.
 > - **Lita, Little Orphan Amphibian** — modal where each mode is once-per-turn ("hasn't been chosen this turn").
 > - **Mondo Gecko** — become-chosen-color + "draw a card for each color among permanents you control" dynamic.
--- a/backlog/sets/teenage-mutant-ninja-turtles/TODO.md
+++ b/backlog/sets/teenage-mutant-ninja-turtles/TODO.md
@@ -29,14 +29,20 @@ commits all carry `flavorText` in metadata.
 >   reflexive "when you do" (HEXPROOF counter + entersWith already exist).
 > - **Purple Dragon Punks** — *any*-ability mana ("artifact spell or activate an ability";
 >   `CardTypeSpellsOrAbilitiesOnly` ties abilities to the card type, too narrow).
-> - **The Last Ronin** — chapters I (`DestroyAll`) and II (`mill` + `ReflexiveTrigger` return)
->   compose cleanly; chapter III installs a turn-scoped delayed `attacks(filter=youControl,
->   requires=Alone, binding=ANY)` trigger via `CreateDelayedTriggerEffect` (Rediscover the Way's
->   tested chapter-III idiom + Beastmaster Ascension's filtered-attack spec). Built it, but a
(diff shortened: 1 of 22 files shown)


bolav:
Review: tmt-4-mana-and-etb (TMT mana & ETB cards)

Workspace: Reviewed in place on the checked-out branch. origin/main (be6fbe83) is already an ancestor of HEAD, so no merge was needed — the diff reflects post-merge reality
as-is.

Verdict

Behavior is right and the SDK shape is excellent. This branch adds ~12 TMT cards plus exactly one new SDK primitive (ManaRestriction.AbilityActivationOnly); every other card
composes primitives that already exist in main. All tests I ran are green (6 rules-engine scenario/unit tests, plus the corpus snapshot + lint nets). The one real gap is test
coverage: 7 cards shipped with only a snapshot entry and no behavioral scenario test.

What's good (keep this)

- ManaRestriction.AbilityActivationOnly is a model SDK addition. It's genuinely novel — "any ability of any source" can't be expressed by CardTypeSpellsOrAbilitiesOnly (which
ties abilities to a card type) — it's a parameterless atom, and Purple Dragon Punks reaches "artifact spell or any ability" by composing it under the existing AnyOf rather than
inventing a bespoke type. Handler wiring (ManaPool.kt:83), @Serializable, and the doc entry (card-sdk-language-reference.md) all land in the same change. This is exactly the
"small generic primitive + composition" the review optimizes for.
- No over-specialized card types. Leatherhead reuses ReflexiveTriggerEffect (Slumbering Walker shape) + Dawning Purist's opponentControls() filter; Donatello uses
ConditionalEffect/BecomeCreature + MoveAllLastKnownCounters; Turtles in Time uses ForEachPlayer + gather/move primitives; Mondo Gecko uses ChooseColorThen driving two
chosen-color readers. Each is a recipe over existing atoms, not a new executor.
- CR citations verify. 605.1b (triggered mana ability resolves off-stack — Groundchuck) and 608.2b (target-legality at resolution — Don & Leo fizzle guard) both match the local
Comp Rules text exactly.
- The three scenario tests that exist are sharp. Krang's "casts only the chosen card, leaves the rest exiled" regression guard, Don & Leo's "decline slot-0, keep slot-1, don't
fizzle" guard, and Groundchuck's off-stack-immediacy + opponent-control negative case all exercise the actual rule corner that drove the change.
- Projection is respected — colorsAmongPermanents() defaults to Player.You and reads projected colors; no base-state battlefield reads introduced.

Issues

Important

- 7 cards have no behavioral test — only a snapshot entry. Mondo Gecko, Tokka & Rahzar, Raphael, Turtles in Time, Donatello, Lita, and Leatherhead were each added with just a
TMT.json snapshot row (confirmed: no scenario test in rules-engine or game-server). The snapshot net pins compiled structure, not behavior — per the project's own add-card rule,
the scenario test is the real gate. The compositions are standard, but several of these have non-trivial wiring that nothing currently verifies:
  - Tokka & Rahzar (TokkaAndRahzarTerribleTwos.kt:42) — the ManaSpent < ManaValue comparison is a subtle condition (fires only under cost reduction / alternative costs). Worth a
test that casts a reduced-cost spell and asserts 3 damage, plus a full-cost spell that does not trigger.
  - Donatello (DonatelloMutantMechanic.kt) — conditional BecomeCreature on a non-creature artifact, plus the LTB "move last-known counters" trigger. Two distinct behaviors, zero
coverage.
  - Mondo Gecko (MondoGecko.kt:33) — one color choice feeding both ChangeColorToChosen and GrantHexproofFromChosenColor; the dual-read is the interesting axis.
  - Turtles in Time (TurtlesInTime.kt) — per-player optional shuffle-then-draw inside ForEachPlayer ("each player who does draws seven"); easy to get the per-player "may" wrong.

Recommend adding scenario tests for at least these four before merge; Raphael/Lita/Leatherhead are lower-risk but should follow. Note Leatherhead's P/T needed a fix in the final
commit (5/4) — a sign it wasn't behaviorally exercised.

Minor

- Purple Dragon Punks is covered only at the restriction layer. ManaSpendRestrictionAnyAbilityTest tests the exact AnyOf(...) shape against ManaPool.canPay directly, which is
good coverage of the logic — but the card's end-to-end path (tap → produce restricted {R} → actually spend it on an artifact spell vs. an ability) is untested. Optional given
the unit test, but a one-shot scenario would close it.
- Commit messages slightly oversell. 8ad21734 ("Fix Groundchuck to add mana off-stack via AdditionalManaOnSourceTap") and b16335e6 ("Cast Krang's Disappear during resolution")
describe switching the card to the right existing primitive, not an engine fix — the engine code is unchanged. Accurate enough, just don't let a future reader hunt for an engine
diff that isn't there.

Recommendation

Mergeable in substance — the SDK addition is clean, the cards compose correctly, and everything that's tested is green. Before merge, add behavioral scenario tests for the four
higher-risk untested cards (Tokka & Rahzar, Donatello, Mondo Gecko, Turtles in Time); backfill Raphael/Lita/Leatherhead/Purple Dragon Punks as a fast follow. No code changes to
the cards or the SDK are required.

PR #1047, opened 2026-06-27
Fix double-fire of graveyard-active self-dies triggers

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/1046

A dies trigger declared with triggerZone = GRAVEYARD (activeZone == GRAVEYARD,
e.g. Paramecia Coloniex, so its effect can reference Self after death) was
detected twice for a single battlefield->graveyard zone change: once by the
graveyard-resident trigger scan (the dead card already sits in the graveyard
and its trigger pattern matches its own death event) and once by the dedicated
detectDeathTriggers pass.

Guard the graveyard-resident scan so it skips a card's own battlefield-exit
event; that case is owned by detectDeathTriggers / detectLeavesBattlefieldTriggers,
which fire regardless of activeZone. Other graveyard-resident reactions (another
creature dying/entering) still fall through to the matcher.

Add regression tests covering a triggerZone = GRAVEYARD dies trigger firing
exactly once on its own death.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/backlog/sets/teenage-mutant-ninja-turtles/TODO.md
+++ b/backlog/sets/teenage-mutant-ninja-turtles/TODO.md
@@ -29,8 +29,14 @@ commits all carry `flavorText` in metadata.
 >   reflexive "when you do" (HEXPROOF counter + entersWith already exist).
 > - **Purple Dragon Punks** — *any*-ability mana ("artifact spell or activate an ability";
 >   `CardTypeSpellsOrAbilitiesOnly` ties abilities to the card type, too narrow).
-> - **The Last Ronin** — a Saga chapter that installs a *this-turn* triggered ability
->   ("whenever a creature attacks alone this turn …"; the attacks-alone trigger exists).
+> - **The Last Ronin** — chapters I (`DestroyAll`) and II (`mill` + `ReflexiveTrigger` return)
+>   compose cleanly; chapter III installs a turn-scoped delayed `attacks(filter=youControl,
+>   requires=Alone, binding=ANY)` trigger via `CreateDelayedTriggerEffect` (Rediscover the Way's
+>   tested chapter-III idiom + Beastmaster Ascension's filtered-attack spec). Built it, but a
+>   combat scenario test left the lone attacker with 0 counters — the delayed attacks-alone
+>   trigger didn't fire/apply. Reverted pending a correct chapter-III combat verification (likely
+>   a test-flow or binding nuance, not a missing capability — attack delayed triggers are
+>   supported per TriggerDetector.detectEventBasedDelayedTriggers).
 > - **Tokka & Rahzar** — "mana spent to cast it was less than its mana value" trigger condition.
 > - **Lita, Little Orphan Amphibian** — modal where each mode is once-per-turn ("hasn't been chosen this turn").
 > - **Mondo Gecko** — become-chosen-color + "draw a card for each color among permanents you control" dynamic.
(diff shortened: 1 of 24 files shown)


bolav:
Verdict

The fix is correct, minimal, and well-targeted. A graveyard-active dies trigger (triggerZone = GRAVEYARD, the Paramecia Coloniex shape) was being detected twice — once by the
graveyard-resident scan (because the dead card now sits in the graveyard and its Dies pattern matches its own death event) and once by the dedicated detectDeathTriggers. The
guard skips the card's own battlefield-exit in the graveyard scan, leaving the dedicated death/leaves detectors as the single owner. No SDK surface changes, so no
card-sdk-language-reference.md update is needed — appropriate for a pure engine-internal fix.

What's good

- Right layer for the fix. The double-fire is a detection-path overlap, and the guard lives exactly where the overlap occurs (the graveyard-resident scan in
detectTriggersForEvent), rather than papering over it downstream in trigger processing or dedup.
- The premise is sound and I verified it. detectDeathTriggers (DeathAndLeaveTriggerDetector.kt:71) and detectLeavesBattlefieldTriggers (:458) both resolve the dying entity's
abilities by cardDefinitionId regardless of activeZone, so a graveyard-active SELF/ANY death-or-leaves trigger is still re-added there. The guard removes the duplicate, not the
trigger — and the companion test asserts exactly that.
- Guard scoping is precise. It keys on event.entityId == entityId (the card's own exit) and fromZone == BATTLEFIELD, so other graveyard-resident reactions — another creature
dying/entering, FuriousForebear's OTHER-bound leaves trigger watching teammates — still fall through. I confirmed FuriousForebear (tdm) is unaffected: its OTHER binding never
matched its own death anyway, and for other deaths the entityId differs.
- Comments earn their place. Both the engine comment and the test comments explain the why (the two competing detection paths) rather than restating the code — consistent with
the project's comment policy.

Issues

Minor

1. Latent gap in the guard's invariant — worth a one-line comment, no card hits it today. The guard assumes "the card's own battlefield-exit is fully owned by the death/leaves
detectors." That holds only for trigger patterns with from == BATTLEFIELD: isDeathTrigger requires from == BATTLEFIELD && to == GRAVEYARD (TriggerMatcher.kt:1014) and
isLeavesBattlefieldTrigger requires from == BATTLEFIELD (:1023). A hypothetical graveyard-active trigger with pattern from = null, to = GRAVEYARD ("whenever [this] is put into a
graveyard from anywhere") would match the card's own death in the graveyard scan but is caught by neither detector — so the guard would drop it to zero fires. I checked every
triggerZone = Zone.GRAVEYARD card in the corpus (Paramecia, RootingKavu, FuriousForebear, BladewingsThrall, FearOfInfinity, PyreZombie, etc.); none use that shape, so there's no
current regression. Suggest noting the assumption in the guard comment ("relies on every self battlefield-exit trigger being a from = BATTLEFIELD death/leaves pattern") so a
future from = null graveyard-active card doesn't silently lose its self-trigger.
2. The two tests are near-duplicates. TriggerDetectorBatchTriggerTest.kt:624 and :646 build identical state/events and both assert own.shouldHaveSize(1); the second adds nothing
the first doesn't already cover (the first already proves "exactly one," which is simultaneously "not zero" and "not two"). Harmless, but the "companion" test is redundant —
consider folding the binding/controller assertions into one test, or dropping the second.

Adjacent (pre-existing, not introduced here — out of scope)

For an ANY/OTHER-bound graveyard-active dies trigger on a creature that itself dies in a multi-creature batch, the graveyard scan and detectSimultaneousDeathTriggers
(TriggerDetector.kt:253) can both add the same "B sees A die" trigger. This predates the fix (the new guard only touches the SELF self-death case) and no current card has that
shape, so it's not blocking — flagging only so it's on record if such a card appears.

Recommendation

Ship it. Optionally add the one-line invariant note (Minor 1) and trim the duplicate test (Minor 2) — both author's discretion. I'll post the test result
(TriggerDetectorBatchTriggerTest) as soon as the background build finishes.

PR #1048, opened 2026-06-27
TMT: 7 legendaries + sneak-from-graveyard, next-spell affinity, attack-your-opponent trigger

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/1047

Adds 7 Teenage Mutant Ninja Turtles cards (188/190) and the SDK primitives they need.

### Cards
- **Don & Raph, Hard Science** — attack grants your next noncreature spell affinity for artifacts
- **Mikey & Don, Party Planners** — cast Mutant/Ninja/Turtle from top of library, +1/+1 if creature
- **Ninja Teen** (Class) — L3 grants graveyard creatures sneak {3}{B}
- **Party Dude** (Class) — L3 pumps an attacker when an opponent is attacked
- **Rat King, Verminister** — reanimate target creature + all same-named cards, tapped
- **The Cloning of Shredder** (Saga) — gather→copy from linked exile
- **The Last Ronin** (Saga) — III: pump a lone attacker each combat this turn

### New SDK primitives
- `GrantNextSpellAffinity(spellFilter, forType)` — one-shot rider (mirrors `MakeNextSpellUncounterable`); cost calculator reduces the next matching spell by the caster's count
of `forType` at cast time
- `GraveyardCreaturesHaveSneak(cost)` — granted sneak alt-cost cast from graveyard; additive graveyard loop in `SneakCastEnumerator`, leaves the printed-Sneak hand path
untouched
- `ReturnSameNamedFromGraveyard(target)` — return target + all same-named cards tapped
- `CreaturesAttackYourOpponent` trigger — opponent-side counterpart of `CreaturesAttackYou`
- Per-attacker fan-out for filter-scoped delayed attack triggers (binds `TriggeringEntity`)

Riders are turn-scoped (cleared by `TurnManager`); grant lookups read projected controller (CR 613). Scenario tests cover each card incl. negative cases (no-grant, not-alone,
cross-turn leak); SDK reference + snapshot golden updated.


bolav:
Verdict

Behavior is correct and the SDK shapes are well-chosen — new types reuse existing patterns (the one-shot rider, the attack-matcher family) rather than growing bespoke machinery.
No blocking issues. A few minor coverage/robustness notes.

What's good (keep it)

- GrantNextSpellAffinity is the right shape: it reuses the MakeNextSpellUncounterable one-shot-rider pattern — PendingNextSpellAffinity state + cost-calculator hook +
consume-on-cast in CastSpellHandler — and is parameterized (spellFilter, forType) so the next "next spell has affinity for X" card composes for free. The reduction is read
dynamically at cast time (countPermanentsOfType), which is correct affinity semantics.
- CreaturesAttackYourOpponent is a faithful mirror of the existing CreaturesAttackYou, wired through every site (EventPattern, TriggerIndex, TriggerMatcher, Triggers facade) and
using the same base AttackingComponent.defenderId read as its sibling. The player-only-not-planeswalker semantics are correct and tested both directions.
- The delayed-attack fan-out (TriggerDetector) correctly gates on the Alone predicate before fanning out — matchesEventForWatchedEntity delegates AttackEvent to
matcher.matchesTrigger, which enforces requires.all{…} → Alone -> attackers.size == 1 — then binds TriggeringEntity per attacker. So "attacks alone" can't misfire on a
multi-attacker swing.
- Uncommitted refinements are exactly right: graveyardSneakGrantCost now reads state.projectedState.getController (projection rule — a control-changed grantor moves correctly),
and TurnManager clears the turn-scoped affinity riders so an unused grant can't leak into a later turn. Both have paired tests.
- Tests carry their weight: Ninja Teen L2 (no grant) vs L3 (grant); affinity consume + cross-turn leak; Rat King leaves the non-matching Snake behind; the {3} instant going free
proves the actual reduction, not just rider presence.
- Doc reference updated for all four new types; snapshot golden refreshed.

Issues

Minor

1. Missing the rule-corner test for The Last Ronin — TheLastRoninScenarioTest.kt. The fan-out + Alone gate exist precisely to handle the multi-attacker case, but the only test
declares a single attacker. Add a test that declares two creatures attacking together and asserts neither gets counters/keywords — that's the corner that proves "alone" actually
works (a decorative-citation risk per the test rubric otherwise).
2. wasSneaked uses a player-level signal — CastSpellHandler.kt:~2453. It keys off graveyardSneakGrantCost(currentState, playerId, …) != null (any active grant) rather than this
card's effective sneak. It's not wrongly reachable today — the enumerator only emits a SNEAK action for printed-Sneak hand cards (covered by the || printed-Sneak branch) or
grant-eligible graveyard creatures — and the comment correctly explains the card has already left the graveyard zone so the precise check isn't available. Fine to ship, but a
one-line note that correctness here leans on the enumerator's gating would help the next reader.
3. ReturnSameNamedFromGraveyardEffect bakes tapped=true / battlefield / "your graveyard" — acceptable for Rat King's single mechanic, and the dynamic "same name as the resolved
target" genuinely isn't expressible as a static GameObjectFilter, so a dedicated effect is justified. Just flagging the target-only constructor as the seam if a non-tapped /
to-hand variant ever appears.
4. Latent: per-attacker fan-out only re-applies the GameObjectFilter, not per-attacker AttackPredicates — TriggerDetector.kt:~196. Fine for Alone (a whole-event property), but a
future delayed trigger using a per-attacker predicate like FirstTimeEachTurn (which binds sourceId, not the attacker) wouldn't filter per-attacker. Out of scope here; worth a
comment so it isn't assumed to work.

Recommendation

Ship as-is after adding the two-attacker negative test for The Last Ronin (#1) — that's the one gap with real value. #2–#4 are author's-discretion comments. I didn't run
check-card-printing since all seven are new TMT cards with TMT as their only/earliest printing, so placement is trivially correct.

PR #1050, opened 2026-06-27
Fix auto-build manabase fixing off-color basics for castability-relax…

bolav:
…ed fill

A fresh auto-build that relaxes castability to reach 23 nonland cards can pull off-color filler (e.g. a lone {W} or {G} card) into a 2-color deck. `DraftsimMana.pipCounts` always counts plain colored pips even when they fall outside the allowed set, so that filler leaked off-color demand into the manabase, and `splashFloor` then fabricated basics for it — producing a 4-color manabase (e.g. {R=4,U=4,G=4,W=4}) under a deck whose reported colors stayed [R,U]. The label and the lands disagreed.

Restrict basic-land allocation (and the splash floor) to the deck's committed colors so basics only ever fix `forcedColors` (plus any intentional splash). Completion mode is unaffected: locking a genuine four-color pile still commits to and fixes all four.

Add DraftsimManabaseColorTest (regression over the real TMT sealed pool that surfaced this) and update SPEC_deckbuild.md §3.4/§3.5.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/draftsim/DraftsimDeckBuilder.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/draftsim/DraftsimDeckBuilder.kt
@@ -314,10 +314,17 @@ class DraftsimDeckBuilder(private val tables: DraftsimSetTables) {
         }.take(DECK_LANDS)
         var basicsCount = max(0, DECK_LANDS - usefulLands.size)
 
+        // Basics only ever fix the deck's committed colors ([deckColorSet]) — never an off-color card
+        // that the castability-relaxed greedy fill dragged in. `pipCounts` always counts plain colored
+        // pips even when they fall outside the allowed set, so a {W} filler in an RU build would
+        // otherwise leak White demand and have `splashFloor` fabricate Plains the deck can't use — a
+        // manabase whose colors no longer match the build's reported [forcedColors].
+        val basicDemand = demand.filterKeys { it in deckColorSet }.ifEmpty { demand }
+
         // §3.4 allocate basics proportional to pip demand.
-        val basics = allocateBasics(demand, basicsCount)
+        val basics = allocateBasics(basicDemand, basicsCount)
         // §3.5 splash floor for demanded colors beyond the top 2.
-        splashFloor(basics, demand, deckCards, usefulLands)
+        splashFloor(basics, basicDemand, deckCards, usefulLands)
         // §3.6 trim to 17 total.
         trimLands(basics, usefulLands.size)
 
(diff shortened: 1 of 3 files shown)


bolav:
Review: fix-autobuild-manabase-offcolor-leak

Verdict. Behavior is correct and the fix is at the right layer. This is a focused, well-targeted one-line correction to the Draftsim manabase port, with a real regression test
that exercises the exact failure and passes. No SDK surface is added — this is AI/deckbuild logic, not engine/card SDK — so the SDK-elegance lens is N/A here. Ready to merge.

What's good

- Right layer for the fix. The root asymmetry is in DraftsimMana.pipCounts (DraftsimMana.kt:152): hybrid sides respect allowed, but plain {X} pips are always counted. Spec §2
(SPEC_deckbuild.md:127) confirms that plain-pip behavior is a faithful port of Draftsim. So patching pipCounts would diverge from the reference and ripple into the a-derivation
(DraftsimDeckBuilder.kt:301). Restricting the demand at the basics-allocation step (basicDemand, line 322) instead is the correct, localized choice.
- Spec kept in sync. §3.4/§3.5 were updated in the same diff to describe o' and why, matching the code's basicDemand and its use in both allocateBasics and splashFloor. This is
exactly the discipline the repo asks for.
- The test pins the real bug. The first test asserts the general invariant (basicsNeeded.keys ⊆ colors) on a fresh build; the second locks a W/U/B/R/G-spanning maindeck whose
plain pips only commit R/U/G/W and asserts no stray Swamp leaks in — i.e. the off-color leak the black-heavy pool would otherwise produce. The comment even names the source pool
(colors=[R,U] but basicsNeeded={R,U,G,W}). Good, motivated coverage on both the forced and fresh paths.
- Reasoning holds on both paths. Forced (completion/refine) path: deckColorSet == forcedColors, so basicDemand strips exactly the relaxed-fill off-color plain pips — the
intended scenario. Non-forced path: a is derived from all plain pips (line 301), so every plain-pip color is already in deckColorSet and the filter is a no-op — no legitimate
color is ever dropped. splashFloor now iterates only on-color demand, while its maxReq (line 360) still reads the true per-card requirement. All consistent.

Issues

Minor

- DraftsimDeckBuilder.kt:322 — the .ifEmpty { demand } fallback re-admits the full (leaky) demand when no committed color has any pip (e.g. an all-colorless locked pile forced
into colors with zero on-color pips). It's a degenerate case and producing some basics beats zero, so this is acceptable — but it's the one path where the leak can still
surface. A one-clause note in the comment ("fallback only when the build has no on-color pips at all") would make that explicit. Author's discretion.
- DraftsimManabaseColorTest.kt:21-44 — the add(BCard(...)) block is indented ~12 spaces deeper than its enclosing fun buildPool, which reads like a stray indentation. Purely
cosmetic; doesn't affect compilation.

No CR rule numbers are cited anywhere in the diff (this is deckbuild AI, not rules), so there's nothing to verify against the Comprehensive Rules.

Recommendation

Merge as-is. Optionally tighten the .ifEmpty comment and fix the test indentation — both are discretionary. The branch already has origin/main merged in (it was a no-op /
already up to date), and the new test is green (:ai:test BUILD SUCCESSFUL, both cases PASSED).

PR #1079, opened 2026-06-28
Complete Teenage Mutant Ninja Turtles (190/190) + reanimation reveal fix

bolav:
Finishes the TMT set and lands the last two deferred cards within the existing
zone model (no sideboard/wishboard zone added), plus three correctness fixes
surfaced during the sweep.

## Cards
- **North Wind Avatar** — cast-gated "wish" with no reveal clause. Adds an
  optional `revealed` param to `Patterns.Sideboard.wish` (defaults `true`, so the
  Burning Wish cycle is untouched); North Wind Avatar passes `revealed = false`.
- **Turtles Forever** — "you assemble, they split" pile, composed from the
  `Effects.Pipeline { }` facade (multi-zone gather → exactly-4/different-names →
  reveal → opponent splits two → hand / shuffle rest into library).

## Fixes
- **Krang & Shredder** — restrict the Disappear cast to nonland cards (lands in
  the linked-exile pile can't be cast; CR 305/601).
- **Party Dude L2** — `TriggerBinding.ANY` so an opponent's artifact leaving play
  triggers the draw (was defaulting to SELF).
- **Reanimation reveal** — `CardsRevealedEvent` revealer is now the controller
  performing the move, not the card's owner, so stealing a graveyard card onto
  your board (Shark Shredder) surfaces the overlay to the victim, not the thief.
- Client: `GraveyardTargetingUI` labels reanimation as "Put onto Battlefield"
  rather than the "Return to Hand" fallback.

## Tests
New scenario tests for all four cards (incl. clamp-below-four search and the
reanimation-reveal assertion). TMT snapshot re-blessed. CR citation `701.20
(Reveal)` verified. `CardLintTest` + affected scenario tests green.

Set marked complete; backlog archived.


bolav:
● Review: tmt-finish (TMT completion — 7 commits)

▎ Note: the two issues below have since been applied and verified green (Turtles Forever → Effects.Pipeline, executor comment reworded; CardLintTest + scenario tests pass, snapshot re-blessed). Re-emitting the review as written.

Branch already has origin/main merged in (git merge reported Already up to date — no merge commit needed). Reviewed in place on the current working tree.

Verdict

Behavior is right and the SDK shape is almost entirely right. The two new cards (North Wind Avatar, Turtles Forever) introduce zero new SDK types — exactly the goal — and the three fixes (Krang nonland filter, Party Dude ANY binding, reanimation reveal controllerId)
are correct and well-tested. One SDK-elegance issue: Turtles Forever hand-threads raw pipeline constructors instead of using the Effects.Pipeline { } facade the project built precisely to replace that. Nothing blocking.

What's good (keep this)

- wish(revealed = …) is the right shape. A single optional param, defaulting true, leaves the entire Burning Wish cycle untouched while letting North Wind Avatar pass revealed = false. Parameterized for the next card, not just this one. Oracle verified against local
Scryfall — North Wind Avatar genuinely has no "reveal that card" clause, so revealed = false is correct.
- CR-citation discipline. The doc/code comment correctly migrated 701.19j → 701.20 (Reveal) — I verified against MagicCompRules_20260417.txt: 701.19 is now Regenerate, 701.20 is Reveal. 305.1 (lands are played, not cast), 601, 100.4, 400.11a all check out.
- Krang nonland filter is correct (Krang exiles "until they exile a nonland card", so the linked-exile pile routinely holds lands; "you may cast" excludes them) and the regression test asserts the land is neither offered nor playable — both halves.
- Party Dude TriggerBinding.ANY matches the oracle ("an artifact an opponent controls"), and the test pairs a positive (Level 2 draws) with a negative (Level 1 doesn't) — good axis coverage.
- controllerId reanimation fix is correctly scoped. controllerId only diverges from ownerId when controllerOverride != null && destination == BATTLEFIELD (MoveToZoneEffectExecutor.kt:78-83), so the comment's claim that hand/library returns are unchanged actually
holds. SharkShredderKillerCloneTest asserts both revealingPlayerId == player and revealToSelf == false.
- Turtles Forever tests are thorough — the full four-card path and the clamp-below-four path, plus the subtle "sideboard pull ends up in library, not back in the sideboard" assertion.

Issues

Important

Turtles Forever bypasses the Effects.Pipeline { } facade — TurtlesForever.kt:18-62. The card composes Effects.Composite(listOf(GatherCardsEffect(...), SelectFromCollectionEffect(...), RevealCollectionEffect(...), ...)), hand-threading the string slots "tfPool" →
"tfFound" → "tfChosen"/"tfRest". That is exactly the pattern Effects.Pipeline { } was built to replace ("the facade-respecting replacement for hand-threading string slot keys between raw pipeline step constructors"), and CLAUDE.md mandates the Effects.* facade over
raw constructors. I confirmed PipelineBuilder covers every step this card needs — multi-zone gather, chooseExactly(restrictions = …), reveal, chooseExactlySplit(chooser = Chooser.Opponent, alwaysPrompt = true), toHand, and run(…) for the trailing shuffle — and it
serializes to the same CompositeEffect tree, so the snapshot stays valid. Rewrite:

spell {
    effect = Effects.Pipeline {
        val pool = gather(
            CardSource.FromMultipleZones(
                zones = listOf(Zone.LIBRARY, Zone.SIDEBOARD),
                player = Player.You,
                filter = GameObjectFilter.Creature.legendary(),
            )
        )
        val found = chooseExactly(
            4, from = pool,
            restrictions = listOf(SelectionRestriction.OnePerCardName),
            prompt = "Search for exactly four legendary creature cards with different names",
        )
        reveal(found)
        val (chosen, rest) = chooseExactlySplit(
            2, from = found, chooser = Chooser.Opponent, alwaysPrompt = true,
            prompt = "Choose two of those cards to put into their hand",
        )
        toHand(chosen)
        move(rest, CardDestination.ToZone(Zone.LIBRARY, Player.You))  // default placement — shuffled next
        run(ShuffleLibraryEffect())
    }
}

Typed slot handles instead of four magic strings; same behavior, fewer ways to mistype a slot key. (Use the generic move for the remainder rather than toLibraryBottom — the latter would inject a ControllerChooses ordering prompt right before a shuffle that makes order
moot.)

Minor

- MoveToZoneEffectExecutor.kt:140-142 — the comment "For hand/library returns there is no control change, so controllerId == ownerId" states the right conclusion for a slightly wrong reason. The invariant actually holds because controllerOverride is only consulted
when destination == Zone.BATTLEFIELD (lines 79-83) — a bounce of an opponent's permanent to its owner's hand is control-relevant, yet controllerId is still forced to ownerId. Consider "returns to hand/library never carry a controller override, so controllerId ==
ownerId" to name the real mechanism.

Recommendation

Optionally refactor Turtles Forever to Effects.Pipeline { } (rewrite above) — that's the one SDK-elegance nit. Everything else is mergeable as-is: oracle text, CR numbers, projection, and reveal scoping all check out.

Test status: affected scenario tests (Krang, NorthWindAvatar, PartyDude, TurtlesForever, SharkShredderKillerClone) + CardLintTest + the TMT snapshot all pass.

PR #1102, opened 2026-06-29
Fix "any target" / planeswalker targeting to read projected types

bolav:
## What

Battlefield targeting in `TargetFinder` filtered candidates by the **printed**
type line (`cardComponent.typeLine.isCreature` / `cardComponent.isPlaneswalker`),
so a permanent whose type comes from a continuous effect was never offered as a
legal target. Concretely: an Earthbend land (e.g. via Earthbending Lesson) becomes
a 0/0 land *creature* in projection, but Cat-Gator's "deal damage to any target"
ETB could not target it.

## Change

All four type checks in `TargetFinder` now read **projected** state
(`projected.isCreature` / `projected.isPlaneswalker`), matching this engine's
projected-vs-base-state rule (CR 115.4 governs what "any target" may be):

- `findAnyTargets`
- `findCreatureOrPlaneswalkerTargets`
- `findOpponentOrPlaneswalkerTargets`
- `findPlayerOrPlaneswalkerTargets`

No SDK types added — pure correctness plumbing.

## Tests

`CatGatorEarthbendTargetingTest` asserts an earthbended Forest is **not** a legal
"any target" / "creature or planeswalker" before animation and **is** after.
Verified green (`:rules-engine:test`, BUILD SUCCESSFUL).

Out of scope: Cat-Gator still lacks a card-level ETB-damage scenario test; "any
target" does not yet include battles (not modeled in this engine).

### DIFF
--- a/manual-scenarios/cards/c/cat-gator-earthbend-target.json
+++ b/manual-scenarios/cards/c/cat-gator-earthbend-target.json
@@ -1,0 +1,41 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Cat-Gator", "Earthbending Lesson"],
+    "battlefield": [
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"}
+    ],
+    "graveyard": [],
+    "library": ["Swamp", "Forest", "Swamp"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Badgermole"},
+      {"name": "Forest"}
+    ],
+    "graveyard": [],
+    "library": ["Forest", "Forest", "Swamp"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "SELF"
+}
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/TargetFinder.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/TargetFinder.kt
@@ -193,10 +193,13 @@ class TargetFinder(
         val battlefield = state.getBattlefield()
         for (entityId in battlefield) {
             val container = state.getEntity(entityId) ?: continue
-            val cardComponent = container.get<CardComponent>() ?: continue
+            if (!container.has<CardComponent>()) continue
             val entityController = container.get<ControllerComponent>()?.playerId
 
-            if (!cardComponent.isPlaneswalker) continue
+            // Read the PROJECTED type line, not the printed one, so a permanent that
+            // becomes a planeswalker via a continuous effect is offered (CR 115.4 /
+            // projection rule), consistent with findAnyTargets.
+            if (!projected.isPlaneswalker(entityId)) continue
 
             // Check hexproof/shroud
             if (projected.hasKeyword(entityId, Keyword.HEXPROOF) && entityController != controllerId) continue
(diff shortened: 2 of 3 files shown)


bolav:
# Review: `fix-anytarget-projected-types` (base: `more-tla`)

**Reviewed in place** — clean tree, branch is exactly one commit ahead of `more-tla`
(merge-base `258f141df`), nothing in `more-tla` is missing. No merge needed.

Lens: SDK elegance + engine correctness (projected vs base state).

## Verdict

Correct, minimal, well-targeted fix. It swaps a base-state type check for a projected
one in `TargetFinder` so "any target" can hit animated lands — exactly the right move
under this engine's projected-state rule. The SDK surface is untouched (no new types —
good; this is pure correctness plumbing). Two follow-up items were raised and have since
been addressed (see **Resolution** below).

## What's good

- **Right root cause, right tool.** `findAnyTargets` (line 327) and
  `findCreatureOrPlaneswalkerTargets` (line 389) now read
  `projected.isCreature(entityId)` / `projected.isPlaneswalker(entityId)` instead of the
  printed `cardComponent.typeLine.isCreature`. This is precisely the "projected vs base
  state" rule for battlefield type reads (`architecture-principles.md` §2.3).
  `isCreature`/`isPlaneswalker` exist on `ProjectedState` (lines 145/147) and delegate to
  `hasType`.
- **`has<CardComponent>()` instead of binding the unused component.** Since the type read
  moved to `projected`, the `cardComponent` local was dead — the diff correctly drops it
  to a presence check rather than leaving an unused binding.
- **Accurate rule citation.** The comment cites **CR 115.4**, verified against
  `MagicCompRules_20260417.txt:853`: it's exactly the "any target → creatures, players,
  planeswalkers, battles" rule. The projection mechanism itself is described by name, not
  a guessed number — matches the CLAUDE.md guidance.
- **Comment accuracy.** Cat-Gator's ETB really does use `AnyTarget()` (`CatGator.kt`), and
  Earthbend animates a *land into a 0/0 creature* (`EarthbendingLesson.kt` /
  `Effects.Earthbend`), so the scenario the comment describes is real.
- **Test isolates the regression cleanly.** It asserts a plain Forest is *not* an
  `AnyTarget` candidate before animation and *is* after `Earthbending Lesson` resolves — a
  true before/after on the projected-vs-base distinction, not a tautology. All driver
  helpers (`putLandOnBattlefield`, `castSpell`, etc.) resolve in `GameTestDriver`.

## Issues

### Important — second changed path (`findCreatureOrPlaneswalkerTargets`) has no test

The diff fixes two methods, but the test only exercises `AnyTarget` (`findAnyTargets`,
line 327). The parallel change at `TargetFinder.kt:389` (`TargetCreatureOrPlaneswalker`)
is never hit. It's an identical one-liner so risk is low, but per the review bar every
changed path should be exercised.

*Fix:* add an assertion calling
`findLegalTargets(state, TargetCreatureOrPlaneswalker(), controllerId = you)` against the
same earthbended Forest.

### Minor — planeswalker siblings still read base state

`findOpponentOrPlaneswalkerTargets` (`TargetFinder.kt:199`) and
`findPlayerOrPlaneswalkerTargets` (line 235) still use `cardComponent.isPlaneswalker`
instead of `projected.isPlaneswalker(entityId)`. After this commit the file is internally
inconsistent — the same concept is projected in two methods and base in two others.
Earthbend only makes creatures so it isn't the trigger here, but a "becomes a
planeswalker" projection (or a face-down permanent) would be mis-handled by those two.

*Fix:* switch both to `projected.isPlaneswalker`, or note as a deliberate follow-up.

### Minor — `findAnyTargets` omits battles

CR 115.4 (the very rule cited) also lists *battles* as legal "any target"s; the method
only adds players/creatures/planeswalkers. Pre-existing, and battles don't appear to be
modeled in this engine, so out of scope — flagged only so the citation isn't read as
fully covered.

### Informational — manual-scenario JSON is a repro fixture, not coverage

`manual-scenarios/cards/c/cat-gator-earthbend-target.json` is a hand-loaded repro fixture
(consumed by `game-server`'s `ScenarioSessionFactory`, like its 48 siblings), not
automated coverage. The board is a valid repro (7 Swamps → 7 damage kills the 4/4
earthbended land). Fine to keep; just not a substitute for the unit test.

## Recommendation

Mergeable after: (1) add the one-line `TargetCreatureOrPlaneswalker` assertion so both
changed methods are covered, and (2) decide on the planeswalker base-vs-projected
inconsistency at lines 199/235 — fix here or file a follow-up.

Separately (out of scope for this targeting fix): `CatGator.kt` is still
mtgish-generated (`// === GENERATED by mtgish-tooling ...`) and lacks a card-level
scenario test that resolves its ETB damage end-to-end. That's the real gate per the
project's auto-gen rules and remains outstanding.

---

## Test result

`./gradlew :rules-engine:test --tests "...CatGatorEarthbendTargetingTest"` (isolated,
`--rerun-tasks`):

```
CatGatorEarthbendTargetingTest > an earthbended land is a legal "any target" PASSED
BUILD SUCCESSFUL
```

Report: `tests="1" skipped="0" failures="0" errors="0"`.

---

## Resolution (committed `a2411b18b`)

Both review items above were addressed in a follow-up commit on the branch:

1. **Planeswalker base→projected consistency** — `findOpponentOrPlaneswalkerTargets` and
   `findPlayerOrPlaneswalkerTargets` switched to `projected.isPlaneswalker(entityId)`
   (dropping the now-unused `cardComponent` binding to a `has<CardComponent>()` presence
   check). All four type checks in `TargetFinder` now read projected state uniformly.
2. **Second changed path now tested** — the regression asserts the earthbended Forest is
   (and, before animation, is not) a legal `TargetCreatureOrPlaneswalker`, covering
   `findCreatureOrPlaneswalkerTargets`.

Re-verified green: `CatGatorEarthbendTargetingTest ... PASSED`, `BUILD SUCCESSFUL`, all
tasks executed (genuine rerun).

The "omits battles" and "Cat-Gator needs a card-level ETB scenario test" notes were left
as out-of-scope follow-ups.

PR #1122, opened 2026-06-30
Add Exhaust keyword + double strike counter (8 TLA cards)

bolav:
Implements the **Exhaust** keyword (CR 702.177) and 8 Avatar: The Last Airbender cards.

## Engine / SDK
- `isExhaust = true` on an `activatedAbility` renders the "Exhaust — " prefix and
  auto-adds `ActivationRestriction.Once`, so the marker and its enforcement can't drift.
  No new Effect/Trigger type — pure composition over the existing once-per-object tracker.
- Fix: `ZoneMovementUtils` now strips `AbilityActivatedEverComponent` on battlefield exit,
  so a blinked/bounced-and-recast permanent is a new object whose exhaust ability resets
  (CR 400.7). This also fixes a latent bug for any pre-existing `Once` ability.
- New **double strike** keyword counter (`CounterType.DOUBLE_STRIKE`), projected via
  `KEYWORD_COUNTER_MAP` (CR 122.1b); web-client counter badge added.

## Cards
Bitter Work, Hog-Monkey, Invasion Submersible (Exhaust—Waterbend), Jeong Jeong the
Deserter, Mai Jaded Edge (double strike counter), Rebellious Captives, Rough Rhino
Cavalry, The Legend of Kuruk // Avatar Kuruk (Saga DFC + Exhaust—Waterbend {20}).
All compose existing primitives; text/P/T verified against Scryfall.

## Tests
14 new scenario tests (once-then-illegal, bounce-resets, waterbend exhaust, Saga
transform → extra turn, copy-next-Lesson). All green. Docs + SDK reference updated.

PR #1125, opened 2026-06-30
Fix two TLA battlefield UI bugs: duplicate granted-ability badge + unclickable stacked tokens

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1122

Two frontend bug fixes surfaced while playing TLA, both at the display layer (no
game-state or SDK changes).

### 1. Duplicate "Granted Ability" badge
Earthbending a land twice grants the identical "return it tapped" self-trigger twice —
two real grant entries, each with its own AbilityId. To the player that's one ability,
but the client painted two stacked "Granted Ability" tiles. `ClientStateTransformer`
now dedupes the *badge* by description across triggered/activated/keyword grants, while
leaving both grants live (both still fire). Added a rules-engine scenario test asserting
two grant entries collapse to a single badge.

### 2. Hidden stacked tokens were unclickable
A stack of N identical tokens renders at most `MAX_VISUAL_STACK_DEPTH` (4) peeked cards
plus a "×N" count badge. Server-driven actions reach every member, but flows needing
direct clicks on individual hidden members (e.g. waterbend — tap N tokens to pay) could
only reach the front 4. Added a ⤢ toggle on `CardStack` that ungroups the stack into a
wrapped row of individually-clickable cards (⤡ collapses it back).

### Tests
- New `EarthbendingLessonTest` case for the badge dedup.
- Two manual repro scenarios under `manual-scenarios/bugs/`.

image

image

Old:
image

New:
image


bolav:
Review: frontend-bugs vs tla-exhaust

Verdict

Two well-scoped frontend bug fixes, both at the right altitude. No new SDK types, no game-state mutation — the granted-ability fix is a pure display-layer dedup and the stack
fix is pure presentation. Behavior is correct. Two robustness concerns on the new ungroup toggle are worth addressing before merge. origin/main is already an ancestor of
frontend-bugs, so no merge was needed and the diff reflects post-merge reality.

What's good

- Right layer for the dedup. The duplicate "Granted Ability" badge is fixed in ClientStateTransformer (the view), not by collapsing the two real grant entries. The comment is
explicit that both grants remain live game state and both still fire — only the badge is deduped. That's exactly correct: earthbending a land twice genuinely is two triggers.
- Dedup key is sound. ability.description is a non-null String on all three ability types (TriggeredAbility, ActivatedAbility, KeywordAbility), so the shared HashSet<String>
can't collapse distinct grants via a null key.
- Good comments. Both new comment blocks explain why (the fresh-AbilityId-per-grant rationale; the "only the front 4 are clickable" rationale), not what the code does.
- New rules-engine test targets the real seam. earthbended twice — client view shows a single Granted Ability badge asserts the premise (grantedTriggeredAbilities.count == 2)
and the outcome (one granted-ability badge), and lives in the correct module (the transformer is rules-engine). Constructor ClientStateTransformer(cardRegistry) and the
activeEffects/icon DTO fields all match.
- Manual scenarios are valid. Both JSON files use real ScenarioDtos fields (player1StopAtSteps, player2OpponentStopAtSteps, mode: SELF) — verified against ScenarioDtos.kt.
- Click-through guarded. StackToggle uses e.stopPropagation() + zIndex: 1000, so the chip never falls through to a card underneath.

Issues

Important

- CardStack.tsx:102-126 — the ungroup view has no size cap, reintroducing the number-explosion risk the stack cap exists to prevent. The collapsed view caps rendered layers at
MAX_VISUAL_STACK_DEPTH (4) precisely because cardGrouping.ts is built around hordes up to ~10,000 identical tokens. The expanded branch does group.cards.map(...) with no bound,
painting one full GameCard per member. Ungrouping a large stack would render thousands of DOM nodes and can freeze the tab. Suggest gating the toggle (or the expanded render) on
a threshold — e.g. only offer ungroup when group.count <= N, or paginate/scroll the expanded row — so a deliberate click on a horde can't lock the UI.
- CardStack.tsx:83 + Battlefield.tsx:242 — expanded is local state but the parent keys the component by group.cardIds[0], so a membership change that shifts the first id
remounts and silently resets the toggle. When any member diverges (counter, tap, combat) and it happens to be cardIds[0], the React key changes, CardStack remounts, and expanded
flips back to false, re-collapsing the stack. The intended waterbend flow likely survives this (cost selection usually doesn't tap the tokens client-side until the server
confirms, so they stay grouped), but it's a real fragility — worth either lifting expanded into a store keyed by group identity, or keying the component on a content-stable
group key rather than cardIds[0]. At minimum, confirm the waterbend tap-selection path doesn't diverge the front member mid-selection.

Minor

- ClientStateTransformer.kt:2599 — shared dedup set spans all three grant types. If a granted activated ability and a granted keyword ability ever shared an identical
description string, one badge would be suppressed. Practically impossible today (descriptions differ across types) and harmless if it ever happened (identical text =
indistinguishable to the player), so just noting it — no change needed.
- No automated coverage for the toggle itself. groupCards.test.ts covers the grouping logic but nothing exercises StackToggle/expanded rendering; the change leans on the two
manual scenarios. Acceptable per the web-client convention (UI → manual scenarios + screenshots), but the number-explosion concern above in particular has no regression guard.
Per web-client/CLAUDE.md, a screenshot of the ungrouped layout in the PR body would let a reviewer judge the visual result.
- Test run not green locally. EarthbendingLessonTest could not complete — both attempts died on the shared Gradle daemon crashing (sibling-container collision per
CLAUDE.local.md), not on test failure. Worth a clean re-run on an uncontended box before merge to confirm green.

Recommendation

Behavior and SDK shape are right; nothing to drop. Before merge: (1) cap or threshold-gate the expanded render so ungrouping a large stack can't freeze the UI; (2) decide
whether expanded should survive a front-member divergence (lift it out of cardIds[0]-keyed local state, or confirm the waterbend path keeps the group intact during selection);
(3) get one clean EarthbendingLessonTest run on an uncontended machine. The dedup fix and its test are good as-is.

(Reviewed in place on frontend-bugs; no worktree created, origin/main already merged.)

PR #1130, opened 2026-06-30
Fix three TLA card bugs: conditional-mana auto-tap, reflexive connive target, token-leave trigger

bolav:
Three TLA card bugs, each with a paired regression test. Adds two small reusable
SDK/engine capabilities and one HandPatterns composition — no card-specific types.

### Fixes
- **Raucous Audience** — auto-tapper now recognizes a *conditional* mana ability as a
  source. `ManaSolver` unwraps the `GatedEffect` that `ConditionalEffect` lowers to,
  evaluating its gate against current state (stable during a payment) and reading the
  branch that will run. Before, the gate fell through unread and the source was skipped.
- **Teo, Spirited Glider** — the +1/+1 counter now lands on a *reflexively chosen*
  target: `Effects.ConniveTargeting` selects the recipient at resolution, inside the
  nonland gate, so the player only picks a creature after actually discarding a nonland
  (not up front, and never on a land). Replaces the up-front `target(...)`.
- **Suki, Courageous Rescuer** — "whenever another permanent you control leaves the
  battlefield" now fires for a leaving **token**. `TriggerMatcher` gets an LKI-safe
  `CardPredicate.IsPermanent` case that reads the event's last-known type line, since the
  token entity is swept by SBA (CR 704.5d) before trigger detection.

### Tests
- `ConditionalManaAbilityAutoTapTest` — both gate branches (1 green / {G}{G}).
- `ConniveTargetingTest` — nonland → target after discard; land → no target.
- `TriggerMatcherIsPermanentTest` — swept token (LKI) + nontoken baseline.

Also de-dups the connive pipeline (shared `connivePipeline` helper), updates the SDK
reference, and re-blesses the TLA snapshot for Teo. CR citations verified locally.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1523,7 +1523,8 @@ one-off pipeline belongs inline in the card file via `Effects.Pipeline { }` (§5
 - `forage(afterEffect?)` — Forage cost; choose card-from-hand to play.
 - `loot(draw?, discard?)` — "draw N, discard M" loop.
 - `rummage(count?)` — discard then draw.
-- `connive(target?)` — draw 1, discard 1, then put a +1/+1 counter on `target` if the discard was a nonland (CR 702.166). Also exposed as `Effects.Connive(target)`.
+- `connive(target?)` — draw 1, discard 1, then put a +1/+1 counter on `target` (default Self) if the discard was a nonland (CR 702.166). Also exposed as `Effects.Connive(target)`.
+- `conniveTargeting(requirement, storeAs?)` — connive whose +1/+1 counter lands on a *reflexively chosen* target: "draw a card, then discard a card. When you discard a nonland card this way, put a +1/+1 counter on target creature you control" (Teo, Spirited Glider). The recipient is selected at resolution via `SelectTargetEffect` *inside* the nonland gate — so the player never chooses up front or when the discard is a land. Pass the recipient's `TargetRequirement` (e.g. `Targets.CreatureYouControl`); do **not** also declare it as a cast-time `target(...)`. Exposed as `Effects.ConniveTargeting(requirement)`.
 - `readTheRunes()` — "draw X cards; for each, discard a card unless you sacrifice a permanent." Composes `RepeatDynamicTimesEffect(XValue, ChooseActionEffect(...))` with feasibility guards. Exposed as `Effects.ReadTheRunes()`.
 - `eachOpponentMayPutFromHand(filter?)` — each opponent may dump a matching card.
 - `putFromHand(filter?, count?, entersTapped?, entersAttacking?)` — you may put N from hand onto
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/dsl/Effects.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/dsl/Effects.kt
@@ -482,6 +482,19 @@ object Effects {
         HandPatterns.connive(target)
 
     /**
+     * Connive whose +1/+1 counter lands on a *reflexively chosen* target rather than the conniving
+     * permanent — "draw a card, then discard a card. When you discard a nonland card this way, put a
+     * +1/+1 counter on target creature you control" (Teo, Spirited Glider).
+     *
+     * The counter recipient is selected at resolution, *after* a nonland card has actually been
+     * discarded — so the player never picks a target up front or when the discard is a land. Pass the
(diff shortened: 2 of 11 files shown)


bolav:
# Review: `bugfixes` vs `tla-small-engine`

**Base:** `tla-small-engine` (`24ed87e90`) · **Head:** `bugfixes` (`f0ccae4fc`)
**Scope:** 2 commits — `187e338ca` (Fix three TLA card bugs) + `f0ccae4fc` (review fixes folded in).
The TLA card additions (Ran and Shaw, Trusty Boomerang, North Pole Patrol, Firebending Student) and
`EffectTarget.GrantingSource` are all in the base, so they are **not** part of this diff.

Files touched (11): `ManaSolver.kt`, `TriggerMatcher.kt`, `HandPatterns.kt`, `Effects.kt`,
`TeoSpiritedGlider.kt`, `CreateTokenCopyOfSourceExecutor.kt` (comment only), the SDK reference,
the TLA snapshot, and three new test classes.

## Verdict

Three well-scoped, well-tested bug fixes. Two of them add small reusable capabilities
(`GatedEffect` mana-unwrap; an LKI-safe `IsPermanent` match) and one composes existing primitives
into a `HandPatterns` recipe (`conniveTargeting`) rather than growing a card-specific type. CR
citations are correct and the connive pipeline is de-duplicated. Nothing blocking; ready to merge.

## What's good (keep this)

- **`ManaSolver` `GatedEffect` unwrap** (`ManaSolver.kt`) — clean recursion over
  `CompositeEffect`/`GatedEffect`, evaluating the gate against current state (correctly justified:
  the battlefield is stable during a single payment, so the branch that will run is knowable).
  `ConditionalManaAbilityAutoTapTest` exercises *both* branches — else = 1 green, then = {G}{G}
  from one tap with a power-4+ creature out. Right fix, right tests.
- **`conniveTargeting`** — modeled as a `HandPatterns` composition of existing atomics (no new
  `Effect` type), with the reflexive target selected via `SelectTargetEffect` *inside* the nonland
  gate. `ConniveTargetingTest` proves the recipient is chosen only after a nonland discard, and
  never on a land. Correct shape and correct timing — exactly the "When you discard a nonland card
  this way…" semantics.
- **`TriggerMatcher` `IsPermanent`** — LKI-safe, mirrors the existing `IsCreature`/`IsLand` cases,
  reading the event's last-known type line so a token swept by SBA (CR 704.5d) still matches.
  `TriggerMatcherIsPermanentTest` asserts both the token (swept entity) and nontoken baselines.
- **`connivePipeline` extraction** (`HandPatterns.kt`) — `connive` and `conniveTargeting` now share
  one private body and differ only in the nonland branch, instead of duplicating the whole
  draw→gather→select→move→conditional sequence. Compiled AST is unchanged
  (`CardDefinitionSnapshotTest` green, no re-bless).

## Issues

None blocking or important. The CR citations referenced by the diff are correct:

- **CR 704.5d** — token ceases to exist (`TriggerMatcher.kt` + its test). ✅ verified against
  `MagicCompRules_20260417.txt`.
- **CR 707.9b** — modify-a-characteristic-during-copy, for the "except it's not legendary" clause
  (`CreateTokenCopyOfSourceExecutor.kt`, comment-only hunk in this diff). ✅ verified.
- **CR 702.166** (connive) and **CR 113.7 / 113.7a** (granted-ability source) are unchanged from
  the base and remain accurate.

### Minor (optional)

- `CreateTokenCopyOfSourceExecutor.kt` appears in this diff for a single comment-line CR-citation
  fix; the `removeLegendary` code it annotates lives in the base. Harmless — just noting the lone
  hunk so it isn't mistaken for a behavioral change.

## Tests

All three fixes ship with a paired regression test, and each exercises the rule-corner that drove
the change:

| Fix | Test | Corner covered |
|-----|------|----------------|
| ManaSolver gated mana | `ConditionalManaAbilityAutoTapTest` | both gate branches (1 green / {G}{G}) |
| Teo reflexive connive | `ConniveTargetingTest` | nonland → target after discard; land → no target |
| Suki token-leave | `TriggerMatcherIsPermanentTest` | swept token (LKI) + nontoken baseline |

Verified green locally: `CardDefinitionSnapshotTest`, `ConniveTargetingTest`,
`TriggerMatcherIsPermanentTest`, `ConditionalManaAbilityAutoTapTest`, plus the two scenario tests
that live in the base.

## Recommendation

Merge as-is. No structural rework needed — the two new SDK capabilities are minimal and reusable,
and the third fix is a pure composition.

PR #1166, opened 2026-07-01
Implement Flanking keyword (+ tests, icon)

wingedsheep:
## What & why

The **Flanking** keyword (CR 702.25) was defined but **inert**: 13 Mirage cards carried the bare `Keyword.FLANKING`, yet nothing in the engine ever applied the −1/−1 to non-flanking blockers. This wires up the actual behavior, adds a scenario test net, and hooks up the icon.

Flanking is a *triggered* ability (CR 702.25b): *"Whenever this creature becomes blocked by a creature without flanking, that blocking creature gets −1/−1 until end of turn."*

## Approach — derive, don't author

Rather than adding a `flanking()` builder and editing 13 (mtgish-generated) card files + the emitter + re-blessing snapshots, flanking is **synthesized from the bare keyword** — the same pattern the engine already uses for ward and suspend:

- **`mtg-sdk`** — new `Flanking.blockedByNonFlankerTrigger`: a `BecomesBlockedEvent(filter = Creature.withoutKeyword(FLANKING))` SELF trigger with `ModifyStatsEffect(-1, -1, TriggeringEntity)`.
- **`rules-engine`** — `TriggerAbilityResolver.getFlankingTriggeredAbilities` grants it to any creature whose **projected** keywords include `FLANKING` (so it respects loses-all-abilities *and* granted flanking), wired into both `getTriggeredAbilities` paths. The engine's existing filtered becomes-blocked path fires the trigger once per non-flanking blocker, targeting that blocker — so each blocker independently takes −1/−1 and a blocker that also has flanking is excluded (CR 702.25c).
- Cards stay bare `keywords(Keyword.FLANKING)` — **no card edits, no snapshot re-bless**; future flanking cards work for free.

## Tests

`FlankingCombatTest` (rules-engine) — each case gives the flanking attacker **0 power** so any blocker death is attributable to the −1/−1 alone, not ordinary combat:

- weakens a non-flanking blocker enough to kill it
- a blocker that also has flanking is **not** weakened (CR 702.25c)
- −1/−1 applied to **each** non-flanking blocker under a gang block
- no trigger when the flanking creature is unblocked (bystander untouched, normal damage)

## Icon

`flanking.svg` added to the keyword icon set (`keywordSvgIcon` + `displayableKeywords`) so flanking creatures render the badge, alongside the existing custom SVGs (persist, banding).

## Verification

- `just test-class FlankingCombatTest` — 4/4 pass
- `just test-rules` — full suite green (core resolver path touched)
- `mtg-sets` `CardDefinitionSnapshotTest` + `CardLintTest` — green (card definitions unchanged)
- `web-client` `npm run typecheck` — clean


wingedsheep:
### Icon preview

The new `flanking.svg` rendered the way the in-app `SvgGlyph` / `KeywordGlyph` draws it (CSS mask → white glyph on the keyword chip):

Flanking keyword badge

This is a static render of the exact asset + badge styling rather than a full in-game capture (spinning up the whole stack and getting a flanking permanent onto the battlefield is heavy for a single badge). The rendering path is identical to the existing custom-SVG keywords (persist, banding). The `flanking-tests-icon-screenshots` branch is throwaway.

PR #1219, opened 2026-07-04
Four-bend event system + Avatar Aang // Aang, Master of Elements

bolav:
## Four-bend event system + Avatar Aang // Aang, Master of Elements

Implements the "Whenever you waterbend, earthbend, firebend, or airbend, …" event family
(CR 701.65b / 701.66b / 701.67c / 702.189b) and uses it to complete **Avatar Aang //
Aang, Master of Elements** (TLA 207).

### What's new

**Four-bend event system (SDK + engine)**
- `BendType` enum (`WATER` / `EARTH` / `FIRE` / `AIR`) with oracle verbs.
- `Triggers.YouBend(types = BendType.ALL, player = You)` → `EventPattern.BendPerformedEvent` —
  fires once per bend of any element in `types`; pass a subset (e.g. `setOf(BendType.EARTH)`)
  for a single-element variant.
- `TurnTracker.DISTINCT_BENDS` (0–4), backed by a per-player `BendsThisTurnComponent`
  (`Set<BendType>`, reset each turn) — powers "if you've done all four this turn".
- `EmitBendEventEffect` (internal marker, executor `EmitBendEventExecutor`) folded into
  `Effects.Earthbend`, `Effects.Airbend` / `AirbendAll`, and `firebendingAttackTrigger`.
- `Effects.AirbendSpell` = `ExileTargetSpell(emitAirbend = true)` — the airbend-a-spell stack
  branch now fires the bend once the spell is exiled.
- Waterbend emits engine-side when the cost is **paid** (in `CastSpellHandler` /
  `ActivateAbilityHandler`), ungated on how it was paid — paying entirely with mana still
  fires it (CR 701.67c).

Every emit routes through a single `BendEvents.record` chokepoint so the per-turn set and the
`BendPerformedEvent` stay consistent regardless of emit site. `BendPerformedEvent` is
internal-only (dropped from the client log via `ClientEventTransformer`); both it and
`BendsThisTurnComponent` are registered for polymorphic serialization.

**Avatar Aang // Aang, Master of Elements (TLA 207)**
- Front: Flying, firebending 2; "Whenever you … bend, draw a card. Then if you've done all
  four this turn, transform." (`Triggers.YouBend()` + `DISTINCT_BENDS >= 4`).
- Back: Flying; "Spells you cast cost {W}{U}{B}{R}{G} less" (`ReduceColoredPerUnit`,
  overflowing to generic); upkeep "you may transform … gain 4 life, draw four, four +1/+1
[...]

PR #1220, opened 2026-07-04
Add Koh, the Face Stealer to TLA (chosen-linked-exile ability grant)

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/1219

## Koh, the Face Stealer (TLA #107)

`{4}{B}{B}` Legendary Creature — Shapeshifter Spirit, 6/6, mythic.

> When Koh enters, exile up to one other target creature.
> Whenever another nontoken creature dies, you may exile it.
> Pay 1 life: Choose a creature card exiled with Koh.
> Koh has all activated and triggered abilities of the last chosen card.

Koh builds a personal pile of "cards exiled with Koh" and, on demand, dons the face of one
of them.

### New engine feature — `HasAbilitiesOfChosenLinkedExiledCard`

The self-scoped, single-card, activated-**and**-triggered sibling of the existing
`HasAllActivatedAbilitiesOfLinkedExiledCard` (Territory Forge). It reads the source's new
`ChosenLinkedExileComponent` — stamped by `RecordChosenLinkedExileEffect` after a
`SelectFromCollection` over the linked-exile pile — and grants the source that one chosen
card's:

- **activated** abilities, surfaced in `ActivateAbilityHandler` + `CastPermissionUtils`
  (source recorded as granter, so `{T}` / self-references bind to Koh), and
- **triggered** abilities, via `TriggerAbilityResolver.getSelfGrantedTriggeredAbilities`
  (merged into Koh's general triggered-ability pool, so the stolen triggers match against
  any event — attacks, ETB, dies, upkeep).

Resolution re-reads the choice live, so re-choosing a different exiled card swaps which
abilities Koh has. It never grants static, keyword, or replacement abilities. The two
`grantActivated` / `grantTriggered` flags parameterize it for the next such card. A
"still in exile" last-known guard stops a chosen card that has left exile from lending
abilities.

### Card wiring

- ETB → `Effects.ExileLinkedToSource` (permanent exile into Koh's pile, no return).
- Dies trigger (`TriggerBinding.OTHER`, nontoken) → `MayEffect` + `ExileLinkedToSource`
  of the dying card from its graveyard.
- "Pay 1 life" → `GatherCards(FromLinkedExile)` → `SelectFromCollection` (creature) →
  `RecordChosenLinkedExile`.
- The grant modeled as a separate `staticAbility` (matches the real card's line break).

### Supporting changes

- `ChosenLinkedExileComponent` registered in `engineSerializersModule`.
- `RecordChosenLinkedExile` classified as a collection reader in `CardLinter`.
- Both new types documented in `docs/card-sdk-language-reference.md`.
- TLA snapshot golden re-blessed.

### Tests

`KohTheFaceStealerScenarioTest` (rules-engine) — all green:
- ETB-exile a creature, choose it, Koh gains its activated `{T}` ability (with a negative
  assertion that Koh lacks the ability before choosing).
- Koh gains the chosen card's triggered ability (attack trigger fires from Koh).
- A dying nontoken creature is exiled with Koh, then chosen for its ability.

Oracle text verified against Scryfall; printing placement verified via
`check-card-printing` (canonical in TLA, its earliest real printing).


bolav:
## Review: Koh, the Face Stealer

**Verdict.** Behavior is correct and the SDK shape is defensible. Oracle text matches
Scryfall exactly, printing placement is right, all three scenario tests pass, and the new
static-ability type is wired consistently across every surfacing path. One SDK-elegance
discussion point (a near-sibling type) and a couple of minor notes — nothing blocking.

### What's good — keep this
- **Oracle fidelity.** Text is byte-for-byte the Scryfall Oracle. Correctly modeled as
  four separate abilities: ETB triggered, dies triggered, "Pay 1 life" activated, and
  "Koh has all…" as a distinct **static** ability (matches the real card's line break —
  the grant is not part of the activated ability's effect). "Up to one other target
  creature" → `TargetPermanent(optional = true, …other())`.
- **Consistent plumbing.** The new `HasAbilitiesOfChosenLinkedExiledCard` mirrors the
  existing `HasAllActivatedAbilitiesOfLinkedExiledCard` pattern in all the right places:
  activated abilities surfaced in *both* `ActivateAbilityHandler` and
  `CastPermissionUtils` (self as granter, so `{T}`/self-refs bind to Koh), triggered
  abilities routed through the self-granted path in `TriggerAbilityResolver`, the
  skip-list entry in `StaticAbilityHandler`, and the `Component` subclass registered in
  `engineSerializersModule`. The `Effect`/`StaticAbility` are sealed so no explicit
  polymorphic registration is needed — correct.
- **Trigger breadth is actually robust.** Because `getSelfGrantedTriggeredAbilities`
  merges into the permanent's general triggered-ability list (both call sites), a stolen
  trigger matches against *any* event — upkeep, ETB, dies, attacks — not just the attack
  case in the test.
- **Tests hit the real axes.** Activated-ability inheritance, triggered-ability firing
  *from Koh* (the rule-corner that justifies the triggered flag), and the graveyard→pile
  dies-exile path. The pre-choose negative assertion (Koh lacks the ability before
  choosing) is a nice touch.
- **Guard + hygiene.** `chosenLinkedExiledCardDef` has a last-known "still in exile"
  guard; doc reference, CardLinter classification, and the TLA snapshot golden are all
  updated in the same change.

### Issues

**Important (SDK elegance — a decision, not a blocker)**
- `HasAbilitiesOfChosenLinkedExiledCard` is a close sibling of
  `HasAllActivatedAbilitiesOfLinkedExiledCard`. They differ on three real axes —
  *chosen-one vs whole pile*, *activated+triggered vs activated-only*, *self-scoped vs
  filtered* — so a separate type is justified **today**, and the `grantActivated` /
  `grantTriggered` flags are correctly parameterized. But you now have two types reading
  the same linked-exile pile with duplicated surfacing branches copy-pasted into four
  files. If a third linked-exile-abilities card appears, collapse both into one
  `HasAbilitiesOfLinkedExiledCard(selection = All | Chosen, grantActivated,
  grantTriggered, filter)` rather than adding a third. Flagging so consolidation is a
  conscious future step, not accreted drift. No change needed for merge.

**Minor**
- `chosenLinkedExiledCardDef` does a full `state.zones` scan on every ability query
  (`zoneType == EXILE && chosenId in cards`). Fine at current scale; an obvious
  O(zones)-per-call hotspot only if ability-legality queries ever get hot.
- `ChosenLinkedExileComponent` is deliberately *not* stripped on zone change
  (documented). The reasoning holds because Koh re-enters as a fresh entity id with an
  empty pile, so a stale stamp can't attach to the new incarnation — the safety rests on
  the "still in exile" guard, which is present.
- Consider running the `mtg-sets` **CardLintTest** and **CardDefinitionSnapshotTest**
  before merge (the golden is already updated in the diff, so this should be green). The
  `rules-engine` scenario suite was run here (3/3 green); the mtg-sets nets were not.

### Recommendation
Merge-ready as-is. The only substantive follow-up is the SDK note: **don't add a third
linked-exile-abilities type** — when the next one lands, unify
`HasAllActivatedAbilitiesOfLinkedExiledCard` + `HasAbilitiesOfChosenLinkedExiledCard` into
a single parameterized static ability. Optionally run the two `mtg-sets` test nets to
close out the snapshot/lint gate.

*Reviewed with `tla-avatar-aang` as base (`origin/main` not merged in, per the requested
base). `rules-engine` scenario suite: 3/3 green.*

PR #1231, opened 2026-07-05
Add Lava Dart to Judgment and MH1 reprint

AndreyItMe:
[Add Lava Dart to Judgment and MH1 reprint](https://github.com/wingedsheep/argentum-engine/commit/f318b7d7ea6c518f9ab82b4b6dc13e70615706d1)


wingedsheep:
Pushed fixes to this branch (`e61b826`) — thanks for the card! It composes existing primitives cleanly and the reprint placement is correct (canonical in JUD 2002, `Printing` row in MH1; `check-card-printing` passes). Three blockers I ran into and fixed:

1. **Compile error** — `LavaDart.kt` imported `com.wingedsheep.sdk.dsl.EffectTarget`, which doesn't exist; the whole `:mtg-sets` module failed to compile. Corrected to `com.wingedsheep.sdk.scripting.targets.EffectTarget`.
2. **JUD snapshot not blessed** — `CardDefinitionSnapshotTest > JUD` failed because the new card wasn't in `JUD.json`. Re-blessed (only Lava Dart added, no churn).
3. **No test** — added `LavaDartScenarioTest` covering the plain damage cast and the no-mana flashback (Sacrifice a Mountain → deals 1, sacrifices the Mountain, exiles the card). Both pass, along with snapshot + lint.

Heads-up (not fixed, pre-existing engine gap, **not blocking this card**): the flashback is only reachable via the engine directly. `CastFromZoneEnumerator.enumerateFlashback` builds `additionalCostInfo` only for `AdditionalCost.Behold`, so a `SacrificePermanent` flashback cost is never surfaced to the client (no sacrifice prompt, and not gated on controlling a Mountain), and cast-time validation then rejects the empty payment. GroupProject's `TapPermanents` flashback has the same limitation. Worth a separate `add-feature` to extend `enumerateFlashback` to surface non-Behold bundled costs (mirroring `CastSpellEnumerator.buildAdditionalCostData`).

PR #1243, opened 2026-07-06
Fire ETB triggers dropped by the legend rule on the decision path

bolav:
## Problem

An enters-the-battlefield trigger on a permanent that a state-based action immediately removes was being **silently dropped** — but only on the decision-resume path.

Concretely: cast Ghalta, Stampede Tyrant. Its ETB ("put any number of creature cards from your hand onto the battlefield") is a decision-driven effect. If it puts a *second* Ghalta onto the battlefield, the legend rule (SBA, CR 704.5j) immediately moves one copy to the graveyard. Per CR 603.6a the duplicate's ability triggers the moment it enters — *before* the legend-rule SBA removes it — so its own ETB must still fire. It didn't.

## Cause

`SubmitDecisionHandler` (the decision-resume path) checked SBAs *before* detecting the resumed resolution's triggers, and on an SBA pause returned without ever detecting them. `PassPriorityHandler.resolveTopOfStack` (the non-decision path) already handled this correctly — detect pre-SBA, defer beneath the legend-rule continuation.

## Fix

Bring `SubmitDecisionHandler` in line with `PassPriorityHandler`:

- Detect the resumed resolution's triggers on the **pre-SBA** state (`result.events`), so ETB triggers on a permanent an SBA is about to remove are captured (CR 603.6a).
- On an SBA pause, queue those triggers as a `PendingTriggersContinuation` inserted **beneath** the SBA's continuation frames, so they fire after the legend-rule decision resolves instead of being dropped.
- Post-SBA detection now covers only the submitted decision + SBA events; the resumed resolution's own events are no longer re-detected there, so the `triggersAlreadyProcessed` modal-cast guard (Riku of Many Paths) still prevents double-queueing.

## Test

`GhaltaStampedeTyrantLegendRuleTest` — casts Ghalta A, uses its ETB to put Ghalta B onto the battlefield, keeps the older Ghalta under the legend rule, and asserts the *removed* Ghalta B's ETB still fires (Grizzly Bears reaches the battlefield). Also asserts exactly one Ghalta survives and the duplicate is in the graveyard.

Regression-checked: `LegendRuleTest`, `ModalCastTimeModeTest`, `ModalCastTimeDeferredTargetsTest` all green.

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/actions/decision/SubmitDecisionHandler.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/actions/decision/SubmitDecisionHandler.kt
@@ -102,12 +102,42 @@ class SubmitDecisionHandler(
 
             // Check SBAs after continuation completes
             if (result.isSuccess && !result.isPaused) {
+                // Detect triggers from the resumed resolution's events BEFORE SBAs run, so an
+                // enters-the-battlefield trigger on a permanent that an SBA immediately removes —
+                // e.g. the legend rule putting a just-entered duplicate legendary into the
+                // graveyard, or a Gishath/Ghalta chain that puts a second copy onto the
+                // battlefield — is still queued (CR 603.6a: the ability triggers the moment the
+                // permanent enters; the SBA that removes it happens afterward). Mirrors
+                // PassPriorityHandler.resolveTopOfStack, which handles the non-decision path.
+                // Skipped when the resumed chain already ran detection on its own emitted events
+                // (see the post-SBA note below).
+                val preSbaTriggers = if (result.triggersAlreadyProcessed) {
+                    emptyList()
+                } else {
+                    triggerDetector.detectTriggers(result.state, result.events)
+                }
+
+                val preSbaStackSize = result.state.continuationStack.size
                 val sbaResult = sbaChecker.checkAndApply(result.state)
 
-                // If SBA needs player input (e.g., legend rule), return paused
+                // If SBA needs player input (e.g., legend rule), return paused — but first queue
+                // preSbaTriggers beneath the SBA's continuation frames so they fire after the SBA
+                // decision resolves, instead of being silently dropped.
                 if (sbaResult.isPaused) {
+                    var pausedState = sbaResult.state
+                    if (preSbaTriggers.isNotEmpty()) {
+                        val pendingTriggers = PendingTriggersContinuation(
+                            decisionId = "submit-sba-deferred-triggers-${java.util.UUID.randomUUID()}",
+                            remainingTriggers = preSbaTriggers
+                        )
+                        val stack = pausedState.continuationStack
+                        val newStack = stack.subList(0, preSbaStackSize) +
+                            pendingTriggers +
+                            stack.subList(preSbaStackSize, stack.size)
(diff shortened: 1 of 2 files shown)


bolav:
## Review — SDK elegance + correctness

**Verdict:** Correct, well-scoped bug fix. The decision-resume path (`SubmitDecisionHandler`) was checking SBAs *before* detecting the resumed resolution's triggers, and dropping them entirely on an SBA pause — so an ETB trigger on a permanent the legend rule immediately kills (Ghalta putting a second Ghalta out) never fired. The fix now detects those triggers pre-SBA and defers them beneath the SBA's continuation frames, faithfully mirroring the already-correct `PassPriorityHandler.resolveTopOfStack`. No SDK surface added — pure engine control-flow, so the SDK-elegance lens is N/A. Behavior is right.

### What's good

- **Exact mirror of the proven pattern.** The pre-SBA detection, `preSbaStackSize` snapshot, and `subList(0, n) + pendingTriggers + subList(n, size)` insertion are line-for-line consistent with `PassPriorityHandler.kt:214-242`. Same stack-ordering assumption (SBA only pushes frames on top), same deferral semantics (CR 603.3 — triggers wait for the next priority).
- **The `triggersAlreadyProcessed` guard is preserved correctly.** `preSbaTriggers` is empty on the modal-cast re-entry path, and `postSbaTriggers` deliberately excludes `result.events`, so Riku-style battlefield triggers still don't double-queue. `ModalCastTimeModeTest` and `ModalCastTimeDeferredTargetsTest` confirmed green.
- **`result.events` are now detected exactly once** (pre-SBA), never re-detected post-SBA, so no double-fire.
- **Comment quality.** The rationale (CR 603.6a, the SBA-pause-drops-them explanation, the cross-reference to the non-decision path) is accurate and load-bearing. The drive-by fix of the stale `"line 108"` reference to `"success branch's trigger-detection above"` is the right call — line numbers rot.
- **Test is a real regression net.** It keeps Grizzly Bears in hand specifically to prove the *dead* Ghalta's ETB fires (not just that the survivor resolves), and asserts all three end states. Passes on this branch. CR citations (603.6a ETB, 704.5j legend rule) both verified against the local Comprehensive Rules.

### Issues

**Minor — parity gap with the mirrored handler (informational).** The comment says this "mirrors `PassPriorityHandler.resolveTopOfStack`," but two steps that handler does around detection are absent here: `trackNonTokenCreatureDeaths(...)` wrapping the state before `detectTriggers` (last-known-info for dies triggers, `PassPriorityHandler.kt:210,251`) and the `stateTriggerPoller.poll` / CR 603.8 state-trigger pass (`:260`). This is **pre-existing** — the old code detected `result.events` post-SBA without tracking either, so the fix introduces no new gap — but since the diff now advertises parity, a reader may assume full equivalence. If a decision-driven resolution ever sacrifices/bounces a nontoken creature mid-chain, its dies/leaves trigger would resolve without last-known power/counters here. Worth a one-line note in the comment or a follow-up, not a merge blocker.

### Recommendation

Merge as-is. Optionally add a half-sentence to the `SubmitDecisionHandler.kt:105-113` comment acknowledging that the `trackNonTokenCreatureDeaths` / state-trigger-poll steps of `resolveTopOfStack` are intentionally *not* carried over, so the "mirrors" claim isn't read as full parity.

PR #1246, opened 2026-07-07
Add Discover N (CR 701.57) and Lost Caverns of Ixalan cards

bolav:
## Discover N (CR 701.57) + LCI cards

Adds the **Discover** mechanic to the engine and the Lost Caverns of Ixalan cards that use it, plus a batch of independent LCI cards.

### Engine feature — `DiscoverEffect` (CR 701.57)

A new SDK primitive `Effects.Discover(amount, storeDiscoveredAs?, thenEffect?)`. It exiles from the top of the controller's library until a nonland card with mana value ≤ N (the *discovered card*), then presents a two-option prompt — **cast it for free** or **put it into hand** (falling back to hand if the cast can't initiate) — and bottom-randomizes the rest.

It is its own primitive rather than a parameter on Cascade because it differs on three axes: explicit threshold (fixed or a `DynamicAmount`, not the triggering spell's MV), `≤` vs strict `<`, and the non-cast branch keeps the card rather than bottoming it. It reuses everything it can — `CascadeExecutor.bottomRandomize`, `CardsRevealedEvent`, `MayPlayPermission` / `PlayWithoutPayingCostComponent`, the yes/no decision handler, and the effect-continuation machinery — and casts the discovered card *from exile* through the normal cast machinery so "whenever you cast a spell from exile" triggers (Quintorius Kand) fire exactly once.

`storeDiscoveredAs` + `thenEffect` support follow-ups keyed on the discovered card (Hit the Mother Lode's Treasures).

### Cards

Discover consumers: Geological Appraiser, Trumpeting Carnosaur, Caparocti Sunborn, Digsite Conservator, Primordial Gnawer, Daring Discovery, Walk with the Ancestors, Hit the Mother Lode, Hurl into History (dynamic threshold = countered spell's MV), Quintorius Kand.

Independent LCI cards: Hidden Cataract / Courtyard / Necropolis / Nursery / Volcano, Buried Treasure, Swashbuckler's Whip, Chimil the Inner Sun, Zoetic Glyph, Etali's Favor.

### Tests

`DiscoverScenarioTest`, `HurlIntoHistoryScenarioTest`, `QuintoriusKandScenarioTest` pin each clause of CR 701.57 (hit / cast / hand / above-threshold skip / whiff / `thenEffect` reading the discovered MV / cast-from-exile drain). The SDK reference, `CardLinter`, the LCI snapshot golden, and the mtgish emitter are updated in the same change.

### CR 701.57c deck-out fix (`9dce10cca`)

Follow-up to a self-review: when a Discover decks out the library without a castable nonland ≤ N, the *final* exiled card is still the discovered card if its own mana value is ≤ N (a bottom land has MV 0). Hit the Mother Lode decking itself out on lands now still makes its Treasures; a genuine whiff (final card a nonland with MV > N) makes none. Includes updated + added tests.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 95 / 286
+**Implemented:** 96 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -117,7 +117,7 @@
 - [x] Hermitic Nautilus
 - [x] Hidden Cataract
 - [x] Hidden Courtyard
-- [ ] Hidden Necropolis
+- [x] Hidden Necropolis
 - [ ] Hidden Nursery
 - [ ] Hidden Volcano
 - [ ] Hit the Mother Lode
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 101 / 286
+**Implemented:** 102 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -127,7 +127,7 @@
 - [ ] Huatli, Poet of Unity
 - [ ] Hulking Raptor
 - [ ] Hunter's Blowgun
-- [ ] Hurl into History
+- [x] Hurl into History
 - [ ] Idol of the Deep King
 - [ ] In the Presence of Ages
 - [ ] Inti, Seneschal of the Sun
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 109 / 286
+**Implemented:** 110 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -77,7 +77,7 @@
 - [ ] Defossilize
 - [ ] Diamond Pick-Axe
 - [ ] Didact Echo
-- [ ] Digsite Conservator
+- [x] Digsite Conservator
 - [x] Dinotomaton
 - [ ] Dire Flail
 - [x] Disruptor Wanderglyph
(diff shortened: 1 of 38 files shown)


bolav:
## Review: Discover N (CR 701.57) + LCI cards

Reviewed `lci-discover` against `lci-backlog` (the meaningful diff: one feature commit for the Discover engine primitive + 21 card commits, ~10 consuming Discover). Primary lens: SDK elegance.

### Verdict

Behavior is right and the SDK shape is **exactly right**. `DiscoverEffect` is a genuinely novel primitive (a new exile-until → two-option cast/hand → bottom-randomize interaction, distinct from Cascade on three real axes) that reuses every existing piece it can — `CascadeExecutor.bottomRandomize`, `CardsRevealedEvent`, `MayPlayPermission` / `PlayWithoutPayingCostComponent`, `DecisionHandler.createYesNoDecision`, the `EffectContinuation` machinery — and all 10 discover cards compose it through a one-line `Effects.Discover(N)`. No blocking issues.

### What's good

- **Primitive parameterization.** `amount: DynamicAmount` (fixed *and* dynamic overloads), optional `storeDiscoveredAs` + `thenEffect` — parameterized for the *next* card, not just Hit the Mother Lode. Hurl into History's "Discover X = countered spell's MV" drops out for free via `EntityProperty(Target(0), ManaValue)`.
- **Cast-from-exile trigger correctness.** The hardest part — casting the discovered card *from exile* must fire "whenever you cast a spell from exile" exactly once. `resumeDiscoverMayCast` synthesizes the cast through `CastSpellHandler` (like Cascade) and propagates `triggersAlreadyProcessed` on every cast branch so the `SpellCastEvent` isn't re-scanned. `QuintoriusKandScenarioTest` proves the drain fires once.
- **Fallback discipline.** The "cast can't initiate → falls back to hand, but Hit the Mother Lode still makes Treasures" path is handled, including popping the pre-pushed `thenEffect` continuation to avoid double-running it.
- **Tests pin the rule clauses**, and docs / linter / snapshot golden / mtgish emitter are all kept in sync in the same change.

### Issues

**Minor**

1. ~~**Deck-out whiff can still have a "discovered card" per CR 701.57c.**~~ **FIXED in `9dce10cca`.** When the library is exhausted without a castable nonland ≤ N, the final exiled card is still the discovered card if its MV ≤ N (a bottom land has MV 0), so a `thenEffect` keyed on it (Hit the Mother Lode) now runs. Whiff-encoding test corrected; genuine-whiff test added.

2. **The "resulting spell's mana value ≤ N" gate (CR 701.57a) isn't enforced.** The executor offers a free cast unconditionally once a card with MV ≤ N is found. For a normal card the resulting spell's MV equals the card's, so it's fine — but a discovered split/adventure/MDFC card whose castable face has MV > N should only be put-to-hand. Vanishingly rare (lands are skipped, so MDFC lands can't be discovered); worth a `// TODO(CR 701.57a)` so it isn't silently lost.

3. **`thenEffect` vs. composing existing primitives.** `thenEffect` + `storeDiscoveredAs` is close to `IfYouDoEffect(..., CollectionNonEmpty)` over a published collection. It's *not* wrong — the "runs after the cast even if the cast pauses" sequencing (pre-pushed as an `EffectContinuation`) is why it's embedded — but a KDoc sentence explaining why it can't just be `Discover(10) then <gated effect>` would stop the next author "simplifying" it. Author's discretion.

### Recommendation

Merge-ready. Issue 1 is fixed; issues 2 and 3 are optional polish (a TODO and a doc sentence). The primitive is the right size and all consuming cards compose it cleanly.

PR #1247, opened 2026-07-08
Add 3 Lost Caverns of Ixalan cards (batch 22)

bolav:
Adds the final 3 Lost Caverns of Ixalan cards from the triage — existing SDK primitives
only, no engine changes, each with a scenario test. This exhausts the triage's
IMPLEMENTABLE and VERIFY buckets.

> **Stacked PR** on `lci-cards-21` (#1276). GitHub shows the full stack vs `main`; please
> review **only the diff vs `lci-cards-21`** (the single batch-22 commit).

## Cards

| Card | Notes |
|---|---|
| Sunken Citadel | Cave land: `EntersTapped` + `EntersWithChoice(COLOR)` replacements; both mana abilities read the stored color via `AddManaOfChosenColor`, the second carrying `ManaRestriction.CardTypeSpellsOrAbilitiesOnly(LAND, abilities only)` — "Spend this mana only to activate abilities of land sources" |
| The Ancient One | Descend 8 as two `ConditionalStaticAbility` gates (`CantAttack`/`CantBlock` while `Not(CardsInGraveyardMatchingAtLeast(8, Permanent))`), mirroring Basking Capybara's gate shape; the {2}{U}{B} ability is draw → mandatory-discard `ReflexiveTriggerEffect` whose mill reads `DynamicAmount.StoredCardManaValue("discarded")` (correct after the card has left the hand) |
| Zoyowa's Justice | The owner of the target — not the caster — shuffles and discovers: `Effects.ForEachPlayer(Player.OwnerOf(target))` rebinds the resolution controller around `ShuffleIntoLibrary` + `Discover(EntityProperty(Target(0), ManaValue))`; no engine change needed since Discover already takes a dynamic threshold |

## Tests

- Scenario tests for all three cards: Sunken Citadel's ETB color pause + both mana
  abilities (incl. the restriction payload on the two-mana ability); The Ancient One's
  attack and block gates on both sides of the 8-permanent-card threshold, plus the
  reflexive mill (MV-3 discard → opponent mills exactly 3); Zoyowa's Justice discover
  performed by the owner in both the opponent-owned and self-owned cases, with X gating
  which cards are discoverable.
- Full run green: the three scenario tests + `mtg-sets` snapshot and lint nets
  (LCI golden extended with exactly the three new cards).
- `just check-card-printing` clean for all three (LCI is the earliest real printing;
  Sunken Citadel's EOS reprint needs no row — EOS isn't scaffolded).
- Manual self-play scenario included: `manual-scenarios/sets/lci/cards-22.json`.


bolav:
## Review: batch 22 (diff vs `lci-cards-21`)

### Verdict

Approve. All three cards match their Scryfall oracle text exactly, compose existing SDK
primitives with zero engine changes, and every ability is covered by a passing scenario
test. No blocking or important issues — only minor untested corners.

### What's good

- **Zoyowa's Justice is the SDK-elegance poster child of the batch**: the tricky part of
  the card — the *owner* of the target, not the caster, shuffles and discovers — falls out
  of composing `Effects.ForEachPlayer(Player.OwnerOf(...))` around `ShuffleIntoLibrary` +
  `Discover`, with X read as `EntityProperty(Target(0), ManaValue)`. No new types, and
  both tests assert the load-bearing thing: *which player* gets the discover decision.
- **Descend 8 reuses the established gate pattern** (mirrors Basking Capybara): two
  `ConditionalStaticAbility`-wrapped `CantAttack`/`CantBlock` statics gated on
  `Not(CardsInGraveyardMatchingAtLeast(8, Permanent))`. `Player.You` in that condition
  resolves to the source's controller, so "your graveyard" is correct.
- **Tests exercise both sides of every threshold**: attack at 7 vs 8 permanent cards,
  block at 7 vs 8, and the reflexive mill with a known MV-3 discard. The `replaceState`
  jump to `DECLARE_ATTACKERS` in the can't-attack test has a genuinely useful comment
  explaining why `passPriorityUntil` would mask the restriction (the step gets skipped,
  and cleanup discards would cross the threshold).
- **Verified externally**: `just check-card-printing` passes for all three (all
  LCI-canonical; Sunken Citadel's EOS reprint correctly needs no row since EOS isn't
  scaffolded), and mana costs / P/T / rarity / collector numbers match Scryfall exactly.
  No CR numbers are cited anywhere in the diff (Descend 8 is an ability word — correctly
  given no rules weight).
- Full test run green: the three scenario tests, LCI snapshot golden, and the card lint
  net (pipeline-variable dataflow: `discardCards` stores as `"discarded"`, matching
  `StoredCardManaValue("discarded")`).

### Issues

**Blocking** — none. **Important** — none.

**Minor** (author's discretion):

1. **Untested corner: discarding a land with The Ancient One (mill 0).**
   `TheAncientOne.kt` explicitly claims "Milling zero (e.g. a discarded land) still
   requires and uses a target player, matching the oracle wording," but no test exercises
   a land discard. A second activation case discarding a Swamp and asserting the
   opponent's graveyard stays empty (and the target decision still appears) would pin it.

2. **Untested corner: Zoyowa's Justice targeting a token.** The card comment argues MV
   "survives the shuffle because the entity keeps its id and CardComponent" — true for
   cards, but a targeted *token copy* (MV ≥ 1, so targetable) ceases to exist after the
   shuffle, and the ruling is the owner still discovers for its last-known mana value.
   Worth one test if `EntityProperty` can read the dead entity; if it can't, that's a
   narrow behavior gap to note in the card comment.

3. **Sunken Citadel's restriction is asserted but never enforced in-test.** The test
   checks the pool holds two restricted mana with the right payload, but never attempts
   to spend it (legally on a land ability, illegally on a spell). Fine if
   `ManaRestriction.CardTypeSpellsOrAbilitiesOnly` enforcement is pinned elsewhere; a
   spend-rejection assertion would make the test self-contained.

### Recommendation

Merge as-is; the minors are follow-up material.

PR #1249, opened 2026-07-08
Add 5 Lost Caverns of Ixalan cards (batch 2)

bolav:
Depends on : https://github.com/wingedsheep/argentum-engine/pull/1248

## Add 5 Lost Caverns of Ixalan cards (batch 2)

Stacked on `lci-cards-01`. This PR is scoped to the difference from that branch only.

Adds five LCI cards, each modeled by composing **existing** SDK primitives — the diff touches no `mtg-sdk` or `rules-engine` engine source, only card definitions, the LCI snapshot, the backlog index, scenario tests, and a manual dev scenario.

| Card | Cost | Notes |
|------|------|-------|
| Another Chance | {2}{B} | Optional mill 2, then return up to two creature cards from graveyard (MayEffect + Gather→SelectUpTo→Move pipeline) |
| Armored Kincaller | {2}{G} | ETB: reveal a Dinosaur from hand *or* control another Dinosaur → gain 3 life (`Conditions.Any` over `CollectionContainsMatch` + `YouControl(excludeSelf)`) |
| Basking Capybara | {1}{G} | Descend 4 — +3/+0 while ≥4 permanent cards in your graveyard (`ConditionalStaticAbility` + `CardsInGraveyardMatchingAtLeast`) |
| Bedrock Tortoise | {3}{G} | Hexproof to your creatures during your turn; creatures with toughness > power assign combat damage equal to toughness (`GrantKeyword` + `AssignDamageEqualToToughness`) |
| Belligerent Yearling | {1}{R} | Trample; when another Dinosaur you control enters, may set base power to that creature's power until EOT (`SetBasePower` reading `Triggering.Power`) |

Backlog updated 116 → 121 implemented.

### Testing
- `AnotherChanceScenarioTest`, `ArmoredKincallerScenarioTest`, `BaskingCapybaraScenarioTest`, `BedrockTortoiseScenarioTest`, `BelligerentYearlingScenarioTest` — 14 scenario tests, all green.
- `CardDefinitionSnapshotTest` (LCI) — additive only, green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1251, opened 2026-07-08
Add 5 Lost Caverns of Ixalan cards (batch 3)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1249

## Lost Caverns of Ixalan — cards batch 3 (5 cards)

Stacked on `lci-cards-02`. This PR contains only the batch-3 diff.

Adds five LCI cards, each with a scenario test. Composes existing SDK primitives — the
only SDK change is a one-line facade parameter surfacing a field that already existed on
`CreateTokenEffect` (no new effect types, no engine changes).

| Card | Cost | Type | Mechanic |
|------|------|------|----------|
| Bonehoard Dracosaur | `{3}{R}{R}` | Dinosaur Dragon 5/5 | Upkeep impulse-2 + conditional Dinosaur/Treasure tokens |
| Brackish Blunder | `{1}{U}` | Instant | Bounce; Map token if the creature was tapped |
| Brazen Blademaster | `{2}{R}` | Orc Pirate 2/3 | +2/+1 on attack while controlling ≥2 artifacts |
| Broodrage Mycoid | `{3}{B}` | Fungus 4/3 | End-step Fungus token (can't block) if descended |
| Burning Sun Cavalry | `{1}{R}` | Human Knight 2/2 | +1/+1 on attack or block while controlling a Dinosaur |

### Implementation notes

- **Bonehoard Dracosaur** — upkeep trigger is `Patterns.Exile.impulse(2)` (exile top two,
  grant may-play until EOT) followed by two independent `ConditionalEffect` gates reading
  the same collection: one `CollectionContainsMatch(Land)` → 3/1 Dinosaur token, one
  `CollectionContainsMatch(Nonland)` → Treasure. Both can fire in the same upkeep, matching
  the oracle's two independent "if … this way" clauses.
- **Brackish Blunder** — `ConditionalEffect(TargetIsTapped)` evaluates the tapped state
  *before* the bounce, so it captures the creature's pre-bounce state (the condition reads
  live battlefield state and would be false once the permanent leaves).
- **Brazen Blademaster / Burning Sun Cavalry** — "while you control …" is a trigger-time
  condition, **not** an intervening-"if" (CR 603.4 applies only to an "if" immediately
  after the trigger event). Modeled via `triggerCondition`, which the engine evaluates at
  trigger detection and does not re-check on resolution — matching "while" semantics. Burning
  Sun includes a regression test proving the bonus still applies when the Dinosaur leaves
  after the trigger fires.
- **Broodrage Mycoid** — "if you descended this turn" is a proper intervening-"if"
  (CR 700.11 for the "descended" definition). The token's "This token can't block" is a
  static ability on the token itself, modeled as `CantBlock(GroupFilter.source())` — the
  same component the decayed keyword uses. It uses the new `Effects.CreateToken(staticAbilities = …)`
  facade parameter.

### SDK change

`Effects.CreateToken` gains an optional `staticAbilities` parameter that forwards to the
already-existing `CreateTokenEffect.staticAbilities` field (covers the static-only inline-token
case; triggered/activated abilities still use the raw constructor). SDK reference updated to match.

### Tests

15 scenario test cases across five classes, all green:
Bonehoard (4 — land+nonland, two lands, two nonlands, one-card library + may-play permission),
Brackish (2 — tapped/untapped), Brazen (2 — two artifacts / one artifact),
Broodrage (3 — no-descend, descend + can't-block static, can't-block action rejected),
Burning Sun (4 — attack, no-Dinosaur, block, Dinosaur-leaves-after-trigger).

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 95 / 286
+**Implemented:** 96 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -117,7 +117,7 @@
 - [x] Hermitic Nautilus
 - [x] Hidden Cataract
 - [x] Hidden Courtyard
-- [ ] Hidden Necropolis
+- [x] Hidden Necropolis
 - [ ] Hidden Nursery
 - [ ] Hidden Volcano
 - [ ] Hit the Mother Lode
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 116 / 286
+**Implemented:** 121 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -17,14 +17,14 @@
 - [x] Ancestors' Aid
 - [x] Ancestral Reminiscence
 - [x] Anim Pakal, Thousandth Moon
-- [ ] Another Chance
-- [ ] Armored Kincaller
+- [x] Another Chance
+- [x] Armored Kincaller
 - [x] Attentive Sunscribe
 - [x] Bartolomé del Presidio
-- [ ] Basking Capybara
+- [x] Basking Capybara
 - [ ] Bat Colony
-- [ ] Bedrock Tortoise
-- [ ] Belligerent Yearling
+- [x] Bedrock Tortoise
+- [x] Belligerent Yearling
 - [x] Bitter Triumph
 - [x] Bloodletter of Aclazotz
 - [ ] Bloodthorn Flail
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 101 / 286
+**Implemented:** 102 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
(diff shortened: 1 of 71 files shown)


bolav:
## Review — `lci-cards-03` vs base `lci-cards-02`

**Verdict:** Behavior is correct and the SDK shape is right. Clean 5-card batch that composes
existing primitives — the only SDK change is a one-line facade parameter surfacing a field
that already existed on `CreateTokenEffect`. No new engine types, no projection bugs, both
cited CR numbers verified. One minor test-symmetry gap.

**Tests:** `BUILD SUCCESSFUL` — 15 scenario cases across the five classes, 0 failures
(Bonehoard 4, Brackish 2, Brazen 2, Broodrage 3, Burning Sun 4).

### What's good

- **The SDK change is exactly the right size.** `Effects.CreateToken(… staticAbilities = …)`
  just forwards to the pre-existing `CreateTokenEffect.staticAbilities` field. No new type, no
  executor change; the doc correctly explains the boundary (facade covers static-only; drop to
  the raw constructor for triggered/activated). Textbook composition-over-monolith.
- **Every card composes existing primitives.** Bonehoard = `Patterns.Exile.impulse(2)` + two
  independent `ConditionalEffect` gates on `Conditions.CollectionContainsMatch`; both can fire
  (land *and* nonland), matching the oracle's two independent "if … this way" clauses. Broodrage's
  can't-block token = `CantBlock(GroupFilter.source())` — the same component the decayed keyword
  uses. Brackish = `ConditionalEffect(TargetIsTapped)`. No bespoke effects anywhere.
- **Trigger-timing reasoning is correct and the rules are verified.** CR **603.4** really is the
  intervening-"if" rule and applies *only* to an "if" immediately after the trigger event — so
  treating "while you control …" as a `triggerCondition` (checked at detection, not re-checked at
  resolution) is right. CR **700.11** for "descended" is also correct.
- **Burning Sun's test 4 is the standout** — it explicitly proves the Dinosaur leaving *after* the
  trigger fires still applies +1/+1 (no resolution re-check), directly exercising the CR 603.4 claim.
- Broodrage tests go all the way to `submitExpectFailure(DeclareBlockers(...))` — proving the
  can't-block static gates a legal action, not just that a flag is set.

### Issues — Minor

- **Brazen Blademaster makes the same CR 603.4 "not re-checked on resolution" claim as Burning Sun
  but has no paired test for it.** Burning Sun's test 4 proves the Dinosaur-leaves case; the
  Blademaster comment asserts identical semantics for artifacts but only tests two-artifacts-vs-one.
  The engine path is the same and already proven by the sibling, so low-risk — but for the "every
  cited rule has a test" discipline, consider adding: two artifacts → attack → destroy one before
  the trigger resolves → Blademaster is still 4/4.
- **Broodrage's "if you descended" is a genuine intervening-"if" (CR 603.4 wants a double-check),
  implemented via `triggerCondition` (detection-only).** Behaviorally correct — "descended this
  turn" is a monotonic latch (CR 700.11) that can't become false before the end-step trigger
  resolves, and you can't descend during the end step in time — so the second check is moot. No
  change needed; flagged only so a future reader doesn't mistake the pattern for the general case
  where the resolution re-check *does* matter.

### Recommendation

Merge-ready. Optionally add the Brazen Blademaster "artifact leaves after trigger fires" test for
parity with Burning Sun's test 4.

PR #1252, opened 2026-07-09
LCI batch 4 — 5 cards

bolav:
## Lost Caverns of Ixalan — cards batch 4 (5 cards)

Stacked on `lci-cards-03`. This PR contains **only** the batch-4 diff against that base.

Adds five LCI cards, all composing existing SDK primitives — **no engine/SDK changes**. Each has a scenario test.

| Card | Cost | Type | Mechanic |
|------|------|------|----------|
| Calamitous Cave-In | `{3}{R}` | Sorcery | X damage to each creature/planeswalker; X = Caves you control + Cave cards in your graveyard |
| Canonized in Blood | `{1}{B}` | Enchantment | End-step descend trigger → +1/+1 on target creature; `{5}{B}{B}`, sac → 4/3 flying Vampire Demon token |
| Captain Storm, Cosmium Raider | `{U}{R}` | Legendary Creature 2/2 | Artifact-you-control ETB → +1/+1 counter on target Pirate |
| Child of the Volcano | `{3}{R}` | Elemental 3/3 | Trample; end-step descend trigger → +1/+1 on itself |
| Coati Scavenger | `{2}{G}` | Raccoon 3/2 | Descend 4 ETB → return target permanent card from graveyard to hand |

### Implementation notes

- **Calamitous Cave-In** — X is modeled as the AST `DynamicAmount.Add(Count(battlefield Caves), Count(graveyard Cave cards))` fed to `Patterns.Group.dealDamageToAll` over `CreatureOrPlaneswalker`.
- **Canonized in Blood / Child of the Volcano** — "if you descended this turn" is a proper intervening-"if" (`Conditions.YouDescendedThisTurn()`); CR 700.11 defines "descended". The sac ability uses `Costs.Composite(Mana, SacrificeSelf)`.
- **Captain Storm** — `Triggers.entersBattlefield(GameObjectFilter.Artifact.youControl(), binding = ANY)`; Treasure/Clue tokens count. Storm is itself a Pirate and a legal target.
- **Coati Scavenger** — "Descend 4" is an ability word (CR 207.2c) with no rules meaning; modeled as `Conditions.CardsInGraveyardMatchingAtLeast(4, Permanent)` intervening-if + mandatory `Effects.ReturnToHand` of a target permanent card from the graveyard.

### Tests

12 scenario test cases across five classes, all green:
Calamitous (2 — X=2 kills 2/2 not 3/3, graveyard Cave raises X to kill the 3/3),
[...]

PR #1253, opened 2026-07-09
LCI cards batch 5 (5 cards) + tokens honor global enters-tapped

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1252

## Lost Caverns of Ixalan — cards batch 5 (5 cards)

Stacked on `lci-cards-04`. This PR contains **only** the batch-5 diff against that base.

Adds five LCI cards plus one engine change (global enters-tapped replacements now apply to created tokens). Each card has a scenario test.

| Card | Cost | Type | Mechanic |
|------|------|------|----------|
| Corpses of the Lost | `{2}{B}` | Enchantment | Skeleton anthem (+1/+0 & haste); ETB 2/2 Skeleton Pirate token; end-step descend → may pay 1 life to bounce itself |
| Cosmium Confluence | `{4}{G}` | Sorcery | Choose three (repeatable): tutor a Cave (tapped) / animate a Cave permanently (+3 counters, 0/0 Elemental, still a land) / destroy target enchantment |
| Council of Echoes | `{4}{U}{U}` | Creature 4/4 | Flying; Descend 4 ETB → return up to one target nonland permanent other than itself to hand |
| Dauntless Dismantler | `{1}{W}` | Creature 1/4 | Artifacts your opponents control enter tapped; `{X}{X}{W}`, sac → destroy each artifact with mana value X |
| Deepfathom Echo | `{2}{G}{U}` | Creature 4/4 | Begin-combat: explores, then may become a copy of another creature you control until end of turn |

### Implementation notes

- **Corpses of the Lost** — split layer-6 anthem (`ModifyStats` +1/+0 and `GrantKeyword` haste over Skeletons you control); end-step intervening-if `Conditions.YouDescendedThisTurn()` (CR 700.11) with `OptionalCostEffect(PayLifeEffect(1), ifPaid = ReturnToHand(Self))`.
- **Cosmium Confluence** — `modal(chooseCount = 3, allowRepeat = true)`; mode 1 = `AddCounters` then `BecomeCreature(Duration.Permanent)` (permanent animation, still a land).
- **Council of Echoes** — Descend 4 is an ability word; modeled as intervening-if `CardsInGraveyardMatchingAtLeast(4, Permanent)` + optional `TargetOther(NonlandPermanent)` (up-to-one).
- **Dauntless Dismantler** — `PermanentsEnterTapped` runtime replacement scoped to `Artifact.opponentControls()`; `{X}{X}{W}, SacrificeSelf` destroys each artifact with `manaValueEqualsX()`.
- **Deepfathom Echo** — target for the copy is selected mid-resolution inside the `MayEffect`, so explore always runs and target selection happens only if the player accepts.

### Engine change — global enters-tapped replacements now apply to created tokens

`PermanentsEnterTapped` (Dauntless Dismantler, Authority of the Consuls, …) previously only affected cast/played permanents, not tokens. Added `EnterTappedReplacements.applyCreatedTokenEntryTap(...)`, a shared helper mirroring `ZoneTransitionService`'s CR-614 resolution order (consult `EnterUntappedReplacements` first, so an applicable "enters untapped" wins), called from every token-minting executor after `BattlefieldEntry.place(...)`. Entering tapped on entry is not a tap transition, so no `TappedEvent` is emitted (`TapEventEnforcementTest` allowlist updated). Fixes the reported Waterwind-Scout-under-Dismantler bug (an opponent's Map token now enters tapped).

### Tests

New scenario test classes for all five cards, all green (verified locally):
Corpses (4 — ETB token is 3/2 under its own anthem, static grant, descend bounce, no-descend no-op),
Cosmium (2 — `allowRepeat` mode 2×2 + mode 1, and mode 1×2 + mode 2),
Council (3 — Descend 4 met bounces, sub-threshold no-fire, up-to-one decline),
Dauntless (6 — opponent artifact + Map token + Construct token enter tapped, controller-scoped negatives, `{X}{X}{W}` X=1 destroy),
Deepfathom (2 — explore + accept copy → 2/2, explore + decline → stays 4/4).

`card-sdk-language-reference.md`, the LCI card snapshot, and the backlog implemented count (131 → 136) are updated.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 126 / 286
+**Implemented:** 131 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -38,10 +38,10 @@
 - [x] Broodrage Mycoid
 - [x] Buried Treasure
 - [x] Burning Sun Cavalry
-- [ ] Calamitous Cave-In
-- [ ] Canonized in Blood
+- [x] Calamitous Cave-In
+- [x] Canonized in Blood
 - [x] Caparocti Sunborn
-- [ ] Captain Storm, Cosmium Raider
+- [x] Captain Storm, Cosmium Raider
 - [x] Captivating Cave
 - [x] Careening Mine Cart
 - [x] Cartographer's Companion
@@ -50,11 +50,11 @@
 - [ ] Cavernous Maw
 - [x] Cenote Scout
 - [x] Chart a Course
-- [ ] Child of the Volcano
+- [x] Child of the Volcano
 - [x] Chimil, the Inner Sun
 - [x] Chupacabra Echo
 - [ ] Clay-Fired Bricks
-- [ ] Coati Scavenger
+- [x] Coati Scavenger
 - [x] Cogwork Wrestler
 - [x] Colossadactyl
 - [x] Compass Gnome
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 131 / 286
+**Implemented:** 136 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -60,19 +60,19 @@
 - [x] Compass Gnome
 - [x] Confounding Riddle
 - [ ] Contested Game Ball
-- [ ] Corpses of the Lost
+- [x] Corpses of the Lost
 - [x] Cosmium Blast
-- [ ] Cosmium Confluence
-- [ ] Council of Echoes
+- [x] Cosmium Confluence
+- [x] Council of Echoes
 - [ ] Curator of Sun's Creation
 - [x] Daring Discovery
-- [ ] Dauntless Dismantler
+- [x] Dauntless Dismantler
 - [x] Dead Weight
 - [x] Deathcap Marionette
 - [ ] Deconstruction Hammer
 - [x] Deep Goblin Skulltaker
 - [x] Deep-Cavern Bat
-- [ ] Deepfathom Echo
+- [x] Deepfathom Echo
 - [ ] Deeproot Pilgrimage
 - [ ] Defossilize
 - [ ] Diamond Pick-Axe
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -6122,7 +6122,8 @@ replacementEffect {
   which is a self-replacement consumed once as the source enters, this is a *runtime* replacement
   stamped into the source's `ReplacementEffectSourceComponent` (`StaticAbilityHandler.isRuntimeReplacementEffect`)
   and consulted from the battlefield against OTHER permanents as they enter — so `appliesTo.filter`
-  describes the *affected* permanents. The entry-tap paths (`PlayLandHandler`, `ZoneTransitionService`)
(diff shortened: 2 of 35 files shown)


bolav:
## Review: batch 5 (diff vs `lci-cards-04`)

### Verdict

Behavior is correct and the SDK stays clean — all five cards compose **existing** primitives, and the one engine change extends the existing `PermanentsEnterTapped` runtime replacement to created tokens via a single shared helper rather than inventing a new type. All six touched/new test classes pass locally. One real gap: five of the seven modified token-minting sites are untested, and one minting site was missed entirely.

### What's good

- **No new SDK types.** The five cards reach for `ModifyStats` + `GrantKeyword` (Corpses' split anthem), `modal(allowRepeat=true)` + `Patterns.Library.searchLibrary` + `BecomeCreature(Duration.Permanent)` (Cosmium), `CardsInGraveyardMatchingAtLeast` + `TargetOther`/optional target (Council's Descend 4), `PermanentsEnterTapped` + `manaValueEqualsX` (Dismantler), and `Explore` + `MayEffect` + `EachPermanentBecomesCopyOfTarget` (Deepfathom). Compose-don't-grow, as intended.
- **The token-tapping fix is the right primitive, not a new one.** Every token routes through `EnterTappedReplacements.applyCreatedTokenEntryTap`, mirroring `ZoneTransitionService`'s CR-614 order (consult `EnterUntappedReplacements` first, so "enters untapped" wins). Filter matching correctly goes through `predicateEvaluator.matches(state, state.projectedState, …)` — projection-aware, consistent with the pre-existing `entersTapped` path.
- **Strong rule-corner tests.** Corpses test 1 asserts the token enters as a 3/2 (base 2/2 + the enchantment's own anthem already live) — a real ETB-vs-static-timing check. Dismantler has cast-artifact and token positives *plus* controller-scoped negatives on both paths; the double-X cost is exercised at X=1 (MV-1 destroyed, MV-6 survives). Council tests both intervening-if branches and the "up to one → decline" no-op.
- **Housekeeping complete:** `TapEventEnforcementTest` allowlist updated with a correct rationale, `card-sdk-language-reference.md` synced, LCI snapshot re-blessed, backlog count 131→136.
- **CR numbers verified** against the local comp rules: 700.11 (descend), 602.2b (activation cost process), 614 (replacement ordering). Printing placement clean for all five (`check-card-printing` ok — all LCI-original).

### Important

- **Five of seven token-minting sites are modified but untested.** `DauntlessDismantlerScenarioTest` exercises only `CreateTokenExecutor` and `CreatePredefinedTokenExecutor`. The change also adds `applyCreatedTokenEntryTap` to `CreateTokenCopyOfChosenPermanentExecutor`, `CreateTokenCopyOfEquippedCreatureExecutor`, `CreateTokenCopyOfSourceExecutor`, `CreateTokenCopyOfTargetExecutor`, `TokenFromDefinition`, and both sites in `TokenCreationReplacementHelper` — none tested. The doc even claims "Authority of the Consuls taps an opponent's token **copy** of a creature" with nothing verifying it. Correctness at each site is ordering-dependent (the helper requires the token to already carry its `ControllerComponent`/`CardComponent`); I read all seven and the ordering looks correct, but that isn't the bar for a cross-cutting entry-tap change. Add at least one token-copy test (e.g. an Authority-of-the-Consuls-style opponent minting a creature-copy token) so the copy executors and the "untapped wins" branch are actually exercised.

### Minor

- **`CreateRoleTokenExecutor.kt:105` was missed** — it calls `BattlefieldEntry.place(...)` with no following `applyCreatedTokenEntryTap`, so the "every token executor" sweep isn't exhaustive. Harmless in practice (Role tokens are Auras you control; tapping one is meaningless), but inconsistent with the doc. Either add the call or note the deliberate exclusion.
- **Cosmium mode 0 (Cave tutor) and "It's still a land" are unasserted** — `drainStack` avoids the library-search mode, and the animation tests check `isCreature` + counters but not that the Cave keeps its land type. A `projectedState.isLand(cave) shouldBe true` assertion would lock the "still a land" clause.
- **Unused binding** — `val enchantment = driver.putPermanentOnBattlefield(...)` in `CorpsesOfTheLostScenarioTest` test 3 is never read.

### Recommendation

Not blocking. Before considering it complete, add one token-**copy** scenario test to cover the untested `CreateTokenCopyOf*` paths and the CR-614 "enters untapped wins" branch — the one place the change's breadth outruns its test coverage. Optionally wire `CreateRoleTokenExecutor` for an exhaustive sweep. Everything else (definitions, CR citations, printings, projection use, docs/snapshot sync) is solid.

PR #1254, opened 2026-07-09
Add 5 Lost Caverns of Ixalan cards (batch 6)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1253

## LCI cards — batch 6 (5 cards)

Sixth batch of Lost Caverns of Ixalan cards, each with a scenario test. Stacked on `lci-cards-05` — review only the diff to that branch; the diff vs `main` includes batches 1–5 until those merge.

- **Defossilize** — return target creature card from your graveyard to the battlefield, then it explores twice (`PutOntoBattlefield` → `Explore` → `Explore`).
- **Diamond Pick-Axe** — indestructible Equipment; equipped creature gets +1/+1 and makes a Treasure whenever it attacks; Equip {2} (`ModifyStats` + `GrantTriggeredAbility`).
- **Didact Echo** — ETB draw a card; has flying as long as there are 4+ permanent cards in your graveyard (Descend 4, `ConditionalStaticAbility` + `CardsInGraveyardMatchingAtLeast`).
- **Disturbed Slumber** — turn target land you control into a 4/4 reach + haste Dinosaur until end of turn (still a land); it must be blocked this turn if able (`BecomeCreature` + `MustBeBlockedEffect`).
- **Eaten by Piranhas** — flash Aura: enchanted creature is a 1/1 black Skeleton and loses all abilities (`TransformPermanent` + `LoseAllAbilities` + `SetBasePowerToughnessStatic`, the Sugar Coat pattern). Includes the FDN reprint `Printing` row.

No new SDK types — all five cards compose existing primitives. The batch surfaced three bugs, fixed here:

- **`MustBeBlockedExecutor` used the base type line** to validate its target, rejecting permanents that are only creatures via a continuous effect (Disturbed Slumber's animated land). Now reads `projectedState.isCreature`.
- **"Must be blocked if able" over-enforced Rule 509.1c:** with two such attackers and one blocker, every declaration was rejected. The validator now computes the maximum satisfiable number of requirements via bipartite matching (provoke-pinned blockers only count for their pinned attacker; blockers claimed by a Lure-style requirement are excluded) and rejects only declarations that obey fewer. Covered by `MustBeBlockedMatchingScenarioTest` (asymmetric blockers) and two new Disturbed Slumber tests.
- **The client reveal overlay dropped all but the first `cardsRevealed` event** in a state update; a Defossilize resolution reveals twice (two explores). Events are now merged into one overlay payload (presentation-only, no game logic in the client).

[...]

PR #1255, opened 2026-07-09
Add 5 Lost Caverns of Ixalan cards (batch 7)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1254

## LCI cards — batch 7 (5 cards) + sacrifice-as-cost trigger fix

Seventh batch of Lost Caverns of Ixalan cards, each with a scenario test. Stacked on `lci-cards-06` — review only the diff to that branch; the diff vs `main` includes batches 1–6 until those merge.

- **Echo of Dusk** — 2/2 Vampire Spirit; Descend 4 — gets +1/+1 and has lifelink while there are 4+ permanent cards in your graveyard (`ConditionalStaticAbility` + `ModifyStats` / `GrantKeyword` on `CardsInGraveyardMatchingAtLeast`).
- **Explorer's Cache** — artifact that enters with two +1/+1 counters; whenever a creature you control with a +1/+1 counter on it dies, put a +1/+1 counter on it; {T}, sorcery-speed: move a +1/+1 counter onto target creature (`EntersWithCounters` + LKI-filtered dies trigger + `MoveCounters`).
- **Fanatical Offering** — {1}{B} instant; sacrifice an artifact or creature as an additional cost, draw two and make a Map token (`additionalCost(SacrificePermanent)` + `Composite(DrawCards, CreateMapToken)`).
- **Frilled Cave-Wurm** — 2/5 Salamander Wurm; Descend 4 — gets +2/+0 while there are 4+ permanent cards in your graveyard.
- **Fungal Fortitude** — flash Aura; enchanted creature gets +2/+0; when it dies, return it to the battlefield tapped under its owner's control (`ModifyStats` + `ATTACHED` leaves-to-graveyard trigger + `PutOntoBattlefield(tapped)`).

No new SDK types — all five cards compose existing primitives.

### Engine fix bundled in (heads-up for reviewers)

Playtesting Explorer's Cache surfaced a real, cross-cutting engine bug that this PR also fixes:

- **Sacrifice-as-an-additional-cost emitted a `ZoneChangeEvent` with `lastKnown = null`.** The four cost-sacrifice sites in `CastSpellHandler` (`CostAtom.Sacrifice`, `SacrificeCreaturesForCostReduction`, `SacrificeOrPay`, `Casualty`) hand-built the event and bypassed `ZoneTransitionService.moveToZone`, so every dies/leaves trigger that reads the dying permanent's last-known information — counters, power/toughness, keywords, token-ness (CR 603.10 / 608.2h) — silently missed. Destruction and sacrifice-*effects* (`SacrificeExecutor`) already worked because they route through `moveToZone`. Explorer's Cache did not gain a counter when the counter-bearing creature was sacrificed to Fanatical Offering.
- **Fix:** a shared `CastSpellHandler.sacrificePermanentAsCost` helper (`trackPermanentSacrifice` → `PermanentsSacrificedEvent` → `moveToZone`), with all four sites routed through it, mirroring `SacrificeExecutor`. This also fixes two latent bugs on that path: exit cleanup + graveyard-replacement redirects now run, and the event is correctly tagged `wasSacrificed = true` (CR 701.21), so "if it wasn't sacrificed" triggers no longer mis-fire on cost sacrifices. The activated-ability sacrifice paths (`CostHandler`, `CostPaymentService`) were audited and already used `moveToZone`.
- **Test:** `SacrificeAsCostDeathTriggerTest` (fails before, passes after) plus the full `rules-engine` suite green.

Also includes `manual-scenarios/sets/lci/cards-7.json`, a manual playtest scenario (all five cards castable) — this is what surfaced the bug.

All scenario tests pass; LCI snapshot re-blessed; linter and `check-card-printing` green for all five cards.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 136 / 286
+**Implemented:** 141 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -74,19 +74,19 @@
 - [x] Deep-Cavern Bat
 - [x] Deepfathom Echo
 - [ ] Deeproot Pilgrimage
-- [ ] Defossilize
-- [ ] Diamond Pick-Axe
-- [ ] Didact Echo
+- [x] Defossilize
+- [x] Diamond Pick-Axe
+- [x] Didact Echo
 - [x] Digsite Conservator
 - [x] Dinotomaton
 - [ ] Dire Flail
 - [x] Disruptor Wanderglyph
-- [ ] Disturbed Slumber
+- [x] Disturbed Slumber
 - [ ] Dowsing Device
 - [ ] Dreadmaw's Ire
 - [x] Dusk Rose Reliquary
 - [x] Earthshaker Dreadmaw
-- [ ] Eaten by Piranhas
+- [x] Eaten by Piranhas
 - [ ] Echo of Dusk
 - [ ] Echoing Deeps
 - [x] Enterprising Scallywag
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 141 / 286
+**Implemented:** 146 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -87,18 +87,18 @@
 - [x] Dusk Rose Reliquary
 - [x] Earthshaker Dreadmaw
 - [x] Eaten by Piranhas
-- [ ] Echo of Dusk
+- [x] Echo of Dusk
 - [ ] Echoing Deeps
 - [x] Enterprising Scallywag
 - [x] Envoy of Okinec Ahau
 - [x] Etali's Favor
-- [ ] Explorer's Cache
+- [x] Explorer's Cache
 - [ ] Fabrication Foundry
 - [x] Family Reunion
-- [ ] Fanatical Offering
+- [x] Fanatical Offering
 - [x] Forgotten Monument
-- [ ] Frilled Cave-Wurm
-- [ ] Fungal Fortitude
+- [x] Frilled Cave-Wurm
+- [x] Fungal Fortitude
 - [ ] Gargantuan Leech
 - [x] Geological Appraiser
 - [ ] Get Lost
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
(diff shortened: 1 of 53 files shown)


bolav:
## Review — LCI batch 7 + sacrifice-as-cost LKI fix

_Scope: the stacked diff `lci-cards-06..lci-cards-07` only (three commits: batch-7 cards, the CastSpellHandler fix, the playtest scenario). The diff vs `main` additionally contains batches 1–6 until those merge._

### Verdict
Behavior is correct and the SDK shape is right: all five cards are pure compositions of existing primitives (no new SDK types), and the engine fix roots a real bug at the canonical zone-transition service rather than patching symptoms. Mergeable; only minor/scope notes below.

### What's good
- **Cards compose existing primitives cleanly.** The two Descend cards (Echo of Dusk, Frilled Cave-Wurm) are `ConditionalStaticAbility` + `ModifyStats` / `GrantKeyword` gated on the shared `Conditions.CardsInGraveyardMatchingAtLeast(4, Permanent)` — correct layer split (7c stats, 6 keyword). Fanatical Offering is `additionalCost(SacrificePermanent)` + `Composite(DrawCards(2), CreateMapToken)`. Fungal Fortitude is a Flash aura with `ModifyStats(2,0)` + an `ATTACHED` leaves-to-graveyard trigger returning the host tapped. No card-specific types introduced.
- **Oracle text / stats / collector numbers verified** against Scryfall LCI data for all five — exact matches.
- **Tests read through projection** (`projectedState.getPower/getToughness/hasKeyword`) and cover both sides of the Descend-4 boundary (3 vs 4 permanent cards), lifelink grant, the aura dies→return-tapped path, and — for Explorer's Cache — ETB counters, dies-with-counter, dies-without-counter, counter-move, and the sorcery-speed gate. The "permanent cards only" nuance of Descend is already pinned by `DescendTrackerTest`.
- **The engine fix is the correct shape.** It routes the four `CastSpellHandler` sacrifice-as-additional-cost sites (`CostAtom.Sacrifice`, `SacrificeCreaturesForCostReduction`, `SacrificeOrPay`, `Casualty`) through `ZoneTransitionService.moveToZone` via a single shared `sacrificePermanentAsCost` helper, mirroring `SacrificeExecutor`. The hand-built `ZoneChangeEvent(lastKnown = null)` that silently dropped LKI-dependent dies/leaves triggers (counters, P/T, keywords, token-ness — CR 603.10 / 608.2h) is gone. Bonus correctness: exit cleanup + graveyard-replacement redirects now run, and the event is tagged `wasSacrificed = true` (CR 701.21) so "if it wasn't sacrificed" triggers stop mis-firing on cost sacrifices. `SacrificeAsCostDeathTriggerTest` is a targeted regression test (fails before, passes after); the full `rules-engine` suite is green.
- **CR citations verified** against the local Comprehensive Rules: 603.10, 608.2h, 701.21, 702.153 all correct; a pre-existing bogus `112.7a` in the Casualty comment was corrected to `608.2h / 113.7a`.

### Issues

**Blocking** — none.

**Important** — none.

**Minor**
- _Scope / PR framing._ This branch carries a cross-cutting **engine** fix on top of the five cards. Its blast radius is every sacrifice-as-cost in the game, not just LCI — called out in the PR body so a reviewer expects the `CastSpellHandler` diff. The activated-ability and non-cast sacrifice paths (`CostHandler`, `CostPaymentService`) were audited and already used `moveToZone` — no change needed there. (Full suite green mitigates the risk.)
- _Untested behavior improvements on the fixed path._ Routing through `moveToZone` newly makes cost-sacrifice honor graveyard-replacement redirects (e.g. Rest in Peace exiling instead) and set `wasSacrificed = true`. Both are correct, but neither has a dedicated test. Optional follow-up: one scenario asserting a cost-sacrificed permanent is exiled under a graveyard-replacement, and one asserting an `excludeSacrifice` trigger does not fire on a cost sacrifice.

### Recommendation
Ship as-is. The two optional follow-up tests above would fully pin the incidental correctness gains but aren't required for merge.

PR #1256, opened 2026-07-09
Add 5 Lost Caverns of Ixalan cards (batch 8)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1255

## LCI cards — batch 8 (5 cards)

Eighth batch of Lost Caverns of Ixalan cards (146 → 151 / 286), each with a scenario test. Stacked on `lci-cards-07` — review only the diff against that branch until the earlier batches merge.

- **Gargantuan Leech** — lifelink; costs {1} less to cast for each Cave you control and each Cave card in your graveyard. Two additive `ModifySpellCost` statics (battlefield count via projected state, graveyard count via base state).
- **Get Lost** — destroy target creature, enchantment, or planeswalker; its controller creates two Map tokens (`CreatePredefinedTokenEffect` with `EffectTarget.TargetController`, the Stroke of Midnight / Beast Within shape).
- **Glimpse the Core** — choose one: search your library for a basic Forest onto the battlefield tapped (fail-to-find per CR 701.23b); or return target Cave card from your graveyard to the battlefield tapped.
- **Glorifier of Suffering** — ETB `ReflexiveTriggerEffect`: you may sacrifice another creature or artifact (resolution-time choice); when you do, put a +1/+1 counter on each of up to two target creatures (true targets via `reflexiveTargetRequirements`).
- **Goblin Tomb Raider** — as long as you control an artifact, it gets +1/+0 and has haste (two `ConditionalStaticAbility`s on `Conditions.ControlArtifact`).

All five cards compose existing SDK primitives — no new SDK types.

### Engine fix — `TargetController` resolves projected + last-known controller (CR 608.2h)

Review of Get Lost surfaced an engine-wide gap: control-changing effects (Threaten, Empress Galina) are Layer-2 continuous effects that never touch the base `ControllerComponent`, so `EffectTarget.TargetController` read the pre-steal controller for a live stolen permanent, and fell back to the card's **owner** once the effect itself had destroyed the target ("Destroy target creature. Its controller creates two Map tokens." — Get Lost, Stroke of Midnight, and Beast Within all shared the bug).

- `ZoneTransitionService` now stashes the battlefield-exit `EntitySnapshot` (already built for `ZoneChangeEvent.lastKnown`) on the entity as a new `LastKnownPermanentComponent`, stripped again on the entity's next zone change (CR 400.7 — a later move makes a new object).
- The `controllerOf` funnel in `TargetResolutionUtils` resolves caster → projected controller → base component → last-known snapshot → owner; the `TargetController` reads in `PredicateEvaluator` and `DamageUtils` gained the same last-known fallback.
- The battlefield-exit controller capture is projected-first, so the exit snapshot and the per-player LTB/death trackers now credit a Threaten-stolen permanent to the thief, as their surrounding comments already claimed.
- `TargetControllerLastKnownInfoScenarioTest` pins all three corners: controller-at-death creates the tokens (CR 608.2h), a live stolen permanent reads the projected controller, and the snapshot is stripped on the next zone change (CR 400.7).
- SDK doc comment + `card-sdk-language-reference.md` updated; no serialized card shape change.

### Verification

Full `:rules-engine`, `:game-server`, and `:mtg-sets` suites pass (card snapshot + lint nets included). `just check-card-printing` passes for all five cards (LCI is the earliest real printing in every case). Manual playtest scenario at `manual-scenarios/sets/lci/cards-8.json`.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 136 / 286
+**Implemented:** 141 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -74,19 +74,19 @@
 - [x] Deep-Cavern Bat
 - [x] Deepfathom Echo
 - [ ] Deeproot Pilgrimage
-- [ ] Defossilize
-- [ ] Diamond Pick-Axe
-- [ ] Didact Echo
+- [x] Defossilize
+- [x] Diamond Pick-Axe
+- [x] Didact Echo
 - [x] Digsite Conservator
 - [x] Dinotomaton
 - [ ] Dire Flail
 - [x] Disruptor Wanderglyph
-- [ ] Disturbed Slumber
+- [x] Disturbed Slumber
 - [ ] Dowsing Device
 - [ ] Dreadmaw's Ire
 - [x] Dusk Rose Reliquary
 - [x] Earthshaker Dreadmaw
-- [ ] Eaten by Piranhas
+- [x] Eaten by Piranhas
 - [ ] Echo of Dusk
 - [ ] Echoing Deeps
 - [x] Enterprising Scallywag
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 146 / 286
+**Implemented:** 151 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -99,15 +99,15 @@
 - [x] Forgotten Monument
 - [x] Frilled Cave-Wurm
 - [x] Fungal Fortitude
-- [ ] Gargantuan Leech
+- [x] Gargantuan Leech
 - [x] Geological Appraiser
-- [ ] Get Lost
+- [x] Get Lost
 - [x] Ghalta, Stampede Tyrant
 - [x] Gishath, Sun's Avatar
-- [ ] Glimpse the Core
-- [ ] Glorifier of Suffering
+- [x] Glimpse the Core
+- [x] Glorifier of Suffering
 - [ ] Glowcap Lantern
-- [ ] Goblin Tomb Raider
+- [x] Goblin Tomb Raider
 - [ ] Goldfury Strider
 - [ ] Grasping Shadows
 - [x] Greedy Freebooter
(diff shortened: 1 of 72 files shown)


bolav:
# Review: batch 8 (diff vs `lci-cards-07`)

## Verdict

Approve. Behavior is right and the SDK shape is right: the batch introduces **zero new SDK or engine types** — all five cards compose existing primitives. Oracle text matches Scryfall exactly, `just check-card-printing` passes for all five (LCI is the earliest real printing in every case), the snapshot diff is purely additive, and all scenario tests plus the mtg-sets snapshot/lint suites pass.

## What's good

- **Pure composition.** Gargantuan Leech's two-source cost reduction is modeled as two additive `ModifySpellCost` statics rather than a new `CostReductionSource` — and `PermanentsYouControlMatching` counts through projected state while `CardsInGraveyardMatchingFilter` uses base state, the correct split for battlefield vs. graveyard.
- **Get Lost follows the established idiom** (Stroke of Midnight and Beast Within use the same Destroy → `CreatePredefinedTokenEffect(..., EffectTarget.TargetController)` shape), and the file comment honestly explains why the `Effects.CreateMapToken` facade couldn't be used (it doesn't expose the controller override — verified).
- **Glorifier of Suffering** gets the trickiest rules shape right: the sacrifice is a resolution-time choice (not a target at trigger time), while the "when you do" counters are true targets chosen when the reflexive trigger goes on the stack (`reflexiveTargetRequirements`), and `ForEachTargetEffect` + `ContextTarget(0)` is the documented multi-target idiom. Test coverage hits the interesting axes: accept, partial targets, decline, and no-legal-sacrifice (no prompt raised).
- **Rule citations are accurate.** CR 701.23 / 701.23b in Glimpse the Core's comment match the Comprehensive Rules text, and the fail-to-find corner they cite is real behavior of the search pipeline.
- Gargantuan Leech's cost tests cover both sources independently, the additive combination, opponent's Caves not counting, and the floor at 0 generic.

## Issues found — and how they were resolved

**Minor**

1. ~~**Get Lost: Map tokens went to the owner, not the last-known controller, when a control-stolen permanent was destroyed.**~~ `TargetController` resolved the live `ControllerComponent` with an owner fallback, and the zone change strips that component — a pre-existing engine-wide gap shared by Stroke of Midnight and Beast Within. Investigation showed it was wider still: control changes are Layer-2 continuous effects that never touch the base component, so even a *live* stolen permanent read its pre-steal controller. **Fixed in this PR** (`Resolve TargetController through projected and last-known controller`): the battlefield-exit `EntitySnapshot` is now stashed on the entity (`LastKnownPermanentComponent`, stripped on the next zone change per CR 400.7), and the `controllerOf` funnel resolves caster → projected controller → base component → last-known snapshot → owner (CR 608.2h). Pinned by the new `TargetControllerLastKnownInfoScenarioTest`.
2. ~~**Glorifier test had a no-op assertion**~~ (`findPermanent(...)?.let { ... }` silently asserted nothing when null). **Fixed**: the test now asserts directly that both counter targets remain on the battlefield and the fodder is in the graveyard.
3. **Glorifier: no test sacrificing an artifact.** The oracle allows "another creature **or artifact**"; all tests sacrifice a creature. Deferred — a one-test addition with e.g. a Millstone as fodder would pin the artifact arm of the filter.
4. **Get Lost: the planeswalker arm is untested** (creature and enchantment are covered). Deferred.
5. **Get Lost's inline `creatureEnchantmentOrPlaneswalker` filter** could live as a `TargetFilter` companion value next to `CreatureOrArtifact` / `ArtifactOrLand` if the wording recurs. Author's discretion; inline is acceptable.
6. ~~**Trailing whitespace** in `manual-scenarios/sets/lci/cards-8.json`.~~ **Fixed.**

## Verification

Issues 1, 2, and 6 are fixed in follow-up commits on this branch; the full `:rules-engine`, `:game-server`, and `:mtg-sets` suites pass after the changes.

PR #1266, opened 2026-07-10
Add 5 Lost Caverns of Ixalan cards (batch 12)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1265

Adds 5 Lost Caverns of Ixalan cards (batch 12), all built from existing SDK primitives — no engine changes.

> **Stacked PR:** this branch is based on `lci-cards-11` (batch 11). Only the top commit — "Add 5 Lost Caverns of Ixalan cards (batch 12)" — is in scope here; the batch 11 commit disappears from the diff once that PR merges.

## Cards

| Card | Cost | Type | Notes |
|------|------|------|-------|
| Petrify | {1}{W} | Enchantment — Aura | Pacifism combat lock + Cursed Totem-style activation lock scoped to the enchanted permanent (`PreventActivatedAbilities` with `attachedToBySource()`); covers artifact hosts too |
| Pirate Hat | {1}{U} | Artifact — Equipment | +1/+1 and granted attack-loot trigger via `GrantTriggeredAbility`; "Equip Pirate {1}" mirrors Dúnedain Blade's restricted-equip idiom |
| Poetic Ingenuity | {2}{R} | Enchantment | Once-per-combat group attack trigger (`YouAttackWithFilter`) + `AggregateBattlefield` attacker count; artifact-cast trigger uses the first-class `oncePerTurn` cap |
| Pugnacious Hammerskull | {2}{G} | Creature — Dinosaur 6/6 | Negated, self-excluding intervening-if (CR 603.4) via `Conditions.YouControl(negate = true, excludeSelf = true)`; stun counter (CR 122.1d) |
| Queen's Bay Paladin | {3}{B}{B} | Creature — Vampire Knight 5/4 | Enters-or-attacks pair sharing one rider (Sentinel of the Nameless City idiom); finality counter + life loss = returned card's mana value |

## Known deviation (documented in the card comment)

Poetic Ingenuity's "create that many Treasure tokens" reads the attacking-Dinosaur count at trigger **resolution**, so a Dinosaur removed from combat in response is not counted, while the ruling fixes the count at trigger time. Same idiom as Choco, Seeker of Paradise; a correct fix needs a trigger-snapshot count primitive (add-feature) that no card uses yet.

## Testing

- 14 scenario tests across 5 new test classes in `rules-engine` — all passing.
- `CardDefinitionSnapshotTest` and `CardLintTest` green; snapshot diff is purely additive (only the 5 new cards).
[...]

PR #1267, opened 2026-07-10
Add 5 Lost Caverns of Ixalan cards (batch 13)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1266

## LCI cards — batch 13 (5 cards)

Thirteenth batch of Lost Caverns of Ixalan cards, existing SDK primitives only (no engine changes), each with a scenario test. Stacked on `lci-cards-12` — please review only the diff vs `lci-cards-12` (the last commit).

- **Ray of Ruin** — {4}{B} sorcery; exile target creature, Vehicle, or nonbasic land, then scry 1.
- **Relic's Roar** — {U} instant; until end of turn, target artifact or creature becomes a 4/3 Dinosaur artifact creature in addition to its other types.
- **Resplendent Angel** — {1}{W}{W} Angel, flying; each end step if you gained 5+ life this turn create a 4/4 flying vigilance Angel; {3}{W}{W}{W}: +2/+2 and lifelink. **Reprint** — canonical `card(...)` authored in M19 (earliest printing); LCI gets a `Printing(...)` row.
- **Restless Anchorage** — creature-land; enters tapped, {T}: {W}/{U}, {1}{W}{U} become a 2/3 flying Bird, attack → create a Map token.
- **Restless Prairie** — creature-land; enters tapped, {T}: {G}/{W}, {2}{G}{W} become a 3/3 Llama, attack → other creatures you control get +1/+1.

All scenario tests pass; LCI + M19 snapshots re-blessed; linter green. Also includes a manual play scenario (`manual-scenarios/sets/lci/cards-13.json`) with all five cards in hand and mana to cast them.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 171 / 286
+**Implemented:** 176 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -196,11 +196,11 @@
 - [x] Quintorius Kand
 - [x] Rampaging Ceratops
 - [x] Rampaging Spiketail
-- [ ] Ray of Ruin
-- [ ] Relic's Roar
-- [ ] Resplendent Angel
-- [ ] Restless Anchorage
-- [ ] Restless Prairie
+- [x] Ray of Ruin
+- [x] Relic's Roar
+- [x] Resplendent Angel
+- [x] Restless Anchorage
+- [x] Restless Prairie
 - [ ] Restless Reef
 - [ ] Restless Ridgeline
 - [ ] Restless Vents
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 161 / 286
+**Implemented:** 166 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -150,7 +150,7 @@
 - [ ] Malamet Battle Glyph
 - [x] Malamet Brawler
 - [x] Malamet Scythe
-- [ ] Malamet Veteran
+- [x] Malamet Veteran
 - [x] Malamet War Scribe
 - [x] Malcolm, Alluring Scoundrel
 - [ ] Malicious Eclipse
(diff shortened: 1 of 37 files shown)


bolav:
## Review (`/review-changes`, diff vs `lci-cards-12` only — stacked PR)

### Verdict

Behavior is right and the SDK shape is right: all five cards compose existing primitives with **zero new SDK types**, and every claim checked against ground truth holds. Approve; one small test gap worth closing at the author's discretion.

### What's good

- **Oracle fidelity verified.** All five cards' mana costs, oracle text, collector numbers, rarities, and artists match the Scryfall LCI data exactly; the M19 Resplendent Angel metadata (CN 34, mythic, Volkan Baǵa) matches the live Scryfall API.
- **Printing discipline.** `just check-card-printing` passes for all five: the canonical Resplendent Angel lives in M19 (its earliest real printing, with the four official rulings) and LCI gets only a `Printing(...)` row.
- **Relic's Roar composition is subtle and correct.** `Effects.BecomeCreature` without `creatureTypes` skips the `SetCreatureSubtypes` modification (confirmed in `BecomeCreatureExecutor.kt:76`), so pairing it with a separate additive `Effects.AddSubtype("DINOSAUR")` correctly implements "in addition to its other types" — and the test pins Centaur/Warrior retention. The doc comment's layer references (Layer 4, Layer 7b) check out against the CR (613.1d, 613.4b).
- **Anchorage/Prairie mirror Raging Ravine's proven structure**, and making the attack trigger an *intrinsic* `triggeredAbility` (rather than granted by the animate, as on Raging Ravine) is rules-accurate for the Restless cycle — the trigger is printed on the land, so it fires even if something else animates it. The comment explains this well.
- **Tests are green and well-aimed.** All 19 scenario tests pass, plus `CardDefinitionSnapshotTest` and `CardLintTest`. Coverage includes the negative target case (Ray of Ruin can't hit a basic land), the Vehicle and nonbasic-land target arms, the intervening-if failure branch (3 life ≠ 5), end-of-turn reversion for both lands, and Prairie's pump correctly *excluding* the attacking land itself.

### Issues

**Blocking** — none.

**Important**

- **Ray of Ruin's Scry 1 is never asserted.** `RayOfRuinScenarioTest.kt:24` — `drainScry()` drains decisions *if* they appear, but asserts nothing; if `Effects.Scry(1)` were dropped from the Composite, all four tests would still pass, so the docstring's "Scry 1 runs after the exile" is currently decorative. One-line fix in the first test: after `resolveStack()`, assert a pending decision exists (e.g. `game.getPendingDecision() shouldNotBe null`) before draining.

**Minor**

- **Anchorage's attack trigger has no description.** `RestlessAnchorage.kt:76` — Prairie sets an explicit oracle-matching `description`; Anchorage's trigger serializes with none (visible in the LCI snapshot). If the runtime auto-description renders as "this creature attacks" or blank, the client text drifts from oracle. Adding `description = "Whenever this land attacks, create a Map token."` would match its cycle-mate.
- **Prairie's mana abilities are untested** (Anchorage's white one is). They're copy-paste of the proven Raging Ravine shape, so author's discretion.

### Recommendation

Merge-ready as-is; ideally add the one-line scry assertion (and optionally the Anchorage trigger description) in a follow-up.

PR #1268, opened 2026-07-10
Add 5 Lost Caverns of Ixalan cards (batch 14)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1267

## LCI cards — batch 14 (5 cards)

Fourteenth batch of Lost Caverns of Ixalan cards, existing SDK primitives only (no engine changes), each with a scenario test. Stacked on `lci-cards-13` — please review only the diff vs `lci-cards-13` (the last commit).

- **Restless Reef** — creature-land; enters tapped, {T}: {U}/{B}, {2}{U}{B} become a 4/4 deathtouch Shark, attack → target player mills 4.
- **Restless Ridgeline** — creature-land; enters tapped, {T}: {R}/{G}, {2}{R}{G} become a 3/4 Dinosaur, attack → another attacking creature gets +2/+0 and untaps.
- **Restless Vents** — creature-land; enters tapped, {T}: {B}/{R}, {1}{B}{R} become a 2/3 menace Insect, attack → rummage (may discard a card, if you do draw a card).
- **Sage of Days** — {2}{U} Human Wizard; ETB look at top three, keep up to one on top, rest to graveyard.
- **Saheeli, the Sun's Brilliance** — {U}{R} Human Artificer; {U}{R},{T}: create a token copy of another target creature or artifact you control (also an artifact), gains haste, sacrificed at the next end step.

All scenario tests pass; LCI snapshot re-blessed; linter green. Also includes a manual play scenario (`manual-scenarios/sets/lci/cards-14.json`) with all five cards in hand and mana to cast them.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 171 / 286
+**Implemented:** 176 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -196,11 +196,11 @@
 - [x] Quintorius Kand
 - [x] Rampaging Ceratops
 - [x] Rampaging Spiketail
-- [ ] Ray of Ruin
-- [ ] Relic's Roar
-- [ ] Resplendent Angel
-- [ ] Restless Anchorage
-- [ ] Restless Prairie
+- [x] Ray of Ruin
+- [x] Relic's Roar
+- [x] Resplendent Angel
+- [x] Restless Anchorage
+- [x] Restless Prairie
 - [ ] Restless Reef
 - [ ] Restless Ridgeline
 - [ ] Restless Vents
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 161 / 286
+**Implemented:** 166 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -150,7 +150,7 @@
 - [ ] Malamet Battle Glyph
 - [x] Malamet Brawler
 - [x] Malamet Scythe
-- [ ] Malamet Veteran
+- [x] Malamet Veteran
 - [x] Malamet War Scribe
 - [x] Malcolm, Alluring Scoundrel
 - [ ] Malicious Eclipse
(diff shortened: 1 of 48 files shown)


bolav:
## Review: batch 14 (diff vs `lci-cards-13` only — stacked PR)

### Verdict

Merge-ready. All five cards compose existing SDK primitives with zero engine/SDK changes, the oracle text matches Scryfall word-for-word, and every check is green: the five scenario test classes, `CardDefinitionSnapshotTest`, `CardLintTest`, and `just check-card-printing` for all five cards (all canonical in LCI, their earliest real printing).

### What's good

- **Cycle consistency.** The three Restless lands mirror the already-merged Restless Anchorage/Prairie modeling exactly — `EntersTapped()` replacement, two separate mana abilities, `BecomeCreatureEffect` animate, and the attack trigger as an intrinsic ability (correct for LCI's cycle, unlike Raging Ravine). Diffing any two cycle members shows only the intended deltas.
- **Precedent reuse.** Vents' rummage is byte-for-byte the `MayEffect(IfYouDoEffect(discard, draw))` shape from Volatile Wanderglyph; Saheeli reuses the `CreateTokenCopyOfTarget` facade the same way The Fire Crystal and Terra, Magical Adept do (`addCardTypes = ARTIFACT`, `sacrificeAtStep = Step.END`), with `TargetOther` correctly excluding herself and no invented sorcery-speed clause.
- **Sage of Days** uses the preferred atomic Gather → Select → Move pipeline with `ChooseUpTo(1)` honoring the "may", and tests cover both the keep-one and decline-all paths, asserting the kept card is back on top.
- **Tests read projected state** (`StateProjector`/`projectedState`) for all type/P/T/keyword assertions. The Ridgeline combat test is particularly good: it verifies the bear gets +2/+0 *and* untaps while the Ridgeline itself stays unbuffed and tapped.
- **Oracle fidelity down to phrasing:** Ridgeline's animate line genuinely reads "This land becomes … until end of turn" on Scryfall while Reef/Vents read "Until end of turn, …" — the files preserve that rather than normalizing it.

### Issues

**Blocking** — none.

**Important** — none.

**Minor** (author's discretion):

1. **Misleading comment in `SaheeliTheSunsBrilliance.kt:33-37`.** The comment justifies permanent haste with "since the token is sacrificed at the next end step it never outlives the haste" — but the printed "It gains haste." has no duration, so a duration-less keyword is simply *correct*, not an approximation needing that defense. Trimming the rationale to "haste is duration-less on the printed card" would stop a future reader from thinking there's a modeling gap.
2. **Redundant re-registrations in tests.** `TestCards.all` already flat-maps every catalog set, so `driver.registerCards(listOf(RestlessReef))` (and the Ridgeline/Vents equivalents) are no-ops. Harmless; dropping them would match the Saheeli/Sage tests, which rely on the catalog.
3. **Untested engine-handled corners:** Ridgeline attacking alone (trigger fizzles via the generic no-legal-targets guard in `TriggerProcessor`), Sage of Days with fewer than three library cards, Reef milling a player with fewer than four. All ride generic engine paths with existing coverage, so tests aren't mandatory.
4. **Inconsistent `description` fields:** the three new attack triggers set `description`, while Anchorage's doesn't and Saheeli's activated ability doesn't. Cosmetic.

### Recommendation

Ship it. Optionally fold in the Saheeli comment trim first; nothing blocks.

PR #1269, opened 2026-07-10
Add 5 Lost Caverns of Ixalan cards (batch 15)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1268

## LCI cards — batch 15 (5 cards)

Fifteenth batch of Lost Caverns of Ixalan cards, existing SDK primitives only (no engine changes), each with a scenario test. Stacked on `lci-cards-14` — please review only the diff vs `lci-cards-14` (the last commit).

- **Sanguine Evangelist** — {2}{W} Vampire Cleric; Battle cry (composed from `ForEachInGroup` + `ModifyStats`, no new keyword, CR 702.91a); enters or dies → create a 1/1 black flying Bat.
- **Shipwreck Sentry** — {1}{U} Human Pirate, Defender; can attack as though it didn't have defender while an artifact entered under your control this turn (`CanAttackDespiteDefender` + `ArtifactEnteredBattlefieldThisTurn`).
- **Sinuous Benthisaur** — {5}{U} Dinosaur; ETB look at top X (X = Caves you control + Cave cards in graveyard), put two into hand, rest on bottom in a random order — fully private, no reveal.
- **Skullcap Snail** — {1}{B} Fungus Snail; ETB target opponent exiles a card from their hand (opponent-chosen, via `Patterns.Hand.exileFromHand`).
- **Song of Stupefaction** — {1}{U} Aura enchanting creature or Vehicle; ETB may mill 2; enchanted permanent gets -X/-0 where X = permanent cards in your graveyard.

All scenario tests pass; LCI snapshot re-blessed; linter green. Also includes a manual play scenario (`manual-scenarios/sets/lci/cards-15.json`) with all five cards in hand and mana to cast them.

The self-review below (posted as a comment) found two issues — an information leak on Benthisaur's dig (`revealed = true` on a look that the printed card keeps private) and a wrong CR citation for battle cry — both already fixed in this commit.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 171 / 286
+**Implemented:** 176 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -196,11 +196,11 @@
 - [x] Quintorius Kand
 - [x] Rampaging Ceratops
 - [x] Rampaging Spiketail
-- [ ] Ray of Ruin
-- [ ] Relic's Roar
-- [ ] Resplendent Angel
-- [ ] Restless Anchorage
-- [ ] Restless Prairie
+- [x] Ray of Ruin
+- [x] Relic's Roar
+- [x] Resplendent Angel
+- [x] Restless Anchorage
+- [x] Restless Prairie
 - [ ] Restless Reef
 - [ ] Restless Ridgeline
 - [ ] Restless Vents
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 161 / 286
+**Implemented:** 166 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -150,7 +150,7 @@
 - [ ] Malamet Battle Glyph
 - [x] Malamet Brawler
 - [x] Malamet Scythe
-- [ ] Malamet Veteran
+- [x] Malamet Veteran
 - [x] Malamet War Scribe
 - [x] Malcolm, Alluring Scoundrel
 - [ ] Malicious Eclipse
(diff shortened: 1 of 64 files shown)


bolav:
## Review: batch 15 (diff vs `lci-cards-14` only — stacked PR)

### Verdict

Merge-ready after two fixes that are already folded into the commit. All five cards compose existing SDK primitives with **zero engine/SDK changes**, oracle text matches Scryfall word-for-word (mana cost, P/T, rarity, collector number, artist, flavor), and every check is green: the five scenario test classes (12 tests), `CardDefinitionSnapshotTest` re-bless, and `just check-card-printing` for all five cards (all canonical in LCI, their earliest real printing).

### What's good

- **Pure composition.** Battle cry is modeled from its reminder text as `Triggers.Attacks` + `ForEachInGroup(GroupFilter(attacking, excludeSelf))` rather than a new keyword type; Shipwreck Sentry reuses the EOE `CanAttackDespiteDefender` + `ArtifactEnteredBattlefieldThisTurn` pair (functional twin of Mechan Shieldmate); Skullcap Snail reuses `Patterns.Hand.exileFromHand` with the chooser derived from the effect target; Song of Stupefaction builds fathomless descent from `GrantDynamicStatsEffect` + `Multiply(Count(GRAVEYARD, Permanent), -1)`; Benthisaur's dig is the preferred atomic Gather → Select → Move pipeline with `ChooseExactly(2)` capping naturally at X < 2.
- **Tests exercise real paths.** Shipwreck Sentry casts Ornithopter through the actual zone-transition path because the gate is an ETB *event* tracker, and covers the negative case with a decoy attacker so the declare-attackers step isn't skipped. Benthisaur pins the decision's options/min/max and the bottom-of-library placement, plus the X = 1 cap edge. Song covers empty graveyard, instant-vs-permanent counting, and the Vehicle host (dormant -X/-0 per CR 208.3, verified against the current Comp Rules). Battle cry asserts the pumped other attacker, the unpumped Evangelist, and an untouched non-attacker.

### Issues found (both fixed in the amended commit)

**Blocking (fixed)**

1. **Sinuous Benthisaur leaked hidden information.** The kept→hand `MoveCollectionEffect` set `revealed = true`, showing the opponent which two cards went to hand — but the oracle text has no reveal ("look at" is fully private). Codebase convention confirms the flag tracks oracle wording: Staunch Crewmate sets it because its text says "reveal … and put it into your hand"; Worldly Counsel and Stargaze (no reveal wording) leave it false. Fixed by dropping the flag and re-blessing the LCI snapshot (the snapshot diff is exactly the one removed field).

**Important (fixed)**

2. **Wrong CR number for battle cry.** `SanguineEvangelist.kt` cited CR 702.92a, which is Living weapon in the current Comprehensive Rules; battle cry is **702.91a**. Corrected.

**Minor** (author's discretion, not addressed):

3. **Skullcap Snail test doesn't pin who chooses.** The decision loop answers whatever `SelectCardsDecision` arises using `pd.playerId`, so it would still pass if the engine wrongly gave the choice to the caster. Adding `pd.playerId shouldBe opp` in that branch would pin the card's actual point.
4. **Controller scoping of the Sentry gate is untested.** `ArtifactEnteredBattlefieldThisTurn` defaults to `Player.You` (verified in `TurnConditions.kt`), but no test here or for Mechan Shieldmate covers "an *opponent's* artifact entering does not enable the attack."

### Recommendation

Ship it. The two real findings are already fixed and re-verified (snapshot re-bless + affected scenario tests green); items 3–4 are optional test-tightening.

PR #1271, opened 2026-07-10
Add 5 Lost Caverns of Ixalan cards (batch 17)

bolav:
## LCI cards — batch 17 (5 cards) + 3 engine bug fixes surfaced by playtest

Seventeenth batch of Lost Caverns of Ixalan cards, each with a scenario test. **Stacked on `lci-cards-16`** — please review only the difference to that branch (the four commits listed below); everything under it is reviewed on the earlier batch PRs.

### Cards (commit `ce9c22a4c`)

- **Sunshot Militia** — {1}{R} Human Soldier; tap two untapped artifacts/creatures you control (sorcery-speed): deals 1 damage to each opponent. The Militia itself may be one of the two (no "other" — pinned by test).
- **Synapse Necromage** — {2}{B} Fungus Wizard; dies → create two 1/1 black Fungus tokens with "This token can't block."
- **Tectonic Hazard** — {R} sorcery; deals 1 damage to each opponent and each creature they control.
- **Tarrian's Soulcleaver** — {1} Legendary Equipment; equipped creature has vigilance; whenever another artifact or creature is put into a graveyard from the battlefield, +1/+1 counter on equipped creature; Equip {2}.
- **Tendril of the Mycotyrant** — {1}{G} Fungus Wizard; {5}{G}{G}: put seven +1/+1 counters on target noncreature land you control, it becomes a 0/0 Fungus with haste (still a land).

Includes a manual self-play scenario (`manual-scenarios/sets/lci/cards-17.json`) with all five in hand and mana to cast them.

Note: **Sunfire Torch** was triaged out of this batch to NEEDS-ENGINE (a granted triggered ability can't sacrifice its granting Equipment — `granterId` isn't threaded to triggered abilities; same family as Deconstruction Hammer) and replaced by Tendril of the Mycotyrant.

### Engine fixes (found by playtesting this batch — latent bugs, not new capabilities)

- **`b6cf3f071` — Fix zone-change trigger filters for composite predicates under LKI.** Union filters (`Artifact or Creature`, collapsed to `CardPredicate.Or`) fell into `TriggerMatcher`'s live-entity path, so Tarrian's Soulcleaver silently missed every **token** death (the token entity is swept by CR 704.5d before matching) and any permanent whose types were only projected at the moment of leaving. Composites now recurse through the LKI-aware evaluator; heterogeneous `anyOf` unions (previously fail-open in this matcher) match per branch. Regression test kills a real Fungus token and asserts the counter lands.
- **`50694928c` — Fix JSON discriminator collision on AddType/RemoveType floating effects.** `SerializableModification.AddType(val type: …)` collided with the persistence Json's `classDiscriminator = "type"`, crashing `RedisGameRepository.save` the first time a permanent animation (Tendril's `BecomeCreature`) was in the saved state. Fields renamed via `@SerialName("cardType")`; a hygiene test now walks every `SerializableModification` leaf so the next collision fails at test time. No migration concern — colliding states could never have been serialized.
- **`a004ed888` — Honor `excludeSelf` in TapPermanents PayCost candidates.** The PayCost prompt/affordability path always excluded the cost's source; per the atom's `excludeSelf` flag (false unless the card says "another"), an untapped source may pay its own "tap an untapped permanent you control" cost. The activation path already honored the flag; the two paths now agree.

### Verification

All 5 scenario test classes + the new regression tests pass; full `:rules-engine:test` (6842 tests) and `:mtg-sets:test` (snapshot + lint) green via `gradle-locked`; `just check-card-printing` clean for all five cards; LCI snapshot re-blessed (flavor-text addition only).

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 191 / 286
+**Implemented:** 196 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -243,13 +243,13 @@
 - [ ] Sunbird Standard
 - [ ] Sunfire Torch
 - [ ] Sunken Citadel
-- [ ] Sunshot Militia
+- [x] Sunshot Militia
 - [x] Swashbuckler's Whip
-- [ ] Synapse Necromage
+- [x] Synapse Necromage
 - [ ] Tarrian's Journal
-- [ ] Tarrian's Soulcleaver
-- [ ] Tectonic Hazard
-- [ ] Tendril of the Mycotyrant
+- [x] Tarrian's Soulcleaver
+- [x] Tectonic Hazard
+- [x] Tendril of the Mycotyrant
 - [x] Terror Tide
 - [ ] The Ancient One
 - [ ] The Belligerent
--- a/manual-scenarios/sets/lci/cards-17.json
+++ b/manual-scenarios/sets/lci/cards-17.json
@@ -1,0 +1,50 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": [
+      "Sunshot Militia",
+      "Synapse Necromage",
+      "Tectonic Hazard",
+      "Tarrian's Soulcleaver",
+      "Tendril of the Mycotyrant"
+    ],
+    "battlefield": [
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Grizzly Bears"}
+    ],
+    "graveyard": [],
+    "library": ["Forest", "Mountain", "Swamp", "Grizzly Bears", "Forest", "Mountain"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": ["Grizzly Bears", "Island"],
+    "battlefield": [
+      {"name": "Cenote Scout"},
+      {"name": "Grizzly Bears"},
+      {"name": "Basking Capybara"}
+    ],
+    "graveyard": [],
+    "library": ["Island", "Swamp", "Forest", "Mountain", "Island", "Swamp"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": ["END"],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "SELF"
+}
--- a/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/lci/cards/SunshotMilitia.kt
+++ b/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/lci/cards/SunshotMilitia.kt
@@ -1,0 +1,49 @@
+package com.wingedsheep.mtg.sets.definitions.lci.cards
+
+import com.wingedsheep.sdk.dsl.Costs
+import com.wingedsheep.sdk.dsl.Effects
+import com.wingedsheep.sdk.dsl.card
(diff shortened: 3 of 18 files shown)


bolav:
# Review: batch 17 (diff vs `lci-cards-16`)

Stacked PR — this review covers only the difference to `lci-cards-16` (the batch-17 card commit plus the three engine-fix commits it motivated); everything below it is reviewed on the earlier batch PRs.

## Verdict

The five card definitions are faithful to their Oracle text and are pure composition over existing SDK primitives — no new SDK types, nothing over-specialized. However, playtesting this batch surfaced **two blocking engine bugs** the cards walk straight into (token deaths invisible to OR-filter triggers; a JSON-discriminator crash on saving animated permanents) plus a cost-payment inconsistency. All three were fixed on this branch with regression tests before this push.

## What's good

- **SDK shape is exactly right.** Every card composes existing primitives: `Costs.TapPermanents` (no-"other" shape, self-eligible per the card's ruling), `Triggers.Dies`, `Triggers.leavesBattlefield(to = GRAVEYARD, binding = OTHER)` for "another artifact or creature", `Effects.AddCounters` + `Effects.BecomeCreature(Duration.Permanent)` for the animate, `Patterns.Group.dealDamageToAll` for the sweep. `TriggerBinding.OTHER` is the correct reading of "another" (excludes the Equipment, not the equipped creature).
- **Verified against ground truth:** oracle text, mana costs, P/T, collector numbers, rarities, artists, and flavor text match Scryfall exactly; `just check-card-printing` passes for all five (LCI is the earliest real printing); CR 700.4 citation verified against CR 2026-04-17; backlog count 191→196 matches.
- **Tests hit the corners:** Sunshot Militia covers the sorcery-speed gate, the cost gate, and self-tap; Soulcleaver covers both branches of the OR filter, any-controller deaths, the unattached no-op, and (new) a token death; Tectonic Hazard proves the caster's side is spared. Kills go through the real damage/SBA path (Lightning Bolt / Doom Blade), not state surgery.

## Issues

### Blocking — **fixed on this branch**

1. **Tarrian's Soulcleaver never triggered on token deaths** (found in playtest: a dying Fungus token never grew the equipped creature). Not a card bug — `Artifact.or(Creature)` collapses to a single `CardPredicate.Or`, which fell into `TriggerMatcher.matchesZoneChangeTrigger`'s generic branch requiring the live `CardComponent`; a token is swept by CR 704.5d before matching runs, so the lookup was null and the trigger silently missed. The same branch read base printed types, missing permanents whose types were only projected at the moment of leaving (e.g. a Tendril-animated land dying as a creature). **Fix (`b6cf3f071`):** composite `Or`/`And`/`Not` predicates recurse through the LKI-aware evaluator against the event's last-known type line; heterogeneous `anyOf` unions, previously ignored (fail-open) by this matcher, now match per branch. Regression test creates a real token via Synapse Necromage's dies trigger, kills it (died + swept in the same SBA pass), and asserts the counter lands.
2. **Saving a game with an animated permanent crashed Redis persistence** (found in playtest: `JsonEncodingException` in `RedisGameRepository.save` after activating Tendril of the Mycotyrant). `SerializableModification.AddType`/`RemoveType` had a property named `type`, colliding with the persistence Json's `classDiscriminator = "type"`. **Fix (`50694928c`):** JSON fields renamed via `@SerialName("cardType")`; a hygiene test walks every `SerializableModification` leaf and asserts no field collides with the discriminator, plus round-trips `ActiveFloatingEffect` for both types. No stored-data migration concern — colliding states could never have been serialized successfully.

### Important — **fixed on this branch**

3. **`excludeSelf` ignored in the TapPermanents PayCost path.** `CostPaymentService` hard-excluded the cost's source from tap candidates, while the activation path (`ActivatedAbilityEnumerator`, `CostHandler.payTapPermanents`) honored the atom's `excludeSelf` flag — so the two paths disagreed, and a no-"other" cost like "tap an untapped permanent you control" (Command Bridge) couldn't be paid by the untapped source itself. **Fix (`a004ed888`):** the PayCost prompt/affordability path passes the exclusion only when `excludeSelf` is set. One existing test had pinned the buggy behavior; it was updated to keep its real intent (unaffordable when *nothing* is untapped) and a companion test pins the corrected self-eligibility.

### Minor — **fixed pre-push (amended into the batch commit)**

- Synapse Necromage used the raw `CreateTokenEffect` constructor instead of the `Effects.CreateToken` facade (project rule); switched, identical compiled output.
- Synapse Necromage was missing its flavor text (present on Scryfall); added, snapshot re-blessed.
- Tectonic Hazard's test carried a meaningless trailing assertion (`oppMinnow != myBears`); removed.

### Notes (no action)

- Tectonic Hazard resolves its two damage sub-effects sequentially within one resolution rather than simultaneously; no SBA runs mid-resolution, so there is no observable rules difference in the supported formats.
- The Soulcleaver trigger correctly fires when the *equipped creature itself* dies ("another" excludes only the Equipment); the counter then finds no equipped creature and no-ops, covered by the unattached test.

## Verification

- Full `:rules-engine:test` (6842 tests) and `:mtg-sets:test` (snapshot + lint nets) pass via `gradle-locked` after the fixes.
- All new regression tests pass: Soulcleaver token-death, discriminator hygiene + two round-trips, PayCost self-eligibility, Sunshot Militia self-tap.
- `just check-card-printing` clean for all five cards.

PR #1272, opened 2026-07-10
Add 5 Lost Caverns of Ixalan cards (batch 18)

bolav:
## LCI cards — batch 18 (5 cards) + a client-view fix surfaced by playtest

Eighteenth batch of Lost Caverns of Ixalan cards, each with a scenario test. **Stacked on `lci-cards-17`** — please review only the difference to that branch (the two commits listed below); everything under it is reviewed on the earlier batch PRs.

### Cards (commit `1028b433e`)

- **The Belligerent** — {2}{U}{R} Legendary Artifact Vehicle 5/5; attacks → create a Treasure and, until end of turn, look at / play lands and cast spells from the top of your library; Crew 3. Modeled as the Lunar Whale pattern: conditional statics gated on "attacked this turn" (the approximation vs the printed until-end-of-turn one-shot is documented in the card comment).
- **The Mycotyrant** — {1}{B}{G} Legendary Elder Fungus \*/\*; trample; P/T = number of Fungi/Saprolings you control (CDA, counts itself); your end step → create X 1/1 black can't-block Fungus tokens, X = times you descended this turn (CR 700.11 count, not the boolean).
- **Thousand Moons Crackshot** — {1}{W} Human Soldier 2/2; attacks → you may pay {2}{W}; when you do, tap target creature (reflexive trigger, target chosen when the reflexive ability goes on the stack).
- **Thousand Moons Infantry** — {2}{W} Human Soldier 2/4; untap during each other player's untap step.
- **Threefold Thunderhulk** — {7} Artifact Creature Gnome 0/0; enters with three +1/+1 counters; enters or attacks → create 1/1 colorless Gnome artifact tokens equal to its power (evaluated at resolution, so post-pump attacks mint more); {2}, sacrifice another artifact → +1/+1 counter.

No new SDK types — all five compose existing primitives. Includes a manual self-play scenario (`manual-scenarios/sets/lci/cards-18.json`) with all five in hand and mana to cast them.

### Client-view fix (commit `3c081a894` — found by playtesting this batch; latent bug, not a new capability)

**Reveal top-of-library behind conditional static gates in client view.** `ClientStateTransformer`'s visibility checks matched statics by bare type, so a `ConditionalStaticAbility`-wrapped `LookAtTopOfLibrary` (The Belligerent is the first such card) never revealed the controller's top card — the play permission was legal server-side (`CastPermissionUtils` already unwraps conditionals) while the client had no visible card to act on. The transformer's four visibility checks (private look at top, public top reveal, look at face-down creatures, opponents' hands revealed) now resolve conditional gates through the same unwrap-and-evaluate helper, keeping what a player *sees* in sync with what the legal-actions layer lets them *do*. View-layer assertions added: top card hidden pre-attack, revealed post-attack to the controller only.

### Verification

All 5 scenario test classes pass, plus regression tests for every other transformer path touched (Goblin Spy, Seer's Vision, Future Sight, The Lunar Whale); `:mtg-sets:test` snapshot + lint green via `gradle-locked`; `just check-card-printing` clean for all five cards; oracle text verified against Scryfall; CR 700.11 citation verified against the Comprehensive Rules.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 191 / 286
+**Implemented:** 196 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -243,13 +243,13 @@
 - [ ] Sunbird Standard
 - [ ] Sunfire Torch
 - [ ] Sunken Citadel
-- [ ] Sunshot Militia
+- [x] Sunshot Militia
 - [x] Swashbuckler's Whip
-- [ ] Synapse Necromage
+- [x] Synapse Necromage
 - [ ] Tarrian's Journal
-- [ ] Tarrian's Soulcleaver
-- [ ] Tectonic Hazard
-- [ ] Tendril of the Mycotyrant
+- [x] Tarrian's Soulcleaver
+- [x] Tectonic Hazard
+- [x] Tendril of the Mycotyrant
 - [x] Terror Tide
 - [ ] The Ancient One
 - [ ] The Belligerent
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 196 / 286
+**Implemented:** 201 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -252,17 +252,17 @@
 - [x] Tendril of the Mycotyrant
 - [x] Terror Tide
 - [ ] The Ancient One
-- [ ] The Belligerent
+- [x] The Belligerent
 - [ ] The Enigma Jewel
 - [ ] The Everflowing Well
 - [ ] The Millennium Calendar
-- [ ] The Mycotyrant
+- [x] The Mycotyrant
 - [ ] The Skullspore Nexus
-- [ ] Thousand Moons Crackshot
-- [ ] Thousand Moons Infantry
+- [x] Thousand Moons Crackshot
+- [x] Thousand Moons Infantry
 - [ ] Thousand Moons Smithy
 - [x] Thrashing Brontodon
-- [ ] Threefold Thunderhulk
+- [x] Threefold Thunderhulk
 - [ ] Throne of the Grim Captain
 - [ ] Tinker's Tote
 - [ ] Tishana's Tidebinder
--- a/manual-scenarios/sets/lci/cards-17.json
+++ b/manual-scenarios/sets/lci/cards-17.json
@@ -1,0 +1,50 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": [
+      "Sunshot Militia",
+      "Synapse Necromage",
+      "Tectonic Hazard",
+      "Tarrian's Soulcleaver",
+      "Tendril of the Mycotyrant"
+    ],
+    "battlefield": [
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Swamp"},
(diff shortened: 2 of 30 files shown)


bolav:
# Review: `lci-cards-18` (vs base `lci-cards-17`)

Stacked PR — this review covers only the difference to `lci-cards-17` (the batch-18 card commit plus the follow-up fix commit). Diff: 5 card definitions, 5 scenario tests, snapshot golden, backlog bookkeeping, one manual self-play scenario — and, after playtest, one `ClientStateTransformer` fix.

## Verdict

The five card definitions are faithful to Oracle text and are pure composition of existing primitives — zero new SDK types. All scenario tests plus the snapshot and lint nets pass. Playtesting the batch surfaced one real (latent, pre-existing) bug — the client never revealed a top-of-library card whose look permission sits behind a conditional gate — which is fixed in `3c081a894`. All review findings below were addressed on the branch.

## What's good

- **Oracle fidelity verified**: all five cards match Scryfall exactly (mana cost, type line, P/T, rules text). `just check-card-printing` passes for all five — LCI is the earliest real printing in each case.
- **CR 700.11 citation verified** against the Comprehensive Rules, and the descend tracker correctly excludes tokens and non-permanent cards, so the rulings quoted in The Mycotyrant's metadata are actually true of the implementation.
- **Good primitive reuse throughout**: The Mycotyrant's CDA via `AggregateBattlefield` + `withAnySubtype`; the descend count via the existing `DynamicAmounts.descendedThisTurn`; Thunderhulk's "enters or attacks" as the established two-trigger split with `sourcePower()` read at resolution (rulings-accurate for post-pump attacks — and tested); Crackshot's "you may pay… when you do" as `ReflexiveTriggerEffect` with targets chosen when the reflexive ability goes on the stack (rulings-accurate).
- **Tests exercise the right corners**: negative controls (Mycotyrant with a non-Fungus creature; the plain-soldier control in the Infantry test; decline-payment path for Crackshot; play-from-top denied before attack), and the Infantry double-untap loop across two turn cycles.

## Issues found → all fixed in `3c081a894`

### Blocking (found by playtest)

**1. The Belligerent's top-of-library card was never revealed in the client.**
`ClientStateTransformer.hasLookAtTopOfLibrary` checked `it is LookAtTopOfLibrary` on the raw static list. The Belligerent's look is wrapped in `ConditionalStaticAbility(SourceAttackedThisTurn, LookAtTopOfLibrary)`, so the check never matched and the top card was never revealed to the controller. The legal-actions side does this correctly — `CastPermissionUtils.activeStaticAbility` unwraps the conditional and evaluates the gate — so the server considered the play legal, but the client had no visible card to attach the action to. The Lunar Whale never hit this because its `LookAtTopOfLibrary` is unconditional; The Belligerent is the first card to gate it. The rules-engine scenario tests passed because they drive actions directly and never go through the view layer.
**Fixed**: the transformer now resolves conditional gates through the same unwrap-and-evaluate helper (`activeStaticAbility` / `hasActiveStaticAbility`), applied to all four bare-type visibility checks (`hasLookAtTopOfLibrary`, `revealsTopOfLibraryPublicly`, `hasLookAtFaceDownCreatures`, `revealsOpponentHandsTo` — the latter three had the same latent gap). View-layer assertions added to `TheBelligerentScenarioTest`: top card hidden pre-attack, revealed post-attack to the controller only (not the defender). Regression suites for the other paths (Goblin Spy, Seer's Vision, Future Sight, The Lunar Whale) stay green.

### Important

**2. "Until end of turn" modeled as "attacked this turn" diverges when The Belligerent leaves the battlefield.** The real card grants the permission as a one-shot effect of the trigger, lasting until end of turn regardless of what happens to the Vehicle; the conditional-static model ends the permission the moment the permanent leaves (and opens the window at attack declaration rather than trigger resolution). For Lunar Whale the pattern is exact — its printed text is "as long as ~ attacked this turn" — but here it's an approximation.
**Fixed** (documented): the card comment now states the divergence explicitly and notes the proper fix (an until-end-of-turn floating permission grant) as future `add-feature` work.

### Minor

**3. Redundant test registrations.** `TestCards.all` already includes every catalog card, so the explicit `TheBelligerent`, `TheMycotyrant`/`BroodrageMycoid`, and `ThousandMoonsInfantry` registrations were no-ops. **Fixed**: dropped; the two registrations that do real work (the Treasure token — tokens aren't in the catalog list — and the test-local vanilla soldier) stay, each with a comment saying why.

**4. `colorIdentity` inconsistency.** Thousand Moons Infantry was the lone outlier among 126 mono-colored LCI cards without an explicit `colorIdentity`. **Fixed**: added `colorIdentity = "W"`; snapshot golden re-blessed (exactly one field added) and re-verified without the update flag.

## Verification

All 5 batch-18 scenario suites pass; transformer regression tests (Goblin Spy, Seer's Vision, Future Sight, The Lunar Whale) pass; `CardDefinitionSnapshotTest` + `CardLintTest` green; `just check-card-printing` clean for all five cards.

PR #1273, opened 2026-07-10
Add 5 Lost Caverns of Ixalan cards (batch 19)

bolav:
## LCI cards — batch 19 (5 cards)

Nineteenth batch of Lost Caverns of Ixalan cards, existing SDK primitives only (no engine changes), each with a scenario test. **Stacked on `lci-cards-18`** (PR #1272) — please review only the difference to that branch (the single commit `b5df0bf6d`); everything under it is reviewed on the earlier batch PRs.

- **Tinker's Tote** — {2}{W} artifact; ETB create two 1/1 colorless Gnome artifact tokens; {W}, sacrifice: gain 3 life.
- **Triumphant Chomp** — {R} sorcery; deals damage to target creature equal to the greater of 2 and the greatest power among Dinosaurs you control — modeled as `DynamicAmount.Max(Fixed(2), battlefield(You, Dinosaur).maxPower())`, no new types.
- **Uchbenbak, the Great Mistake** — {3}{U}{B} Skeleton Horror, vigilance/menace; Descend 8 — {4}{U}{B}: return it from your graveyard to the battlefield with a finality counter (sorcery-speed graveyard activation gated on 8+ permanent cards in your graveyard; the Balustrade Wurm shape).
- **Unlucky Drop** — {3}{U} instant; target artifact or creature's owner puts it on the top or bottom of their library (owner's choice — the tests pin that the decision goes to the owner, not the caster).
- **Vito, Fanatic of Aclazotz** — {2}{W}{B} Vampire Demon, flying; whenever you sacrifice another permanent, escalate by this ability's resolution count this turn: 1st gain 2 life, 2nd each opponent loses 2 life, 3rd create a 4/3 W/B flying Vampire Demon token; 4th+ does nothing (the Victor, Valgavoth's Seneschal template — and this batch adds the family's first 4th-resolution regression test).

Includes a manual self-play scenario (`manual-scenarios/sets/lci/cards-19.json`) with all five in hand, mana to cast them, sacrifice fodder for Vito's three tiers, a Dinosaur for Chomp, and a pre-stocked graveyard for Uchbenbak's Descend 8.

### Verification

All 14 scenario tests pass — including negative gates (Descend 8 refused at 7 permanents, refused when 8 cards are mostly instants, refused outside a main phase) and Vito's full 1st→4th escalation; `:mtg-sets:test` snapshot + lint green via `gradle-locked`; `just check-card-printing` clean for all five cards; oracle text verified against Scryfall.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 191 / 286
+**Implemented:** 196 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -243,13 +243,13 @@
 - [ ] Sunbird Standard
 - [ ] Sunfire Torch
 - [ ] Sunken Citadel
-- [ ] Sunshot Militia
+- [x] Sunshot Militia
 - [x] Swashbuckler's Whip
-- [ ] Synapse Necromage
+- [x] Synapse Necromage
 - [ ] Tarrian's Journal
-- [ ] Tarrian's Soulcleaver
-- [ ] Tectonic Hazard
-- [ ] Tendril of the Mycotyrant
+- [x] Tarrian's Soulcleaver
+- [x] Tectonic Hazard
+- [x] Tendril of the Mycotyrant
 - [x] Terror Tide
 - [ ] The Ancient One
 - [ ] The Belligerent
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 196 / 286
+**Implemented:** 201 / 286
 - [x] Abrade
 - [ ] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -252,17 +252,17 @@
 - [x] Tendril of the Mycotyrant
 - [x] Terror Tide
 - [ ] The Ancient One
-- [ ] The Belligerent
+- [x] The Belligerent
 - [ ] The Enigma Jewel
 - [ ] The Everflowing Well
 - [ ] The Millennium Calendar
-- [ ] The Mycotyrant
+- [x] The Mycotyrant
 - [ ] The Skullspore Nexus
-- [ ] Thousand Moons Crackshot
-- [ ] Thousand Moons Infantry
+- [x] Thousand Moons Crackshot
+- [x] Thousand Moons Infantry
 - [ ] Thousand Moons Smithy
 - [x] Thrashing Brontodon
-- [ ] Threefold Thunderhulk
+- [x] Threefold Thunderhulk
 - [ ] Throne of the Grim Captain
 - [ ] Tinker's Tote
 - [ ] Tishana's Tidebinder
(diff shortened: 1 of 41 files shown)


bolav:
# Review: `lci-cards-19` (vs base `lci-cards-18`)

Stacked PR — this review covers only the difference to `lci-cards-18` (the single batch-19 commit). Diff: 5 card definitions, 5 scenario tests, snapshot golden, backlog bookkeeping (201→206), one manual self-play scenario. No SDK or engine changes.

## Verdict

All five cards are faithful to Oracle text and pure composition of existing primitives — zero new SDK types. Tests are green, printings check out. All review findings below were addressed and amended into the commit.

## What's good

- **Triumphant Chomp is a model citizen of the AST-composition principle**: "2 or greatest Dinosaur power, whichever is greater" built as `DynamicAmount.Max(Fixed(2), battlefield(You, Dinosaur).maxPower())` instead of a new effect type. The test matrix covers all three interesting cases (no Dinosaurs → 2; power 5 → 5; power 1 → floor still wins) plus controller scoping (opponent's Dinosaur ignored), each read off a surviving 0/20 wall's `DamageComponent` so the exact amount is asserted, not inferred from a death.
- **Uchbenbak reuses the Balustrade Wurm reanimation shape verbatim** (graveyard-zone activated ability, `CardsInGraveyardMatchingAtLeast(8, Permanent)` gate, sorcery timing, Move + finality counter). The gate is tested in both refusal directions — 7 permanents, and 8 cards of which only 1 is a permanent — through the legal-action enumerator (the right layer), and the "counts itself" corner is implicitly pinned (Uchbenbak + 7 Forests = 8, offered).
- **Unlucky Drop's tests pin the semantics that matter**: the top/bottom decision goes to the *owner* (`pendingDecision.playerId shouldBe player2Id`), not the caster, exercised for both a creature (top) and a non-creature artifact (bottom, with a pre-seeded library card so bottom is distinguishable from top).
- **Vito's trigger is correctly self-excluding**: `Triggers.YouSacrificeAnother` carries `TriggerBinding.OTHER`, so Vito being sacrificed never fires his own ability. The `{0}` sacrifice-sorcery driver keeps the escalation test clean.
- Oracle text matches Scryfall exactly for all five; `just check-card-printing` passes for all five (LCI earliest real printing); backlog bookkeeping correct.

## Issues found → all fixed (amended into the commit)

### Important

**1. Vito's "4th+ resolution does nothing" was untested — here and everywhere else in the Victor-template family.** The card comment asserts it, but the scenario test stopped at three sacrifices, and Victor, Valgavoth's Seneschal's own test also stops at three. If `Conditions.SourceAbilityResolvedNTimes` ever regressed from `==` to `>=`, a 4th sacrifice would mint a second token and no test in the repo would catch it. **Fixed**: the test now runs a fourth sacrifice and asserts life totals unchanged and still exactly one Vampire Demon token — the family's first 4th-resolution regression coverage.

### Minor

**2. Redundant target filter** — `TriumphantChomp.kt` used `TargetCreature(filter = TargetFilter.Creature)`, which restates the factory default. **Fixed**: now `Targets.Creature` like the rest of the corpus; compiled JSON is identical, so no snapshot change.

**3. Redundant registrations** — `TriumphantChomp` and `VitoFanaticOfAclazotz` were explicitly registered on top of registries that already auto-discover the catalog. **Fixed**: dropped; only the test-local cards (wall, dinosaurs, fodder, the `{0}` sacrifice sorcery) keep explicit registration, with comments saying why.

**4. Untested sorcery-speed restriction** — Uchbenbak sets `TimingRule.SorcerySpeed` but no test attempted the activation outside a main phase. **Fixed**: new test proves the activation is not offered at `DECLARE_ATTACKERS` with the graveyard gate satisfied and mana available (a decoy attacker keeps the combat steps from being skipped). Note: the finality counter lands via `AddCountersEffect` after the move rather than "enters with" — the same accepted shape as Balustrade Wurm.

## Verification

All 14 tests pass across the five suites (including the new sorcery-speed and 4th-resolution tests); `CardDefinitionSnapshotTest` + `CardLintTest` green; `just check-card-printing` clean for all five cards.

PR #1277, opened 2026-07-11
Match becomes-target trigger filters against projected types

bolav:
## Summary

"Becomes the target" triggers (Pawpatch Recruit, Surrak Elusive Hunter, King of the Oathbreakers, …) never fired when the targeted permanent only matched the trigger's type filter via a continuous effect — e.g. an animated land.

Observed in play: Singing Bell Strike legally enchanted an animated Forest (target validation uses the projection), but the defender's Pawpatch Recruit — "whenever a **creature you control** becomes the target of a spell or ability an opponent controls" — never triggered, because `TriggerMatcher.matchesBecomesTargetTrigger` checked the target's card predicates against the **base** type line (`cardComponent.typeLine.isCreature`). The two sides of the engine disagreed about whether the Forest was a creature.

## Fix

`matchesBecomesTargetTrigger` now evaluates the whole target filter through the projection-aware `PredicateEvaluator.matches()`, replacing its hand-rolled per-predicate loops (card predicates, state predicates, and a bespoke controller check):

- Battlefield permanents are matched against their **projected** types, so type-changing effects count (an animated land *is* "a creature you control" while the effect lasts). Changeling permanents now also correctly match subtype filters, and face-down permanents match only their 2/2-creature projection (Rule 708.2a) instead of their base card types.
- Stack objects have no battlefield projection and fall back to base card data — and to the spell's caster for the controller predicate — preserving the Surrak "…or a creature spell you control" behavior (`includeSpellTargets`).
- The collapse deletes ~30 lines of duplicated predicate evaluation and picks up filter-level `anyOf` support, which the old loops silently ignored.

The zone-change/LKI trigger path intentionally keeps `TriggerMatcher`'s own predicate matcher — for permanents that already left the battlefield, last-known info is the correct source, not the current projection.

## Testing

- New regression test in `PawpatchRecruitTest`: Kamahl, Fist of Krosa animates a Forest, the opponent bolts it → Pawpatch Recruit's trigger must fire and must exclude the targeted Forest from its "other than that creature" counter targets. Fails without the fix, passes with it.
- Related becomes-target suites pass: Surrak (permanent + spell-target + non-creature cases), Cactarantula, Manifold Mouse (Valiant), King of the Oathbreakers, both Elrond, Master of Healing target-trigger tests, Kamahl.
- Full `just test-rules` suite green.

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/event/TriggerMatcher.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/event/TriggerMatcher.kt
@@ -1194,48 +1194,20 @@ class TriggerMatcher(
         // Valiant: check if this is the first time this turn
         if (trigger.firstTimeEachTurn && !event.firstTimeByThisController) return false
 
-        // Check targetFilter against the targeted entity
+        // Check targetFilter against the targeted entity. The target is live (a battlefield
+        // permanent, or a spell on the stack for includeSpellTargets), so evaluate the whole
+        // filter through PredicateEvaluator: card predicates read projected types when the
+        // entity has a battlefield projection — an animated land IS "a creature you control"
+        // while the effect lasts — and fall back to base card data for stack objects; the
+        // controller predicate likewise falls back to the spell's caster (Surrak, Elusive
+        // Hunter). Targeted abilities on the stack carry no card data and never match.
         if (trigger.targetFilter != GameObjectFilter.Any) {
-            val projected = state.projectedState
             val targetContainer = state.getEntity(event.targetEntityId) ?: return false
-            val targetCard = targetContainer.get<CardComponent>() ?: return false
-
-            // Check card predicates
-            for (predicate in trigger.targetFilter.cardPredicates) {
-                if (!matchesCardPredicate(predicate, targetCard, projected, event.targetEntityId)) return false
-            }
-
-            // Check state predicates (e.g. "with a +1/+1 counter on it" — Elrond, Master of
-            // Healing). The targeted permanent is on the battlefield, so projected/base state
-            // applies. (A spell-on-stack target has no such state, so the filter simply won't match.)
-            if (trigger.targetFilter.statePredicates.isNotEmpty()) {
-                val predicateContext = com.wingedsheep.engine.handlers.PredicateContext(
-                    controllerId = controllerId, sourceId = sourceId
-                )
-                val statePredicateEvaluator = PredicateEvaluator()
-                for (predicate in trigger.targetFilter.statePredicates) {
-                    if (!statePredicateEvaluator.matchesStatePredicate(
-                            state, event.targetEntityId, predicate, predicateContext
-                        )) return false
-                }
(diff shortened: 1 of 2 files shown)


bolav:
## Code review (`/review-changes`)

### Verdict

Correct fix, right shape. The becomes-target trigger matcher checked card predicates against the base type line while target validation used the projection, so an animated land could be legally targeted as "a creature you control" yet never fire Pawpatch Recruit. Routing the whole filter through the projection-aware `PredicateEvaluator` fixes the disagreement and aligns with the engine's "projected state for battlefield filters" rule. No new SDK types — nothing to compose differently.

### What's good

- **The fix converges two evaluators instead of patching one.** Trigger matching and target validation now agree on what "a creature you control" means; the previous per-predicate reimplementation in `TriggerMatcher` was exactly the kind of duplicated evaluator that drifts.
- **Free correctness upgrades ride along:** battlefield subtype filters at this site now honor Changeling and projected subtypes (a Changeling permanent matches "Spirit you control" for King of the Oathbreakers), and face-down permanents match only their 2/2-creature projection (CR 708.2a, verified against the current Comp Rules) instead of leaking base card types.
- **The test is the real thing:** it animates a Forest with Kamahl, Fist of Krosa's actual activated ability rather than a harness shortcut, and asserts both halves — the trigger fires, and `otherThanTriggeringEntity` excludes the animated Forest from the counter targets while other creatures remain legal.

### Issues found and resolved during review

All were addressed in the final commit before this PR was opened:

1. **Important — duplicated evaluation logic.** The first version of the fix swapped only the card-predicate loop to `PredicateEvaluator.matchesCardPredicate`, leaving hand-rolled loops for state predicates and a bespoke controller check that reimplemented a subset of `matchesControllerPredicate`'s fallback chain (projected controller → base `ControllerComponent` → `SpellOnStackComponent.casterId` → ability stack components). Resolved by collapsing the whole block into a single `PredicateEvaluator.matches()` call — ~30 fewer lines, and it closes a latent gap: the old loops silently ignored `targetFilter.anyOf`, so a filter built with the `or` infix lost that constraint.
2. **Minor** — a `get<CardComponent>() ?: return false` guard that discarded its value became `if (!targetContainer.has<CardComponent>()) return false`.
3. **Minor** — a fully-qualified `com.wingedsheep.engine.handlers.PredicateContext` now uses the existing import.

### Verification

- New `PawpatchRecruitTest` scenario passes (and exercises exactly the reported bug — trigger firing on an animated land plus the "other than that creature" exclusion).
- Becomes-target regression sweep green: Surrak, Elusive Hunter (spell / permanent / non-creature / permanent-only-vs-spell cases — these guard the stack-object base-data fallback), King of the Oathbreakers, Cactarantula, Altanak, Manifold Mouse (Valiant), Elrond, Master of Healing (state predicates), Ward counter tests.
- Full `:rules-engine:test` suite green.

The zone-change/LKI trigger path intentionally keeps `TriggerMatcher`'s own matcher — last-known information is the correct source there, not the current projection.

PR #1278, opened 2026-07-11
fix: auto-tap prefers free colorless over painful colored mana for generic costs

bolav:
## Problem

Auto-tap routed generic mana pips through a mixed source's *painful* colored ability even when the same source has a free `{T}: Add {C}` ability. Casting Surrak with two Starting Towns paying the generic portion cost 2 life for no reason; the same happened with painlands like Battlefield Forge (pain modeled as a self-damage side effect rather than a `PayLife` cost atom).

The solver's existing `hasPainCost`/`painAmount` fields are source-level and only flag a source whose *every* ability pains — a mixed source (free colorless alongside painful colored) looked entirely pain-free.

## Fix

- **`ManaSource.colorPainCost` / `colorlessPainCost`** — per-color (and colorless) life cost from the *cheapest* ability producing that mana, mirroring the existing `colorActivationManaCost` pattern. Built during source analysis.
- **`selfDamageAmount`** — pain modeled as a fixed self-damage effect in the ability's effect chain (Battlefield Forge: `AddMana(RED).then(DealDamage(1, You))`) now counts the same as a `PayLife` cost atom, feeding both the new per-color fields and the existing source-level pain fields.
- **Generic pass** — picks the cheapest production: prefers unrestricted colors, but never pays a color's extra cost (pain *or* activation mana) when a cheaper colorless ability exists. This also fixes filter-land shapes (free `{T}: Add {C}` alongside `{1}, {T}: Add {W}{U}`) wasting an extra tap on generic.
- **Colored / hybrid / `{C}` pip passes** — add a `painPenalty` (15 + pain, matching the pain-land rule in `calculateTapPriority`, with a guard against double-counting `hasPainCost` sources) so pips prefer pain-free sources when alternatives exist, while still using the painful ability when it's the only way to pay.

Execution was already consistent: `ManaAbilitySideEffectExecutor` matches the activated ability by produced color, so colorless production runs the free ability (no side effects) and colored production runs the painful one (side effects + life payment).

## Tests

New in `AutoTapPainAndBonusReproTest` (all green, alongside the existing pain/bonus repros):

- Starting Town ×2 paying `{2}` of a `{2}{G}` spell → free colorless ability, no life loss.
- Battlefield Forge ×2 paying `{2}` → free colorless ability, no damage (effect-modeled pain).
- Battlefield Forge as the only white source for a `{W}` spell → painful ability still used, damage still lands.

Also verified green: `MonocolorHybridManaSolverTest`, `SmartTapTest`, `ConditionalManaAbilityAutoTapTest`, `MultiManaSourceTest` (no pinned auto-tap ordering shifted).

A self-review (per the repo's review-changes flow) is attached as a PR comment.

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/mechanics/mana/ManaSolver.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/mechanics/mana/ManaSolver.kt
@@ -34,6 +34,7 @@ import com.wingedsheep.sdk.scripting.effects.AddDynamicManaEffect
 import com.wingedsheep.sdk.scripting.effects.AddManaEffect
 import com.wingedsheep.sdk.scripting.effects.AddManaOfChoiceEffect
 import com.wingedsheep.sdk.scripting.effects.CompositeEffect
+import com.wingedsheep.sdk.scripting.effects.DealDamageEffect
 import com.wingedsheep.sdk.scripting.effects.Effect
 import com.wingedsheep.sdk.scripting.effects.Gate
 import com.wingedsheep.sdk.scripting.effects.GatedEffect
@@ -48,6 +49,8 @@ import com.wingedsheep.sdk.scripting.DampLandManaProduction
 import com.wingedsheep.sdk.scripting.GrantActivatedAbility
 import com.wingedsheep.sdk.scripting.GameObjectFilter
 import com.wingedsheep.sdk.scripting.predicates.CardPredicate
+import com.wingedsheep.sdk.scripting.references.Player
+import com.wingedsheep.sdk.scripting.targets.EffectTarget
 import com.wingedsheep.engine.state.components.identity.FaceDownComponent
 import com.wingedsheep.sdk.scripting.values.DynamicAmount
 
@@ -122,6 +125,22 @@ data class ManaSource(
      * Colors not present in this map can be produced for free (or not at all).
      */
     val colorActivationManaCost: Map<Color, Int> = emptyMap(),
+    /**
+     * Life required to produce each color, from the *cheapest* ability producing that color.
+     * Covers both pain modeled as a cost atom (Starting Town's "{T}, Pay 1 life: Add one mana
+     * of any color") and pain modeled as a self-damage side effect (Battlefield Forge's
+     * "{T}: Add {R} or {W}. This land deals 1 damage to you."). Colors not in this map are
+     * pain-free. Unlike the source-level [hasPainCost]/[painAmount] pair — which only flags a
+     * source whose *every* ability pains — this lets the solver see that a mixed source (free
+     * "{T}: Add {C}" alongside a painful colored ability) charges life for its colors but not
+     * for colorless, so auto-pay can route generic pips through the free ability.
+     */
+    val colorPainCost: Map<Color, Int> = emptyMap(),
+    /**
+     * Life required to produce colorless via the cheapest colorless-producing ability
+     * (0 = free, the overwhelmingly common case). Only meaningful when [producesColorless].
+     */
+    val colorlessPainCost: Int = 0,
     /**
      * Tapping this source also requires sacrificing it (e.g. Treasure tokens —
      * "{T}, Sacrifice this artifact: Add one mana of any color"). The auto-pay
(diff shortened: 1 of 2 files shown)


bolav:
# Self-review (review-changes flow)

Reviewed the branch in place against `main` (branch is `main` + 1 commit, no merge needed).

## Verdict

The behavior is right and the shape is right. Auto-tap now routes generic pips through a mixed source's free colorless ability instead of its painful (or activation-taxed) colored ability, while colored pips still pay pain when the painful ability is genuinely required. Engine-internal solver work — no new SDK types, no card changes.

## What's good

- **Right precedent, not a new mechanism.** `colorPainCost` / `colorlessPainCost` mirror the existing `colorActivationManaCost` per-color-cheapest-ability pattern on `ManaSource`, and the doc comments explain why the source-level `hasPainCost`/`painAmount` pair can't express a mixed source. `painPenalty` documents and enforces the no-double-count guard against `calculateTapPriority`'s existing pain-land rule.
- **Unifying both pain modelings is the correct scope.** `selfDamageAmount` makes effect-modeled pain (Battlefield Forge's `AddMana.then(DealDamage)`) count the same as a `PayLife` cost atom — and because it feeds `abilityHasPainCost` before the `anyAbilityHasNoPainCost` bookkeeping, it correctly participates in the source-level pain fields too. Side benefit: classic painlands' *colored* production now carries a pain penalty in `findBestSourceForColor`, so colored pips prefer basics over painland colored abilities — previously effect-modeled pain was invisible to the solver entirely.
- **The generic-pass fix is broader than the repro.** Including `colorActivationManaCost` in `coloredExtraCost` means filter-land shapes (free `{T}: Add {C}` alongside `{1}, {T}: Add {W}{U}`) also stop burning an extra tap on generic.
- **Tests cover the interesting axes**: both pain modelings for the generic case, plus the negative control (Battlefield Forge as the only white source must still pain), on top of the pre-existing Starting Town colored-pip pain test. The execution side (`ManaAbilitySideEffectExecutor` matching abilities by produced color) is already consistent with the solver's routing.

## Issues

**Minor**

1. **Hybrid pips: pain influences source choice but not color choice within the source.** At `ManaSolver.kt:582-585` the pain penalty picks *which source* to tap for a hybrid symbol, but at `ManaSolver.kt:590-592` the color used is still `color1` whenever available, ignoring pain. A source producing painful `color1` and free `color2` would pay pain unnecessarily. No implemented card has that asymmetry (Starting Town pains all colors equally; painlands pain both halves), so nothing breaks today — a `minByOrNull(::coloredExtraCost)` over the two candidate colors, like the generic pass now does, would close it. Fine as a follow-up.

2. **Solver assumes cheapest-ability, executor picks first-match.** `colorPainCost` records the *cheapest* pain per color, but `ManaAbilitySideEffectExecutor.kt:97-99` runs the side effects of the *first* mana ability matching the produced color. A card with two same-color abilities — painful one declared first, free one second — would be charged pain the solver priced at 0. This mismatch pre-exists for activation costs and no implemented card has the shape; noted so the assumption is on record.

3. **Colorless restrictions aren't consulted when switching to colorless.** The new `colorToUse = null` path (`ManaSolver.kt:763-768`) assumes the colorless ability is unrestricted-for-this-spell. Colorless restriction tracking is coarse throughout the solver (pre-existing — the `{C}` pip pass doesn't check it either), and no implemented mixed source has a restricted colorless ability alongside a painful color. Theoretical; author's discretion.

Other checks: no CR rule numbers cited anywhere in the diff or commit; no SDK surface changed (so `card-sdk-language-reference.md` correctly untouched); no new cards (no printing-placement check needed).

## Tests — run and verified green

- `AutoTapPainAndBonusReproTest` — all 6 tests (3 new + pre-existing pain/bonus repros).
- `MonocolorHybridManaSolverTest` (4), `SmartTapTest` + `ConditionalManaAbilityAutoTapTest` + `MultiManaSourceTest` (9) — all passed; no pinned auto-tap ordering shifted (including the Mana Confluence five-color+pain case).

## Recommendation

Merge as-is. Optional follow-up: apply the same cheapest-color logic to the hybrid-pip color selection (issue 1) — a three-line change in the same style as the generic pass.

PR #1279, opened 2026-07-11
Fix discover/cascade free-casting of targeted spells

bolav:
## Summary

Fixes discover (CR 701.57a) and cascade (CR 702.85a) free-casting a non-modal **targeted** spell. The resumers synthesized a `CastSpell` with no targets and invoked `CastSpellHandler.execute` directly — bypassing `validate`, which is what normally rejects a targeted spell with `targets = []`. A discovered Zombify therefore went on the stack untargeted and resolved doing nothing.

### Commit 1 — surface the target prompt

- Extracts `CastFromCollectionWithoutPayingCostExecutor`'s target-selection logic into a shared `prepareTargetSelection` (companion, sealed `TargetPrep` result) and calls it from both the cascade and discover may-cast resumers.
- Non-modal targeted spells now pause with the same `ChooseTargetsDecision` the synthesized-free-cast executor uses; modal spells still fall through to `CastSpellHandler`'s own per-mode target flow.
- If a required slot has no legal targets the cast can't initiate (CR 601.2c): the discovered card goes to the controller's hand ("If you don't cast it, put that card into your hand"), the cascade card is bottomed.
- `triggersAlreadyProcessed` is propagated through the targets-continuation resumer so "whenever you cast a spell (from exile)" triggers (Quintorius Kand) aren't double-fired on the new path.

### Commit 2 — self-review follow-ups

- **Free-cast grant leak**: target legality is now checked *before* granting. The grant (`PlayWithoutPayingCostComponent` + `MayPlayPermission`) is zone-agnostic and lives until end-of-turn cleanup, so granting first meant a discovered spell with no legal targets reached the hand still castable for {0} that turn. Applies to both resumers and to the executor's own pre-existing `NoLegalTargets` exit in exile.
- The grant is extracted into a shared `grantFreeCast` / `revokeFreeCast` helper pair (executor companion) instead of three copy-pasted blocks, and every cast-couldn't-initiate fallback revokes the grant it made.
- `CastFromCollectionTargetsContinuation` carries `grantedPermissionId` + `onCastFailure` (`LEAVE` / `HAND` / `BOTTOM_OF_LIBRARY`, defaulted for serialization compatibility) so a cast that fails after target selection revokes exactly its own permission and sends the card to the owning flow's fallback zone instead of stranding it in exile.
- Cascade's direct sync/paused cast paths now propagate `triggersAlreadyProcessed`, matching discover.

## Tests

- New `DiscoverCastTargetedSpellScenarioTest` (rules-engine):
  - Trumpeting Carnosaur's discover 5 hitting Zombify prompts for a target and reanimates it.
  - No legal targets → the discovered card goes to hand **with no leftover free-cast stamp or may-play permission** (the leak assertions fail against commit 1 alone).
  - Cascade (Quandrix, the Proof) prompts for a target through the same shared path.
- Full `just test-rules` and `just test-server` suites pass.
- Three manual scenarios under `manual-scenarios/bugs/` for UI-level verification of the discover/cascade target prompt and the free-cast leak.

No SDK types were added or changed, so no `card-sdk-language-reference.md` update is needed.

### DIFF
--- a/manual-scenarios/bugs/cascade-targeted-spell-target-prompt.json
+++ b/manual-scenarios/bugs/cascade-targeted-spell-target-prompt.json
@@ -1,0 +1,34 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Quandrix, the Proof"],
+    "battlefield": [
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Island"},
+      {"name": "Island"}
+    ],
+    "graveyard": ["Grizzly Bears"],
+    "library": ["Zombify", "Forest", "Forest"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [],
+    "graveyard": [],
+    "library": ["Forest", "Forest"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "SELF"
+}
--- a/manual-scenarios/bugs/discover-no-legal-targets-free-cast-leak.json
+++ b/manual-scenarios/bugs/discover-no-legal-targets-free-cast-leak.json
@@ -1,0 +1,41 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Trumpeting Carnosaur"],
+    "battlefield": [
+      {"name": "Grizzly Bears"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"}
+    ],
+    "graveyard": ["Grizzly Bears"],
+    "library": ["Zombify", "Mountain", "Mountain"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Hill Giant"}
+    ],
+    "graveyard": [],
+    "library": ["Mountain", "Mountain"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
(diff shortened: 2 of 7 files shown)


bolav:
## Code review (`/review-changes`, run pre-merge on the branch)

### Verdict

The core fix is right and well-shaped: extracting `prepareTargetSelection` into the executor's companion and reusing it from the cascade/discover resumers is exactly the composition this codebase wants, and all three cited CR numbers (601.2c, 701.57a, 702.85a) verify against the Comprehensive Rules. The review found one blocking issue — the new "no legal targets" fallback paths leaked the pre-granted free-cast stamp out of exile — which is fixed in the follow-up commit on this branch.

### What's good

- **Right refactor direction**: one shared `prepareTargetSelection` with a small sealed `TargetPrep` result, instead of copy-pasting the 70-line target-decision block into two resumers. The KDoc explaining *why* synthesized casts can't carry targets (no client action; `execute` bypasses `validate`) is genuinely useful.
- **The `triggersAlreadyProcessed` propagation in `resumeCastFromCollectionTargets`** is a real catch — targeted discover/cascade casts now route through that resumer, and without the flag Quintorius-style "cast from exile" triggers would double-fire.
- **Correct continuation ordering** in the discover branch: the `thenEffect` frame is pushed *below* the targets continuation, so Hit the Mother Lode's Treasures still resolve after a paused cast, and the fallbacks route through `runDiscoverThenEffect` so the follow-up isn't lost.
- **Good tests**: real decision-protocol walking (YesNo → ChooseTargets → resolve), assertion clues, and pinning both discover and cascade on the shared path.

### Issues found → resolution

1. **[Blocking] Free-cast grant leaked out of exile on the `NoLegalTargets` fallbacks.** Both resumers granted `PlayWithoutPayingCostComponent` + `MayPlayPermission` *before* checking target legality. On `NoLegalTargets` the card was bottomed / moved to hand still stamped: zone moves never strip the component, `CastZoneResolver.hasPlayWithoutPayingCost` has no zone gate, and cleanup only expires the stamp at end of turn. Concrete exploit: discover 5 hits Zombify with empty graveyards → Zombify to hand; a creature dies in combat that turn → cast Zombify from hand for {0}. **Fixed** in the follow-up commit: legality is checked before granting (also fixing the executor's pre-existing in-exile variant), and every cast-couldn't-initiate fallback revokes its grant. Pinned by new no-leak assertions in the scenario test.
2. **[Important] Cascade `NoLegalTargets` → bottom path untested.** Deliberately left for a follow-up; discover's no-legal-targets path is covered, and the cascade branch shares the same `prepareTargetSelection`/fallback shape.
3. **[Minor] Grant block copy-pasted three times.** **Fixed**: extracted into `grantFreeCast` / `revokeFreeCast` in the executor companion.
4. **[Minor] Stale test comment** ("Annoyed Altisaur" in the Quandrix cascade test). **Fixed.**
5. **[Minor] Pre-existing asymmetry**: cascade's targetless sync/paused cast paths didn't propagate `triggersAlreadyProcessed` while discover's did. **Fixed**: aligned with discover.
6. **[Minor] Unreachable-in-practice fallback gap**: a cast failing in `resumeCastFromCollectionTargets` (targets became illegal between pick and cast) stranded the card in exile. **Fixed**: the continuation now carries `onCastFailure` (discover → hand per CR 701.57a, cascade → bottom per CR 702.85a) and `grantedPermissionId` so exactly its own grant is revoked.

### Verification

- New `DiscoverCastTargetedSpellScenarioTest` passes, including the no-leak assertions (which fail against commit 1 alone).
- Full `just test-rules` and `just test-server` suites pass with zero failures.
- No new SDK types; no card-SDK reference update needed. No new cards, so no printing-placement checks apply.

PR #1280, opened 2026-07-11
fix: AI soft-lock casting Spree spells — gate cast-time mode picks on stacked mana affordability

bolav:
## Summary

Fixes an AI soft-lock when casting a Spree spell it can't fully pay for.

**Repro (from a live game):** the AI cast Unfortunate Accident ({B}, Spree: + {2}{B} destroy / + {1} token) with 4 mana available, chose **both** modes ({3}{B}{B} total — unpayable), picked targets, and the cast dead-ended at payment with `Not enough mana to auto-pay`. The pending `ChooseTargetsDecision` could never be answered legally, `PassPriority` is blocked while a decision is pending, and the AI retried the same response forever:

```
WARN  AI action failed: Not enough mana to auto-pay — trying safe fallbacks
WARN  AI fallback PassPriority also failed: Cannot pass priority while there's a pending decision …
(repeats indefinitely)
```

## Root cause

Cast-time mode selection for choose-N modal spells (`CastSpellHandler.pauseForCastTimeModeSelection` → `presentCastModalModeDecision`) filtered mode options only by **target legality**, never by mana. Per-mode additional costs stack when several modes are chosen (CR 700.2h), and each mode is only affordability-checked *individually* at enumeration time — nobody checked the cumulative combination until payment, which runs after target selection. By then the only legal way out is cancelling, which the AI never does.

## Changes

**rules-engine** — `presentCastModalModeDecision` now offers only modes the caster can still pay for *on top of* the modes already picked (base effective cost + stacked `additionalManaCost`s, validated through the same `validatePayment` path the final cast uses). "Done" remains available once `minChooseCount` is satisfied. Early-returns when no chosen/candidate mode carries a mana cost, so Confluence-style choose-N picks are completely unaffected. Free casts (Cascade-style, which also hit this pause) still owe the stacked additional costs per CR 601.2b/601.2f and are gated on those alone.

**rules-engine (follow-up commit)** — the gate prices picks through the *full* cost pipeline, not a bare `calculateEffectiveCost`. `validate()`'s cost computation (alternative-cost bases — flashback/warp/evoke/harmonize/…, kicker, Or-Pay additional costs, waterbend, airbend fixed-alternative + runtime cost increases, sacrifice-for-reduction, convoke/delve/waterbend alternative-payment reductions, and the harmonize/waterbend payment-X adjustment) is extracted into a shared `computeTotalCastCost` helper used by both `validate()` and the gate. Without this, a mode affordable only through a cost-reduction static or a cheaper alternative cost would have been silently withheld (and vice versa the gate could over-offer). Mode offers now can't diverge from what payment actually charges, and validate's ~230-line pipeline is deduplicated rather than copied a third time.

**game-server** — `GamePlayHandler.safeFallbackActions` now tries `SubmitDecision(CancelDecisionResponse)` first whenever the AI holds the pending decision. A pending decision blocks every priority action, and a rejected decision response is deterministic — re-submitting can't succeed — so cancelling is the only fallback that can unwedge the game. Cancellable decisions (cast-time mode/target pauses) abort cleanly with no side effects; non-cancellable ones reject the cancel without state change and fall through to the existing fallbacks. Belt-and-braces net for any future soft-lock of this class.

## Tests

New `ModalCastTimeModeAffordabilityTest` (4 tests):
- synthetic stacking-cost modal: an unpayable stacked mode is withheld on the next pick; the cast still completes via "Done"
- same card with enough mana: the mode is still offered
- a mode affordable only through a `ModifySpellCost` cost-reduction static is still offered (pins the full-pipeline pricing)
- **Unfortunate Accident end-to-end repro**: 4 mana, destroy mode picked, token mode withheld, target chosen, creature destroyed — no payment dead-end

All existing modal/Spree suites pass: `ModalCastTimeCancelTest`, `ModalCastTimeDeferredTargetsTest`, `ModalPerModeAdditionalCostTest`, `ModalMinChooseCountTest`, `OtjUnfortunateAccidentScenarioTest`, `SmugglersSurpriseScenarioTest`, `ThreeStepsAheadScenarioTest`, `TrashTheTownScenarioTest`, `GreatTrainHeistTest`, `LivelyDirgeScenarioTest`. Full `:rules-engine:test` run green after each commit (6919 passed, 0 failed on the follow-up).

### DIFF
--- a/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/GamePlayHandler.kt
+++ b/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/GamePlayHandler.kt
@@ -1251,6 +1251,24 @@ class GamePlayHandler(
         gameSession: GameSession,
         aiPlayerId: EntityId
     ): List<com.wingedsheep.engine.core.GameAction> {
+        // A pending decision blocks every priority action (pass included), and a decision
+        // response that was just rejected is deterministic — re-submitting it can't succeed.
+        // Cancelling the decision is the only fallback that can unwedge the game: cancellable
+        // decisions (cast-time mode/target pauses) abort cleanly with no side effects, and
+        // non-cancellable ones reject the cancel without state change, falling through to the
+        // remaining fallbacks. Without this, an AI whose decision answer keeps failing (e.g. a
+        // mid-cast pause whose eventual payment is impossible) loops forever.
+        val pendingDecision = gameSession.getStateForTesting()?.pendingDecision
+        val cancelPending = if (pendingDecision?.playerId == aiPlayerId) {
+            listOf<com.wingedsheep.engine.core.GameAction>(
+                com.wingedsheep.engine.core.SubmitDecision(
+                    aiPlayerId,
+                    com.wingedsheep.engine.core.CancelDecisionResponse(pendingDecision.id)
+                )
+            )
+        } else {
+            emptyList()
+        }
         val pass = com.wingedsheep.engine.core.PassPriority(aiPlayerId)
         // If PassPriority is currently a legal action, we're in a priority window (e.g. after
         // blockers/attackers were already declared), not the actual declare-step decision. In that
@@ -1261,7 +1279,7 @@ class GamePlayHandler(
         val passIsLegal = gameSession.getLegalActions(aiPlayerId).any {
             it.action is com.wingedsheep.engine.core.PassPriority
         }
-        return when (gameSession.getStateForTesting()?.step) {
+        return cancelPending + when (gameSession.getStateForTesting()?.step) {
             com.wingedsheep.sdk.core.Step.DECLARE_BLOCKERS -> {
                 val declare = com.wingedsheep.engine.core.DeclareBlockers(aiPlayerId, emptyMap())
                 if (passIsLegal) listOf(pass, declare) else listOf(declare, pass)
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/actions/spell/CastSpellHandler.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/actions/spell/CastSpellHandler.kt
@@ -3317,6 +3317,47 @@ class CastSpellHandler(
     }
 
     /**
+     * Mana-affordability gate for cast-time mode selection: can the caster still pay
+     * the spell's total cost if [chosenIndices] end up being the chosen modes? Chosen
+     * modes' additional mana costs stack (rule 700.2h), so a pick that is affordable
+     * alone can become unpayable combined with earlier picks — and by the time payment
+     * runs (after target selection) the only way out is cancelling the whole cast.
+     *
+     * Uses the plain effective-cost pipeline for the base cost; a "without paying its
+     * mana cost" cast still owes the stacked per-mode additional costs (CR 601.2b,
+     * 601.2f — additional costs apply on top of an alternative cost).
+     */
+    private fun canPayModeSelection(
+        state: GameState,
+        action: CastSpell,
+        modes: List<com.wingedsheep.sdk.scripting.effects.Mode>,
+        chosenIndices: List<Int>
+    ): Boolean {
+        val extraCosts = chosenIndices.mapNotNull { modes.getOrNull(it)?.additionalManaCost }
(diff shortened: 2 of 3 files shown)


bolav:
# Review (`/review-changes`, in-session)

## Verdict

The fix is correct, well-placed, and solves the real production soft-lock: the affordability gate filters mode offers at the decision layer where the dead-end is created, and the AI cancel-fallback is a sound second line of defense. No new SDK types — pure engine/server plumbing, so SDK elegance isn't in question. The one Important finding (gate cost-model divergence) was fixed in the follow-up commit on this branch.

## Verified

- All `ModalCastTimeModeAffordabilityTest` tests pass, and the pre-existing modal suites (`ModalCastTimeModeTest`, `ModalCastTimeCancelTest`, `ModalMinChooseCountTest`, `ModalAllowRepeatTest`, `ModalPerModeAdditionalCostTest`, serialization round-trip, `MiseryCharmTest`, `LongstalkBrawlTest`) stay green. Full `:rules-engine:test`: 6919 passed, 0 failed. `:game-server:compileKotlin` succeeds.
- Rule citations check out against CR 2026-04-17: 700.2h (chosen modes' additional costs stack), 601.2f (total cost = mana or alternative cost *plus* all additional costs — so a free cast still owes mode costs).
- Unfortunate Accident's definition matches Oracle Spree text; the test reproduces the production wedge at the exact mana boundary.

## What's good

- **Right layer for the gate.** Filtering `offerIndices` in `presentCastModalModeDecision` fixes both the initial pause and every iterative re-entry through `CastModalContinuationResumer` with one predicate, and the continuation's `offeredIndices` index-mapping stays consistent automatically.
- **`validatePayment` reuse** gives the gate payment-strategy awareness for free (solver over untapped sources for AutoPay, floating pool for FromPool, chosen-sources logic for Explicit), so it won't hide modes from a player with untapped lands.
- **The AI fallback is safe by construction**: `CancelDecisionResponse` is only accepted by the `ChooseTargets`/`ChooseOption` validators when `canCancel` is set; everything else rejects it without state change and falls through to the existing fallbacks. Trying cancel before pass is correct since a pending decision blocks priority actions anyway.
- **Tests cover both directions** (mode withheld when the stacked cost is unpayable; same pick still offered with one more mana) plus the real card end-to-end through target selection to resolution — the boundary assertion ({2}{B}{B} exactly affordable, {3}{B}{B} not) is the interesting axis.

## Issues

### Important — **fixed in follow-up commit**

1. **Gate cost model diverged from the real total-cost pipeline.** The original gate rebuilt the base cost with a plain `calculateEffectiveCost`, ignoring alternative-cost bases (flashback/warp/evoke/harmonize/…), kicker, Or-Pay additional costs, convoke/delve/waterbend alternative payments, airbend fixed-alternative, and runtime cost increases. The too-strict direction was a new regression: a modal spell cast via a cheaper alternative cost or under a cost-reduction static would have modes silently withheld that the player can legally pay for. **Fixed:** `validate()`'s cost computation is extracted into a shared `computeTotalCastCost` (also returning the harmonize/waterbend-adjusted payment X) and the gate now prices picks through it; a new test pins the cost-reduction case.

### Minor (follow-ups, author's discretion)

2. **Retroactive dead-end for `minChooseCount ≥ 2`** (`presentCastModalModeDecision` error branch): if pick 1 is accepted but no pick-2 candidate is affordable and the minimum isn't met, the already-submitted choice is rejected after the fact and the decision stays pending until cancelled. No current card has min ≥ 2 with per-mode mana costs (Spree is min 1), so this is theoretical.
3. **The fallback comment slightly overstates cancel coverage** (`GamePlayHandler.safeFallbackActions`): the additional-cost pause (`CastSpellAdditionalCostContinuation`) uses a `SelectCardsDecision`, and `validateSelectCards` rejects `CancelDecisionResponse` outright (the type has no `canCancel` field), so that resumer's cancel branch is unreachable via `SubmitDecision` and that wedge family stays unwedgeable. Pre-existing gap — adding `canCancel` to `SelectCardsDecision` would close it.
4. **Pre-existing payment-time ordering quirk** (`CastSpellHandler.execute`): per-mode costs are added to `effectiveCost` *before* the airbend fixed-alternative branch replaces it wholesale, so an airbended modal cast would silently drop the stacked mode costs — contradicting CR 601.2f. Execute still has its own copy of the cost pipeline; converging it onto `computeTotalCastCost` is the natural next deduplication (out of scope here since execute interleaves actual alternative-payment application).

PR #1281, opened 2026-07-11
Fix enters-with-counter cards using a trigger instead of a replacement effect

bolav:
## Summary

"This creature **enters with** a +1/+1 counter on it if \<condition\>" is a **replacement effect (CR 614.1c)**, not an ETB trigger: nothing goes on the stack, there is no response window, and the counters (and any "and with \<keyword\>" rider) are present from the moment the permanent enters. Two groups of cards implemented this as a condition-gated ETB trigger instead:

- **3 conditional enters-with-counter cards** — Goblin Boarders and War-Name Aspirant (Raid), Cindering Cutthroat (opponent lost life). With the trigger version, the creature entered as its base size and opponents could respond before the counter arrived.
- **17 Invasion/Dominaria kicker cards** — "If this creature was kicked, it enters with N +1/+1 counters on it [and with \<keyword\>]". A kicked Kavu Titan was killable as a 2/2 in response to its own trigger, and Torpor-Orb-style effects would wrongly suppress the counters.

## Changes

**Commit 1 — conditional (Raid / lost-life) cards**

Convert Goblin Boarders, War-Name Aspirant, and Cindering Cutthroat to the existing `EntersWithCounters(condition = …, selfOnly = true)` replacement, following the Frilled Sparkshooter precedent. No SDK changes.

**Commit 2 — kicker cards, plus the SDK/engine capability the keyword riders needed**

- **New SDK replacement effect `EntersWithKeywords(keywords, condition?, selfOnly?, appliesTo?)`** — the keyword counterpart of `EntersWithCounters`. Keywords are granted as an entry-timestamped, permanent Layer-6 floating effect: a later "loses all abilities" removes it, it does not re-apply if stripped, and it is cleaned up when the permanent leaves the battlefield (CR 400.7). Applied on both entry paths (stack resolution and non-stack zone moves) and by the global battlefield scan, mirroring `EntersWithCounters` semantics including `selfOnly`. Reuses the existing `KeywordGrantedEvent` — no new client/server surface.
- **Renamed `EntersWithCountersHelper` → `EntersWithReplacements`** (it now applies counters *and* keywords, matching the `EnterTappedReplacements` sibling naming) and collapsed `StackResolver`'s duplicated self-path into the shared `applyFromDefinition` core.
- **12 counter-only cards** → `EntersWithCounters(condition = WasKicked)`, following the Gnarlid Colony precedent: Ardent Soldier, Kavu Aggressor, Llanowar Elite, Pincer Spider, Prison Barricade, Urborg Skeleton, Vodalian Serpent (INV); Academy Drake, Baloth Gorger, Grunn the Lonely King, Stronghold Confessor, Untamed Kavu (DOM).
- **5 keyword-rider cards** → `EntersWithCounters` + `EntersWithKeywords`: Benalish Lancer (first strike), Duskwalker (fear), Faerie Squadron (flying), Kavu Titan (trample), Pouncing Kavu (haste).
- SDK reference documents the new type; the mtgish bridge note explains why the emitter still declines `EntersWithLayerEffect` (every corpus card with that tag carries SetPT / activated-grant / until-EOT riders beyond the plain-keyword shape — no calibrated card to render).

`WasKicked` needed no engine change: the KICKED cast-choice is stamped on the spell entity before the entry replacements apply, and a token copy or reanimated body (never kicked) correctly gets nothing.

## Tests

- `GoblinBoardersScenarioTest` — Raid satisfied: enters as a 4/3 with the counter already on it and **an empty stack** (the discriminating assertions the trigger version fails); no attack: plain 3/2.
- `KickedEntersWithReplacementScenarioTest` — kicked Kavu Titan enters as a 5/5 with trample and an empty stack; unkicked baseline 2/2; kicked Pouncing Kavu has haste from the moment it enters and attacks the same turn; bounced and recast without kicker enters as a fresh 2/2 (CR 400.7 new-object reset).
- Full `rules-engine` suite, `mtg-sdk` suite, `CardDefinitionSnapshotTest` (BLB/FDN/KTK/INV/DOM re-blessed — the diff is exactly the 20 trigger→replacement conversions) and `CardLintTest` all green.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -6225,6 +6225,19 @@ replacementEffect {
     `DistinctColorsManaSpent` reads the new creature's own cast (and a token / reanimated creature that
     wasn't cast spent no mana → 0). Player-scoped counts (`TurnTracking(Player.You)`, Gev) still read
     the replacement source's controller.
+- `EntersWithKeywords(keywords, condition?, selfOnly?, appliesTo?)` — "[permanent] enters with
+  [keywords]" (CR 614.1c), the keyword counterpart of `EntersWithCounters`. The grant happens as the
+  permanent enters — no trigger, no stack, no response window — as a permanent, entry-timestamped
+  Layer-6 floating effect: a later "loses all abilities" removes it, it does not re-apply if stripped,
+  and it is cleaned up when the permanent leaves the battlefield (new object, CR 400.7). Kicker riders
+  are the canonical use (Kavu Titan "If this creature was kicked, it enters with three +1/+1 counters
+  on it and with trample" = an `EntersWithCounters` **plus** an `EntersWithKeywords`, both
+  `selfOnly = true, condition = WasKicked`; also Benalish Lancer / Duskwalker / Faerie Squadron /
+  Pouncing Kavu). `condition` is evaluated at the moment of entry against the entering permanent
+  (cast-choice conditions like `WasKicked` read the durable cast-choices bag, so a token copy or
+  reanimated body — never kicked — correctly gets nothing). Like `EntersWithCounters`, a
+  non-`selfOnly` instance stamped on a battlefield permanent applies to *other* permanents matching
+  `appliesTo` as they enter.
 - `EntersWithDevour(multiplier, sacrificeFilter, counterType, variant)` — Devour (CR 702.82) and its
   printed variants. As the permanent resolves from the stack, the controller is prompted to pick any
   number of their own permanents matching `sacrificeFilter`. Those permanents are sacrificed and the
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/ReplacementEffect.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/ReplacementEffect.kt
@@ -443,6 +443,50 @@ data class EntersWithDynamicCounters(
     }
 }
 
+/**
+ * Permanent enters the battlefield with [keywords] (CR 614.1c) — the keyword counterpart of
+ * [EntersWithCounters]. Example: Kavu Titan "If this creature was kicked, it enters with three
+ * +1/+1 counters on it and with trample" — an [EntersWithCounters] plus an [EntersWithKeywords],
+ * both gated on the same [condition].
+ *
+ * The grant happens as the permanent enters: no trigger, no stack, no response window. It is
+ * entry-timestamped for Rule 613 layer ordering (a later "loses all abilities" effect removes
+ * it) and lasts as long as the permanent remains on the battlefield — it is cleaned up when the
+ * permanent leaves (a new object per CR 400.7) and does NOT re-apply if the keyword is removed.
+ *
+ * @param keywords The keywords the entering permanent has from the moment it enters.
+ * @param condition When non-null, the keywords are only granted if this condition evaluates true
+ *                  at the moment the permanent enters (e.g. [conditions.WasKicked], read from the
+ *                  durable cast-choices bag).
+ * @param selfOnly When true, only applies to the permanent carrying this effect as it enters,
+ *                 never to other permanents matching [appliesTo] (mirrors [EntersWithCounters]).
+ */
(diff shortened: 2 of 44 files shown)


bolav:
# Review (`/review-changes`)

Reviewed in place with `main` already merged in (the branch is linear on top of it). Both commits reviewed; the second commit implements this review's own follow-up finding, so it is covered here as well.

## Verdict

Correct fix, correct SDK shape. Twenty "enters with a +1/+1 counter if …" cards move from an ETB triggered ability to `EntersWithCounters(condition = …)` replacement effects — which is what CR 614.1c says they are, and which changes observable gameplay (the creature has its counters/keywords the moment it enters; nothing goes on the stack for opponents to respond to). The one new SDK type, `EntersWithKeywords`, is genuinely novel, parameterized for the next card, and named for the mechanic.

## What's good

- **Minimal new SDK surface.** Commit 1 is pure adoption of the existing `EntersWithCounters.condition` primitive (Frilled Sparkshooter precedent); the 12 counter-only kicker cards in commit 2 follow the Gnarlid Colony precedent. `selfOnly = true` is set everywhere it must be — without it the default `appliesTo` filter would let the stamped replacement add counters to *other* entering creatures via the global scan.
- **`EntersWithKeywords` passes the reuse test.** No existing primitive grants a keyword at entry time (`OnEnterRunEffect` is land-only and unconditioned; a counter-gated static would lie about semantics). It is parameterized over keyword list, condition, and filter; the engine grants it as an entry-timestamped permanent Layer-6 floating effect, so "loses all abilities" removes it (correct per rulings) and `removeFloatingEffectsTargeting` cleans it up on leave (CR 400.7). It reuses `KeywordGrantedEvent`, so no new client/server wiring is needed.
- **The helper refactor removes real duplication.** `StackResolver`'s ~60-line copy of the self-path now delegates to `EntersWithReplacements.applyFromDefinition`, and the rename fixes a name that would otherwise lie (a "CountersHelper" applying keywords).
- **Rule citations verified.** CR 614.1c ("Effects that read '[This permanent] enters with…' are replacement effects") and CR 400.7 both checked against the Comprehensive Rules; Oracle text for all 20 cards verified against Scryfall — including that the remaining Raid/kicker cards left on triggers (Mardu Skullhunter, Wingmate Roc, Josu Vess, Verix Bladewing, …) are genuine "When … enters" triggers, correctly untouched. The four BLB Offspring cards that grep-matched are false positives (their triggers are correct).
- **Tests discriminate trigger from replacement.** Both scenario tests assert `stackSize == 0` *plus* projected P/T/keywords immediately after resolution — assertions the old trigger implementations fail. Edge coverage includes the unkicked/no-raid baselines, haste being usable the same turn, and the CR 400.7 bounce-and-recast reset proving the floating grant and cast-choices bag don't survive the zone round trip.
- **`WasKicked` timing verified in the engine**: the KICKED cast-choice is stamped into the durable bag (StackResolver) *before* `applyEntersWithReplacements` runs, and non-cast entries (reanimation, token copies) correctly evaluate it false.

All verified green by running them (not trusting the description): full `rules-engine` suite, `mtg-sdk` suite, the new scenario tests, `CardDefinitionSnapshotTest` (the BLB/FDN/KTK/INV/DOM re-bless diff is exactly the 20 conversions plus serialization reordering), and `CardLintTest`.

## Issues

**Blocking** — none.

**Important** — none.

**Minor**

1. **The `OpponentLostLifeThisTurn` variant has no scenario test.** Cindering Cutthroat (and the pre-existing Frilled Sparkshooter) are exercised only by snapshots; the machinery is shared with the tested `YouAttackedThisTurn`/`WasKicked` paths, so this is discretionary — but a positive case would pin the "an opponent lost life" condition-context semantics for the second condition family.
2. **mtgish emitter deliberately declines the new capability.** All four corpus cards tagged `EntersWithLayerEffect` carry riders (SetPT composites, activated-ability grants, until-EOT) beyond the plain-keyword shape, so there is no calibrated card to render and the bridge note documents the decline. Revisit if a keyword-only corpus card appears.

## Recommendation

Merge as-is. Optionally add one scenario test for the `OpponentLostLifeThisTurn` variant as a follow-up.

PR #1282, opened 2026-07-12
Implement VOW cards with 1:1 scenario tests

rwaldron:
Bring Innistrad: Crimson Vow (VOW) to 109/272 implemented cards. Adds the new card definitions, registers them in InnistradCrimsonVowSet.kt, and pairs every non-reprint / non-basic-land card with a Kotest scenario test — the set now has verified 1:1 card-to-scenario-test coverage.

- New VOW card definitions (creatures, spells, artifacts, duals) plus the Hero's Downfall / Sure Strike / Valorous Stance reprints and basic lands.
- ~97 *ScenarioTest.kt classes in rules-engine covering ETB/dies triggers, modal spells, activated abilities, and combat interactions.
- Fix Arm the Cathars: its two "up to one other target creature" slots shared one binding name, so the name-keyed target map (EffectContext.buildNamedTargets) collapsed them last-write-wins and the middle creature got no bonus. Rename the slots to distinct names so each optional target resolves independently.
- Re-bless the VOW card-definition snapshot golden after the fix.
- Refresh the VOW backlog checklist + mechanics notes to match implemented state.

### DIFF
--- a/backlog/sets/innistrad-crimson-vow/cards.md
+++ b/backlog/sets/innistrad-crimson-vow/cards.md
@@ -1,0 +1,277 @@
+# Innistrad: Crimson Vow (VOW) - Card Checklist
+
+**Set Size:** 272 booster cards (excluding basic lands and tokens)
+**Release Date:** November 19, 2021
+**Implemented:** 109 / 272
+- [x] Abrade
+- [x] Adamant Will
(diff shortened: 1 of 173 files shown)


rwaldron:
Claude is working on the failure now

PR #1283, opened 2026-07-12
Fix warp end-step exile hitting a blinked (new-object) permanent

bolav:
@wingedsheep please check if you think this is the right implementation for this problem.

## Problem

A permanent cast for its warp cost that left the battlefield and returned before the end step (blink — e.g. **Daydream** on **Starfield Shepherd**) was still exiled by warp's "exile it at the beginning of the next end step" delayed trigger. Per CR 603.7c / 400.7 the returned permanent is a new object the delayed trigger no longer tracks, so it must stay on the battlefield permanently. Entity ids survive zone round-trips in this engine, so the trigger could still find the blinked permanent by id.

## Fix

An object-identity stamp for battlefield entries:

- **`BattlefieldEntryTimestampComponent`** — stamped in `PermanentEntryTracker.record`, the chokepoint both entry pipelines (`BattlefieldEntry.place` and `ZoneTransitionService.moveToZone`) funnel through; stripped on leave alongside the other CR 400.7 "new object" components. `record` ticks the global timestamp after stamping: resolutions themselves don't advance the timestamp (only puts on the stack, land plays, and similar actions do), so two back-to-back resolutions could otherwise share a value — stamp-then-tick makes every entry stamp unique unconditionally.
- **`WarpExileEffect.enteredBattlefieldTimestamp`** (nullable, default `null` so previously serialized states keep working) — snapshotted when the delayed trigger is created, in both creation paths (`StackResolver`'s warp path and `CreateDelayedTriggerExecutor`'s generic path), and compared against the permanent's current stamp at resolution. A mismatch means the entity re-entered as a new object and the exile does nothing.
- `TimestampComponent` (Rule 613.7 layer ordering) and the new component describe the same notional moment; their KDocs now cross-reference each other and document why they stay separate (stamping the layer one on every entry would change layer ordering engine-wide).

## Tests

- `WarpMechanicTest` — new driver-level regression: a warped creature blinked before the end step survives the trigger. The blink card is composed inline from existing `Effects.Move` primitives; the test fails without the fix.
- `StarfieldShepherdDaydreamScenarioTest` — full scenario with the real EOE/SOS cards.
- `PermanentEntryTrackerStampTest` — pins the stamp-uniqueness invariant: distinct entries always get distinct stamps, even with no cast/land-play tick in between.
- Manual repro scenario: `manual-scenarios/bugs/warp-blink-daydream-not-exiled.json`.

Full `:rules-engine:test` suite green locally (includes the engine-wide stamp-then-tick change).

A self-review (per the repo's review-changes flow) is attached as a PR comment; its findings are addressed in the second commit.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -6605,7 +6605,10 @@ Card authors rarely reference these directly; they are created/updated by the ma
   when only one (or neither) is payable; the unpayable side appears grayed out (CR 118.9a, the caster
   chooses which cost to use). The warp action is enumerated by `CastFromZoneEnumerator`, which also
   emits the grayed-out normal-cast placeholder when the normal cost is unaffordable (mirroring
-  `MorphCastEnumerator`).
+  `MorphCastEnumerator`). The end-step exile is a delayed trigger whose `WarpExileEffect` snapshots
+  the permanent's battlefield-entry timestamp (`enteredBattlefieldTimestamp`); at resolution it only
+  exiles the *same object* — a warped permanent that left the battlefield and returned before the
+  end step (blink, e.g. Daydream) is a new object (CR 603.7c / 400.7) and stays permanently.
 - **Evoke** — `evoke = "{U}"`; pay alt cost, sacrifice on ETB.
 - **Sneak** — `sneak("{1}{U}")`; declare-blockers-step alt cost (pay mana + return an unblocked attacker you control to hand); a resolving permanent enters tapped and attacking the same defender. `Conditions.SneakCostWasPaid` reads the rider flag.
 - **Ninjutsu** — `ninjutsu("{1}{U}{B}")`; the canonical CR 702.49 keyword that **Sneak** reflavors. Same declare-blockers alt cost and tapped-and-attacking entry, shared via `KeywordAbility.ninjutsuStyleCost`. *Kaito, Bane of Nightmares* (DSK).
--- a/manual-scenarios/bugs/warp-blink-daydream-not-exiled.json
+++ b/manual-scenarios/bugs/warp-blink-daydream-not-exiled.json
@@ -1,0 +1,30 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Starfield Shepherd", "Daydream"],
+    "battlefield": [
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Plains"}
+    ],
+    "graveyard": [],
+    "library": ["Plains", "Plains"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [],
+    "graveyard": [],
+    "library": ["Plains", "Plains"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "SELF"
+}
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/effects/RemovalEffects.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/effects/RemovalEffects.kt
@@ -545,11 +545,18 @@ data class PutOnLibraryPositionOfChoiceEffect(
  * from exile for its warp cost on a later turn.
  *
  * @property target The permanent to exile (resolved to SpecificEntity by delayed trigger creation)
+ * @property enteredBattlefieldTimestamp The tracked permanent's battlefield-entry timestamp,
+ *   snapshotted when the delayed trigger is created. At resolution the executor exiles the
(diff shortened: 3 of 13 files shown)


bolav:
# Self-review: `fix-warp-exile-blinked-new-object`

Review of the first commit (`9ffa13d`), per the repo's review-changes flow. All findings below are addressed in the follow-up commit (`f1472c7`) — resolution notes at the end.

## Verdict

The behavior is right and the shape is mostly right: a generic object-identity stamp at the single battlefield-entry chokepoint, compared by the warp executor. Both CR citations are accurate — verified against the local Comprehensive Rules; CR 603.7c's own example is literally the warp-exile wording, and 400.7 is the new-object rule. Two things should be tightened before merge: a comment that states a false engine invariant (with a real, if contrived, soundness corner behind it), and a second `WarpExileEffect`-creation path that silently loses the guard.

## What's good

- **Right chokepoint.** Stamping in `PermanentEntryTracker.record` covers both entry pipelines (`BattlefieldEntry.place` and `ZoneTransitionService.moveToZone`), and the strip in `ZoneMovementUtils` sits exactly among its CR 400.7 siblings. Phasing was checked: it's a marker component with no zone movement, so a phased-out-and-in permanent correctly keeps its stamp (same object, CR 702.26).
- **Fail-safe null semantics.** Old serialized states and unstamped entries skip the check and keep prior behavior; a re-entry that misses `record` (the known bypass-bug class) yields `null ≠ snapshot` → no exile, which is the *correct* outcome for a new object.
- **Serialization done right**: nullable field with a default (backward-compatible decode), component registered in `engineSerializersModule`.
- **Two-level tests**: a driver-level test with a minimal inline blink card composed from `Effects.Move` (no new SDK surface for the test), plus a full scenario with the real EOE/SOS cards, plus the manual repro JSON. Docs updated per the repo hard rule.

## Issues

### Blocking

None.

### Important

1. **`PermanentEntryTracker.kt` — the comment's uniqueness argument is false, and there's a corner where it matters.** It claims "the global timestamp ticks at the end of every resolution." It doesn't: `tick()` is called only when objects are *put on the stack* (`castSpell`, `putTriggeredAbility`, `putSpellCopy`, `putActivatedAbility`) and on land plays/special actions — resolutions never tick. The guarantee actually holds because any blink of the warped permanent requires a stack object cast/put *after* it entered, and that put ticks. But two *consecutive resolutions* can share a timestamp: cast a mass-blink sorcery, hold priority, flash in the warp creature (e.g. Vedalken Orrery) — the creature resolves at stamp T, the mass blink resolves next also at T, and the re-entered new object collides with the snapshot and gets wrongly exiled. Fix the comment to state the true invariant; for robustness, have `record` stamp-then-`tick()` so every entry gets a unique stamp unconditionally.

2. **`CreateDelayedTriggerExecutor.kt` — the generic delayed-trigger path creates `WarpExileEffect` without a snapshot.** It resolves the target to `SpecificEntity` but leaves `enteredBattlefieldTimestamp` null, so any card that ever builds a warp exile via `CreateDelayedTrigger` silently loses the new-object guard (null skips the check). No current card uses it, but the branch exists and will be the path a future "warp another permanent" card takes. Snapshot the entity's `BattlefieldEntryTimestampComponent` there too.

3. **`BattlefieldComponents.kt` — relationship to `TimestampComponent` is undocumented, and they're conceptually the same fact.** The new component sits directly below `TimestampComponent` ("Timestamp for ordering effects (Rule 613)"), which per CR 613.7 *is* the permanent's battlefield-entry time — and which, notably, is never written anywhere in main code (only read with a `?: state.timestamp` fallback in `StateProjector` and stripped on leave). Keeping them separate is defensible — actually stamping `TimestampComponent` would change layer-ordering behavior engine-wide, way beyond this bugfix — but the KDoc should say so explicitly, and there's a follow-up opportunity: one stamped component could serve both purposes *and* fix the currently degenerate 613.7 timestamp ordering.

### Minor

- **Fully-qualified names in code bodies** (`WarpExileExecutor`): the file imports everything else normally — import the component instead of the inline FQN.
- **Scope note, not a defect:** CR 603.7c applies to every delayed trigger tracking a specific object (sacrifice-it-at-end-step, return-it, token-copy triggers — all present in `CreateDelayedTriggerExecutor`), and they all still have this bug class. The primitive introduced here is the right general building block; wiring only warp is fine for this fix, but it's worth a tracked follow-up to hoist the stamp/compare into `DelayedTriggeredAbility` itself rather than per-effect fields.

---

## Resolution (second commit)

- **Important 1 — fixed.** `record` now stamps then ticks; the comment states the real invariant (resolutions don't tick — puts on the stack and land plays do). New `PermanentEntryTrackerStampTest` pins uniqueness directly: distinct entries and same-entity re-entries get distinct stamps even with no cast/land-play tick in between. `state.timestamp` consumers were audited before adding the tick — all are capture-into-stored-value sites (floating effects, permissions, shields) whose relative orderings a strictly-more-often-ticking counter preserves.
- **Important 2 — fixed.** `CreateDelayedTriggerExecutor`'s `WarpExileEffect` branch snapshots the entry stamp when resolving the target, mirroring the `StackResolver` path.
- **Important 3 — fixed.** Both KDocs now cross-reference each other and document why the two stamps stay separate.
- **Minor (FQN) — fixed**; the follow-up on generalizing the guard to other object-tracking delayed triggers remains open by design.
- **Verification:** full `:rules-engine:test` suite green after the stamp-then-tick change (engine-wide behavior), including `WarpMechanicTest` (15/15), `StarfieldShepherdDaydreamScenarioTest` (1/1), `PermanentEntryTrackerStampTest` (2/2).

PR #1286, opened 2026-07-13
Implement Exploit keyword (CR 702.110) + three VOW cards

rwaldron:
## Summary

Implements the **Exploit** keyword (CR 702.110, design home *Dragons of Tarkir*; reprinted in
*Innistrad: Crimson Vow*) end-to-end — SDK, engine, and client — plus the three VOW reference cards
that use it (paper printings, **not** the Alchemy A- rebalances).

Exploit is a **triggered ability**: "When this creature enters, you may sacrifice a creature"
(CR 702.110a). A creature "exploits a creature" when its controller sacrifices a creature as that
ability resolves (CR 702.110b) — and the sacrificed creature may be the exploiter itself, since the
sacrifice is scoped only to the controller.

Rather than add a bespoke executor, the keyword is **composed from existing primitives**. The
`CardBuilder.exploit(onExploit, onExploitTargets)` helper installs an ETB `ReflexiveTriggerEffect`
whose *action* is "you may sacrifice a creature" and whose *reflexive* half is the "when this
creature exploits a creature, …" payoff:

- **Untargeted / self-bound payoffs** (Stitched Assistant) are baked straight into the reflexive.
- **Targeted payoffs** (Fell Stinger) ride the reflexive with `onExploitTargets`, so the target is
  chosen *after* the sacrifice resolves — matching "target player …".
- **Broadcast payoffs** (Skull Skaab) declare their own `ExploitedEvent` watcher instead, keyed on
  the observable event the reflexive emits on a real sacrifice.

Because the reflexive is established while the source is on the battlefield, self-bound payoffs
survive the exploiter **sacrificing itself** with no gone-source detection. A separate watcher on the
exploiter (broadcast case) is caught by a sacrifice look-back pass (CR 603.10a — abilities that
trigger when a player sacrifices a permanent look back in time).

### SDK (`mtg-sdk`)
- `Keyword.EXPLOIT`.
- `EventPattern.ExploitedEvent` — an observable trigger pattern with `player` (controller scope) and
  `requireNontokenExploited` filters; matched via `TriggerBinding` SELF / OTHER / ANY.
- `EmitExploitedEvent` effect (`RemovalEffects.kt`) — fires the observable event, reading the
  sacrificed creature's last-known info threaded in from the sacrifice.
- `CardBuilder.exploit(onExploit, onExploitTargets)` DSL helper (`dsl/mechanics/ExploitDsl.kt`) —
  composes the ETB reflexive from `SacrificeEffect` + `EmitExploitedEvent`, with the payoff on the
  reflexive half.

### Engine (`rules-engine`)
- `GameEvent.ExploitedEvent` (`exploiterId`, `exploiterControllerId`, `sacrificedId`,
  `sacrificedWasToken`, `sacrificedName`, `sourceName`) + serialization registration.
- `SacrificeExecutor` snapshots the sacrificed creature's token-ness before it leaves and emits
  `ExploitedEvent` **only on a real sacrifice** (declining the optional sacrifice emits nothing);
  `EntitySnapshot` carries the token-ness so `EmitExploitedEventExecutor` can populate the event.
- `TriggerMatcher` / `TriggerIndex` match the new event (SELF/OTHER/ANY binding, controller-player
  scope, `requireNontokenExploited` filter).
- `TriggerDetector.detectExploitedSelfSacrificeTriggers` — a look-back pass (CR 603.10a) so a
  self-exploiting creature's own watcher still fires after it has moved to the graveyard, guarded
  against double-firing with the battlefield index scan.
- `SacrificeAndPayContinuationResumer` wiring for the exploit reflexive path.

### Client (`web-client`)
- `Keyword.EXPLOIT` enum + display name (`enums.ts`); EXPLOIT ability badge (`keywordManaClass` +
  `displayableKeywords`) using the mana-font `ms-ability-exploit` glyph.

### Cards (`mtg-sets`)
| Card | Cost | P/T / Type | Exploit payoff |
|------|------|------------|----------------|
| **Stitched Assistant** | {2}{U} | 3/2 Zombie | Scry 1, then draw a card (self-bound, untargeted — baked into the reflexive). |
| **Fell Stinger** | {2}{B} | 3/2 Zombie Scorpion | Deathtouch; target player draws two cards and loses 2 life (self-bound, targeted — target chosen after the sacrifice). |
| **Skull Skaab** | {U}{B} | 2/2 Zombie | Whenever a creature you control exploits a **nontoken** creature, create a 2/2 black Zombie token (broadcast `ExploitedEvent` watcher, `player = You`, `requireNontokenExploited = true`). |

All three are canonical in VOW (their earliest real printing); later reprints (DBL/PLST/PIO/PRM) are
correctly left without rows since those sets aren't scaffolded.

### Tests
- Consolidated scenario coverage in `rules-engine`
  (`com.wingedsheep.engine.scenarios.ExploitTest`, 8 tests), a coverage matrix over the shared
  mechanic rather than per-card files:
  - declining the optional sacrifice fires no payoff and emits no `ExploitedEvent`;
  - Stitched Assistant scries + draws sacrificing another creature (net +1 card) **and** sacrificing
    itself;
  - Fell Stinger's targeted payoff draws two and drains the chosen player, **and** still fires when
    it sacrifices itself;
  - Skull Skaab makes a Zombie when another exploiter sacrifices a nontoken creature, makes **no**
    Zombie when the exploited creature was a token, and fires its watcher when it exploits itself
    (a nontoken).
- `VOW.json` card snapshot re-blessed — a clean pure-addition diff (the three new cards only).

### Docs
- `docs/card-sdk-language-reference.md`: new **Exploit** keyword section (the reflexive composition
  model, `ExploitedEvent` / `EmitExploitedEvent`, `exploit(onExploit, onExploitTargets)`, and the
  self-exploit look-back under CR 603.10a).

## Verification
- `just build` — green across all modules.
- All 8 Exploit scenario tests pass; `CardDefinitionSnapshotTest` and `CardLintTest` pass against the
  re-blessed golden; `npm run typecheck` clean.


rwaldron:
@wingedsheep claude is resolving this now

PR #1287, opened 2026-07-13
Add 5 LCI Craft DFCs (batch 2)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1285

Second batch of LCI Craft transform-DFCs (5 cards, one per commit, scenario test each). Stacked on `lci-craft-01` — **review only the diff vs that branch**; the commits below `lci-craft-01`'s tip belong to that PR.

| Card | Front | Back |
|---|---|---|
| Spring-Loaded Sawblades // Bladewheel Chariot | Flash, ETB 5 dmg to tapped opp creature; Craft with artifact | Vehicle 5/5; tap-two-other-artifacts animate + Crew 1 |
| Lodestone Needle // Guidestone Compass | Flash, ETB tap + 2 stun counters (up to one target); Craft with artifact | {1},{T}: target creature you control explores (sorcery) |
| Waterlogged Hulk // Watertight Gondola | {T}: mill 1; Craft with Island | Vehicle 4/4 vigilance, Descend 8 unblockable, Crew 1 |
| Idol of the Deep King // Sovereign's Macuahuitl | Flash, ETB 2 dmg any target; Craft with artifact | Equipment: ETB auto-attach, +2/+0, Equip {2} |
| Master's Guide-Mural // Master's Manufactory | ETB 4/4 WU Golem token; Craft with artifact | {T}: Golem token, only if an artifact ETB'd your side this turn |

Notes:
- **No new SDK types** — every card composes existing primitives (`craft(...)` DSL, `Effects.DealDamage/Composite/Tap/AttachEquipment/CreateToken/Explore`, `Patterns.Library.mill`, `Costs.TapPermanents(excludeSelf = true)`, `ConditionalStaticAbility(CantBeBlocked, CardsInGraveyardMatchingAtLeast)`, `ActivationRestriction.OnlyIfCondition(ArtifactEnteredBattlefieldThisTurn)`, `ModifyStats`, `equipAbility`, `KeywordAbility.crew`).
- All five crafts are exact-one-material, using the `maxCount` support introduced in the `lci-craft-01` stack. Waterlogged Hulk crafts with an **Island** (subtype filter; materials from battlefield or graveyard).
- Master's Manufactory's activation gate ("this artifact or another artifact entered ... this turn") reads the permanent-entry tracker, which records the craft return itself — so the {T} ability works the same turn it's crafted, and the scenario test pins that.
[...]

PR #1288, opened 2026-07-13
Add 5 LCI Craft DFCs (batch 3) + crafted-material amounts + while-tapped activation lock

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1287

Third batch of LCI Craft transform-DFCs (5 cards, one per commit, scenario test each) plus the two engine features they need. Stacked on `lci-craft-02` — review the diff vs that branch.

## Engine/SDK features (first two commits)

**Crafted-material dynamic amounts + per-color mana.** `DynamicAmount.CraftedMaterialsTotalManaValue` and `CraftedMaterialsColorCount` mirror the existing `CraftedMaterialsTotalPower` over `CraftedFromExiledComponent` (Jadeheart Attendant's MV lifegain, Sunbird Effigy's `*/*` CDA). `AddOneManaOfEachColorAmongEffect` gains a `colorSource` selector so crafted materials' printed colors drive Bloom Tender-shape mana (`Effects.AddOneManaOfEachCraftedMaterialColor()`).

**While-tapped activation lock.** Braided Net's "its activated abilities can't be activated for as long as it remains tapped": `PreventActivatedAbilities` becomes grantable to a specific permanent (`CastPermissionUtils.isActivationPrevented` now also consults `grantedStaticAbilities` — one read site covers the activate handler and both enumerators), with the new one-way `Duration.WhileAffectedTapped` (CR 611.2b: pruned by the SBA expiry check on untap/leave; re-tapping never restarts it). Adds `StatePredicate.IsSource` / `sourceItself()`, `Targets.OtherNonlandPermanent`, and the passive `net` counter type. **Bonus fix caught by the new tests:** `TargetValidator` never enforced `excludeSelf` on directly-submitted battlefield targets — every "another target ..." card accepted a self-target on manual submission (enumeration was correct).

## Cards

| Card | Front | Back |
|---|---|---|
| Braided Net // Braided Quipu | 3 net counters; {T}+counter: tap + while-tapped ability lock | draw per artifact, self-tuck third from top |
| Visage of Dread // Dread Osseosaur | targeted reveal-hand discard (Divest pipeline); **craft with two creatures** | 5/4 menace, enters/attacks may-mill 2 |
| Dire Flail // Dire Blunderbuss | +2/+0 Equipment | +3/+0 + granted attack trigger: sac artifact → power damage (reflexive) |
| Jade Seedstones // Jadeheart Attendant | distribute 3 +1/+1 counters | 7/7, ETB gain life = crafted material's MV |
| Sunbird Standard // Sunbird Effigy | {T}: any color; **craft with one or more (any)** | `*/*` = crafted colors, flying/vigilance/haste, {T}: one mana of each crafted color |

Known, documented approximations/limitations (all pre-existing patterns, flagged in KDoc):
- Dire Blunderbuss's "an artifact other than ~" uses a name-exclusion filter — the granted-triggered-ability channel doesn't thread `granterId` (only granted *activated* abilities do), so a second copy of the card can't be sacrificed either. KDoc names the proper fix.
- Color indicators aren't representable (`Dread Osseosaur` prints black, `Sunbird Effigy` W/U on Manufactory precedent) — same as the existing Mastercraft Raptor.
- Distribute-counters auto-splits evenly (engine-wide primitive behavior, same as Armament Corps).

`manual-scenarios/sets/lci/craft-03.json`: all five front faces in hand, exactly enough untapped lands, craft materials (two creatures, artifacts, colored permanents) on battlefield and in graveyard, and an opposing hand with both a choosable and a non-choosable card for Visage.

Verified: full `:mtg-sets:test` (snapshot + lint), 25 scenario tests across the 5 new classes + Saheeli/exact-count/Cursed Totem regressions green.


bolav:
## Review: LCI Craft-transform DFCs — batch 3 (diff vs `lci-craft-02`)

### Verdict
Ship it. Five Craft DFCs plus the supporting engine primitives. All Oracle text matches
Scryfall verbatim, all eight cited CR numbers verify against the comp rules, and the new
SDK types are genuinely novel and well-parameterized — existing primitives are reused
wherever one exists. Tests green: **14 scenario cases across the 5 new classes, 0 failures.**

### What's good
- **Reuse over new types, consistently.** `AddOneManaOfEachCraftedMaterialColor` is a
  `colorSource` discriminator on the *existing* `AddOneManaOfEachColorAmongEffect` (reuses the
  executor, no new effect). `Targets.OtherNonlandPermanent` rides the existing `excludeSelf`
  flag. Only genuinely-novel primitives were added.
- **`StatePredicate.IsSource` / `sourceItself()` fills a real gap** — the `GameObjectFilter`
  counterpart of `GroupFilter`'s `Scope.Self`, which didn't exist. Correctly classified in all
  four predicate lists (BeginningPhaseManager, TriggerMatcher, `AffectsFilterResolver`→false,
  PredicateEvaluator). Traced: a self-scoped grant locks only the victim, never bystanders.
- **`Duration.WhileAffectedTapped` is fully wired** across every site its sibling
  `WhileAffectedHasCounter` touches — per-frame gate in `StateProjector` + activation legality,
  cleanup in `CleanupPhaseManager`, and the one-way latch in `EndedDurationExpiryCheck`
  (cleanly generalized `pruneCounterKeyedGrants` → `pruneAffectedKeyedGrants`, covering both the
  granted-activated and granted-static ability maps). The 611.2b one-way behavior is tested.
- **Loyalty carve-out is filter-driven, not hardcoded**, so Braided Net's `sourceItself()`
  filter correctly locks whatever it taps — no accidental exemption bug.
- **Tests hit the rule corners, not just happy paths:** 712.8a tuck-as-front-face, 611.2b
  one-way re-tap latch, mana-abilities-included lock, "another"+lands rejected, exactly-N craft
  materials (both directions), colorless→0/0 dies, decline-optional paths, battlefield+graveyard
  material mix. Battlefield reads go through projection.
- **The one real limitation is documented, not hidden.** Dire Blunderbuss's "artifact other
  than Dire Blunderbuss" is a name-filter approximation of an entity reference (CR 201.5a), with
  the exact mis-handled corner (a second Blunderbuss) spelled out and the proper fix named.
- **Bonus correctness fix** caught by the new tests: `TargetValidator` never enforced
  `excludeSelf` on directly-submitted battlefield targets, so every "another target …" card
  accepted a self-target on manual submission (enumeration was already correct).

### Issues
None blocking or important. Minor / discretionary:
- **Three discrete `CraftedMaterials*` DynamicAmounts** now exist (`TotalPower`,
  `TotalManaValue`, `ColorCount`), each reading `CraftedFromExiledComponent` and aggregating one
  property. A parameterized aggregator would scale better for the next craft card — but this is
  consistent with the pre-existing `TotalPower` precedent (the generalization point was that
  earlier card), and the color-count's distinct-set aggregation differs from the sums. Names
  match semantics precisely; fine to leave.
- **`VisageOfDreadScenarioTest` has no empty-eligible-hand case** (opponent holds no
  artifact/creature card → nothing discarded). It's the proven Divest / Thought-Stalker Warlock
  idiom so behavior is almost certainly correct, but the corner is untested — one extra `test`
  would close it.
- **Granted-triggered-ability `granterId` gap** (Dire Blunderbuss) is a genuine engine
  limitation worth a backlog note for a future `add-feature`, not this PR.

### Recommendation
Merge as-is. Reviewed against `lci-craft-02` as the stacked base; the branch was rebased onto
the current `lci-craft-02` tip (dropping the 6 duplicated batch-2 commits) with no conflicts,
and its own diff is the 8 batch-3 commits.

Issue #41, opened 2026-05-05
Essence Channeler - When this creature dies, put its counters on target creature you control - bug


bolav:
Essence Channeler has this text:
`When this creature dies, put its counters on target creature you control.`

How ever the code only moves the +1/+1 counters:

```
    triggeredAbility {
        trigger = Triggers.Dies
        target = Targets.CreatureYouControl
        effect = Effects.AddDynamicCounters(
            Counters.PLUS_ONE_PLUS_ONE,
            DynamicAmount.LastKnownCounterCount,
            EffectTarget.ContextTarget(0)
        )
    }

```

PR #1289, opened 2026-07-13
Add Unstable Glyphbridge // Sandswirl Wanderglyph to LCI + fix ControlledByActivePlayer
  floating filter

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1288

## Unstable Glyphbridge // Sandswirl Wanderglyph (LCI #41)

Stacked on `lci-craft-03` — this PR shows only the diff between `lci-craft-03` and
`lci-unstable-glyphbridge` (2 commits): the card + a small engine fix it depends on.

### Card

**Front — Unstable Glyphbridge** (Artifact, {3}{W}{W})
- ETB: *"When this artifact enters, if you cast it, for each player, choose a creature with
  power 2 or less that player controls. Then destroy all creatures except creatures chosen
  this way."* — the Winnowing shape: `ForEachPlayerEffect` (active player first) →
  `GatherCardsEffect` → `SelectFromCollectionEffect` (`Chooser.SourceController`, so *you*
  choose one per player, filtered to power ≤ 2) → `MoveCollectionEffect(Destroy)`.
- *"if you cast it"* via `Conditions.WasCast` as the trigger's intervening-if, so the wipe
  never fires on the craft return / reanimation / blink.
- **Craft with artifact {3}{W}{W}** — exact-count craft (`minCount = maxCount = 1`) via the
  `craft(...)` helper.

**Back — Sandswirl Wanderglyph** (Artifact Creature — Golem, 5/3, Flying)
- *"Whenever an opponent casts a spell during their turn, they can't attack you or
  planeswalkers you control this turn."* — `Triggers.OpponentCastsSpell` gated on
  `IsNotYourTurn`, applying a floating `CantAttackGroup` over the active player's creatures.
- *"Each opponent who attacked you or a planeswalker you control this turn can't cast
  spells."* — `PlayersCantCastSpells` scoped to opponents, conditioned on
  `PlayerAttackedPlayerThisTurn(attacker = AnOpponent, defender = You)`.

Both back abilities are exact for two-player games (the engine's scenario scope); the
multiplayer over-approximation is documented in the card KDoc. No new SDK types — the card is
pure composition of existing primitives.

### Engine fix

`AffectsFilterResolver.resolveGenericFilter`'s controller-predicate branch had no case for
`ControllerPredicate.ControlledByActivePlayer`, so it returned `null` ("unknown leaf"), and
`evaluateWith` fails unknown leaves **open** — a `CantAttackGroup` / anthem / any floating
effect scoped to the active player's permanents was silently applied to **everything**. Now
resolves against `state.activePlayerId`, matching the activated/spell targeting path that
`PredicateEvaluator` already handled. Sandswirl Wanderglyph's back trigger is the first card
to exercise this path.

### Tests

`UnstableGlyphbridgeScenarioTest` (rules-engine, 5 cases, all green):
1. Cast from hand: controller picks one power-≤2 creature per player (active player first);
   all other creatures destroyed; chosen creatures + the artifact survive.
2. Intervening-if: returned from graveyard via a real reanimation spell (not
   `putPermanentOnBattlefield`) — the ETB wipe does not fire.
3. Craft end-to-end: exactly-one-material cost (two rejected), returns transformed as a
   5/3 flying Golem, material in exile.
4. Back trigger: an opponent casting during their turn locks their creatures (incl. ones
   cast later that turn, Rule 611.2c) out of attacking; your creatures unaffected.
5. Back static: an opponent who attacked you this turn can't cast; one who didn't still can.

`LCI.json` card snapshot re-blessed for the new card.


bolav:
## Code review — Unstable Glyphbridge // Sandswirl Wanderglyph (diff vs `lci-craft-03`)

### Verdict
Behavior is right and the SDK shape is excellent — this card is pure composition of existing
primitives (no new `Effect`/`Trigger`/`Condition`/`StaticAbility` type), and the only engine
change is a genuine, well-scoped bug fix. All 5 scenario tests pass, printing placement is
correct. Ship-ready; findings are test-coverage gaps and doc nits, nothing blocking.

### What's good (keep it)
- **Zero SDK sprawl.** The front-face wipe reuses the Winnowing shape (`ForEachPlayerEffect` →
  `GatherCardsEffect` → `SelectFromCollectionEffect` → `MoveCollectionEffect`); the back face
  reuses `CantAttackGroup`, `PlayersCantCastSpells`, `PlayerAttackedPlayerThisTurn`. Exactly
  the "compose, don't add a type per card" goal.
- **The engine fix is correct and confirmed.** `AffectsFilterResolver.kt:313` adds the
  `ControlledByActivePlayer` leaf that was missing from the floating-filter path. Verified the
  fail-open claim: `ControllerPredicate.evaluateWith` resolves an unknown leaf (`null`) to
  `true` at the root (`?: true`, `ControllerPredicate.kt:176`), so the missing branch really
  did match every creature. The new branch (`entityController == state.activePlayerId`) matches
  the already-correct targeting path in `PredicateEvaluator.kt:748`. Good catch — it would have
  mis-scoped any active-player anthem/restriction, not just this card.
- **The intervening-if negative test is real, not vacuous.** It reanimates via a spell
  (`Effects.PutOntoBattlefield`) rather than `putPermanentOnBattlefield`, so the "entered but
  not cast" path actually fires an ETB event and proves `Conditions.WasCast` gates it.
- **Honest KDoc.** The card documents its two-player-exactness assumptions (multiplayer
  over-approximation on both back abilities) and the interleaved choose/destroy approximation
  instead of pretending they're exact — correct call for a two-player-scoped engine.

### Issues (all Minor)
- **Missing "zero eligible creatures" rule-corner test.** The front face uses
  `SelectionMode.ChooseExactly(1)` with a `powerAtMost(2)` filter, and the KDoc claims a player
  with no power-≤2 creature "auto-selects nothing, everything is destroyed." No test exercises
  that path — every test creature set gives each player an eligible creature. Add a case where a
  player controls only a 5/5: the choose must not hang/error and all of that player's creatures
  must die.
- **`powerAtMost(2)` eligibility is never tested against *projected* P/T.** All test creatures
  have printed powers with no continuous modifiers, so a base-vs-projected bug in the selection
  filter would pass unnoticed. A single test with an anthem/debuff in play (a 1/1 pumped to 3/3
  must be ineligible; a 5/5 shrunk to 2/2 must be eligible) would nail down the KDoc's
  "evaluated against projected state" claim. Low risk — `SelectFromCollectionEffect` is a shared
  primitive presumed projection-correct.
- **Snapshot/lint nets not re-run in review.** Only the scenario test (`:rules-engine`) was run
  and is green. The `LCI.json` snapshot diff is included (re-blessed), and the pipeline-variable
  dataflow (`playerCreatures`→`spared`/`doomed`) is the standard Winnowing shape, but a
  `:mtg-sets:test` (CardDefinitionSnapshotTest + CardLintTest) before merge would confirm both
  nets are green.

### Recommendation
Merge as-is after a quick `:mtg-sets:test`. Optionally add the two front-face corner tests
(zero-eligible destruction; projected-P/T eligibility) — both cheap, and they close the gap
between the KDoc's claims and what's exercised. No SDK changes or rewrites needed.

PR #1290, opened 2026-07-14
Add The Enigma Jewel // Locus of Enlightenment (LCI #55)

bolav:
## The Enigma Jewel // Locus of Enlightenment (LCI #55)

Stacked on top of `lci-unstable-glyphbridge` — this PR is only the diff against that branch.

Implements the full DFC plus the two engine features it needs, built entirely from composable
primitives (no card-specific engine code).

### Front — The Enigma Jewel ({U}, Legendary Artifact)
- **Enters tapped** — `EntersTapped` replacement effect.
- **`{T}`: Add `{C}{C}`. Spend this mana only to activate abilities.** — `Effects.AddColorlessMana`
  carrying `ManaRestriction.AbilityActivationOnly` (funds any ability activation, rejects spells).
- **Craft with four or more nonlands with activated abilities `{8}{U}`** (CR 702.167a/b) — the
  `craft(...)` DSL over `GameObjectFilter.Nonland.withCardPredicate(CardPredicate.HasActivatedAbility)`.

### Back — Locus of Enlightenment (Legendary Artifact)
- **Has each activated ability of the exiled cards used to craft it, each usable only once each turn**
  (CR 702.167c) — the `HasAllActivatedAbilitiesOfExiledCards` static with `source = CRAFTED,
  oncePerTurnEach = true`. Each granted ability is re-stamped with an exiled-card-derived `AbilityId`,
  so duplicate materials don't collapse and the existing once-per-turn tracker gives each exiled card
  its own budget — no new tracker component.
- **Whenever you activate a non-mana ability, copy it (you may choose new targets)** — reuses
  `Triggers.YouActivateAbility` + `Effects.CopyTargetSpellOrAbility` (CR 707.10c; CR 605.1a excludes
  mana abilities).

### New / changed SDK surface
- `CardPredicate.HasActivatedAbility` — "has an activated ability of any kind, mana included" (the
  mana-counting sibling of `HasNonManaActivatedAbility`). Flag-backed on `CardComponent`.
- `GameObjectFilter.withCardPredicate(p)` — general-purpose combinator for any `CardPredicate`.
- `HasAllActivatedAbilitiesOfExiledCards(source, filter, creatureCardsOnly, oncePerTurnEach)` — the
  single primitive behind both "exiled with this" (linked-exile) and "exiled to craft this" grants.
  Replaces the former `HasAllActivatedAbilitiesOfLinkedExiledCard` and
[...]

PR #1291, opened 2026-07-14
Throne of the Grim Captain // The Grim Captain (LCI #266)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1290

## Throne of the Grim Captain // The Grim Captain (LCI #266)

Implements the full card, plus the two reusable capabilities it needs beyond the existing
craft/pipeline vocabulary.

> **Stacked PR.** Base is `lci-enigma-jewel`; this PR is only the diff on top of that branch.

### Card

- **Front — Throne of the Grim Captain** ({2}, Legendary Artifact): `{T}: Mill two cards.` +
  `Craft with a Dinosaur, a Merfolk, a Pirate, and a Vampire {4}`.
- **Back — The Grim Captain** (7/7 Legendary Skeleton Spirit Pirate): menace, trample, lifelink,
  hexproof; on attack each opponent sacrifices a nonland permanent of their choice, then you may put
  an exiled creature card used to craft it onto the battlefield tapped and attacking.

### New SDK / engine capabilities (both reusable, not card-specific)

- **Heterogeneous slot-based Craft** — `craft(slots = ...)` DSL + `AbilityCost.Craft.slots`. Naming
  one material of *each* of several kinds makes validation a bipartite perfect-matching problem, not a
  per-subtype count (a Merfolk Pirate fills only one slot; four Vampires can't cover four subtypes).
  The union of the slot filters stays in `filter`, so candidate gathering, `canPay`, enumeration, and
  the client overlay are unchanged; the matching check is layered on top.
- **`CardSource.CraftedMaterials`** — the gather-pipeline twin of `ExiledCardsSource.CRAFTED`, reading
  the source's `CraftedFromExiledComponent` filtered to cards still in exile. Powers the attack
  trigger's second clause via the existing `GatherCards → SelectFromCollection → MoveCollection`
  (`ZonePlacement.TappedAndAttacking`).

### Review-feedback fixes (second commit)

- **Shared bipartite matcher** — extracted `BipartiteMatching` (generic Kuhn's augmenting-path);
  `CraftSlotMatching` and `BlockPhaseManager`'s must-be-blocked matching now share one implementation
  instead of two copies of the same algorithm.
- **Real color indicator (CR 204)** — added `CardDefinition.colorIndicator` / `card { colorIndicator }`.
  The Grim Captain's black back face (empty mana cost) now projects as **black**, not colorless. Colors
  combine mana-cost + indicator (CR 202.2) and fold into color identity (CR 903.4).
- **Projection consistency** — the enumerator's slot-matching edge predicate uses the same projected
  state as the battlefield-material gather.

### Tests

- 7 scenario tests (mill; craft happy path + dual-type Merfolk Pirate; count-can't-cover rejection;
  chosen-subset-unmatchable rejection; edict; put-crafted-creature tapped+attacking) — plus a black
  color-indicator projection assertion.
- `MustBeBlockedMatchingScenarioTest` re-run green after the matcher extraction.
- Card snapshot golden re-blessed (additive, Throne-only). SDK reference doc updated.


bolav:
## Review — Throne of the Grim Captain // The Grim Captain (LCI #266)

Reviewed the diff vs the stacked base `lci-enigma-jewel`. Primary lens: SDK elegance.

### Verdict

Behavior is correct and the SDK shape is right. The card needed two capabilities beyond the existing
craft/pipeline vocabulary, and both were added as small, well-parameterized, genuinely-novel
primitives rather than card-specific types. All scenario tests pass. **The three minor findings below
were addressed in the follow-up commit on this PR** — this comment records the review and how each was
resolved.

### What's good

- **`CardSource.CraftedMaterials` is textbook composition.** Gather-pipeline twin of
  `ExiledCardsSource.CRAFTED`, reads an existing component, and the whole clause-2 behavior falls out
  of existing primitives (`GatherCards → SelectFromCollection(ChooseUpTo 1, Creature) →
  MoveCollection(BATTLEFIELD, TappedAndAttacking)`). No new effect type, no new event, no client
  wiring. Correctly filters to materials still in exile (owner-scoped), so a material already returned
  by a prior attack isn't re-offered.
- **`AbilityCost.Craft.slots` is additive and backward-compatible.** `slots = emptyList()` default,
  union kept in `filter`, `min == max == slots.size` — candidate gathering, `canPay`, enumeration, and
  the client overlay all keep working; only the matching check is layered on. Snapshot diff confirms
  zero other cards changed.
- **Matching gated in all three paths** — `canPay`, the legal-action enumerator, and payment — sharing
  one edge predicate that uses the same projected-state matcher as candidate gathering.
- **Tests hit the corners that motivate the feature**: dual-type Merfolk Pirate filling one slot, four
  Vampires rejected (count would pass, matching fails), and a legal-count-but-unmatchable chosen subset
  rejected at payment. Oracle text matches Scryfall exactly on both faces.

### Findings (all resolved in the follow-up commit)

**1. (Minor) Duplicated bipartite matcher.** `CraftSlotMatching` reimplemented the same Kuhn's
augmenting-path routine that `BlockPhaseManager` inlines — the "reuses the same routine" claim was
shared *pattern*, not shared *code*.
→ **Fixed:** extracted a generic `BipartiteMatching` (`maximumMatching` / `canSaturateLeft`);
`CraftSlotMatching` and `BlockPhaseManager` now delegate to it. `MustBeBlockedMatchingScenarioTest`
re-run green.

**2. (Minor) Back-face black color not represented.** The back used `colorIdentity = "B"` only, so the
projected face read as *colorless*, not black — a documented approximation.
→ **Fixed:** added a real color indicator (CR 204): `CardDefinition.colorIndicator` +
`card { colorIndicator = "B" }`. The transformed face now projects black; colors combine mana-cost +
indicator (CR 202.2) and fold into color identity (CR 903.4). Added a projection assertion.

**3. (Minor) Projection inconsistency in the enumerator.** The slot-matching edge predicate used
`state.projectedState` while the battlefield-material gather used the local `projected` (same object,
so no bug, but inconsistent).
→ **Fixed:** the edge predicate now uses the same `projected`.

### Non-blocking notes (left as-is by design)

- Heterogeneous craft ships a flat union material list to the client (min = max = slot count); an
  unmatchable set is rejected at payment time — no per-slot selection UI. Reasonable for one card;
  documented in the card comment.

### Recommendation

Ship it. The design is exactly the "small generic primitive + reuse" the SDK wants, and the tests
prove the hard cases.

PR #1297, opened 2026-07-15
Thread granting permanent through granted triggered abilities (Dire Blunderbuss)

bolav:
## Thread the granting permanent through granted triggered abilities (Dire Blunderbuss)

Adds a first-class way for a *granted* triggered ability to reference the permanent that
granted it (its Equipment/Aura), so filters can express "other than [this granting
Equipment]" precisely — implementing **Dire Blunderbuss** (LCI #145 back face) correctly.

### The card

> **Dire Blunderbuss** — Equipped creature gets +3/+0 and has "Whenever this creature
> attacks, you may sacrifice an artifact **other than Dire Blunderbuss**. When you do, this
> creature deals damage equal to its power to target creature." Equip {1}.

The exclusion is CR 201.5a: the name refers only to *that specific* granting Equipment. The
tricky part is that a granted *triggered* ability resolves with the **equipped creature** as
its source, not the Equipment — so the granter can't be reached as "the source".

### What's added

- **SDK:** `StatePredicate.IsGrantingPermanent` + filter builder `notGrantingPermanent()`.
  It matches the granting permanent read from the evaluation context's `granterId`, so
  `GameObjectFilter.Artifact.youControl().notGrantingPermanent()` excludes *only* the
  specific granter — a second same-named copy, and any other Equipment on the same creature,
  stay legal.
- **Engine:** the granter is threaded onto the stack and into filter evaluation via
  `PendingTrigger.granterId` → `TriggeredAbilityOnStackComponent.granterId` →
  `EffectContext.granterId` → `PredicateContext.granterId`. `TriggerDetector.assignGranterIds`
  stamps it uniformly across every event-based path (`detectTriggers`, including the
  batch/duplicate detectors) and phase/step triggers (`detectPhaseStepTriggers`).
  `ReflexiveTriggerEffectExecutor` and `SelectTargetPipelineExecutor` both carry `granterId`
  so the optional "may" feasibility check and the actual selection agree.

This replaces an earlier proximity approximation (`notAttachedToSource()`, which also excluded
any *other* Equipment co-attached to the same creature) with an exact, entity-based exclusion.

### Coverage & edge cases

- **Identical co-attached granters:** two copies of the same Equipment on one creature each
  grant a trigger with the same `abilityId`. `resolveGranterIds` returns all granting
  attachments in index order and `assignGranterIds` uses a per-`(source, ability)` cursor to
  map the N fired triggers 1:1 onto the N granters, so each trigger excludes only its own copy.
- **Not covered (documented):** leaves-the-battlefield granted triggers (a granted "dies"
  ability) — the source is gone before the trigger fires, so the granter can't be recovered
  from the attachment index. Noted in the SDK reference.

### Tests

`DireFlailScenarioTest` (rules-engine) — 9 scenarios, all green: front/back equip and stat
grants, craft end-to-end, the granted attack-trigger happy path, declining the optional
sacrifice, the self-exclusion (no prompt when the only artifact is the Blunderbuss itself), a
second unattached Blunderbuss staying sacrificable, a co-attached *different* Equipment staying
sacrificable, and two identical Blunderbusses on one creature each excluding only themselves
(so both can be sacrificed).

### SDK reference

`docs/card-sdk-language-reference.md` updated for `IsGrantingPermanent` / `notGrantingPermanent()`,
including the threading path and the precise coverage note.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -2580,7 +2580,13 @@ work for abilities-on-stack (which carry no `CardComponent`).
   excluding *only* the specific granter (a second same-named copy or another attachment stays legal). The
   granter is threaded to granted triggered abilities via `PendingTrigger.granterId` →
   `TriggeredAbilityOnStackComponent.granterId` → `EffectContext.granterId` → `PredicateContext.granterId`;
-  false with no granter context (an ungranted ability).
+  false with no granter context (an ungranted ability). `TriggerDetector.assignGranterIds` stamps the
+  granter on every event-based trigger (`detectTriggers`, including the batch/duplicate detectors) and on
+  phase/step triggers (`detectPhaseStepTriggers`), so a granted attack *or* upkeep trigger is covered;
+  identical co-attached granters are told apart by a per-`(source, ability)` cursor so N fired triggers map
+  1:1 onto the N granters. Not populated for leaves-the-battlefield granted triggers (a granted "dies"
+  ability): the source is already gone when the trigger fires, so the granter can't be recovered from its
+  attachment index — reference the granter by other means there, or scope with `notAttachedToSource()`.
 - `HasGreatestPower` (filter builder `hasGreatestPower()`) / `HasLeastPower` (filter builder
   `hasLeastPower()`) — has the greatest / least projected power among creatures *its controller*
   controls (ties all qualify). Used for "creature with the greatest/least power" target and edict
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -2570,9 +2570,17 @@ work for abilities-on-stack (which carry no `CardComponent`).
   `AttachedToComponent.targetId == sourceId`. Use it to scope a static ability on the *host* to its own
   attachments — Cloud, Midgar Mercenary's "an Equipment attached to it" via
   `GameObjectFilter.Artifact.withSubtype("Equipment").attachedToSource()`. Source-relative; inert with no
-  source context. Negated builder `notAttachedToSource()` — excludes the source's own attachments, backing
-  "an artifact other than [this granting Equipment]" on a granted triggered ability whose source is the
-  equipped creature (Dire Blunderbuss's "sacrifice an artifact other than Dire Blunderbuss").
+  source context. Negated builder `notAttachedToSource()` — excludes all of the source's own attachments.
+- `IsGrantingPermanent` (negated builder `notGrantingPermanent()`) — matches the *granting permanent* of the
+  resolving ability: the Equipment/Aura/permanent whose `GrantActivatedAbility`/`GrantTriggeredAbility` static
+  granted the ability, read from the evaluation context's `granterId`. For a granted triggered ability the
+  ability's source is the equipped creature but the granter is the Equipment attached to it. Negate for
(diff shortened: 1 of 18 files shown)


bolav:
## Code review — granter threading for granted triggered abilities

**Reviewed:** this branch vs its base. Both commits included (initial threading +
follow-up broadening/disambiguation). `DireFlailScenarioTest` run green (9/9).

### Verdict

Behavior and SDK shape are both right. This replaces a documented proximity hack
(`.notAttachedToSource()`) with a proper entity-based exclusion by threading the granting
permanent's id from trigger detection to the target filter. The new
`StatePredicate.IsGrantingPermanent` is genuinely novel (it reads grant-relative state no
existing primitive exposes), well-parameterized, and honestly named.

### What's good

- **Clean, layered threading** with a single source of truth: `PendingTrigger.granterId →
  TriggeredAbilityOnStackComponent.granterId → EffectContext.granterId →
  PredicateContext.granterId`, documented identically at every hop.
- **Exhaustive `when` discipline.** The new `StatePredicate` variant is handled explicitly
  (returning `false`) in every non-resolution evaluator — `BeginningPhaseManager`,
  `TriggerMatcher`, `AffectsFilterResolver` — each with a one-line rationale
  ("granter-relative exclusion is resolution-time only"). No `else` fallthrough, so the next
  predicate is a compile-time decision.
- **Feasibility/execution agreement.** `ReflexiveTriggerEffectExecutor`'s "may" check and
  `SelectTargetPipelineExecutor`'s selection both build a minimal
  `PredicateContext(controllerId, granterId)`, so the offer and the legal-target set can't
  disagree.
- **`resolveGranterIds` is cheap and correct-by-construction:** O(1) for the unattached
  majority via the `AttachmentsComponent` reverse index, and its `GrantTriggeredAbility` /
  `ConditionalStaticAbility` + `Scope.AttachedTo` matching mirrors
  `getAttachedGrantedTriggeredAbilities`.
- **Tests genuinely exercise the improvement** — the co-attached and two-identical-copies
  cases are exactly the corners the old proximity filter got wrong.
- Snapshot re-blessed consistently and the SDK reference doc updated per the repo's hard rule.

### Findings (both raised as Minor — resolved in the follow-up commit)

1. **✅ Resolved — coverage beyond the attack-trigger path.** `granterId` was originally
   stamped only inside `detectTriggersForEvent`, so granted phase/step and batch-detected
   triggers wouldn't get it. Moved to a shared `TriggerDetector.assignGranterIds` post-pass
   applied at the return of both `detectTriggers` and `detectPhaseStepTriggers`. The one
   genuinely-unresolvable path (leaves-the-battlefield granted triggers — source gone before
   the trigger fires) is now documented in the SDK reference rather than silently uncovered.

2. **✅ Resolved — identical co-attached granters couldn't be disambiguated.**
   `resolveGranterId` matched by per-`cardDef` `abilityId`, so two copies of the same
   Equipment on one creature both resolved to the first attachment. Replaced with
   `resolveGranterIds` (all granters in index order) + a per-`(source, ability)` cursor in
   `assignGranterIds` that maps N fired triggers 1:1 onto the N granters (doubled copies clamp
   to the last granter). Added a scenario test that fails under the old first-match logic.

### Recommendation

Merge-ready. No blocking or important issues; both minor findings addressed with tests.

PR #1301, opened 2026-07-18
Add the last 6 LCI cards (Deeproot Pilgrimage, Dreadmaw's Ire, Kutzil's Flanker, Malicious
  Eclipse, Preacher of the Schism, Scytheclaw Raptor)

bolav:
## The Lost Caverns of Ixalan — the last 6 cards

Implements the six remaining LCI cards (244 → 250 / 286), plus the small SDK/engine
primitives each one needed, full scenario tests, snapshot updates, and the doc reference.

### Cards

| Card | Cost | What it exercises |
|------|------|-------------------|
| **Deeproot Pilgrimage** | `{1}{U}` | `OneOrMoreBecomeTapped` batch trigger (CR 603.2c) — tapping several nontoken Merfolk at once makes one token, not one per Merfolk |
| **Dreadmaw's Ire** | `{R}` | Grants a triggered ability whose target scopes to "that player controls" via `ControllerPredicate.ControlledByTriggeringPlayer` (the damaged player) |
| **Kutzil's Flanker** | `{2}{W}` | New `CREATURES_LEFT_BATTLEFIELD` turn tracker; modal ETB with a targeted graveyard-exile mode |
| **Malicious Eclipse** | `{1}{B}{B}` | Floating, controller-scoped death→exile `RedirectZoneChange` (Forgotten Cellar idiom) applied from a resolving sorcery |
| **Preacher of the Schism** | `{2}{B}` | New `PlayerHasMostLife` condition on both axes (attacked player + controller), including the tied case |
| **Scytheclaw Raptor** | `{2}{R}` | New `IsPlayersTurn(player)` condition resolved against `Player.TriggeringPlayer` |

### SDK / engine additions

- `Triggers.OneOrMoreBecomeTapped` + `TapEvent(batch = true)` and `TriggerDetector.detectTapBatchTriggers`
- `Conditions.PlayerHasMostLife` and `Conditions.IsPlayersTurn`
- `ControllerPredicate.ControlledByTriggeringPlayer` (FQL `ctrl:triggering-player`) wired into the predicate evaluator + `TargetFinder`
- `TurnTracker.CREATURES_LEFT_BATTLEFIELD` backed by `CreatureLeftBattlefieldThisTurnComponent`
- `GrantReplacementEffect` extended to allow a resolving spell to grant a floating, controller-scoped global replacement
- Card hover preview now surfaces granted/active-effect text (web client)

### Review fixes folded in (last commit)

A review of the branch surfaced two engine-level issues, both fixed here:

1. **May/pay target pre-checks dropped `triggeringPlayerId`.** The three legal-target guard
   sites in `TriggerProcessor` built their `PredicateContext` with only `triggeringEntityId`,
   unlike the canonical on-stack targeting path. A "may"/pay trigger targeting "that player
   controls" (where the player rides on `triggeringPlayerId`) could have been silently skipped.
   All three now thread the same context as the on-stack path.
2. **Redundant `triggeringPlayerId` copy in `DamageTriggerDetector`** removed — `fromEvent`
   already carries the damaged player as `triggeringEntityId`, and both resolution paths use the
   `triggeringPlayerId ?: triggeringEntityId` fallback, so the copy was a no-op.

### Tests

Per-card Kotest scenario tests in `rules-engine` (batch semantics, granted-trigger targeting +
clean fizzle, the new turn tracker, floating-replacement lifetime, most-life on both axes, and the
triggering-player turn check). The six affected scenario tests pass after the review fixes.


bolav:
## Code review — high effort (8 finder angles, dedup, no verify)

Traced the full diff and the engine paths it touches: target-context threading, the batch-tap
detector, the granted-replacement executor, the redirect read path, condition/predicate
evaluators, and the zone-transition tracker. The six cards match their oracle text and the core
mechanics are wired correctly end-to-end. No crashing or wrong-output bugs surfaced; the items
below are lower-severity consistency/altitude findings.

### Findings

**1. `TriggerProcessor` may/pay legal-target pre-checks omit `triggeringPlayerId` — `TriggerProcessor.kt` (batchKeyOf / processMayThenTargetTrigger / pay path)** ✅ fixed in this branch

The three fizzle-guard sites called `findLegalTargets` with only `triggeringEntityId`, while the
canonical on-stack targeting path also threads `triggeringPlayerId` via `pipelineContext`. A future
optional (`may`) or pay-cost trigger whose target filter uses `controlledByTriggeringPlayer()` /
`ControlledByReferencedPlayer(Player.TriggeringPlayer)` on a trigger where the associated player
rides on `triggeringPlayerId` (any non-damage trigger) would resolve to 0 legal targets in the
guard and silently skip asking, even though the real path would have found targets. Not hit by the
6 cards here (Dreadmaw's Ire is mandatory, and its `triggeringEntityId` already equals the damaged
player), but the guards were inconsistent with the primary path.
→ **Fixed:** all three pre-checks now pass the same `PredicateContext(controllerId,
triggeringEntityId, triggeringPlayerId)` as the on-stack path.

**2. Redundant `triggeringPlayerId` copy in `DamageTriggerDetector.kt` (SELF damage-source path)** ✅ fixed in this branch

`fromEvent(DamageDealtEvent)` already sets `triggeringEntityId = event.targetId` (the damaged
player), and both `Player.TriggeringPlayer` and `ControlledByTriggeringPlayer` resolve as
`triggeringPlayerId ?: triggeringEntityId`, so the added `.copy(triggeringPlayerId = …)` was a
no-op for Dreadmaw's Ire while quietly setting `triggeringPlayerId` on every other SELF
damage-to-player trigger.
→ **Fixed:** removed the copy; the trigger context is back to plain `fromEvent(event)`.

**3. Scytheclaw Raptor hard-codes `description` to paper over `Not(IsPlayersTurn(...))` text — `ScytheclawRaptor.kt`** — left as-is (low-severity altitude)

`Not(IsPlayersTurn(...))` auto-generates awkward double-"if" text ("if not (if it's their turn)"),
so the card pins a manual `description`. The deeper fix belongs in condition text rendering so
future cards wrapping `IsPlayersTurn` in `Not` don't each need the same override. Not blocking;
noted for a follow-up on text generation.

### Verification

The six affected `rules-engine` scenario tests pass after the fixes (`BUILD SUCCESSFUL`). The full
`rules-engine` suite was not run in-session; the `TriggerProcessor` change is additive (for the
common `triggeringPlayerId == null` case the pre-check context is byte-identical to before), so no
regression to existing may/pay cards is expected.

PR #1303, opened 2026-07-19
Visualize when a Defender can attack + fix Mistbreath Elder upkeep text

bolav:
image

image

## What

Two small fixes plus the UX affordance that ties them together:

1. **Visualize when a Defender can attack** — a green "Can Attack" badge on any creature that
   has Defender but whose attack restriction is currently lifted (e.g. Shipwreck Sentry /
   Mechan Shieldmate once an artifact has entered under your control this turn, or a
   Krotiq-style temporary "attack as though it didn't have defender" grant). The badge appears
   the moment the condition is met — already in the main phase, before declare-attackers — so
   the player can see the Defender can now attack right after playing the artifact.

2. **Fix Mistbreath Elder's garbled upkeep text** — the `Exists → Gather → Select → Move`
   effect tree auto-rendered as garbled prose ("…look at all other creature permanents…Put
   those cards into a hand…"). Pinned the ability description to the oracle wording.

## How

- Extracted the "is the Defender restriction currently lifted?" logic out of `DefenderAttackRule`
  into a shared `DefenderBypass` helper (`rules-engine/.../combat/rules/DefenderBypass.kt`).
  Both the attack-legality rule (enforcement) and `ClientStateTransformer` (the display badge)
  now consult it, so **what the player sees stays exactly in sync with what the rules allow**.
  No new SDK type — the rule file gets smaller.
- The badge follows the existing Akawalli descend-8 block-restriction precedent in
  `ClientStateTransformer`: these restrictions are read directly by combat, never through the
  layer system, so they never reach the projected keyword set / `abilityFlags` and get no
  keyword chip — hence a badge. Gated on **projected** `hasKeyword(DEFENDER)`.
- Web client: `can-attack` icon gets a green badge style + tooltip border (`CardOverlays.tsx`).
- `MistbreathElder.kt` gets a `description` override matching the Scryfall oracle text; `BLB.json`
  snapshot re-blessed accordingly.
- `docs/card-sdk-language-reference.md` updated to describe the badge behavior.

## Tests

`rules-engine` `ShipwreckSentryScenarioTest` — all 4 green (`just test-class
ShipwreckSentryScenarioTest`, BUILD SUCCESSFUL):

- cannot attack when no artifact entered this turn
- can attack once an artifact entered this turn
- **no badge when no artifact entered this turn** (new)
- **badge reveals the Sentry can attack in the main phase, right after an artifact enters**, and
  the opponent sees it too (public info) (new)

Manual self-play scenarios added for both fixes under `manual-scenarios/cards/`.

CR 702.3b ("A creature with defender can't attack") verified against the Comprehensive Rules;
Mistbreath Elder oracle text verified against Scryfall.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1055,6 +1055,12 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
   turn as though it didn't have defender. Adds a transient `CanAttackDespiteDefenderThisTurnComponent`
   honored by the defender attack-restriction rule and cleaned up at end of turn. The
   activated/temporary counterpart to the static `CanAttackDespiteDefender` ability (Krotiq Nestguard).
+  Both the static ability and this transient grant are read directly by the attack-restriction rule
+  (via the shared `DefenderBypass` helper), never through the layer system — so while a Defender's
+  restriction is lifted, `ClientStateTransformer` surfaces a "Can attack despite defender"
+  `activeEffects` badge on the card (mirroring how Akawalli's descend-8 block restriction is shown),
+  letting the player see the creature can attack the moment the condition is met (e.g. after an
+  artifact enters for Shipwreck Sentry / Mechan Shieldmate).
 - `Effects.Goad(target = ContextTarget(0))` (`GoadEffect`) — goad target creature (CR 701.15).
   Tags the creature with `GoadedComponent(goaderIds: Set<EntityId>)`; the effect's controller at
   resolution is recorded as the goader. While goaded the creature (a) must attack each combat if able
--- a/manual-scenarios/cards/m/mistbreath-elder-bounce-and-counter.json
+++ b/manual-scenarios/cards/m/mistbreath-elder-bounce-and-counter.json
@@ -1,0 +1,32 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Mistbreath Elder", "summoningSickness": false},
+      {"name": "Grizzly Bears", "summoningSickness": false},
+      {"name": "Grizzly Bears", "summoningSickness": false},
+      {"name": "Grizzly Bears", "summoningSickness": false},
+      {"name": "Forest"}
+    ],
+    "graveyard": [],
+    "library": ["Forest", "Forest", "Forest"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [],
+    "graveyard": [],
+    "library": ["Forest", "Forest", "Forest"]
+  },
+  "phase": "BEGINNING",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "SELF"
+}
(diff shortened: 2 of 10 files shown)


bolav:
## Review (`/review-changes`)

**Verdict:** Behavior is correct and the SDK shape is right. This change fixes a real UX gap and a
text bug **without introducing a single new SDK type**, extracting shared logic into one helper so
enforcement and display can never disagree. No blocking issues.

### What's good

- **`DefenderBypass` is the correct refactor, not a new SDK type.** It lifts the *existing*
  `DefenderAttackRule` logic verbatim (`cardDef.staticAbilities` → `filterIsInstance<CanAttackDespiteDefender>()`
  → `Scope.Self` → condition eval, plus the `CanAttackDespiteDefenderThisTurnComponent` short-circuit)
  into one `object`, and both the rule and `ClientStateTransformer` call it — a single source of
  truth so what the player SEES matches what the rules let them DO. `AttackRestrictionRules.kt` gets
  *smaller* (net −23 lines there).
- **The badge follows the established Akawalli descend-8 precedent** line-for-line: same
  `restrictionController = state.projectedState.getController(...)`, same `seenGrantDescriptions`
  dedupe, same "read directly by combat, never through the layer system, so no keyword chip"
  rationale. It correctly gates on **projected** `hasKeyword(entityId, Keyword.DEFENDER)`.
- **Tests exercise the actual point of the change.** The new `ShipwreckSentryScenarioTest` cases
  assert the badge is absent with no artifact, present after one enters, **in `PRECOMBAT_MAIN`
  before declare-attackers** (the early-reveal that justifies the feature), and visible to the
  opponent (public info). They drive the artifact in via the real cast path (correct — the gate is
  an ETB-event tracker, not battlefield population). All 4 tests green locally.
- **Mistbreath Elder text fix is exactly right.** The `description` override, the `oracleText`
  field, the doc comment, and Scryfall's oracle text are all four identical (verified against
  Scryfall). Pinning the description is the right call — the `Exists → Gather → Select → Move` tree
  auto-renders garbled prose. Snapshot re-blessed as `descriptionOverride` in `BLB.json`.
- **Web-client stays a dumb terminal.** `getBadgeStyle` + `getTooltipBorderColor` both get a
  `can-attack` case; the badge renders `effect.name` as text, so no icon-glyph switch is missed.

### Verified

- **CR 702.3b** ("A creature with defender can't attack") cited in `DefenderBypass.kt` — matches
  the Comprehensive Rules exactly.
- No new `GameEvent` / client-visible state type → no `ClientEvent.kt` / `data-contracts` wiring
  needed. `ClientCardEffect` is the existing DTO.
- Printing placement N/A — no `card(...)` / `Printing(...)` rows added or moved; both cards already
  exist.
- `docs/card-sdk-language-reference.md` updated alongside the SDK-adjacent behavior (CLAUDE.md
  SDK-sync rule satisfied).

### Issues

**Minor (author's discretion):**

- `DefenderBypass.isActive` reads base `container.get<CardComponent>()` for `cardDefinitionId` —
  matching the original code. Fine: `cardDefinitionId` is identity, not a projected battlefield
  property. Noted only so it isn't mistaken for a projected-state miss.
- Two `ConditionEvaluator()` instances now exist (the old `DefenderAttackRule.companion` one is
  deleted; `DefenderBypass` and the transformer path each hold their own). Consistent with the rest
  of the codebase; not worth plumbing a shared one for this PR.

### Recommendation

Ship as-is. Nothing to drop or restructure. This is the shape future "printed restriction lifted,
read outside the layer system" badges should copy — the `DefenderBypass` / `ClientStateTransformer`
pairing is now a reusable template alongside the Akawalli block-restriction one.

PR #1346, opened 2026-07-20
Add mana-source provenance + Bat Colony, Brass's Tunnel-Grinder, Thousand Moons Smithy, The
  Everflowing Well (LCI)

bolav:
## Mana-source provenance + the 4 LCI cards it unlocks

Generalizes the engine's mana tracking so payoffs can ask **which kind of source produced the mana
spent to cast a spell** — by producing-source **subtype** and by specific producing **source
permanent**. This is the parameterized generalization of the existing Treasure-only counter
(`ManaPoolComponent.treasureMana`), which becomes just one subtype (`manaBySubtype[Subtype.TREASURE]`).

### Feature

- **Pool provenance** — `ManaPoolComponent` / engine `ManaPool` replace `treasureMana: Int` with
  `manaBySubtype: Map<Subtype,Int>` + `manaBySource: Map<EntityId,Int>`, snapshotted at production by
  the renamed `ManaProvenanceTracker` (so a Treasure sacrificed to tap for its own mana still counts).
- **Payment** — `CastPaymentProcessor` consumes provenance proportionally to the unrestricted mana
  pulled from the pool (the same greedy rule the Treasure counter used) **and** captures the
  provenance of mana freshly tapped by the solver (Caves / mana-source lands are tapped directly,
  unlike Treasure which is filtered out of the solver). Returns a `SpentManaProvenance`.
- **Readback** — the spent provenance rides `SpellCastEvent` (`spentManaSubtypes` /
  `spentManaSourceIds`, replacing `paidWithTreasureMana`), the resolving spell/permanent
  (`SpellOnStackComponent` / `CastRecordComponent.manaSpentBySubtype`, read via `ManaSpentReader`), and
  the turn history (`CastSpellRecord.spentManaSubtypes`).
- **Provenance survives ability activation** — `ActivateAbilityHandler` now carries the provenance maps
  across its `ManaPool`↔`ManaPoolComponent` conversions and consumes them proportional to the floating
  mana an activation spends (mirroring `CastPaymentProcessor`). Without this, tapping a second mana
  source wiped the first source's tag (the old `treasureMana` counter had the same latent gap), so a
  "cast using mana produced by this land" trigger silently missed after any multi-source float.
- **SDK** — `DynamicAmount.ManaSpentFromSubtype(subtype)` (count form, Bat Colony);
  `SpellCastPredicate.PaidWithManaFromSubtype` generalized to any subtype; new
  `SpellCastPredicate.PaidWithManaFromSource` (the DFC-land triggers); `Subtype.CAVE`; `Counters.BORE`
  passive counter. SDK reference doc updated.

### Cards (LCI, +4 → 265/286 booster)

- **Bat Colony** — ETB "a 1/1 Bat for each mana from a Cave spent to cast it" + Cave-you-control-enters
  counter.
- **Brass's Tunnel-Grinder // Tecutlan, the Searing Rift** — front loot + descend bore-counter
  transform; back "cast a permanent using Tecutlan's mana → discover X (spell's MV)".
- **Thousand Moons Smithy // Barracks of the Thousand** — Gnome Soldier token with a
  characteristic-defining P/T; first-main tap-five transform; back "cast an artifact/creature using
  Barracks' mana → make a Gnome".
- **The Everflowing Well // The Myriad Pools** — front mill-2-draw-2 + descend-8 transform; back "cast
  a permanent using its mana → up to one other target permanent you control becomes a copy of that
  spell until end of turn".

### Tests

- `BatColonyScenarioTest` — subtype-count → Bat count (full / none / partial).
- `ManaSourceProvenanceScenarioTest` — real production path: tap a source, cast a permanent with its
  mana → the source's own cast trigger fires; mana from elsewhere does not.
- `AlchemistsTalentTest` — updated to the generalized representation (Treasure provenance still works).
- Card snapshot golden re-blessed; `CardLintTest` green.


bolav:
## Review — mana-source provenance + 4 LCI cards

**Verdict.** Correct behavior, right SDK shape. The change *generalizes* the existing Treasure-only
mana counter instead of adding a parallel system, and the two new SDK atoms read state no existing
primitive can. Full rules-engine + mtg-sdk suites green; LCI golden purely additive.

**What's good**
- `treasureMana: Int` → `manaBySubtype` / `manaBySource`; Treasure collapses to
  `manaBySubtype[TREASURE]` with no special-casing left in `TriggerMatcher` / `ConditionEvaluator`.
- Captures provenance from **solver-tapped** sources, not just the pool — Treasure was pool-only
  because it's filtered out of the solver, but Caves / mana-lands are tapped directly, so pool-only
  consumption would have silently missed them.
- `ManaSpentReader.subtypeSpent` reuses the on-stack/as-permanent dual-read `DistinctColorsManaSpent`
  already relies on, so Bat Colony's ETB count works whether read on the stack or as the permanent enters.
- Tests exercise both halves through real machinery (production → payment → event), plus partial-Cave
  scaling and the Treasure regression.

**Provenance now survives ability activation (found + fixed in this branch)**
- Threading review turned up a real gap: `ActivateAbilityHandler` rebuilt the mana pool positionally
  when a mana ability resolved, dropping the `manaBySubtype`/`manaBySource` maps — so floating mana
  from **two** sources (tap Tecutlan for {R}, then tap another land) wiped the earlier source's tag and
  its trigger silently missed. (The old `treasureMana: Int` was dropped at these same sites, so Treasure
  had the identical latent gap — this wasn't a new regression, just newly exercised.)
- **Fix:** carry the provenance maps across every `ManaPool`↔`ManaPoolComponent` conversion in
  `ActivateAbilityHandler` (build at ~L753; primary writeback ~L1004; repeat-activation ~L1425/L1463;
  auto-tap enriched pool ~L1835; Damping Sphere replacement ~L1180) and **consume proportionally to the
  floating mana actually spent** — the same rule `CastPaymentProcessor` already applies for spells
  (`pay()`/`spend()` leave the maps untouched, so a mana-only ability keeps all prior tags and a
  mana-spending ability consumes what it used). Regression test added:
  *"floating mana from a second source does not wipe the first source's provenance."*
- **Out of scope (same latent pattern, rarer paths):** a handful of other positional
  `ManaPoolComponent(…)` reconstructions still omit the maps — `CostPaymentService.payMana` (morph
  unmorph / pay-or-suffer) and the Cycle/Plot/Foretell/continuation-resumer cost paths. They only
  matter for the niche "float tagged mana → pay a non-cast cost with other mana → cast a payoff"
  sequence and are worth a follow-up sweep, but don't affect the four cards here.
- mtgish Step 8b skipped — the provenance predicates carry a subtype/source argument the emitter
  can't recover cleanly.

Merge-ready.

PR #1347, opened 2026-07-21
Add "whenever you discover" trigger + Curator of Sun's Creation (LCI)

bolav:
## Curator of Sun's Creation + `WheneverYouDiscover` trigger (LCI)

Adds the last **Discover**-referencing LCI card, **Curator of Sun's Creation**, and the small engine
feature it needs — completing Discover (CR 701.57) coverage for the set. LCI is now 266 / 286.

### Feature: `Triggers.WheneverYouDiscover` (CR 701.57)

A "whenever you discover" watcher, built by mirroring the existing scry/surveil watcher plumbing end
to end:

- **SDK:** `EventPattern.DiscoveredEvent(player)`, `Triggers.WheneverYouDiscover`,
  `EmitDiscoveredEventEffect(value)`, `ContextPropertyKey.TRIGGER_DISCOVER_VALUE`.
- **Engine event:** `DiscoveredEvent(playerId, value, sourceName)` + serialization; routed as an
  internal watcher signal (no client render) in `ClientEvent`.
- **Emission timing (CR 701.57b):** a player has "discovered" only after the *whole* process —
  including the "cast for free or put into hand" decision — completes. Discover pauses for that
  decision, and an event emitted in a paused batch does not reliably fire watcher triggers, so
  `DiscoverExecutor` bundles an `EmitDiscoveredEventEffect` into the discover's follow-up tail. It
  therefore fires from a completed resolution at every exit path (hand, free cast, whiff, deck-out),
  and inline on the empty/exhausted branches.
- **Cast-for-free path:** casting the discovered card returns `triggersAlreadyProcessed = true` to
  protect its own `SpellCastEvent` from a re-scan — which also suppressed the tail's brand-new
  `DiscoveredEvent`. The discover continuation resumer now explicitly scans "whenever you discover"
  triggers from that event at both the synchronous and targeted-spell cast-resume completion points
  (no-op for the non-discover cascade/suspend callers of the shared helper).
- **Value payload:** the discover value N (the mana-value threshold) is threaded to the resolving
  effect context via a new `triggerDiscoverValue` field (mirroring `triggerScryCount`) and surfaced
  by `DynamicAmount.ContextProperty(TRIGGER_DISCOVER_VALUE)`, so "discover again for the same value"
  reuses N. Pairs with the existing `oncePerTurn` flag for "triggers only once each turn."
- **TriggerIndex:** a new `DISCOVERED` category routes both the trigger and the runtime event.

### Card

**Curator of Sun's Creation** — {3}{R} 3/3 — `WheneverYouDiscover` + `oncePerTurn = true` +
`Discover(ContextProperty(TRIGGER_DISCOVER_VALUE))`. The once-per-turn flag is the recursion guard:
Curator's own re-discover re-emits the event but can't re-trigger itself, capping the chain at exactly
two discovers.

### Tests

`CuratorOfSunsCreationScenarioTest` pins each clause:

1. re-discovers after you discover, and once-per-turn caps the self-chain at exactly two discovers;
2. the re-discover reuses value N, discriminated by an over-threshold card the second discover must
   skip (a threshold-0 default would whiff);
3. Curator still fires when the discovered card is cast **for free** (regression for the
   `triggersAlreadyProcessed` suppression above);
4. Curator re-triggers even when the discover **whiffs** — all-land library, no nonland ≤ N found
   (CR 701.57b).

Scry / Surveil / Discover / Quintorius / cascade scenario regressions green; LCI card snapshot
re-blessed (Curator only); `CardLintTest` green. `docs/card-sdk-language-reference.md` updated. Manual
playtest scenario at `manual-scenarios/cards/c/curator-of-suns-creation.json`.

https://claude.ai/code/session_01ENc5UfCNoLDA9CKckn5ntj


bolav:
## Review: `WheneverYouDiscover` trigger + Curator of Sun's Creation

**Verdict:** behavior is right and the SDK shape is right. The only genuinely new SDK surface is a
discover-*completion* event no primitive could previously read; the card composes entirely from
existing primitives (`Effects.Discover`, `DynamicAmount.ContextProperty`, `oncePerTurn`). All four
scenario tests pass locally; compile clean. Ready to merge.

### What's good

- **Minimal, justified new SDK.** `WheneverYouDiscover` + `EventPattern.DiscoveredEvent` +
  `TRIGGER_DISCOVER_VALUE` expose a state transition (discover finished) no primitive surfaces — a
  genuinely novel event, not a 1:1 wrapper. The discover value N flows through the exact generic
  channel `TRIGGER_SCRY_COUNT` uses (`ContextPropertyKey` → `TriggerContext.discoverValue` →
  `EffectContext.triggerDiscoverValue` → `DynamicAmountEvaluator`), so the next "discover payoff"
  card needs zero new types. No baked-in constants; the card reuses `oncePerTurn` rather than
  inventing a discover-specific cap.
- **Faithful mirror of the scry/surveil watcher.** `EmitDiscoveredEventEffect` is the deliberate twin
  of `EmitSurveiledEventEffect`, and the reasoning is documented where it diverges (value is a
  resolved integer, so baked in; surveil reads from a pipeline collection). Cross-layer wiring is
  complete: serialization subclass, `TriggerIndex` category (both SDK-event and engine-event
  dispatch), `TriggerMatcher`, `StackComponents`/`StackResolver`, `TriggeredAbilityContinuation`,
  both continuation resumers, `ClientEvent` exhaustive `when` (correctly a no-render internal
  signal), `ClientStateTransformer`.
- **All three discover exit paths emit exactly once** — empty library inline; library-exhausted
  whiff as a tail; discovered-card-found in the continuation `thenEffect` — each citing CR 701.57b,
  each preserving the pre-existing `thenEffect` gating (`finalCardMv <= threshold`).
- **The free-cast edge is handled precisely.** `scanDiscoveredEventTriggers` threads the needle: the
  cast-for-free branch sets `triggersAlreadyProcessed = true` to protect the discovered spell's own
  `SpellCastEvent`, which would otherwise swallow the tail's new `DiscoveredEvent`. The helper's
  guards are correct — `isPaused` early-return (tail hasn't run yet), empty-list no-op for the shared
  cascade/suspend callers, and it re-asserts the flag so there's no double-fire downstream. Uses
  `detectTriggers` (battlefield path), the right entry point for an `ANY`-bound creature trigger.
- **Rules verified exact.** CR 701.57 / 701.57a / 701.57b / 701.57c all match the current Comp Rules
  verbatim as cited. Oracle text, cost, P/T, and metadata match Scryfall exactly. SDK reference and
  snapshot updated in-change.

### Addressed during review

- **701.57b whiff path now tested.** The empty-library and library-exhausted branches emit
  `DiscoveredEvent` specifically to satisfy CR 701.57b ("a player has discovered … even if some or
  all of those actions were impossible"), but no test originally drove a whiffing discover that still
  re-triggers. Added a fourth scenario — `Discover 4` into an all-land library — asserting the
  watcher still fires twice (original whiff + one Curator re-discover, once-per-turn capping the
  chain) with the hand unchanged and the game not left paused.

### Notes

- Verified locally by running `CuratorOfSunsCreationScenarioTest` (all four pass, `BUILD
  SUCCESSFUL`). The broader Discover / Quintorius / cascade regressions were not re-run this round;
  the change is additive and mirrors existing watcher paths, so regression risk is low.

PR #1354, opened 2026-07-23
Add Matzalantli, the Great Door (LCI #256)

bolav:
## Matzalantli, the Great Door // The Core (LCI #256)

Implements the LCI double-faced permanent and the small SDK capability its transform gate needs.

### Front — Matzalantli, the Great Door ({3} Legendary Artifact)
- `{T}: Draw a card, then discard a card.`
- `{4}, {T}: Transform Matzalantli. Activate only if there are four or more permanent types among cards in your graveyard.`

### Back — The Core (Legendary Land)
- `Fathomless descent — {T}: Add X mana of any one color, where X is the number of permanent cards in your graveyard.`

Built from existing primitives — `CardDefinition.doubleFacedPermanent`, `TransformEffect`, `AddAnyColorMana(DynamicAmount.Count)`, `Composite(Draw, Discard)`.

### SDK addition: `Aggregation.DISTINCT_PERMANENT_TYPES`

The transform gate counts **permanent types** (CR 110.4: artifact, battle, creature, enchantment, land, planeswalker), which is *not* the same as distinct card types among permanent cards: a **kindred** permanent (e.g. a kindred enchantment) carries card types `{KINDRED, ENCHANTMENT}`, and KINDRED is a card type but not a permanent type — a kindred card is a permanent only via its *other* type (CR 300.2b). Counting distinct card types would over-count such a card by one.

- New `Aggregation.DISTINCT_PERMANENT_TYPES` — like `DISTINCT_TYPES` but restricted via `CardType.isPermanent`, so instant/sorcery/kindred never contribute. Wired into both `DynamicAmountEvaluator` paths (battlefield + zone) and both SDK description branches.
- New `Conditions.DistinctPermanentTypesInGraveyard(count)` — Delirium's permanent-only sibling; Matzalantli's gate routes through it. `Delirium`'s serialized output is unchanged (only `LCI.json` changed across the whole snapshot corpus).
- `docs/card-sdk-language-reference.md` updated for the new aggregation + condition.

### Tests
`MatzalantliTheGreatDoorScenarioTest` (rules-engine), all green:
- transform blocked at 3 permanent types (instant + sorcery excluded);
- **kindred regression** — creature + artifact + kindred enchantment = 3 permanent types but 4 distinct *card* types → still blocked (fails under the old logic);
- transform allowed at 4 permanent types → flips to The Core, a land;
- The Core's fathomless descent taps for X mana of one color, X = permanent cards in the graveyard.

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 268 / 286
+**Implemented:** 269 / 286
 - [x] Abrade
 - [x] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -157,7 +157,7 @@
 - [x] Marauding Brinefang
 - [ ] Market Gnome
 - [x] Master's Guide-Mural
-- [ ] Matzalantli, the Great Door
+- [x] Matzalantli, the Great Door
 - [x] Mephitic Draught
 - [x] Merfolk Cave-Diver
 - [x] Might of the Ancestors
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -5680,14 +5680,15 @@ default to "you" so card authors don't need to pass it explicitly.
   printed threshold is always four, but `count` is parameterized. Works as a static "as long as …"
   gate (`staticAbility { ability = ModifyStats(...); condition = Conditions.Delirium() }` —
   Spineseeker Centipede) and as an activated-ability `ActivationRestriction.OnlyIfCondition`
-  ("Activate only if there are four or more card types …" — Balustrade Wurm). `Delirium(count)`
-  delegates to `DistinctCardTypesInGraveyard(count, filter = GameObjectFilter.Any)` (below).
-- `DistinctCardTypesInGraveyard(count, filter = GameObjectFilter.Any)` — generalizes `Delirium`:
-  "there are `count` or more distinct card types among the cards in your graveyard that match
-  `filter`." Pass `GameObjectFilter.Permanent` for "N or more **permanent** types among cards in
-  your graveyard" (Matzalantli, the Great Door's transform gate — a permanent card only ever carries
-  permanent card types, so distinct card types among the graveyard's permanent cards is exactly its
-  permanent-type count: artifact, battle, creature, enchantment, land, planeswalker).
+  ("Activate only if there are four or more card types …" — Balustrade Wurm).
+- `DistinctPermanentTypesInGraveyard(count)` — Delirium's permanent-only sibling: "there are `count`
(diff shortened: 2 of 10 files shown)


bolav:
## Review — Matzalantli, the Great Door (LCI #256)

Reviewed the full branch (both commits) against `main`. Ran the scenario suite locally — all green.

### Verdict
Behavior is right and the SDK shape is clean. The card is authored entirely from existing primitives, and the one capability it needs — counting **permanent types** in a zone — is added as a small, correctly-scoped aggregation rather than a one-off condition. Merge-ready.

### What's good
- **Composition-first card.** `CardDefinition.doubleFacedPermanent` (artifact front / land back), `TransformEffect`, `AddAnyColorMana(DynamicAmount.Count(You, GRAVEYARD, Permanent))` for The Core's "any one color" fathomless descent, and `Composite(Draw, Discard)` — no bespoke card code.
- **Faithful oracle text**, verified verbatim against Scryfall including the "any one color" wording (mapped to `AddManaOfChoiceEffect(AnyColor, …)`, not the any-combination variant).
- **The SDK addition is scoped correctly.** `Aggregation.DISTINCT_PERMANENT_TYPES` mirrors the existing `DISTINCT_BASIC_LAND_SUBTYPES` precedent (a distinct-count restricted to a type subset) and reuses the `AggregateZone`/`Compare` AST — no new `Condition`/`Effect` type. `Conditions.DistinctPermanentTypesInGraveyard(count)` reads as Delirium's permanent-only sibling. The exhaustive `when (aggregation)` sites forced full coverage (both engine evaluator paths + both SDK description branches).

### Correctness — the permanent-vs-card-types distinction
The subtle case here is that "permanent types" ≠ "distinct card types among permanent cards". A **kindred** permanent (e.g. a kindred enchantment) carries card types `{KINDRED, ENCHANTMENT}`; KINDRED is a card type but not a permanent type (CR 110.4; CR 300.2b — a kindred card is a permanent only via its *other* type). The gate now counts only permanent card types (via `CardType.isPermanent`), so instant/sorcery/kindred never contribute — matching the printed "permanent types" semantics rather than a card-type approximation. `The Core`'s mana counts permanent *cards* (`GameObjectFilter.Permanent`), which is independently correct.

### Tests
`MatzalantliTheGreatDoorScenarioTest` exercises every clause:
- blocked at 3 permanent types (instant + sorcery excluded);
- **kindred regression** — creature + artifact + kindred enchantment = 3 permanent types but 4 distinct *card* types → still blocked. This case fails under a distinct-card-types-among-permanents implementation, so it genuinely pins the fix;
- allowed at 4 permanent types → flips to The Core, a land;
- The Core taps for X mana of one color, X = permanent cards in the graveyard.

Snapshot re-blessed: only `LCI.json` changed (Matzalantli → `DISTINCT_PERMANENT_TYPES`, default filter); every Delirium card is byte-identical, confirming that refactor is a no-op for existing callers.

### Notes
- No mtgish IR tag maps to "permanent types in graveyard" (the corpus only has `NumCardTypesInGraveyardIs` for delirium/card-types), so there's no bridge/emitter entry to add — a justified skip, not an omission.
- Reminder text lists *battle* as a permanent type; the engine models no `BATTLE` card type, so it's harmless flavor, not a gap.

**Recommendation:** merge as-is.

PR #1375, opened 2026-07-24
Add LCI graveyard-caves batch (Echoing Deeps, Intrepid Paleontologist, Pit of Offerings,
  Starving Revenant)

bolav:
## LCI graveyard-caves batch — four cards + engine support

Implements four Lost Caverns of Ixalan cards (LCI 269 → 273 / 286) sharing a graveyard/exile
and Caves theme, plus the engine/SDK primitives they need. Each card has a scenario test; a
manual self-play scenario bundles all four.

### Cards

- **Starving Revenant** (#123) — `{2}{B}{B}` 4/4. ETB surveil-2-then-draw-per-kept + Descend 8
  draw drain. **Composed entirely from existing primitives** — no new SDK type (Surveil →
  `DistinctEntitiesInCollections("toTop")` driving the draw and `Multiply(...,3)` life loss; a
  `YouDraw` trigger gated by an intervening-if on eight permanent cards in the graveyard).
- **Pit of Offerings** (#278) — Land — Cave. ETB exiles up to three cards from graveyards
  (linked to the land); a mana ability adds one of the exiled cards' colors. New
  `ManaColorSet.AmongLinkedExiledCards`.
- **Intrepid Paleontologist** (#193) — `{1}{G}` 2/2. `{2}` exile-and-link a graveyard card;
  cast Dinosaur creatures from that linked exile, entering with a finality counter. New
  `entersWithCounter` rider on `GrantMayCastFromLinkedExile` (mirrors
  `MayCastFromGraveyard.entersWithCounter`) + subtype-aware cast filtering.
- **Echoing Deeps** (#271) — Land — Cave. Enters tapped as a copy of a graveyard land, plus
  Cave. New `EntersAsCopy.tappedIfCopied` and a direct-entry clone path
  (`CloneEntersOnBattlefieldContinuation`) for lands played onto the battlefield.

### Engine / SDK

- `ManaColorSet.AmongLinkedExiledCards` — colors among the source's still-in-exile linked pile.
- `EntersAsCopy.tappedIfCopied` + `PlayLandHandler` / `CloneEntersOnBattlefieldContinuation` —
  as-enters copy (CR 707.2) applied to a permanent already placed on the battlefield directly.
- `GrantMayCastFromLinkedExile.entersWithCounter` — cast-this-way finality-counter rider.
- `CastZoneResolver.matchesCardFilter` unified for both legal-action enumeration and cast-time
  validation; made exhaustive over `CardPredicate` (fixes subtype gating and a latent
  `PowerOrToughnessAtMost` widening).
- SDK reference and the per-set card snapshot golden updated.

### Review + follow-up

The branch was self-reviewed (SDK-elegance + correctness lens). One important item (duplicated
clone-enters resumers) and three minor items were raised and **all fixed in the final commit**
(`Address review feedback …`): shared `applyCopyToEntity` helper, exhaustive `matchesCardFilter`,
a single enriched entry `ZoneChangeEvent` for the copy path, and a direct owner-keyed exile
lookup. The full review is posted as a PR comment.

### Testing

`rules-engine` scenario tests for all four cards pass, plus a regression net over the touched
clone/copy and cast-from-zone paths (`CloneTest`, `SharkShredderKillerCloneTest`,
`ValgavothTerrorEaterScenarioTest`, `DawnhandDissidentTest`, `GraveyardCastGrantChoiceTest`).

### DIFF
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 272 / 286
+**Implemented:** 273 / 286
 - [x] Abrade
 - [x] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -236,7 +236,7 @@
 - [x] Squirming Emergence
 - [x] Staggering Size
 - [x] Stalactite Stalker
-- [ ] Starving Revenant
+- [x] Starving Revenant
 - [x] Staunch Crewmate
 - [x] Stinging Cave Crawler
 - [x] Subterranean Schooner
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -88,7 +88,7 @@
 - [x] Earthshaker Dreadmaw
 - [x] Eaten by Piranhas
 - [x] Echo of Dusk
-- [ ] Echoing Deeps
+- [x] Echoing Deeps
 - [x] Enterprising Scallywag
 - [x] Envoy of Okinec Ahau
 - [x] Etali's Favor
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 269 / 286
+**Implemented:** 270 / 286
 - [x] Abrade
 - [x] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -131,7 +131,7 @@
 - [x] Idol of the Deep King
 - [x] In the Presence of Ages
 - [x] Inti, Seneschal of the Sun
-- [ ] Intrepid Paleontologist
+- [x] Intrepid Paleontologist
 - [x] Inverted Iceberg
 - [x] Ironpaw Aspirant
 - [x] Itzquinth, Firstborn of Gishath
--- a/backlog/sets/the-lost-caverns-of-ixalan/cards.md
+++ b/backlog/sets/the-lost-caverns-of-ixalan/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 286 booster cards (excluding basic lands and tokens)
 **Release Date:** November 17, 2023
-**Implemented:** 270 / 286
+**Implemented:** 272 / 286
 - [x] Abrade
 - [x] Abuelo's Awakening
 - [x] Abuelo, Ancestral Echo
@@ -183,7 +183,7 @@
 - [x] Pathfinding Axejaw
 - [x] Petrify
 - [x] Pirate Hat
-- [ ] Pit of Offerings
+- [x] Pit of Offerings
 - [x] Plundering Pirate
 - [x] Poetic Ingenuity
 - [x] Poison Dart Frog
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -7451,7 +7451,10 @@ levels, saga chapters, faces — is covered automatically. What it checks:
   `ChooseOption` / a cast-time additional cost, …) in the same resolution scope. A read written
   *nowhere* on the card is an **error** (typo → silent no-op); read-before-write and
   cross-resolution reads are warnings, as are stores nothing reads. A collection write `x`
-  also satisfies the numeric read `x_count`.
+  also satisfies the numeric read `x_count`. Macro effects that the engine expands into a
+  pipeline count as writers of the collections that expansion seeds: `Scry` writes `toTop` /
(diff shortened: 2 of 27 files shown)


bolav:
## Review — LCI graveyard-caves batch (SDK-elegance + correctness)

**Verdict.** Strong batch. Behavior is correct across all four cards and the SDK shape is mostly
elegant. **Starving Revenant is exemplary** — zero new SDK types, pure composition. One important
item (duplicated clone-enters resumers) and three minor items were raised in review and have all
been addressed in the follow-up commit (`Address review feedback …`).

### What's good — keep this

- **Starving Revenant** composes entirely from existing primitives: `Surveil(2)` →
  `DistinctEntitiesInCollections("toTop")` driving both `DrawCards` and `Multiply(...,3)` life
  loss, plus a `YouDraw` trigger gated by an intervening-if
  `CardsInGraveyardMatchingAtLeast(8, Permanent)`. No new `Effect`/`DynamicAmount`.
- **CastZoneResolver dedup** is a genuine bug fix: the enumerator previously had its own 4-branch
  `when` with `else -> true`, so a `Creature.withSubtype("Dinosaur")` grant silently matched
  *every* card. Unifying both call sites on the shared `matchesCardFilter` companion fixes it, and
  the "non-Dinosaur cannot be cast" test exercises exactly that.
- **Rider reuse**: `GrantMayCastFromLinkedExile.entersWithCounter` mirrors
  `MayCastFromGraveyard.entersWithCounter` and rides the existing `GraveyardCastRiderComponent`
  plumbing rather than inventing a new path.
- Both exhaustive `when(colorSet)` sites (`ManaColorSetResolver`, `LandManaColorInspector`) updated
  for the new sealed variant; `card-sdk-language-reference.md` updated; snapshot re-blessed. Cited
  CR numbers (603.4 intervening-if, 707.2 copiable values, 400.7 zone-change, 106.7
  lands-could-produce) all verified.
- Tests hit the real corners: copy declined / no-candidate (Echoing Deeps), colorless / empty /
  left-exile pools (Pit), descend boundary 7-vs-8 and keep-none-vs-both (Revenant).

### Issues raised — all fixed in the follow-up commit

- **[Important] Resumer duplication** (`ModalAndCloneContinuationResumer.kt`).
  `resumeCloneEntersOnBattlefield` re-implemented ~50 lines near-identical to `resumeCloneEnters`
  (copy `CardComponent` → additional subtypes/keywords → name/P-T overrides → `CopyOfComponent`
  snapshot). **Fixed:** extracted a shared `applyCopyToEntity(...)` helper used by both paths, each
  keeping its own owner semantics via an explicit `newOwnerId` param.

- **[Minor] `matchesCardFilter` kept `else -> true`** for unknown predicates. As the shared cast
  gate, an unhandled predicate silently widened a grant. **Fixed:** made the `when` exhaustive over
  all `CardPredicate` variants — everything derivable from a `CardComponent` is now evaluated
  (this also fixed a latent `PowerOrToughnessAtMost` bug where `Creature.powerOrToughnessAtMost(1)`
  matched every creature); context-dependent predicates fail closed. A new predicate now forces a
  decision here.

- **[Minor] Land-copy entry event** — the emitted `ZoneChangeEvent` was the printed-name one
  carried from `PlayLandHandler`, missing `copyOfOriginalName` (consumed at `ClientEvent.kt:847`),
  while the resumer's enriched event was used only for trigger detection. **Fixed:** consolidated to
  a single enriched entry event emitted by the resumer, matching the spell-clone path.

- **[Minor] `amongLinkedExiledCards`** scanned every zone per exiled id. **Fixed:** direct
  owner-keyed `state.getExile(ownerId)` lookup (non-battlefield zones are keyed by owner).

### Recommendation

Merge-ready. All review items are resolved; the regression net (four new cards + `CloneTest`,
`SharkShredderKillerCloneTest`, `ValgavothTerrorEaterScenarioTest`, `DawnhandDissidentTest`,
`GraveyardCastGrantChoiceTest`) is green.

bolav:
Fixed a `CardLintTest` failure surfaced by this branch.

**Symptom** — `CardLintTest > every card's pipeline variables, target bindings, and choice slots resolve` FAILED with two `UnresolvedPipelineRead` errors on **Starving Revenant** (LCI #123):

> `DistinctEntitiesInCollections.collections reads collection 'toTop', but nothing on this card writes it — the step would silently no-op.`

**Root cause — a linter false positive.** Starving Revenant's ETB reads the surveil remainder (`toTop`) to drive its draw and life-loss. That collection is written by `Effects.Surveil`'s expansion (`LibraryPatterns.surveilPipeline`, `storeRemainder = "toTop"`), but that expansion only happens in the engine at resolution time. On the static card tree the linter walks, `Surveil` is an opaque node, so the writer is invisible and the reads look unwired. The card itself is correct — `StarvingRevenantScenarioTest` asserts the runtime behavior (keep both surveiled cards on top → draw 2, lose 6 life).

**Fix.** Extended `CardLinter.implicitAccesses` — the existing hook for participation not visible as a string field (already used so token executors register a `createdTokens` write) — to record the collections the expandable macros seed:

- `Scry` → writes `toTop`, `toBottom`
- `Surveil` → writes `toTop`, `toGraveyard`

Collection names mirror `scryPipeline`/`surveilPipeline`. The writes are implicit (field `"(implicit)"`), so surveils that never read `toTop` are not flagged as orphan writes. The fix is generic — it covers any future card that inspects a scry/surveil result — rather than an allowlist entry. Also documented the macro-seeded collections in `card-sdk-language-reference.md` §21.

`CardLintTest` now passes (`BUILD SUCCESSFUL`).

PR #1376, opened 2026-07-24
Add LCI gw-power batch (Kutzil, Sovereign Okinec Ahau, The Skullspore Nexus, Kellan)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1375

## LCI gw-power batch — four cards + engine support (stacked on `lci-graveyard-caves`)

**Stacked PR** — base is `lci-graveyard-caves`; this diff is only the gw-power delta (the other batch cards live in the base branch). Adds four Lost Caverns of Ixalan cards plus the SDK/engine primitives they need. Each card has a scenario test; a manual self-play scenario bundles them.

### Cards

- **Kutzil, Malamet Exemplar** (#232) — `{1}{G}{W}` 3/3. `PlayersCantCastSpells` (opponents, your turn) + a combat-damage batch trigger over `Creature.powerGreaterThanBase()`, drawing once per combat-damage step. Reuses the established `OneOrMoreDealCombatDamageToPlayerEvent` idiom.
- **Sovereign Okinec Ahau** (#240) — `{2}{G}{W}` 3/4. Ward {2}; attack trigger puts +1/+1 counters equal to `power − base power` on each qualifying creature, composed from `ForEachInGroup` + the new `EntityNumericProperty.BasePower`.
- **The Skullspore Nexus** (#212) — `{6}{G}{G}` Legendary Artifact. Cost reduction by greatest power among your creatures, a nontoken creatures-died batch payoff (token base P/T = summed last-known power), and a "double target creature's power" ability.
- **Kellan, Daring Traveler // Journey On** (#231) — attack reveal/branch; the Adventure creates `X` Map tokens via the dynamic `CreateMapToken(DynamicAmount)`.

### Supporting SDK / engine

- `CardPredicate.PowerGreaterThanBase` self-relative filter + `EntityNumericProperty.BasePower`/`BaseToughness` reads (kept consistent so the Sovereign "difference" is exact).
- `ContextPropertyKey.DIED_BATCH_TOTAL_POWER` — summed last-known power of a creatures-died batch, captured at detection (CR 603.2c / LKI).
- `CostReductionSource.GreatestPropertyAmongPermanentsYouControl(property, filter)` — one property-parameterized cost source replacing the `GreatestManaValue…`/`GreatestPower…` twins; Sunderflock migrated.
- `Effects.CreateMapToken(DynamicAmount)` sugar over the existing dynamic-count token effect.

### Tests

Scenario tests for all four cards (`rules-engine` `scenarios/`) covering the rule corners: batch once-per-combat-step, nontoken exclusion + power sum, power == base no-op, per-creature differences, and the Adventure Map count. Card snapshot goldens re-blessed (LCI + ECL for the migrated cost source). All green.


bolav:
## Review — LCI gw-power batch (stacked on `lci-graveyard-caves`)

Scope: the delta over `lci-graveyard-caves` only — **Kutzil, Malamet Exemplar**, **Sovereign Okinec Ahau**, **The Skullspore Nexus**, **Kellan, Daring Traveler // Journey On**, and the SDK/engine primitives they add. Reviewed through the SDK-elegance lens, then correctness/tests. Two issues were caught and **fixed in this branch** (commit `Fix review findings on LCI gw-power batch`); details below.

### Verdict

Strong card modeling — almost everything composes existing primitives, oracle texts match Scryfall exactly, and the tests hit the real rule corners. The branch initially did **not compile** (a missing exhaustive-`when` branch) and shipped a pair of near-duplicate cost sources; both are now resolved. After the fixes, `rules-engine` compiles and all four cards' scenario tests pass.

### What's good

- **`BasePower`/`BaseToughness` + `PowerGreaterThanBase` kept mutually consistent.** Sovereign's per-creature `Subtract(Power, BasePower)` is anchored to the *same* base its filter tests, so the "difference" is provably ≥ 1 for every group member. The right way to introduce a paired primitive.
- **`diedBatchTotalPower` captured at detection time (LKI).** Correct per the 2023-11-10 ruling ("as they last existed on the battlefield") — a graveyard read would drop counters/buffs. Token deaths are excluded from both the fire condition and the sum.
- **Kutzil's combat-damage batch trigger** follows the established `OneOrMoreDealCombatDamageToPlayerEvent` idiom (Kastral, Invasion Tactics) — "you control" is handled by the batch grouping, no bespoke filter needed. Fires once per combat-damage step, not per creature.
- **Kellan / Journey On** is pure composition: `CreateMapToken(DynamicAmount)` sugar over the existing dynamic-count token effect; the attack reveal branches via `CollectionContainsMatch` with a proper "may"-decline path.
- **Tests exercise the rule-corners:** batch once-per-step (Kutzil/Skullspore), nontoken exclusion + power sum, power == base no-op, distinct per-creature differences (Sovereign), and the Adventure Map count with/without an opponent artifact.
- Oracle texts verified against local Scryfall LCI — all four match exactly, including collector numbers. SDK reference updated for every new type.

### Issues

**Blocking — FIXED.** `rules-engine` failed to compile: `CastZoneResolver.matchesCardPredicate`'s now-exhaustive `when` (the fail-closed rewrite lives in the base branch) was missing the `CardPredicate.PowerGreaterThanBase` case this batch adds, and has no `else`. Fixed by adding it to the fail-closed group — a card sitting in a zone has no projected pump, so its power never exceeds its own base (`false`, mirroring `CostCalculator`'s static treatment). Without this, no tests could run.

**Important — FIXED.** `CostReductionSource.GreatestPowerAmongPermanentsYouControl` was an over-specialized twin of the existing `GreatestManaValueAmongPermanentsYouControl`, differing only in which numeric property they read. Consolidated into one property-parameterized `GreatestPropertyAmongPermanentsYouControl(property: EntityNumericProperty, filter)`, collapsing the two `CostCalculator` helpers into one `greatestPropertyAmongMatching(...)`. Migrated The Skullspore Nexus (`Power`) and the existing Sunderflock (`ManaValue`); re-blessed the LCI + ECL snapshot goldens. The next "greatest \<property\> among X" card now reuses the type instead of adding a third.

**Minor — not changed (author's discretion).** `TriggerDetector.deathMatchesFilter` honors only `IsCreature`/`HasSubtype`/`IsNontoken`/`IsToken`; any other predicate in a death-batch trigger filter falls through `else -> true` and would inflate both the fire condition and the power sum. No current card hits it (Skullspore uses `Creature.nontoken()`), but it's a latent trap worth a fail-closed default the day a fancier "creatures you control die" filter appears.

### Result

Both blocking/important findings fixed in-branch; `rules-engine` compiles and the four scenario tests are green. Good to merge once the base PR lands.

PR #1390, opened 2026-07-24
Store replays only in the database, in a format that survives deploys

wingedsheep:
## Why

Two problems, one subsystem.

**Replays lived in three places.** The `game_replays` table, an in-memory ring buffer of the last 100 games (`GameHistoryRepository`), and — for games still in progress — the Redis session blob. Three lifetimes, three eviction rules, three answers to "is this game replayable".

**And they stopped replaying.** The record is an *input log* (seed + decks + ordered actions), re-simulated on demand. That reproduces the original game only while the engine folding it behaves as it did on the day. In this engine cards are **data the engine folds through**, so editing a card rewrites the past. `ReplayReconstructor` truncated at the first action that no longer applied — and worse, when actions kept applying but produced a different board, it silently rendered a game that never happened.

## Storage: one home

- `GameHistoryRepository` deleted; `PersistentGameSession` carries no replay data.
- `ReplayStore` is the single home for finished records *and* in-flight recordings — `JdbcReplayStore` when accounts (and a database) are on, a bounded `InMemoryReplayStore` when there isn't one.
- `ReplayCheckpointFlusher` flushes in-progress recordings every few seconds; `RedisGameRepository` resumes them on restart.
- The flush is on a timer, so a crash can lose a recording's tail. Appending the rest of the game onto that short prefix would produce a record of a game nobody played, so each flush writes a position fingerprint and `GameSession` **refuses to resume** when the recovered state disagrees — keeping the shorter honest replay. Abandoned recordings are finalized rather than left mid-write.
- Player history is now an indexed join on a seat table instead of a scan of the in-memory cache.

## Durability: format v2

| | What | Cost |
|---|---|---|
| `pinnedCards` | Compiled `CardDefinition` of every card in the decks, overlaid on the live corpus via a child `CardRegistry`. Card edits stop mattering; ability ids also stay stable, so recorded yields keep matching. | ~7 KB gzipped |
| `checkpoints` | A cheap position fingerprint every 20 actions. Turns *silent* drift into a detected, located divergence. `ReplayFidelity` = `EXACT` / `UNVERIFIED` / `DIVERGED`. | ~30 B each |
| `presentation` | The `{initialSnapshot, deltas}` stream, materialized just after game over (background thread, off the game-over path) and stored gzipped. A **result** rather than a recipe, so it renders regardless of what the engine does next. | ~160 KB gzipped |

`ReplayService` serves the re-simulation while it is faithful — current view code, and "share frame as scenario" works because a real `GameState` exists — and the archive when it isn't, flagged `degraded`. The viewer shows a **From archive** badge and hides the scenario buttons when no state can be rebuilt.

Measured on POR at ~1650 actions/game: input log **~11 KB** stored, archive **~160 KB**. The recipe is 14x cheaper than the result, which is exactly why the recipe is the record and the result is insurance.

## Compatibility

v1 records decode and reconstruct unchanged. Every added field defaults to empty and decoding ignores unknown keys, so records round-trip in **both** directions across a rolling deploy. `engineVersion` (the deploy's git sha, via a new `COMMIT_HASH` build arg mirroring the frontend) is stamped on every record, so a replay that stops re-simulating can be traced to the build that recorded it.

`V10__replay_durability.sql` is additive. Rows written before it keep working and stay reachable by share link, but have no seat children and so drop out of a player's own replay list — noted in the migration.

## Research

The hybrid here is the standard answer to "replays break on patch": inputs for size, recorded output for longevity ([input vs state recording tradeoffs](https://www.gamedeveloper.com/design/instant-replay-building-a-game-engine-with-reproducible-behavior), [why Dota 2 / SC2 replays break](https://us.forums.blizzard.com/en/sc2/t/old-5012-version-replays-are-broken-after-5013-patch/28404)), plus event-sourcing's versioned-record + tolerant-decode discipline ([schema evolution and upcasting](https://artium.ai/insights/event-sourcing-what-is-upcasting-a-deep-dive)). Forge's macro replay is prompt-aware and simply stops when the state diverges — the same failure we had. The twist that fits *this* engine is that content is pure data, so pinning it is cheap and buys most of the durability outright.

## Tests

- `ReplayDurabilityTest` — a recorded game re-simulates **byte-identically after its cards are changed underneath it** (and demonstrably does *not* without the pin); pinning preserves ability ids; a corrupted checkpoint stops the replay instead of rendering a wrong game; archive fallback vs. faithful re-simulation vs. degraded-with-no-archive; v1 decode; codec round-trip.
- `ReplayStorageTest` — store queries, in-progress isolation, resume gate accepting a matching fingerprint and refusing a stale one, abandoned-record finalization.
- `FlywayMigrationTest` — V10 schema, seat join, cascade, in-progress lookup (verified against real Postgres via testcontainers).
- `just build` / `just test` green; `npm run typecheck` green.

## Note on the SDK-reference rule

`AGENTS.md` requires `card-sdk-language-reference.md` updates for SDK additions. The two SDK changes here (`CardExporter.exportToCompactJson`, `CardLoader.fromJsonPreservingIds`) are serialization utilities, not authoring vocabulary — that document catalogs the card DSL and has no section for the loader/exporter, so there was no catalog entry to keep in sync. `data-contracts.md`, `accounts-and-persistence.md` and `architecture-principles.md` are updated.


wingedsheep:
## Review

**Verdict:** the design is right and the SDK stays clean — this is a storage/durability change that adds no card-facing vocabulary, and the two SDK additions it does make are genuinely new capabilities rather than one-offs. I found two ways the flush sweep could lose a replay outright, both in the one new class that had no test. **I've fixed both and pushed** (`159bac2497`); details below.

### What's good

- **No new SDK types where composition would do.** No `Effect`, `Condition`, `Trigger` or `Modification` variant in a ~1900-line diff. `CardRegistry(parent = …)` is the right shape for the overlay — a generic layered lookup rather than a replay-specific hook — and it's parameterized for the next caller, not just this one.
- **`fromJsonPreservingIds` earns its existence.** Id-preserving decode is semantically distinct from `fromJson`, the KDoc says exactly when *not* to use it, and both share one private `decode`. The reasoning for skipping `card-sdk-language-reference.md` checks out: `CardExporter`/`CardLoader` appear nowhere in that document, so there was no catalog entry to keep in sync.
- **The recipe-vs-result framing is the correct diagnosis** and the layering (pin the cards → fingerprint the rest → archive as last resort) attacks the failure modes in cost order rather than reaching straight for frame storage.
- **`ReplayDurabilityTest` tests the actual claim.** "Re-simulates byte-identically after its cards are changed underneath it" *and* demonstrably not without the pin, in one test, is the right way to write that — the negative control is what makes it mean something. The v1-decode and fingerprint-mismatch cases are equally well-aimed.
- Deleting `GameHistoryRepository` and stripping replay data out of `PersistentGameSession` genuinely removes a class of "two stores disagree" bug rather than papering over it. `V10` is additive and the migration comment is honest about what pre-V10 rows lose.
- Yields don't perturb the fingerprint (`withYield` only touches `yieldsByPlayer`, none of the digested fields), and `recordAction` is called immediately after `gameState = result.state` at every one of its seven call sites — so the live stamp and the reconstructed stamp describe the same position. That's load-bearing and it holds.

---

### Blocking — fixed in `159bac2497`

**1. A flush sweep overlapping game over overwrote the finished record.** `GamePlayHandler.handleGameOver` stores the final record at :685 but only removes the session at :781 — with `matchResultSink`, ranked ELO and the LLM callback in between. Throughout that window the session is still in `gameRepository.findAll()` and still returns a non-null `getReplaySetup()`. `flushSession`'s only guard is `flushed[sessionId] == actions.size`, and the game's last actions land after the previous flush, so the count *has* moved — the sweep writes `StoredReplay(status = IN_PROGRESS, presentation = null, endedAt = "")` straight over the finished record. That loses the archive, `winnerName`, `endedAt` and `tournamentName`, and drops the game out of its own players' history for good, because `forget()` has already removed the id that the abandoned-record sweep uses — nothing ever revisits a record no session is being flushed for.

Fix: `ReplayService.saveInProgress` now refuses to downgrade a `FINISHED` record, making the ordering irrelevant. `flushSession` additionally returns early for `session.isGameOver()`, so the encode isn't paid for either.

**2. A crash stranded every recording it was mid-write on.** The abandoned-record sweep is `flushed.keys - liveIds`, and `flushed` is per-process and empty at boot. So an `IN_PROGRESS` row whose session didn't come back (Redis TTL, Redis wiped, session swept while the server was down) has no live session, no flusher entry, and `IN_PROGRESS` is excluded from `findRecentForPlayer` — a silently unwatchable replay, forever. `ReplayService.inProgress()` was added for exactly this and had no caller.

Fix: the first sweep reconciles the store's in-progress records against the live set — adopting the ones that came back (seeding `flushed` so the next flush continues the record), finalizing the ones that didn't. `SessionRecoveryService` runs on `@PostConstruct` and `@Scheduled` starts after context refresh, so recovery has provably finished by then. Noted in the KDoc that this shares the single-instance assumption `SessionRecoveryService` already bakes in — if the server is ever scaled out, both this and the existing abandoned sweep need revisiting, since `RedisGameRepository.findAll()` only sees locally cached sessions.

Both are covered by new cases in `ReplayStorageTest` ("a flush that lands after game over cannot downgrade the finished record", "an in-progress record stranded by a crash is finalized on the first sweep"). `ReplayCheckpointFlusher` previously had no test at all, which is where both bugs were.

---

### Important — your call, not fixed

**`ReplayFingerprint` digests two engine-internal counters, so it false-positives on behaviour-neutral changes.** `nextEntityId` and `timestamp` move whenever anything anywhere mints an extra entity or bumps the clock, including changes with no player-visible effect. Such a replay re-simulates to *the same game* but reports `DIVERGED`, which costs real things: "share frame as scenario" goes away, and serving switches to the 160 KB archive instead of the ~11 KB input log. `nextEntityId` is also the sharpest genuine drift detector you have, so this is a deliberate trade rather than a bug — but it means `EXACT` will be rare after any core-engine work, and the archive becomes the effective serving path much sooner than the "14× cheaper, so the recipe is the record" framing suggests. If you'd rather bias toward keeping re-simulation: drop the two counters and lean on zone sizes + life + turn/phase/priority, which are what a player would notice.

### Minor

- `restoreReplayRecording` resumes without verifying when `expectedFingerprint` is null (`GameSession.kt:1510`). In practice the flusher always writes one, so the branch is unreachable — but it silently takes exactly the risky path the rest of the design exists to prevent. Worth a `logger.warn` so it can't go unnoticed if it ever becomes reachable.
- `PublicReplayController.getReplay` splices with `payload.body.removePrefix("{").removeSuffix("}")`, coupling the endpoint to `ReplayPresentation.compose`'s exact brace shape — and, for archived rows, to whatever an old build happened to store. Exposing the inner fragment on `ReplayViewerPayload` (or storing it that way) would make the contract explicit instead of positional.
- `CardRegistry.size` is now `allCardNames().size` for a child registry — a full set allocation behind a property that reads O(1). Only used in tests today, so harmless, but easy to misuse later.

### Housekeeping

I merged `origin/main` into the branch before reviewing (clean, no conflicts) and pushed it along with the fix, so the branch now carries a merge commit — pull before you continue.

**Verified:** `:game-server:test --tests "*ReplayStorageTest" --tests "*ReplayDurabilityTest"` green (14 tests, module compiles), `tsc --noEmit` clean on the web client. Full-module and Postgres/testcontainer suites not re-run. No CR rule numbers cited anywhere in the diff, and no card definitions touched, so §2b printing placement doesn't apply.

### Recommendation

Merge once you've decided on the fingerprint-sensitivity trade. The two blocking bugs are fixed and tested; the minors are discretionary.

wingedsheep:
## Correction: retracting the `ReplayFingerprint` finding

The "Important" item in my review above — that digesting `nextEntityId` and `timestamp` makes the fingerprint over-sensitive, and that you could bias toward re-simulation by dropping them — **is wrong. Please disregard it.** Acting on it would have introduced the exact bug the checkpoints exist to prevent.

**Why.** `GameState.newEntity()` mints ids as `EntityId("e$nextEntityId")` (`GameState.kt:876`), and recorded actions carry raw `EntityId` values — `cardId`, `sourceId`, `damageDistribution` keys, `conspiredCreatures`, `casualtyCreature`. So a `nextEntityId` shift of one re-numbers every subsequently minted object, and every later recorded action that names `e47` binds to a *different* object. Those actions mostly still apply — the types line up — while doing something else entirely. There is no such thing as a behaviour-neutral entity-counter shift for an input-log replay: a counter shift **is** target rebinding **is** a different game. `nextEntityId` isn't a noisy proxy for drift, it's the earliest and most reliable detector of it, and `DIVERGED` on a mismatch is the correct and required verdict.

`timestamp` I also mischaracterized as an internal counter. `EffectSorter.kt:61,64` sorts same-layer continuous effects by it — it's the Rule 613.8 tiebreaker. A timestamp difference can change real outcomes, so it belongs in the digest too.

**On the guideline question this came from:** neither field is an architectural violation. `docs/architecture-principles.md` §2.1 states the contract as determinism of `(GameState, GameAction) -> ExecutionResult`, and both are declared `GameState` fields — digesting them checks exactly that contract. `ReplayFingerprint` picked the right fields for the right reasons, and its KDoc already names both and explains why.

Same-build determinism holds and is covered: `ReplayDurabilityTest` reconstructs a real recorded game to `EXACT` and compares `entities`/`zones` byte-for-byte. So a `DIVERGED` verdict can only mean something genuinely changed between record and replay — which is the honest answer, and the one the archive fallback exists to serve.

Nothing else in the review changes: the two blocking bugs and their fixes (`159bac2497`) stand, as do the three minors. Net effect of this correction is that `ReplayFingerprint` needs no change at all.

wingedsheep:
## Minors fixed — `addcaf0063`

**Splice contract.** `PublicReplayController` was stripping braces off the payload inline (`body.removePrefix("{").removeSuffix("}")`), which made "the body is a brace-wrapped JSON object" a positional assumption at the call site — an archive row written by another build that violated it spliced into corrupt JSON that only failed in the client. Moved to `ReplayViewerPayload.bodyFields()`, which asserts the shape once and fails where the cause is visible. Still a string operation: splicing genuinely does beat decode-and-re-encode at these sizes.

**Unverified resume.** `restoreReplayRecording` resumed silently when `expectedFingerprint` was null — the one path that can still extend a stale prefix, which is the failure the whole fingerprint gate exists to stop. Unreachable today (the flusher always writes one for a recorded session), now logged and documented so it can't become reachable quietly.

**`CardRegistry.size`.** Built a full union set on every read for an overlay; it now counts the parent's unshadowed names directly, and the KDoc states the complexity rather than leaving an O(names) property looking like O(1).

**Plus `CardRegistryTest`** (new). The overlay branches this PR added to `CardRegistry` had no direct coverage — they were only exercised incidentally through the pin tests. It now pins the cases that actually bite for a *partial* overlay, where almost every lookup falls through: name shadowing vs. fall-through, `size`/`allCardNames` counting a shadowed name once, `landCardNames` losing a name when the overlay redefines it as a non-land, `getCardsByName` keeping parent printings the overlay never pinned (overlay first), and the KDoc's "registering into a child never mutates the parent" claim, which nothing checked.

**Verified:** `:game-server:test --tests "*ReplayStorageTest" --tests "*ReplayDurabilityTest"` (20 tests) and `:rules-engine:test --tests "*CardRegistryTest" --tests "*PrintingRegistryTest"` (25 tests) all green.

That clears everything from my review. `ReplayFingerprint` needs no change (see the retraction above) — nothing outstanding from me.

wingedsheep:
## Second pass — concurrency and failure paths

I went back over this looking specifically at what I'd skipped first time: thread interleavings between the sweep and the game loop, and what happens when a write fails. Two more real problems, both fixed in `503a87e6b2`. Full `:game-server:test` suite green.

### The resume gate could be defeated by its own read

This is the one I'd want caught before production. The gate compares a stored fingerprint against the state recovered after a restart and refuses to keep recording when they disagree — sound. But `flushSession` sampled the action log and the fingerprint through *separate* getters, and the game thread advances between them under `stateLock`. So a flush could store N actions next to the fingerprint of position N+2. A crash at exactly N+2 then makes the recovered state **match** the stored fingerprint, the gate passes, and recording resumes with a two-action hole in the log — the fictional replay the gate exists to prevent, reached through the gate itself. Checkpoints only catch it if one happens to land after the hole; a hole in the final 20 actions is invisible.

`GameSession.replayRecordingSnapshot()` now reads actions, yields, checkpoints, the fingerprint and game-over state under one lock, and the flusher consumes that snapshot. `getReplayResumeFingerprint()` is gone — it couldn't be used safely on its own. The new test pins the invariant directly rather than trying to race a thread: *a snapshot's fingerprint is the position its own action log re-simulates to.*

### Game over had no failure isolation

`handleGameOver` has no `try`/`catch` anywhere in its ~190 lines, and `replayService.save` sits two thirds of the way down it. A throw there skips the match-history row, the ranked ELO update, the LLM tournament callback, `gameRepository.remove(gameSessionId)` and `aiGameManager.cleanupGame` — so a database hiccup on the *least* important write in the method silently drops a player's stats and leaks the session as a zombie.

The exposure predates this PR but the PR widens it: every game is now stored synchronously, where AI-only games previously skipped the DB entirely (`durable = hasHumanSeat`), and the write gained four columns plus a child collection. Wrapped it so everything after still runs. Nice side effect of leaving `forget` inside the wrapper: on failure the id stays in the flusher's map, so the next sweep finalizes the partial record and the game stays watchable anyway.

### Smaller

`getPinnedCards()` serialized ~34 card definitions while holding `stateLock` — tens of ms of JSON blocking the game thread on each game's first flush. Moved off the lock; it only reads decklists, which are immutable once the game starts.

`GameReplayPlayerRow` was the only `@MappedCollection` child in the schema without `@Id` (`MatchParticipantRow` and `MatchCardRow` both have one). Added, matching the precedent.

### The one thing I'd want before calling this production-ready

**No test exercises the Spring Data JDBC path at all** — and that is the production path. `accounts.enabled` is set to `true` in no test in the repo, so `JdbcReplayStore`, `GameReplayRepository`, the new columns and the new child collection have never been round-tripped through Spring Data JDBC. `FlywayMigrationTest` proves the *schema* (raw JDBC), not the *mapping*.

I checked the mapping by hand and it looks right — `tournamentRound`→`tournament_round`, `engineVersion`→`engine_version`, `resumeFingerprint`→`resume_fingerprint`, child `@MappedCollection(idColumn = "replay_id")` against the FK, all consistent with the default naming strategy — and it now matches the two existing precedents exactly. So this is untested-ness rather than a suspected defect, and with the isolation above the failure mode is a loud log plus no replays, not a broken game-over path.

I deliberately didn't write that harness: doing it faithfully means `@SpringBootTest` with `accounts.enabled=true` and `@DynamicPropertySource` pointed at the existing Postgres testcontainer, which is new test infrastructure for this repo, and hand-wiring a `JdbcAggregateTemplate` instead would risk a test that passes against my config rather than the app's — worse than no test. It's also a gap that predates this PR. Worth its own change; the deps (`testcontainersPostgresql`, and the driver via `testRuntimeOnly`) are already there.

### Verdict

With the four commits on this branch I would ship it. The design is right, the durability reasoning is sound, and the failure modes now degrade honestly instead of silently. The JDBC coverage gap is the one thing I'd want followed up — a first deploy where you tail the logs for `Failed to save replay` and confirm a `game_replay_players` row appears would close it empirically in five minutes.

wingedsheep:
## Verified against a live accounts-enabled server

Ran this branch with `ACCOUNTS_ENABLED=true` against real Postgres and played a full game (263 actions, ECL, human vs AI). **The JDBC gap I flagged earlier is closed — it works end to end.**

- `V10` applied cleanly from `V9`; both tables exactly as designed (FK + cascade + all three indexes).
- Periodic `IN_PROGRESS` flushes carrying a live `resume_fingerprint`, with **2 seat children** — the `@MappedCollection` round-trip works, which was the one untested production path.
- At game over, the full designed sequence fired: synchronous `FINISHED` write on the game-over thread, then the async write from the `replay-archiver` thread with the archive attached. `presentation` non-null, `winner_name` set, `engine_version` stamped, `seats = 2`.
- Zero errors or warnings in the log across the whole run.
- The two pre-V10 rows show `seats = 0` and no `engine_version` — exactly the back-compat behaviour `V10`'s comment predicts.

So: no defect in the persistence layer. But the measurement contradicts the documented sizing, and that's worth correcting because the numbers carry the design rationale.

### Sizing: POR is the cheap end, not the typical case — `481e1f10e9`

| | Documented (POR, 1650 actions) | Measured (ECL, 263 actions) |
|---|---|---|
| Pinned definitions | 34 | **113** |
| Pins, gzipped | ~7 KB | **~30 KB** (80% of the raw payload, median ~1.3 KB each) |
| `data` stored | ~11 KB | **~46 KB** |
| archive ÷ input log | **14×** | **1.0×** |

Portal's cards are simple and few, so it understates the pins by roughly 4×. The important part is *which term scales with what*: pins scale with deck variety and card complexity and are a fixed per-game floor; only the archive scales with game length. That inverts the headline — the two payloads are the same order of magnitude, so "the recipe is 14× cheaper, which is exactly why the recipe is the record" doesn't hold on its own terms.

The design is still right, for a sharper reason: the input log is the record because it's the only thing that can rebuild a real `GameState` for "share frame as scenario". I've documented both measurements and swapped the argument onto that footing rather than the size ratio.

### Write amplification — flagged, deliberately not fixed

The flusher re-encodes the whole blob every 5s, and ~80% of it is byte-identical pin data. A 30-minute game is on the order of 360 flushes × ~46 KB ≈ 16 MB written, ~13 MB of it the same 113 definitions repeatedly — plus TOAST churn, since every one of those updates rewrites a >2 KB TOASTed column and leaves a dead tuple for autovacuum.

**I did not fix this, deliberately.** At this project's scale it's wasteful rather than dangerous, and the fix is a schema change to the exact code path we just verified working — poor trade in an already-large PR. The clean version is a separate `pinned_cards` column written on the first flush only. Recorded in the doc so it's a known, deliberate cost rather than a surprise.

One trap worth writing down, which is why I didn't reach for the cheaper shortcut: **pins cannot simply be deferred to game over.** A restart onto a new build mid-game would then pin the *new* definitions for a game that started on the old ones — the precise failure pinning exists to prevent. (Arguably such a game is already unreproducible, since two engines folded it and the checkpoints would flag it — but that's a judgement call about a rare case, and it's yours, not mine.)

### Where things stand

Five commits from me on this branch: two blocking bugs (finished-record clobbering, stranded recordings), two concurrency/robustness fixes (atomic flush snapshot, game-over failure isolation), the minors, and this doc correction. Everything green, and now confirmed against real Postgres rather than only in tests. Nothing outstanding from my side.

wingedsheep:
## Write-once pins, implemented and verified on a live server — `c7f8d8dec6`, `3dec4a85fc`

`V11` moves the pinned definitions into their own column, and the flush takes a narrow
`GameReplayRepository.updateRecording` UPDATE naming only the columns that move.

**The narrow UPDATE is the load-bearing part, not the column.** Postgres preserves a TOASTed value's
storage only for a column an UPDATE doesn't name — re-assigning identical bytes through the full
aggregate `save()` would still pay the rewrite. So a separate column alone would have fixed nothing.

I did *not* take the cheaper route of deferring pins to game over, even though it's a one-liner: a
restart onto a new build mid-game would then pin the *new* definitions for a game that started on the
old ones, which is the precise failure pinning exists to prevent (and the corpus may no longer hold a
card the decks used). That reasoning is in the migration so nobody re-derives the shortcut.

### Verified end to end against real Postgres

Full lifecycle, observed on a live accounts-enabled server on this branch:

| Leg | Result |
|---|---|
| `V11` applies (V10 → V11) | clean, no errors |
| First flush | INSERT carries the pins (39,820 chars, md5 `657fbb5d`) |
| Flushes 2–N | `frame_count` 1 → 22 → 38 → 56, `data` 2,624 → 3,176 chars, **pins md5 unchanged** |
| Game over | `FINISHED`, `data` 4,780, pins **still md5 `657fbb5d`**, `presentation` 62,152, seats 2 |
| Serve a V11 record (pins only in the column) | `fidelity: EXACT`, `stateReproducible: true`, all 358 frames |
| Serve a pre-V11 record (pins only in `data`) | `fidelity: EXACT`, all 264 frames |

The stable md5 across flushes while `data` advances is the direct evidence that the running server
takes the narrow path; the two serve results prove the read path recombines the split payload *and*
still falls back to blob pins for older rows.

### The saving is larger than I estimated

At 123 frames, `data` was 3,624 chars; the same flush previously wrote 3,624 + 39,820 ≈ **43,400 —
12× more**. I had guessed ~16 KB per flush, having conflated the end-of-game action log with the
whole-game average: the log starts near zero and grows, so early flushes were almost entirely pin
data. A 30-minute game drops from roughly 16 MB written to ~3 MB, and the fixed per-flush pin term is
gone.

### Docs re-measured

The sizing table I added earlier described the pre-V11 layout and is now corrected against a full
357-action ECL game under V11:

| Column | Stored | Scales with |
|---|---|---|
| `data` — input log + checkpoints | ~4.8 KB | game length |
| `pinned_cards` — 113 definitions | ~40 KB | deck variety (fixed per game) |
| `presentation` — archived frames | ~62 KB | game length |

~107 KB per finished game. Worth internalising: **the pins are the largest single cost and are
unrelated to game length**, so a 20-turn concession costs nearly as much as a 40-minute grind — budget
per game, not per action. Also note this game's pins were 40 KB against 30 KB for the previous one,
which is further evidence that the original POR-derived figure (~7 KB) was not a usable planning
number.

Full replay + migration suite green (31 tests), including a new `FlywayMigrationTest` case that proves
a flush-shaped UPDATE leaves `pinned_cards` byte-identical, and a codec test for the split encoding.

PR #1398, opened 2026-07-24
Promise a permanent's gift as you cast it, not after it enters

wingedsheep:
## The bug

The gift choice should be made **while casting** — CR 702.174a makes it an additional cost ("as an additional cost to cast this spell, you may choose an opponent"), and on a permanent the gift itself is the triggered ability "when this permanent enters, if its gift cost was paid, \[effect\]" (CR 702.174b).

Our three **permanent** gift cards — Kitnap, Scrapshooter, Starforged Sword — modelled the promise as a modal choice on their enters-the-battlefield trigger, so the player was asked whether to give a gift *after* the permanent had already entered. Worse, for Scrapshooter and Starforged Sword the printed "if the gift was promised" ability collected its target inside the same mode, i.e. before the promise existed.

Instant/sorcery gift cards were already correct: their two-mode `ModalEffect` is resolved by the engine's cast-time mode picker. I verified **Long River's Pull** in the running client — both "Counter target creature spell" and "Gift a card — counter target spell" appear as cast options in the action menu, and nothing about the gift is asked afterwards. (Screenshot-verified, plus a Playwright spec below.)

## The fix

Gift becomes a real cast-time election:

- **SDK** — `KeywordAbility.Gift(GiftKind)` and `card { gift(kind) }` (`dsl/mechanics/GiftDsl.kt`). The helper adds the keyword *and* derives the CR 702.174b enters-ability: an intervening-if trigger (CR 603.4) gated on `Conditions.GiftWasPromised`, closing with `Effects.GiftGiven()` so "whenever you give a gift" still fires (CR 702.174c). `GiftKind` covers the rules' entire list (CR 702.174d–i — card, Food, tapped Fish, Treasure, Octopus, extra turn), each mapped to its defined effect, so the next gift permanent only names its kind.
- **No new condition type** — `Conditions.GiftWasPromised` is a facade over `CastChoiceMade(ChoiceSlot.GIFT_PROMISED)`. Cards gate printed abilities on it (Scrapshooter's destroy, Starforged Sword's attach) or on `Conditions.Not(...)` for "if the gift wasn't promised" (Kitnap's stun counters).
- **Engine** — `CastSpell.giftRecipient` + `SpellOnStackComponent.giftRecipient`; `CastSpellEnumerator.expandGiftPromise` clones each normal cast into a `CastWithGift` variant per opponent (gift changes neither cost nor targets, so cloning keeps cost/target enumeration in one place, and multiplayer picks the recipient as part of the cost); the handler validates the recipient is an opponent of a card that actually has gift; the resolving permanent carries `GIFT_PROMISED` + `OPPONENT` durably in its cast-choices bag.
- **Client-visible** — the stack's `giftPromised` badge reads the new flag, auto-pass counts `CastWithGift` as an instant-speed response, and the action menu already renders multiple cast variants, so the gift shows up as "Cast Kitnap (Gift a card)" with no client changes beyond the action type.
- **CardLinter** — the gift keyword declares both choice slots it writes.

## Tests

`GiftPromisedAtCastTimeTest` (rules-engine) — the cast-time offer alongside the plain cast, both branches of all three cards, the intervening-if (an unpromised Scrapshooter never triggers and never targets), "whenever you give a gift" through Jolly Gerbils, a countered gift spell giving nothing, recipient validation, and multiplayer per-opponent variants where only the promised opponent draws.

`e2e-scenarios/tests/general/gift-promised-at-cast-time.spec.ts` — pins the UI timing in a real browser for Kitnap (gift is a cast option; nothing asked once the Aura enters) and Long River's Pull (both modes offered while casting).

Verified: `:mtg-sdk:test`, `:mtg-sets:test` (BLB snapshot re-blessed — diff is exactly the three cards' reshape), `:rules-engine:test`, `:game-server:test`, `:ai:test`, `:mtgish-tooling:test`, `web-client` typecheck, and the two Playwright specs.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1697,7 +1697,9 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
   single opponent, so 2-player games see no extra decision. The source may be a spell on the stack
   (the choice lives on the spell entity for its resolution) or a permanent (recorded durably).
   `Patterns.Mechanic.giftSpell` prefixes its gift mode with this automatically — gift recipients
-  address `Player.ChosenOpponent`. The *as-enters* analogue is `EntersWithChoice(ChoiceType.OPPONENT)`,
+  address `Player.ChosenOpponent`. (Permanent gift cards don't need it: `gift(kind)` records the
+  promised opponent in the same slot as part of the cast — see § 11 "Gift".) The *as-enters* analogue
+  is `EntersWithChoice(ChoiceType.OPPONENT)`,
   which writes the same slot (Jihad, The Rack).
 - `GrantHexproofFromChosenColorEffect(target)` — hexproof from chosen color.
 - `GrantProtectionFromChosenColorEffect(target)` — protection from chosen color. Must run inside `ChooseColorThen`; wrap in `ForEachInGroup` for the group case (Akroma's Blessing: "Creatures you control gain protection from the chosen color").
@@ -4852,7 +4854,7 @@ copy of it (CR 707.10e). The activated-ability analogue of the spell-level `cant
 > mechanics — `mayBeginGameOnBattlefield()`, `flurry { }`, `mobilize(…)`, `firebending(n)`, `sneak(cost)`, `decayed()`,
 > `vividEtb { }` / `vividCostReduction()`, `convergeEntersWithCounters(counterType?)`,
 > `impending(time, cost)`, `renew(cost) { }`, `enduring()`,
-> `craft(filter, cost)`, `station()`, `jobSelect()` — are `CardBuilder` **extension functions** in
+> `craft(filter, cost)`, `station()`, `jobSelect()`, `gift(kind)` — are `CardBuilder` **extension functions** in
 > `mtg-sdk/.../dsl/mechanics/` (one file per mechanic), not methods on the core `CardBuilder`. They
 > stay in package `com.wingedsheep.sdk.dsl`, so the call syntax is unchanged, but a card file that
 > uses one needs the matching import (e.g. `import com.wingedsheep.sdk.dsl.station`). Evergreen /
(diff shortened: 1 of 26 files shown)


wingedsheep:
## Review + fixes applied (`14386a05dd`)

Reviewed the mechanic end to end and pushed the follow-ups onto this branch. Also merged `origin/main` in (`3546714927`) — it was clean, no conflicts, so you can either keep the merge commit or rebase it away.

### Verdict

The diagnosis and the model are both right. Every CR citation in the diff checks out against the 2026-06-19 rules — 702.174a/b/c, d–i for each `GiftKind`, and 603.4 — and all three cards match Scryfall Oracle 1:1, including Kitnap's subtle shape where the tap is unconditional and only the stun counters are gated. `Conditions.GiftWasPromised` as a facade over `CastChoiceMade` rather than a new condition type is the right call, and `expandGiftPromise` sits correctly alongside the existing `expandChoiceAdditionalCosts` / `applySpellWaterbendMetadata` passes.

Worth adding to the PR body: **CR 702.174m** is the rule that makes the old modelling illegal rather than merely awkward — "if part of a spell's ability has its effect only if its gift was promised, and that part of the ability includes any targets, the spell's controller chooses those targets only if the gift was promised." That's exactly what Scrapshooter and Starforged Sword were violating. It's now cited in the SDK reference and both card KDocs.

I verified the "no client changes needed" claim rather than taking it: `ActionMenu.tsx:163` renders one button per variant once `castActions.length > 1`, and `ClientMessage.SubmitAction(action: GameAction)` round-trips `giftRecipient` through kotlinx, so the anti-corruption layer genuinely needs nothing. I also confirmed all 19 remaining `giftSpell` users are instants or sorceries — the permanent migration is complete.

### What I fixed

**The one real bug.** The promise was only offered on the plain `CastSpell`, so it disappeared whenever the mana cost was paid another way. Reachable today: with an Omniscience out and no Forests, the only playable action is "Cast Scrapshooter (Free)" and the gift can't be promised at all — but gift is an *additional* cost, chosen and paid whichever cost pays the mana cost (CR 601.2b, 601.2f–h, the exact rules 702.174a points at). `expandGiftPromise` now expands every cast variant except `CastFaceDown` (a face-down spell has no abilities to promise with, CR 708.2), and is gated on `affordable` so an unpayable cast no longer spawns a greyed-out gift twin per opponent. `ActionMenu`'s impending branch keeps the extra variants too instead of dropping any cast option that isn't normal-or-impending.

**A footgun closed.** `CardValidator` now rejects `KeywordAbility.Gift` on a non-permanent. There it can never trigger, yet the enumerator would still offer `CastWithGift`, `validate()` would accept it, `StackResolver` would never stamp it, and `GiftWasPromised` would read `false` — a card that quietly does nothing. Related: `GiftWasPromised` now documents that it's permanent-only, unlike `SneakCostWasPaid` / `WaterbendWasPaid`, which do fall back to the resolution context.

**Smaller.** `CardLinter.slotDeclarers` maps a node type to a *list* of slots, so gift declares its two from the same table every other mechanic uses instead of a hardcoded `if` beside it. The derived ability now names the permanent the way the printed card does ("When this Aura enters", "When this Equipment enters"), falling back to the rule's own generic wording if `gift()` runs before `typeLine`.

**Tests.** New `GiftDslTest` pins all six `GiftKind` → effect mappings (four had no card and no coverage — all four were correct, but nothing was holding them to it), the derived ability's shape, the wording, and the permanent-only guard. `GiftPromisedAtCastTimeTest` gains the Omniscience free-cast case. BLB snapshot re-blessed: the diff is the three cards' ability wording and nothing else.

### One thing I deliberately did *not* do

I initially wanted to collapse `giftRecipient` — the fifth typed cast-time field threaded through `CastSpell` → `SpellOnStackComponent` → the `StackResolver` bag, after `wasKicked` / `wasSneaked` / `wasBlightPaid` / `wasWaterbendPaid` — into one generic `castChoices: Map<ChoiceSlot, ChoiceValue>`, since `ChoiceSlot`/`ChoiceValue` already are that vocabulary. Working through it, that shape is worse: `CastSpell` is client-supplied, so an open map merged into the permanent's cast-choices bag would let a client stuff in `KICKED` or `COLOR` and help itself to a free kicker. Making it safe needs a per-card whitelist — which is the per-mechanic branch the carrier was supposed to remove. The typed field is the right call here; a validated carrier is a separate piece of work if it's ever worth it.

Also: the `E2E_SERVER_URL` override in `scenarioApi.ts` is unrelated drive-by test infra (it pairs with the existing `E2E_BASE_URL` in `playwright.config.ts`). Useful, just worth a line in the PR body.

### Verification

`:mtg-sdk:test`, `:mtg-sets:test`, `:rules-engine:test`, `:game-server:test`, `:ai:test`, `:mtgish-tooling:test` and `web-client` typecheck all green after the changes. I did not re-run the Playwright specs.

PR #1416, opened 2026-07-25
Add Bargain (CR 702.166) to the engine

wingedsheep:
## What

Implements **Bargain** (CR 702.166, Wilds of Eldraine) end to end: *"Bargain (You may sacrifice an artifact, enchantment, or token as you cast this spell.)"* — an optional additional cost declared as the spell is cast (702.166a), whose declaration makes the spell *bargained* (702.166b) for its own linked abilities (702.166c), including a bargain-only target (702.166d).

## Design — generalise the rail, don't add a ninth boolean

Bargain is shaped exactly like kicker-with-a-sacrifice-cost, but it must be a *different fact*: a bargained spell must not trigger "whenever you cast a kicked spell", and vice versa. So instead of a new `wasBargained` boolean next to `wasKicked` / `wasBlightPaid` / `wasWaterbendPaid` / `wasSneaked`, the shared optional-additional-cost rail now carries **which mechanic** declared:

- `KeywordAbility.OptionalAdditionalCost` gains `declaredSlot: ChoiceSlot` (default `KICKED`).
- `CastSpell.wasKicked`, `SpellOnStackComponent`, `SpellCastEvent` and `EffectContext` replace that boolean with `declaredCostSlot: ChoiceSlot?`. `wasKicked` semantics become `== ChoiceSlot.KICKED`; kicker/multikicker/offspring/flash-kicker behaviour is unchanged.
- The enumerator emits one cast variant per mechanic on the rail (grouped by slot), labelled from the keyword — "Cast X (Bargained)".

No new `Effect`, executor, or `Condition` type. The new SDK vocabulary is `Keyword.BARGAIN`, `ChoiceSlot.BARGAINED`, `Conditions.WasBargained` (a facade over `CastChoiceMade`), `GameObjectFilter.ArtifactEnchantmentOrToken`, and the `bargain()` CardBuilder helper that composes them.

Two supporting changes make the card pool authorable:

- `CastChoiceMade` also answers from the declaration a spell carries **on the stack**, so a spell's own *"If this spell was bargained, …"* rider resolves with no permanent to read from (Torch the Tower, Archon's Glory).
- `CostCalculator` threads the declaration into `CostGating.OnlyIf`, so `ModifySpellCost(SelfCast, ReduceGeneric(2), OnlyIf(WasBargained))` prices only the branch it gates — *"This spell costs {2} less to cast if it's bargained"* (Hamlet Glutton, Ice Out, Johann's Stopgap).

## UX

The stack badge is now a server-derived `optionalCostLabel` ("Kicked" / "Bargained" / "Offspring") instead of a `wasKicked` boolean the client had to name itself; the action-menu fallbacks no longer hardcode "Kicked". The sacrifice reuses the existing additional-cost selection flow, so the player picks which artifact/enchantment/token to sacrifice with no new decision component.

## Tests

`BargainMechanicScenarioTest` — 15 scenarios pinned to the CR clauses:

- what may pay the cost: artifact, enchantment, **token** (not a card type), and *not* a nontoken creature
- the cost is paid on announcement (already in the graveyard while the spell is still on the stack)
- the on-stack rider (bargained deals 4, unbargained deals 2)
- the durable flag on the resolving permanent + its intervening-if enters trigger (CR 603.4)
- both directions of the kicker/bargain separation, including a "whenever you cast a kicked spell" payoff staying silent on a bargained spell
- the conditional cost reduction (plain cast {5}{G} unaffordable, bargained {3}{G} affordable)
- the bargain-only target: offered only on the bargained variant (702.166d)

Full `just test` green; `web-client` typecheck green.

## mtgish

Bridge entry + `bargain()` emitter rendering + `SpellPassesFilter(ThisSpell, WasBargained)` → `Conditions.WasBargained` recovery, so the generator scores the mechanic as covered corpus-wide (WOE's leaderboard no longer lists Bargain as blocking) and drafts the keyword line.


wingedsheep:
## Review follow-up — fixes pushed (`a624b034f0`)

Reviewed on the branch with `origin/main` merged in (merge commit `e0a7eec469`, no conflicts). Full `just test` green before and after.

The SDK shape here is the right one and I didn't touch it. Replacing `wasKicked: Boolean` with `declaredCostSlot: ChoiceSlot?` on the shared optional-additional-cost rail is the correct generalisation — Bargain lands with no new `Effect`, executor, `Condition`, or `StaticAbility`, `Conditions.WasBargained` is a `val` facade over `CastChoiceMadeCondition`, and the next mechanic on the rail now costs one enum entry plus a DSL helper. I verified 702.166a–d, 603.4, 601.2b and 601.2c against the 2025-07-25 CR; every citation in the code is accurate, including the 702.166c linked-abilities point that motivates the separate slot.

Two real defects and some cleanup, all fixed in the pushed commit.

### Enumerator regression (was blocking)

Grouping `enumerateKicker` by slot moved the sorcery-speed and cast-restriction gates *inside* the per-slot loop, so `continue` skipped one variant instead of the card. Execution then reached the unaffordable-normal-cast fallback at the bottom of the card loop, which previously was unreachable in that case. Reproduced with a `{9}{G}` kicker creature while a spell was on the stack:

```
[CastSpell|Cast Probe Kicker Bear|false]   // expected: []
```

No illegal play is possible (`affordable = false`), but the authoritative legal-action list was advertising a greyed cast the timing rules forbid, and for unmet `castRestrictions` too. Card-level gates moved back above the loop; the per-slot `flashKicker` re-check stays inside, since a flash unlock belongs to the mechanic that prints it and a bargain variant shouldn't inherit a kicker's instant-speed permission. Regression test added in `CastSpellEnumeratorTest` (Stronghold Confessor at upkeep).

### Stored replays decoded wrong (was important)

`CompactReplay.actions` is persisted and `persistenceJson` sets `ignoreUnknownKeys`, so `"wasKicked": true` in every existing record was being dropped and `declaredCostSlot` defaulted to `null` — the cast re-simulated *unkicked* and everything after it was a game nobody played. Silent by construction; the v2 checkpoint fingerprints would surface it as a reconstruction error rather than a fix. `ReplayCodec.decode` now rewrites the legacy boolean to `KICKED`, gated on a substring test so untouched records skip the extra parse. Three tests in `ReplayLegacyKickerDecodeTest`.

### Dead code and naming

- `CardDefinition.bargainKeyword()` had zero callers anywhere, including tooling and tests, despite a KDoc saying otherwise — removed.
- `BargainSacrificeFilter` had one caller in its own file, so it's inlined as `GameObjectFilter.ArtifactEnchantmentOrToken`.
- `kickerSpellEffect` / `kickerTargetRequirements` / `kickerEffect` / `kickerTarget` now serve any slot on the rail, not kicker specifically — `BargainDsl` even points authors at them for a 702.166d bargain-only target. **I documented rather than renamed them**, deliberately: the `CardScript` field names are serialised into pinned replay card definitions, so a rename would reintroduce exactly the silent-decode-drift bug fixed above. If you want the rename, it needs the same kind of migration on the pin side and is better as its own change. Flagging so it's your call, not a silent narrowing.
- Minor: the "unpayable bargained variant" assertion used `forEach`, which passed vacuously on an empty list — now asserts the variant is enumerated, which is what the comment claimed. Dropped an unused import and corrected a comment saying `description` emits reminder text (it emits the bare keyword).

### Left alone

The slot → display-word mapping is now in three places (`CastSpellEnumerator.kickLabel`, `CastFromZoneEnumerator.kickLabel`, `ClientStateTransformer.optionalCostLabel`) and they already disagree — `"with Flash"` only exists in the first. Pre-existing divergence that this PR extends rather than causes; worth collapsing into one helper, but not in this change.

Review worktree left in place at `~/.claude/worktrees/argentum-engine/pr-1416-review` for further iteration. Note the branch now carries the `origin/main` merge commit, so pull before pushing from `.claude/worktrees/feature-bargain`.

PR #1474, opened 2026-07-27
Resolve token art from the set that printed the creating card

wingedsheep:
## The bug

Arahbo, the First Fang (Foundations) creates a 1/1 white Cat, and the Cat showed **Dominaria Remastered** art.

`Effects.CreateToken(...)` in `ArahboTheFirstFang.kt` passes no `imageUri`, so the token entity gets `imageUri = null`. The client then falls through to `getScryfallFallbackUrl("Cat Token")` → `api.scryfall.com/cards/named?exact=Cat`, and renders whatever printing Scryfall happens to return. Nothing was *missing*, so nothing failed — the art was just silently wrong. That applied to essentially every creature token in the game, not just this one.

## The system

A token has no `CardDefinition` and no `Printing` row, so the per-printing art override that `Printing` gives a real card has nothing to hang on. Sets now declare their own token printings via **`MtgSet.tokenArt: List<TokenPrinting>`**, and the token executors resolve art in three layers:

1. an explicit `imageUri` on the effect — a deliberate per-card override, always wins;
2. the `TokenPrinting` from the set the **creating card was printed in** — `TokenArtRegistry`, keyed off the entity's `Name#SET-CN` definition id;
3. the engine-wide generic fallback — `TokenArt.IMAGES` by creature type, or the canonical `PredefinedTokens` printing.

```kotlin
object FoundationsSet : MtgSet {
    override val tokenArt = listOf(
        // tfdn #1 — Arahbo, the First Fang's 1/1 white Cat.
        TokenPrinting(name = "Cat", imageUri = "https://cards.scryfall.io/art_crop/front/2/8/2885d54c-….jpg"),
    )
}
```

Layer 2 is the reprint answer, and it's why the art goes on the set rather than the card: a card that bakes `imageUri` into its `CreateToken` mints the same art from **every** printing, which is wrong the moment it's reprinted into a set with its own token. **LCI is migrated off exactly that pattern** — 14 cards drop `imageUri = LciTokenArt.TREASURE` in favour of one row on `LostCavernsOfIxalanSet`, and `LciTokenArt` is deleted.

## Visual change worth flagging

Layer 3 previously ran **only** for tokens whose creature type is chosen at resolution time (Riptide Replicator), leaving every other creature token with a null image. Making the "every token has an image" test pass requires it to apply to all of them — so creature tokens now render in the client's generated token frame (name bar / art box / type bar, the path `styles.ts` was already written for and predefined tokens already use) instead of a name-guessed full-card Scryfall image. `TokenArt.IMAGES` URLs are `art_crop`, which is what that art box wants.

`TokenArt.IMAGES` gains the 7 creature types that turned out to be reachable but uncovered: Ally, Gremlin, Homunculus, Kavu, Otter, Raccoon, Robot.

## Wiring fix

`EngineServices` now hands its own `EffectExecutorRegistry` to `EffectHandler`. Without it, effects resolving off the stack ran through a *second*, bare registry built inside `EffectHandler` and never saw the token art registry at all — the Arahbo test caught this.

## Tests

- **`TokenArtCoverageTest`** — the requested corpus gate. Walks every registered card's *serialised* tree (so a `CreateToken` nested in a composite, mode, pipeline, reflexive trigger, granted ability or `ConvertCountersToTokens` is still found) and fails the build if any token it can create resolves to no image. Tokens with a resolution-time chosen type are skipped, as there's nothing statically to look up.
- **`SetScopedTokenArtScenarioTest`** — casts Arahbo, asserts the Cat carries `tfdn` art and specifically *not* the generic Cat; plus a non-declaring set still falls back to the generic table.
- `LCI.json` card snapshot re-blessed: the only movement is the 14 Treasure sites losing their inline `imageUri`. FDN snapshots are untouched — token art lives on the set, not the card.

`just build` green across all modules.


wingedsheep:
## Bulk registration + gap report (2nd commit)

The first commit shipped the mechanism with two data rows, so everything except FDN's Cat and LCI's Treasure still rendered generic stand-in art. Now:

**All sets registered.** `tokens.json` — **1131 token printings across 94 sets**, synced from Scryfall's token sets (`t<code>`), loaded by `TokenArtData`. Registration is `set.tokenArt + TokenArtData.forSet(code)`: hand-authored rows first, so declaring one overrides synced art or supplies art Scryfall doesn't have.

**Self-hosted art preserved.** Invasion's Saproling/Reflection (`/images/tokens/inv-*.jpeg`) moved off 7 card files onto `InvasionSet.tokenArt`, where a resync can't clobber it. The OTJ rabbit stays inline — it's a `BecomeCreature` display override on an animated creature, not a token.

**Gap report for the remaining art.** 57 sets predate token cards and have nothing to sync. `just token-art-gaps` writes `backlog/token-art-gaps.md`: every token with no set-scoped art, the cards that create it, a suggested `web-client/public/images/tokens/<set>-<token>.jpeg` path, and a paste-ready `TokenPrinting(...)` row. **Currently 34 tokens across 23 sets** — e.g.

> **2/1 black Cat** — created by 1 card(s): Penumbra Bobcat
> - image: `web-client/public/images/tokens/apc-cat.jpeg`
> - `TokenPrinting(name = "Cat", imageUri = "/images/tokens/apc-cat.jpeg", power = 2, toughness = 1, colors = setOf(Color.BLACK)),`

It shares one matcher (`TokenPrinting.bestMatch`) and one card-walker (`TokenCreationSites`) with the engine and the coverage test, so the report can't claim coverage the engine wouldn't deliver.

**Sync robustness.** An earlier revision of `just token-art-sync` silently recorded rate-limited sets as having no tokens — the same silent-wrong-data shape as the original Cat bug. It now distinguishes 404 (genuinely no token set) from transport failure, backs off on 429/503, leaves a failed fetch *absent* rather than empty, and flushes after every set so a rate-limited run resumes instead of restarting.

**Matching.** Synced rows pin P/T and colors, so `bestMatch` tries exact identity first (a set printing two Soldiers resolves correctly), then name alone (a variable-P/T printing, or a colour granted rather than printed, still gets the set's art rather than generic). Name comparison is order-insensitive across words, since a token minted from `setOf("Army", "Zombie")` can serialise either way.

`just build` green; `INV.json` snapshot re-blessed for the inline-art removal.

PR #1510, opened 2026-07-29
Pr mh3 couple cards

jonubuu:


wingedsheep:
Reviewed all four cards against the CR (2025-06-06) and Scryfall Oracle.

**What's solid:** Exert is a genuinely new primitive and it's modeled well — a per-object marker rather than a stun counter or a `DOESNT_UNTAP` static, cleared scoped to the active team, and correctly *not* applied to the Seedborn-Muse "each other player's untap step" paths (exert only blocks *your* next untap step — easy to get wrong). Event → serializer → `ClientEvent` → `ClientCard` → transformer → TS type plumbing is complete. Oracle text, mana costs, P/T, collector numbers and all five image URIs (incl. the `tmh3` Cat token) verified exact; `just check-card-printing` clean for all four.

Both shipped test classes pass: `ArenaOfGloryScenarioTest` 5/5, `GuideOfSoulsScenarioTest` 4/4.

---

## Blocking

### 1. Ocelot Pride's Ascend never fires

CR 702.131b: Ascend on a *permanent* is a **static** ability — "**Any time** you control ten or more permanents and you don't have the city's blessing, you get the city's blessing." `OcelotPride.kt:56-62` models it as an ETB trigger gated on `ControlPermanentsAtLeast(10)`, and `Keyword.ASCEND` is inert in the engine (bare enum entry at `Keyword.kt:340`, no handler anywhere in `rules-engine` or `mtg-sdk`). So the threshold is sampled once, at ETB, and never again.

Ocelot Pride is a one-mana 1/1 — you essentially never control ten permanents on the turn it enters, so the "Then if you have the city's blessing…" half of the card is dead in real games. It also contradicts the card's own quoted ruling #3 ("If the creature token created by Ocelot Pride's last ability is your tenth permanent, you'll get the city's blessing before the ability would check"), which needs a continuous check *mid-resolution*.

Repro — tests 1 and 2 fail, test 3 passes, which pins the defect on Ascend alone (the intervening-if trigger, resolution-time `ConditionalEffect` and snapshotting `ForEachInGroup` are all fine):

```kotlin
class OcelotPrideAscendScenarioTest : FunSpec({

    fun driver(): GameTestDriver {
        val d = GameTestDriver()
        d.registerCards(TestCards.all)
        return d
    }

    fun hasCitysBlessing(d: GameTestDriver, playerId: EntityId): Boolean =
        d.state.getEntity(playerId)?.has<PlayerCitysBlessingComponent>() == true

    fun catTokens(d: GameTestDriver, playerId: EntityId): Int =
        d.getCreatures(playerId).count { d.getCardName(it) == "Cat Token" }

    /** "if you gained life this turn" reads LifeGainedAmountThisTurnComponent — seed it directly. */
    fun seedLifeGained(d: GameTestDriver, playerId: EntityId, amount: Int) {
        d.replaceState(
            d.state.updateEntity(playerId) { it.with(LifeGainedAmountThisTurnComponent(amount)) }
        )
    }

    fun grantCitysBlessing(d: GameTestDriver, playerId: EntityId) {
        d.replaceState(d.state.updateEntity(playerId) { it.with(PlayerCitysBlessingComponent) })
    }

    /**
     * Drain the stack and any pending decisions, then stop. Deliberately does *not* pass on an
     * empty stack — an unconditional `bothPass()` loop walks the turn forward and skips the very
     * end step these tests are trying to observe.
     */
    fun settle(d: GameTestDriver, maxSteps: Int = 20) {
        repeat(maxSteps) {
            when {
                d.pendingDecision != null -> d.autoResolveDecision()
                d.getTopOfStack() != null -> d.bothPass()
                else -> return
            }
        }
    }

    test("crossing ten permanents after Ocelot Pride entered still grants the city's blessing (CR 702.131b)") {
        val d = driver()
        d.initMirrorMatch(deck = Deck.of("Plains" to 40), skipMulligans = true)
        val active = d.activePlayer!!
        d.passPriorityUntil(Step.PRECOMBAT_MAIN)

        // Cast it for real so the ETB ascend trigger genuinely fires and finds < 10 permanents,
        // exactly as a turn-one Ocelot Pride would.
        d.giveMana(active, Color.WHITE, 1)
        val prideCard = d.putCardInHand(active, "Ocelot Pride")
        d.castSpell(active, prideCard)
        settle(d)

        d.findPermanent(active, "Ocelot Pride") shouldNotBe null
        hasCitysBlessing(d, active) shouldBe false // fine so far — you control ~2 permanents

        // Now climb well past ten permanents, as any real game does a few turns later.
        repeat(12) { d.putLandOnBattlefield(active, "Plains") }
        settle(d)
        d.getPermanents(active).size shouldBeGreaterThanOrEqual 10

        // Ascend is static and continuous: the moment the tenth permanent is there, the blessing
        // is yours. Today the engine sampled the count once at ETB and never looks again.
        hasCitysBlessing(d, active) shouldBe true
    }

    test("with ten-plus permanents reached later, the end-step ability still doubles the Cat") {
        val d = driver()
        d.initMirrorMatch(deck = Deck.of("Plains" to 40), skipMulligans = true)
        val active = d.activePlayer!!
        d.passPriorityUntil(Step.PRECOMBAT_MAIN)

        d.giveMana(active, Color.WHITE, 1)
        val prideCard = d.putCardInHand(active, "Ocelot Pride")
        d.castSpell(active, prideCard)
        settle(d)

        repeat(12) { d.putLandOnBattlefield(active, "Plains") }
        seedLifeGained(d, active, 1)
        catTokens(d, active) shouldBe 0

        d.passPriorityUntil(Step.END)
        settle(d)

        // One Cat from the first clause, plus one copy of it from the city's-blessing clause
        // (per the 2024-06-07 ruling, the Cat made earlier in this same resolution has already
        // "entered this turn" by the time the second clause is reached).
        catTokens(d, active) shouldBe 2
    }

    test("control: granted the city's blessing directly, the doubling works — the defect is Ascend alone") {
        val d = driver()
        d.initMirrorMatch(deck = Deck.of("Plains" to 40), skipMulligans = true)
        val active = d.activePlayer!!
        d.passPriorityUntil(Step.PRECOMBAT_MAIN)

        d.putCreatureOnBattlefield(active, "Ocelot Pride")
        grantCitysBlessing(d, active)
        seedLifeGained(d, active, 1)
        catTokens(d, active) shouldBe 0

        d.passPriorityUntil(Step.END)
        settle(d)

        catTokens(d, active) shouldBe 2
    }
})
```

```
crossing ten permanents ... still grants the city's blessing (CR 702.131b)   FAILED  expected:<true> but was:<false>
with ten-plus permanents reached later, the end-step ability still doubles the Cat  FAILED  expected:<2> but was:<1>
control: granted the city's blessing directly, the doubling works           PASSED
```

Fix: make `Keyword.ASCEND` a real continuous / state-based check in the engine — that fixes Tendershoot Dryad (BLC) for free, which has the same gap but survives it better as a five-drop. If you'd rather defer, drop ruling #3 from the metadata and note the limitation in the PR body.

### 2. Ocelot Pride and Thraben Charm ship with no tests

Neither card is referenced by any test in the repo. Ocelot Pride chains an intervening-if trigger, a resolution-time `ConditionalEffect` and a snapshotting `ForEachInGroup`. Thraben Charm mode 3 is the **first use of `TargetPlayer(unlimited = true)` inside a modal mode** — every prior user (Riverchurn Monument, The Death of Gwen Stacy, Kaboom) is a top-level ability or saga chapter. Worth covering each mode, and mode 3 with two target players *and* with zero.

---

## Important

### 3. Exert is CR 701.43, not 701.39 — 701.39 is Bolster

20 occurrences across 12 files: `Costs.kt`, `ActivatedAbility.kt`, `CostHandler.kt`, `BattlefieldComponents.kt`, `GameEvent.kt`, `BeginningPhaseManager.kt`, `ClientDTO.kt`, `ArenaOfGlory.kt`, `ArenaOfGloryScenarioTest.kt`, `gameState.ts`, `GameCard.tsx`, `docs/card-sdk-language-reference.md`. `701.39a`→`701.43a`, `701.39b`→`701.43b`, `701.39d`→`701.43d`. The sub-rule letters all line up under 701.43, so it's a clean find-and-replace.

### 4. Reflexive triggered abilities are CR 603.12, not 603.2

603.2 is just "an ability triggers when its event happens", and it has no sub-rules at all. Four sites in this PR: `GuideOfSouls.kt:22`, `CounterEffects.kt:207`, `mtgish-tooling/.../Energy.kt:49`, `docs/card-sdk-language-reference.md:906`.

### 5. The "plain ability's mana does not grant haste" test doesn't test that

`ArenaOfGloryScenarioTest.kt:104-121` adds a colorless *and* a green before casting Grizzly Bears `{1}{G}`, so the rider-tagged red is never spent. Probed the pool directly through that exact sequence:

```
POOL BEFORE CAST: R=1 G=1 C=1
POOL AFTER CAST:  R=1 G=0 C=0
RED WAS SPENT: false
```

Green paid the `{G}`, colorless paid the `{1}`, and the tagged red sat untouched — so `hasKeyword(bears, HASTE) shouldBe false` passes for a reason unrelated to the rider, and would pass even if the plain ability *did* carry one. Dropping `d.giveColorlessMana(active, 1)` on line 114 forces the red to pay the `{1}` and makes the assertion load-bearing.

### 6. The exert corners the comments dwell on are untested

CR 701.43b gets ~30 lines of prose across five files — "always payable even when untapped", "re-exerting doesn't stack", "the marker expires having done nothing". But `exertAbilityId` is activated exactly twice in the whole test file, once per test, never on an untapped permanent and never twice before an untap step. Two short cases would make those citations load-bearing.

wingedsheep (changes requested):
See above

jonubuu:
1. The ascend mechanic was already present and not created when I made ocelot pride, it triggered in the test I had Claude make based on the contribute document, I can definitely go back over it though. *EDIT* Are you referring to the City's Blessing token missing? If so I do believe that was indeed missing from the test, I was thinking of the tokens doubling because of it, will dig back into this in the morning. *EDIT*

2. Both Ocelot Pride and Thraben Charm passed the scenario I was presented with, Thraben Charm worked for all 3 modes when I ran it through some extra testing. 

3. Claude is the one citing specific rules, I'm not sure where he got the wrong reference, I'll work on getting him to replace the wrong info for rule codes. Is there a best practice for making sure his stated rule codes are correct? If not I'm not sure I have the time to verify them every time he puts them in there, so I'll have to figure that part out.

4. I'm going to have to figure out why his rule numbers are incorrect, are they important to be in there? I can't remember if I read somewhere in the documentation if it was best to have them in there or not, I didn't realize a bunch of the rule ids would be wrong with where the info is gathered from by Claude.

5. I had red mana in the scenario I ran, is there a reason as to why they appear to be different then the ones you had, I assume they would be captured properly in the commit, I'll definitely take a look at my test vs the one on the main, I know I did add some extra lands as a test afterwards in a separately built scenario, so maybe I am confusing them.

6. I'm not sure I understand this part, its late and I'm tired, so I'll reread it in the morning, just didn't want to appear rude like I was omitting this point,  :)

jonubuu:
Pic of the scenario claude made for me to test based on the prompt in the contribute write up, if Im to be using a different one, please let me know and I'll switch to that for better accuracy.
arenascenario

jonubuu:
I have the manual scenarios that were created in my folder, but I see they're not in the branch I pushed the changes to...

I have argentum-engine/manual-scenarios/cards/o/ocelot-pride-ascend-doubles-cat.json not sure why its not pushed with all the other changes

PR #1514, opened 2026-07-29
SPM Batch 2: damage-trigger / replacement / redirect features

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1513

## SPM Batch 2 — damage-trigger / replacement / redirect features

**Stacked on `spm-goblins`** — this PR is only the diff against that branch.

Three SPM cards, each unlocked by a small, composable damage-system capability. Two of the
capabilities also repair already-shipped cards with the identical shape.

### Cards
- **Anti-Venom, Horrifying Healer** [1] — `{W}{W}{W}{W}{W}` 5/5. "If he was cast" ETB reanimation
  (existing `Conditions.WasCast`) + "prevent damage to him, put that many +1/+1 counters on him"
  self-replacement.
- **Spider-Slayer, Hatred Honed** [175] — `{2}` 2/1. "Deals damage to a Spider → destroy it" +
  `{6}`, exile-from-graveyard → two tapped 1/1 flying Robot artifact tokens.
- **With Great Power . . .** [24] — `{3}{W}` Aura. "+2/+2 for each Aura/Equipment attached" +
  Pariah-style "all damage to you is dealt to enchanted creature instead."

### Engine capabilities (composition-first — no new Effect/executor types)
1. **Deals-damage-to-a-filtered-creature trigger** — added the missing `RecipientFilter.Matching`
   case to `TriggerMatcher.matchesDealsDamageTrigger`, evaluated against the recipient in projected
   state with the ability's real controller (threaded through `DamageTriggerDetector` /
   `AttachmentTriggerDetector`). Also repairs the identically-shaped, silently-broken **East-Mark
   Cavalier** and **Mauhur, Uruk-hai Captain** (LTR).
2. **Prevent-damage-to-counters on creatures** — added `RecipientFilter.Self` to
   `DamageUtils.applyReplaceDamageWithCounters` and invoked it on both creature-damage paths
   (`DamageUtils` noncombat + `CombatDamageManager.applyDamageToCreature`).
3. **Pariah-style static redirect** — `resolveRedirectTarget` now handles
   `EnchantedCreature`/`EquippedCreature`/`EnchantedPermanent`; the buff uses the new
   `DynamicAmounts.attachmentsOnEnchantedCreature()` over the existing `GrantDynamicStatsEffect`.

### Tests
Scenario tests for all three cards plus a regression test for East-Mark Cavalier, covering both
abilities per card: cast→reanimate and the `WasCast` gate; noncombat *and* combat damage→counters;
[...]

PR #1515, opened 2026-07-29
SPM Batch 3 — land-play trigger / turn-condition / Mayhem land-play

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1514

## SPM Batch 3 — land-play trigger / turn-condition / Mayhem land-play

**Stacked on `spm-damage-triggers`.** Please review only the **2 commits on top of that branch** (`Add SPM Batch 3…` + `Address review…`); the earlier commits belong to the parent PR and will drop out of this diff once it merges.

Adds three Marvel's Spider-Man cards and the engine/SDK capability they need: a real "you play a land" trigger and matching turn-scoped conditions, plus the no-cost Mayhem land-play from the graveyard.

### Cards
- **Oscorp Industries** (#182) — Land. Enters tapped; "when this land enters from a graveyard, you lose 2 life"; `{T}: {U}/{B}/{R}`; **Mayhem** (CR 702.187c no-cost form — play it from your graveyard if you discarded it this turn).
- **Shadow of the Goblin** (#87) — `{1}{R}` Enchantment. First-main loot; *Undying Vengeance* — "whenever you play a land or cast a spell from anywhere other than your hand, deal 1 to each opponent."
- **Spider-Man 2099** (#150) — `{U}{R}` Legendary Creature, double strike + vigilance. *From the Future* cast restriction (can't cast during your first three turns); end-step intervening-if "if you've played a land or cast a spell this turn from anywhere other than your hand, deal power to any target."

### Engine / SDK
- **`LandPlayedEvent`** (CR 305.1) — emitted by `PlayLandHandler` alongside the entry `ZoneChangeEvent` on **every** play path, carrying the zone it was played from. Distinct from an effect *putting* a land onto the battlefield, so "whenever you play a land…" does not over-trigger on fetch / ramp / reanimation. Exposed via `Triggers.youPlayLand(fromZoneOtherThan = …)`, indexed under a new `LAND_PLAYED` trigger category. Emits no separate client event (the `ZoneChangeEvent` already renders the land entering).
- **Land-play provenance** — `LandsPlayedThisTurnComponent(List<Zone>)` records the origin zone of each land played this turn (mirroring the spell path's `castFromZone`), backing `Conditions.YouPlayedLandThisTurn(fromZone? / fromZoneOtherThan?)`, with `YouPlayedLandFromNonHandThisTurn` as a convenience alias.
- **`fromZoneOtherThan`** added to `YouCastSpellsThisTurn` / `PlayerCastSpellsThisTurn` — "cast a spell from anywhere other than your hand." Mutually exclusive with `fromZone` (enforced by `require`).
- **Mayhem land-play** — `PlayLandEnumerator` + `PlayLandHandler` offer a discarded-this-turn Mayhem *land* as a play-from-graveyard action, and stamp `EnteredFromGraveyardComponent` on graveyard land-plays so the "enters from a graveyard" ETB fires.

### Tests
Scenario coverage for all three cards — Oscorp Mayhem land-play + enters-from-graveyard life loss; Shadow's Undying Vengeance firing on a non-hand land-play (and not on a hand play); Spider-Man 2099's end-step trigger and its "From the Future" cast restriction (uncastable turns 1–3, castable turn 4) — plus a dev scenario JSON. `card-sdk-language-reference.md` updated for every new SDK primitive.

Full `:mtg-sdk` + `:rules-engine` suites green; SPM card snapshot re-blessed.


bolav:
## Review — SPM Batch 3 (land-play trigger / turn-condition / Mayhem land-play)

Scope: the delta this branch adds on top of its parent `spm-damage-triggers` (2 commits). Recent `main` is already merged in via the parent.

### Verdict

The behavior is right and the SDK shape is well-judged. This closes a genuine gap — the engine had no way to distinguish *playing* a land (CR 305.1 special action) from a land an effect *puts* onto the battlefield, so "whenever you play a land…" couldn't be modeled without over-triggering on fetch/ramp/reanimation. The new `LandPlayedEvent` + `Triggers.youPlayLand` do this cleanly, and the `fromZoneOtherThan` parameter is parameterized rather than hardcoded to HAND. All scenario tests pass; CR citations verified; printing gate green; docs in sync.

### What's good

- **The new event is genuinely novel, not a wrapped primitive.** `LandPlayedEvent` reads a game action no existing pattern could — `ZoneChangeEvent(→BATTLEFIELD)` can't tell a play from a put. The `EventPattern.kt` comment explaining exactly this gap is excellent.
- **`PlayLandHandler` wiring is complete.** `landPlayedEvent` rides alongside the entry `ZoneChangeEvent` on **all five** success paths (OnEnter, EntersAsCopy, shock-land pay-life, EntersWithChoice, normal finish), mirroring the existing `riderPlayEvent` pattern. The easy thing to get wrong, done right.
- **Both anti-corruption seams handled:** `ClientEvent.kt` has the `LandPlayedEvent -> null` branch (correct — the ZoneChangeEvent renders entry) and `Serialization.kt` registers both event and component.
- **Shadow's two-ability decomposition is correct and documented** — the land-play and spell-cast sub-events are mutually exclusive per game event, so no double-fire.
- **Tests exercise the real corners** — positive and negative for all three cards, and the Mayhem land-play-from-graveyard path end-to-end.
- CR citations (**305.1**, **702.187c**) verified against the rules; printing placement passes for all three cards.

### Issues (all addressed — see Resolution)

**Important**
1. *Spider-Man 2099's "From the Future" cast restriction had no test.* The `castOnlyIf(Not(ControllerTurnsTakenAtMost(3)))` logic is correct (turns are 1-indexed: `AtMost(3)` = turns 1–3, `Not` = turn 4+), but was untested.

**Minor**
2. *`fromZoneOtherThan` mutual-exclusivity documented but unenforced* on `PlayerCastSpellsThisTurn` — setting both silently ANDs rather than erroring.
3. *Asymmetry between land-play and spell-cast provenance.* Spells got full per-zone records; land plays got only a boolean `PlayedLandFromNonHandThisTurnComponent` baking in "non-hand," so a future "from library" card would force a new component. Defensible YAGNI, but `LandPlayedEvent` already carries `fromZone`, so symmetry was cheap.

### Resolution — fixed in `acb789bfc`

1. Added a scenario test asserting Spider-Man 2099 is uncastable on turn 3 and castable on turn 4 (with untapped `{U}`/`{R}` sources so affordability is not the confound).
2. Added `require(fromZone == null || fromZoneOtherThan == null)` on both `PlayerCastSpellsThisTurn` and the new land condition.
3. Reshaped the tracker to `LandsPlayedThisTurnComponent(List<Zone>)` (appended on every land play) backing `PlayerPlayedLandThisTurn(fromZone / fromZoneOtherThan)`, mirroring the spell path; `YouPlayedLandFromNonHandThisTurn` kept as a convenience alias so the card definition is unchanged. SPM snapshot re-blessed; docs updated.

Full `:mtg-sdk` + `:rules-engine` suites green after the fixes.

PR #1521, opened 2026-07-29
Add three Wilds of Eldraine cards

wingedsheep:
Three Wilds of Eldraine cards, all composed from existing primitives.

## Cards

- **Rowan's Grim Search** `{2}{B}` Instant — bargain; if bargained, look at the top four, put up to
  two back on top in any order and the rest into the graveyard; then draw two and lose 2 life.
  Composes `bargain()`, `ConditionalEffect(Conditions.WasBargained, …)`, an `Effects.Pipeline`
  (`gather(TopOfLibrary(4))` → `chooseUpToSplit(2)` → `toLibraryTop` with `CardOrder.ControllerChooses`
  / `toGraveyard`), `Effects.DrawCards`, `Effects.LoseLife`.

- **Thunderous Debut** `{6}{G}{G}` Sorcery — bargain; look at the top twenty, may reveal up to two
  creature cards, bargained puts them onto the battlefield and otherwise into hand, then shuffle.
  Composes `bargain()`, an `Effects.Pipeline` (`gather(TopOfLibrary(20))` →
  `chooseUpTo(2, filter = Creature, showAllCards)` → `reveal`), a `ConditionalEffect(WasBargained)`
  over two `MoveCollectionEffect`s on the *same* collection (so the look-and-pick is identical either
  way), and `ShuffleLibraryEffect`. Nothing puts the unpicked cards back — a top-of-library gather
  only reads ids, so the other eighteen are still in the library when the shuffle runs.

- **The Goose Mother** `{X}{G}{U}` Legendary Creature — Bird Hydra 2/2 — flying; enters with X +1/+1
  counters; enters trigger creates half X Food rounded up; attack trigger may sacrifice a Food to
  draw. Composes `keywords(FLYING)`, `EntersWithDynamicCounters(DynamicAmount.CastX)`,
  `Effects.CreateFood(Divide(CastX, 2, roundUp = true))`, and `MayEffect(Sacrifice(Food).then(Draw))`
  with a `FeasibilityCheck.ControlsPermanentMatching` gate (Provisions Merchant's shape) so a flier
  that attacks every turn stops asking an unanswerable question. Both X clauses read `CastX` rather
  than `XValue` because the enters trigger is a separate resolution from the spell.

## SDK

One facade overload: `Effects.CreateFood(count: DynamicAmount, controller)`, mirroring the existing
`CreateBlood(DynamicAmount)`. No new effect type and no new executor — `CreatePredefinedTokenEffect`
already carries `dynamicCount` and the executor already evaluates it.
`docs/card-sdk-language-reference.md` is updated in the same change.

## Dropped from this unit

Two of the five queued cards were dropped because they need engine capability, not card wiring:

- **Feral Encounter** — needs a delayed triggered ability with **two** target requirements ("target
  creature you control deals damage equal to its power to **up to one target** creature you don't
  control", both chosen when the delayed ability goes on the stack).
  `CreateDelayedTriggerEffect.targetRequirement`, `DelayedTriggeredAbility.targetRequirement`, and
  the `TriggeredAbility.create(...)` call in `TriggerDetector` are all singular. Widening that to a
  list is a cross-layer SDK + engine change. The impulse-exile half of the card is fully supported
  today; only the combat clause blocks it.

- **Faunsbane Troll** — the activated ability's **cost** is "Sacrifice an Aura attached to this
  creature". The natural filter is `Enchantment.withSubtype("Aura").attachedToSource()`, but the cost
  path never populates `sourceId` in its predicate context: both
  `CostHandler.findMatchingPermanentsUnified` and `CostHandler.paySacrificeList` build
  `PredicateContext(controllerId = controllerId)`, and `StatePredicate.IsAttachedToSource` returns
  `false` with no source — the ability would be permanently unactivatable. No existing card uses
  `.attachedToSource()` in a cost (only in effect-time `SelectTargetEffect` and static filters), so
  this is untested ground rather than a regression, but threading `sourceId` through changes shared
  cost matching for every card and wants its own PR with tests.

## What was checked

- `just build` — green (compile + full test suite).
- `just rebless-cards` — the only snapshot movement is three pure insertions into
  `mtg-sets/src/test/resources/snapshots/cards/WOE.json`; no existing card moved.
- `just check-card-printing` for all three — WOE is the earliest real printing in every case, so all
  three canonicals live here and no `Printing(...)` rows are needed.
- Card art URIs verified against Scryfall (HTTP 200) and card fields re-checked field-by-field
  against the `&set=woe` API response.

## What was NOT checked

- No manual playthrough in the web client. In particular nobody has eyeballed Thunderous Debut's
  twenty-card selection UI or Rowan's Grim Search's "put up to two on top, then order them" flow.
- No UX pass from both seats, and no e2e test.
- No new scenario tests — these three cards add no new engine behaviour, so they're covered by the
  snapshot and lint nets only.

This PR came out of an autonomous loop and **independent review is still pending**: a reviewer agent
will comment its findings here and a corrector agent may push follow-up commits, so this is not final
at open time.


wingedsheep:
**Verdict: request changes — 1 blocking, 1 important, 4 minor.** The three cards look behaviourally right and the SDK change is a clean facade twin, but the PR ships **zero scenario tests** for three cards, two of which use shapes the author's own notes flag as untrodden. Do not merge until the tests exist.

---

## Verdict

The card modelling is correct as far as static reading goes: oracle text, mana costs, P/T, rarity, collector numbers, artists and image URIs all match Scryfall exactly, `just check-card-printing` is clean for all three (WOE is the earliest real expansion for each; `pwoe`/`pmei`/`soc` aren't scaffolded so no `Printing(...)` rows are owed), and the one SDK addition — `Effects.CreateFood(count: DynamicAmount)` — is a facade overload over machinery that already exists, with `docs/card-sdk-language-reference.md` updated in the same commit. The SDK shape is right; nothing here warranted a new effect type.

What's missing is proof. The PR body states "No new scenario tests — these three cards add no new engine behaviour, so they're covered by the snapshot and lint nets only." The snapshot net proves serialization, not behaviour, and the claim is contradicted by `LOOP_UNIT.md`, which flags three specific shapes as the least-trodden in the change. That is the blocking issue.

`origin/main` was merged into the branch cleanly before review (merge commit `2c8123855d`, local to the worktree, not pushed). Main's "split batched card tests into one file per card" landed underneath and does not touch anything this PR adds. No gate re-run was needed.

---

## What's good

- **No new SDK type.** `Effects.CreateFood(count: DynamicAmount, controller)` at `mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/dsl/Effects.kt:2416` is exactly the shape the existing `CreateTreasure` / `CreateBlood` / `CreateMap` / `CreateMutagen` dynamic overloads already use — `CreatePredefinedTokenEffect.dynamicCount` and `CreatePredefinedTokenExecutor` needed no change, and the executor's `coerceAtLeast(0)` already handles the X=0 case correctly. Doc line updated in the same commit. This is the right bar.
- **`CastX` over `XValue` on The Goose Mother is correct**, and better than its neighbours. I traced it: `EntersWithReplacements.applySelf` builds `EffectContext(sourceId = entityId, xValue = xValue)` and `DynamicAmountEvaluator` line 149 reads `CastChoicesComponent.x` → `SpellOnStackComponent.xValue` → `context.xValue`, so `CastX` is a strict superset of `XValue` on the replacement path and the only reading that works for the separate ETB-trigger resolution. The uncertainty raised in `LOOP_UNIT.md` resolves in the change's favour.
- **The global/self split on `EntersWithDynamicCounters` is safe.** The default `appliesTo` is `Creature.youControl()`, but `EntersWithReplacements.applyGlobal:226` early-`continue`s unless `otherOnly` is set, so the Goose Mother can't accidentally counter every creature you play. Same as Morlun/Marketback Walker.
- **The attack trigger is Provisions Merchant verbatim**, including the `FeasibilityCheck.ControlsPermanentMatching` gate that suppresses the prompt with no Food — the right precedent, correctly reused.
- **Both drop justifications check out.** `CreateDelayedTriggerEffect.targetRequirement` is singular (`CompositeEffects.kt:735`), and `CostHandler` does build `PredicateContext(controllerId = controllerId)` with no `sourceId` at :622/:839/:920/:968/:1013 while `PredicateEvaluator.kt:1229` `IsAttachedToSource` returns `false` without one. Feral Encounter and Faunsbane Troll are genuinely `add-feature` work, not card wiring.
- **CR 702.166c is cited correctly** in both bargain cards' KDoc (verified against the 2026-06-19 Comprehensive Rules: 702.166c is "Objects with bargain have additional abilities that specify what happens if they were bargained…"). Every `[CardName]` KDoc cross-reference resolves to a real file.

---

## Issues

### Blocking

**1. No scenario tests for any of the three cards.**
`mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/woe/cards/RowansGrimSearch.kt:38`, `ThunderousDebut.kt:42`, `TheGooseMother.kt:42`

A repo-wide search for the three card names finds hits only in the card files, the SDK doc-comment, the WOE snapshot, and the static legalities/coverage resources — there is no test anywhere. `add-card` requires a scenario test per card and AGENTS.md's hard rule is "one card, one test file — `<CardName>ScenarioTest.kt`". The PR body's rationale ("no new engine behaviour … covered by the snapshot and lint nets") doesn't hold: the snapshot is a serialization golden, and `LOOP_UNIT.md` itself names three shapes as the least-trodden in the change. Review rule §5 also applies directly — all three cards cite CR 702.166c and nothing exercises the bargained-vs-unbargained branch, which makes the citation decorative.

Add three files under `rules-engine/src/test/kotlin/com/wingedsheep/engine/scenarios/` — **not** a batched file, and note main has since deleted every `WoeCardsBatch*ScenarioTest.kt` in favour of per-card files, so a batch file would now be actively regressive.

`RowansGrimSearchScenarioTest.kt`
- Unbargained cast: draws 2, loses 2 life, top of library untouched, graveyard gains only the spell.
- Bargained cast (sacrifice a token/artifact): 4 looked at, 2 chosen back on top **in the chosen order**, other 2 in the graveyard — then assert the two cards drawn are the two just placed on top. That ordering (dig before draw) is the specific claim `RowansGrimSearch.kt:35-36` makes and nothing proves it.
- Bargained, choose zero: all 4 to graveyard, draw comes off the new top.
- Library with fewer than 4 cards: no deadlock (the `ChooseUpTo`-caps-at-collection-size claim at `RowansGrimSearch.kt:31-33`).

`ThunderousDebutScenarioTest.kt`
- Unbargained: two chosen creature cards land in hand; assert the other 18 are still in the library afterwards (this is the "gather only reads ids" implicit contract flagged in `LOOP_UNIT.md`, and the only thing that makes "Then shuffle" correct).
- Bargained: the same two picks land on the battlefield under your control, and their ETB triggers fire.
- Pick zero: nothing moves, shuffle still happens.
- Non-creature cards are not selectable — assert the decision's legal options contain only the creature cards out of the twenty.

`TheGooseMotherScenarioTest.kt`
- X=4 → enters as a 6/6 with four +1/+1 counters and creates 2 Food; X=3 → three counters and **2** Food (rounds up); X=0 → 2/2 with no Food.
- Attacks with a Food: prompt appears, accept → Food sacrificed, one card drawn.
- Attacks with no Food: assert **no pending decision at all** (the feasibility gate).
- Enters without being cast (put onto the battlefield by an effect): X = 0, so no counters and no Food. This is the `CastX` corner the KDoc argues about, and Thunderous Debut's bargained mode can produce it for real.

### Important

**2. The one shape with real risk is the one with no coverage: Thunderous Debut reads a pipeline collection across a gate frame, after a decision pause.**
`mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/woe/cards/ThunderousDebut.kt:65-77`

`MoveCollectionEffect(from = chosen.key, …)` sits inside a `ConditionalEffect` (i.e. a `GatedEffect` with `Gate.WhenCondition`) nested inside the pipeline, and `chosen` is produced by a `chooseUpTo` that pauses for a player decision. I traced the path and it should work — `CompositeEffectExecutor:107-120` merges `updatedCollections` into `currentContext`, `GatedEffectExecutor:81-89` passes that same context straight through to `then`/`otherwise`, and the pre-pushed `EffectContinuation` carries the whole `EffectContext` (including `declaredCostSlot`, which is what `ConditionEvaluator:498` reads for `WasBargained`) across the pause. But "I read the executors and it should work" is not the standard; this is the card most likely to silently move zero cards, and both branches need an actual passing assertion. Treat the bargained and unbargained Thunderous Debut cases from finding 1 as the must-have, not the nice-to-have.

### Minor

**3. `reveal(chosen)` should pass `revealToSelf = false`.**
`ThunderousDebut.kt:64`

The controller just hand-picked these cards from a twenty-card pile; showing them a reveal overlay of their own choice is the redundant-overlay case the SDK explicitly documents against (`PipelineEffects.kt:1076-1080`: "Set to false when the controller chose the cards themselves — they already know what's being 'revealed'; only opponents see the overlay"). The existing plain `reveal(found)` precedents (Skyserpent Seeker, Guidelight Pathmaker, Dina's Guidance) are all engine-found cards, where the default is right. Change to `reveal(chosen, revealToSelf = false)`.

**4. Truncated bargain rulings.**
`RowansGrimSearch.kt:87-90`, `ThunderousDebut.kt:97-100`

The "If you copy a bargained spell, the copy is also bargained." ruling is cut off mid-thought — Scryfall's text continues "If a card or token enters the battlefield as a copy of a permanent that's already on the battlefield, the new permanent isn't bargained, even if the original was", which is the half that actually resolves a corner case. Both cards also carry three of the five rulings Scryfall lists (missing "Bargain means 'As an additional cost…'" and the additional-targets one). Note the already-merged Farsight Ritual has the identical truncation, so this is a repo-wide convention rather than a new defect — fix or leave at the corrector's discretion, but don't let the truncated sentence spread further.

**5. The Rowan's Grim Search dig is a near-twin of `Patterns.Library.lookAtTopAndKeep` and the KDoc doesn't say why it isn't used.**
`RowansGrimSearch.kt:54-65`

The inline pipeline is structurally identical to `LibraryPatterns.lookAtTopAndKeep` (gather top N → split select → kept destination → rest destination); the only thing that doesn't fit is that the pattern hard-codes `SelectionMode.ChooseExactly` and this card needs `ChooseUpTo`. Given the project's "inline until the second user" preference, the inline pipeline is the right call today — but the KDoc at :28-33 compares it to surveil and never mentions the pattern it's shadowing, which is the first question a maintainer will ask. One sentence pointing at `lookAtTopAndKeep` and naming the `ChooseExactly` blocker would keep the `Patterns` index discoverable; if a second "look at top N, keep up to M" card shows up, that's the moment to add a selection-mode parameter rather than a second inline copy.

**6. Card-specific name in an SDK doc-comment.**
`mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/dsl/Effects.kt:2413-2415`

"…(The Goose Mother: 'create half X Food tokens, rounded up')" bakes one card's name into the shared SDK surface. The adjacent `CreateBlood` overload does exactly the same with Lacerate Flesh, so this is consistent and I'd leave it — flagging only so it's a deliberate choice rather than drift.

---

## Recommendation

Add the three `<CardName>ScenarioTest.kt` files with the cases listed in finding 1, prioritising the Thunderous Debut bargained/unbargained pair (finding 2) since that's the shape with real failure risk. Flip `reveal(chosen, revealToSelf = false)` while you're in the file. Findings 4–6 are discretionary. Nothing about the card modelling, the printing placement, or the SDK overload needs to change — this is a coverage gap, not a design problem.

wingedsheep:
Corrections pushed as `0a9455b` (on top of the existing history).

**1 — Blocking: no scenario tests.** Fixed. Three per-card files, no batching:

- `rules-engine/src/test/kotlin/com/wingedsheep/engine/scenarios/RowansGrimSearchScenarioTest.kt` — unbargained cast (no prompt at all, top of library untouched, only the spell in the graveyard, life 18); bargained cast asserting the *dig runs before the draw* (the two cards put back on top are the two drawn, in the order submitted to the `ReorderLibraryDecision`, and the cards below the top four never move); keep-zero (all four binned, no second prompt, draw comes off the new top); three-card library (the `ChooseUpTo` offers what's there and resolves).
- `.../ThunderousDebutScenarioTest.kt` — unbargained (picks to hand, the other eighteen still in the library) and bargained (same picks onto the battlefield) as the must-have pair for finding 2; the decision's `options` are asserted to contain only the creature cards out of the twenty; reveal-nothing asserts a `LibraryShuffledEvent` is still emitted.
- `.../TheGooseMotherScenarioTest.kt` — X=4 → 6/6, four counters, two Food; X=3 → three counters, still two Food; X=0 → 2/2, nothing; attack with a Food (prompt, accept, Food sacrificed, card drawn); attack with no Food (no pending decision at all); and the `CastX` corner — put onto the battlefield by Thunderous Debut's bargained mode, so X is 0 and there are neither counters nor Food.

All 13 pass.

**2 — Important: the gate-frame collection read.** Covered by the bargained/unbargained Thunderous Debut pair above; both branches move the picked cards for real across the decision pause, so the shape is now asserted rather than traced.

**3 — Minor: `revealToSelf`.** Fixed — `reveal(chosen, revealToSelf = false)` in `ThunderousDebut.kt`.

**4 — Minor: truncated bargain ruling.** Fixed on both cards. Completed against Scryfall's actual text ("… If a card or token enters the battlefield as a copy of a permanent that's already on the battlefield, the new permanent isn't bargained, even if the original was."). I did not add the other two rulings Scryfall lists, and did not touch the already-merged Farsight Ritual — that's outside this unit.

**5 — Minor: `lookAtTopAndKeep` not mentioned.** Fixed — one paragraph in `RowansGrimSearch.kt`'s KDoc naming the pattern, the `SelectionMode.ChooseExactly` blocker, and the "second such card is when to widen the pattern" trigger.

**6 — Minor: card name in an SDK doc-comment.** Declined, as the review suggested: it matches the adjacent `CreateBlood`/Lacerate Flesh overload, so it's the existing convention rather than drift.

**Gate:** `just build` — passed. The ruling-text edit moved the WOE golden, so `just rebless-cards` was run; the snapshot diff is three lines, all on this unit's two cards (the two ruling texts plus `"revealToSelf": false`). No printing placement changed, so `check-card-printing` was not re-run.

Nothing left open.

PR #1522, opened 2026-07-29
Add three Wilds of Eldraine cards

wingedsheep:
Three Wilds of Eldraine cards, all built from primitives that already exist — no new effect, executor, condition, or keyword.

## Cards

- **Rowdy Research** (#312) — `{6}{U}` instant: draw three cards, costs `{1}` less for each creature that attacked this turn. Composes `ModifySpellCost(SelfCast, ReduceGenericBy(CostReductionSource.CreaturesThatAttackedThisTurn()))` with `Effects.DrawCards(3)` — the same turn-history count Witchstalker Frenzy uses, so a copy cast after combat still sees the attackers that already died.
- **Hylda's Crown of Winter** (#247) — legendary artifact. `{1}, {T}: Tap target creature` with `genericCostReduction = DynamicAmount.Conditional(Conditions.IsYourTurn, 1, 0)` (Starport Security's shape with a turn condition instead of a board one); `{3}, Sacrifice: Draw a card for each tapped creature your opponents control` via `Effects.DrawCards(DynamicAmount.AggregateBattlefield(Player.EachOpponent, Creature.tapped()))`.
- **Faerie Slumber Party** (#311) — `{4}{U}{U}` mass bounce with a per-opponent Faerie payoff. The count has to be taken *before* the bounce (afterwards there's nothing left to count) and the tokens created *after* it (created first they'd be returned themselves and cease to exist), so it's an inline `Effects.Pipeline`: `gather(Creature)` → `filterSplit(opponentControls())` → `captureControllers(theirs)` → `move(all, ToZone(HAND))` → `woeFaerieToken(2 × DistinctEntitiesInCollections(controllers))`. Reuses the existing WOE Faerie token helper.

Also adds `blc/cards/RowdyResearchReprint.kt` — Bloomburrow Commander is scaffolded and reprints Rowdy Research in 2024, so it gets a presentation-only `Printing(...)` row while WOE stays canonical — and ticks the card off both BLC backlogs.

## Dropped from the unit (2 of 5)

Both needed new engine vocabulary, which `CONTRIBUTING.md` keeps to one new-primitive card per PR:

- **Eerie Interference** — "Prevent all damage that would be dealt to you and creatures you control this turn by creatures." No existing shield combines a recipient scope (a player *plus* a group) with a source filter; the only source-filtered modification (`PreventCombatDamageFromGroup`) is combat-only, and `PreventDamageExecutor` short-circuits on `recipientGroup != null` before reading `sourceFilter` at all.
- **Troyan, Gutsy Explorer** — "Spend this mana only to cast spells with mana value 5 or greater or spells with `{X}` in their mana costs." Needs a new `ManaRestriction` variant (`SpellsMV4OrGreater` has no `{X}` branch; `CreatureMV4OrXCost` is creature-only) plus a branch in `ManaPool.isSatisfiedBy`.

## What was checked

`just build` — passed. `just rebless-cards` re-blessed `snapshots/cards/WOE.json`; only the three new cards were added, nothing pre-existing moved. `just check-card-printing` is clean for all three. 15 new scenario tests (7 + 5 + 3), one file per card, all green.

## What was **not** checked

This PR came out of an autonomous loop. None of the following happened:

- **No manual playthrough in the web client.** Nobody clicked Hylda's Crown's tapper, watched the Faerie tokens land, or confirmed the ability button text reads sensibly.
- **No UX pass from both seats** — in particular Hylda's Crown's targeting flow and whether the turn-conditional discount is legible in the action menu.
- **No e2e / Playwright run.** No `e2e-scenarios` spec was added or executed.
- No multiplayer (>2 player) verification of Faerie Slumber Party's per-opponent count; the tests are two-player.

One modelling note: Faerie Slumber Party snapshots the *controllers* of the opponents' creatures before the bounce and counts the distinct ones, rather than counting opponents who control a creature at resolution. (An earlier revision of this PR claimed no primitive existed for "distinct opponents among a gathered collection" — that was wrong; `captureControllers` + `DynamicAmount.DistinctEntitiesInCollections` compose into exactly that, and the card now uses them.) The one remaining divergence from "controlled a creature returned this way" is a creature that fails to leave the battlefield at the move step.

## Not final

Independent review is still pending: a reviewer agent will comment its findings on this PR and a corrector agent may push follow-up commits. Please don't treat the diff as settled at open time.

### DIFF
--- a/backlog/sets/bloomburrow-commander/cards.md
+++ b/backlog/sets/bloomburrow-commander/cards.md
@@ -2,7 +2,7 @@
 
 **Set Size:** 312 cards
 **Release Date:** August 2, 2024
-**Implemented:** 112 / 312
+**Implemented:** 113 / 312
 | Color      | Total | Done |
 |------------|-------|------|
 | White      | 37    | 0    |
@@ -232,7 +232,7 @@
 - [x] Rootbound Crag
 - [ ] Rootcast Apprenticeship
 - [ ] Rose Room Treasurer
-- [ ] Rowdy Research
+- [x] Rowdy Research
 - [ ] Rugged Prairie
 - [x] Sakura-Tribe Elder
 - [ ] Saw in Half
--- a/backlog/sets/bloomburrow-commander/decks/family-matters.md
+++ b/backlog/sets/bloomburrow-commander/decks/family-matters.md
@@ -56,7 +56,7 @@
 - [ ] 1 Path to Exile
 - [ ] 1 Pull from Tomorrow
 - [ ] 1 Rapid Hybridization
-- [ ] 1 Rowdy Research
+- [x] 1 Rowdy Research
 
 ## Sorceries (8)
 
--- a/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/blc/cards/RowdyResearchReprint.kt
+++ b/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/blc/cards/RowdyResearchReprint.kt
@@ -1,0 +1,21 @@
+package com.wingedsheep.mtg.sets.definitions.blc.cards
+
+import com.wingedsheep.sdk.model.Printing
+import com.wingedsheep.sdk.model.Rarity
+
+/**
+ * Rowdy Research reprint in Bloomburrow Commander (BLC). The canonical
+ * [com.wingedsheep.sdk.model.CardDefinition] lives in Wilds of Eldraine's `cards/` package;
+ * this file contributes only the BLC presentation row.
+ */
+val RowdyResearchReprint = Printing(
+    oracleId = "1a33a5c2-5aa1-48e5-95a8-d52cdce980de",
+    name = "Rowdy Research",
+    setCode = "BLC",
+    collectorNumber = "173",
+    scryfallId = "22737b82-bb81-4e4d-8ce7-7f06c5692d9b",
+    artist = "Bram Sels",
+    imageUri = "https://cards.scryfall.io/normal/front/2/2/22737b82-bb81-4e4d-8ce7-7f06c5692d9b.jpg?1783910682",
+    releaseDate = "2024-08-02",
+    rarity = Rarity.UNCOMMON,
+)
--- a/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/woe/cards/FaerieSlumberParty.kt
+++ b/mtg-sets/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/woe/cards/FaerieSlumberParty.kt
@@ -2,13 +2,10 @@ package com.wingedsheep.mtg.sets.definitions.woe.cards
 
 import com.wingedsheep.sdk.core.Zone
 import com.wingedsheep.sdk.dsl.Effects
-import com.wingedsheep.sdk.dsl.Patterns
 import com.wingedsheep.sdk.dsl.card
 import com.wingedsheep.sdk.model.Rarity
 import com.wingedsheep.sdk.scripting.GameObjectFilter
-import com.wingedsheep.sdk.scripting.conditions.Exists
-import com.wingedsheep.sdk.scripting.filters.unified.GroupFilter
-import com.wingedsheep.sdk.scripting.references.Player
+import com.wingedsheep.sdk.scripting.effects.CardDestination
 import com.wingedsheep.sdk.scripting.values.DynamicAmount
 
 /**
@@ -22,15 +19,23 @@ import com.wingedsheep.sdk.scripting.values.DynamicAmount
  *
  * Ordering is the whole difficulty. The tokens are creatures, so they must be created *after*
  * the mass bounce (created first, they would be returned themselves and cease to exist), but the
- * payoff counts a board state that the bounce has already erased. The pipeline therefore
- * snapshots the count into a number slot first, bounces, then reads the slot back:
+ * payoff counts controllers that the bounce has already erased — `ControllerComponent` is stripped
+ * when a permanent leaves the battlefield, and `move` to a hand routes by *owner*, so reading
+ * controllers afterwards would answer the wrong question for a stolen creature.
  *
- *   storeNumber(CountPlayersWith(EachOpponent, "controls a creature")) → returnAllToHand → tokens
+ * So the pipeline snapshots the controllers while the creatures are still in play (the Builder's
+ * Bane / Break the Spell shape):
  *
- * [DynamicAmount.CountPlayersWith] rebinds the controller per candidate, so `Player.You` inside
- * the [Exists] refers to the opponent being tested — the Bandit's Talent idiom. Because *every*
- * creature is returned, "opponents who controlled a creature returned this way" and "opponents
- * who controlled a creature as this resolved" are the same set.
+ *   1. `gather(Creature)` — every creature, any controller; this is the set that gets returned.
+ *   2. `filterSplit(… opponentControls())` — the opponents' half, evaluated pre-bounce.
+ *   3. `captureControllers(theirs)` — a parallel list of controller ids, snapshotted pre-bounce.
+ *   4. `move(all, ToZone(HAND))` — battlefield → hand always collapses to the owner's hand.
+ *   5. `DistinctEntitiesInCollections(controllers)` de-duplicates that list down to *how many
+ *      distinct opponents*, doubled for the two Faeries each.
+ *
+ * Because the gather is unconditional, "opponents who controlled a creature returned this way" and
+ * "opponents who controlled a creature as this resolved" only diverge if some creature fails to
+ * leave the battlefield at step 4.
  */
 val FaerieSlumberParty = card("Faerie Slumber Party") {
     manaCost = "{4}{U}{U}"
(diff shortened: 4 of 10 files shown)


wingedsheep:
**Verdict: approve with follow-ups — 0 blocking, 2 important, 6 minor.** Nothing here should hold the merge.

Three cards, zero new SDK surface, all metadata verified byte-exact against Scryfall. The behaviour is right; the two Important items are a doc-comment claim that is factually wrong about what the SDK can already express, and one missing rule-corner test.

`origin/main` was already merged (`git merge origin/main` → *Already up to date*), so no conflict resolution and no merge commit. Reviewed in place in `.claude/worktrees/agent-a631d01f1b5a704a4`. Per the loop protocol the full suite was not re-run (Stage A gated green and the merge changed nothing); `scripts/check-card-printing.py` was run fresh for all three cards and exits 0.

## What's good

- **No new primitives across three cards.** Hylda's Crown is a line-for-line reuse of Starport Security's `genericCostReduction = DynamicAmount.Conditional(...)` shape with a turn condition swapped for a board one — exactly the reuse the SDK bar asks for. Rowdy Research rides Witchstalker Frenzy's `CostReductionSource.CreaturesThatAttackedThisTurn`, which is the right choice and for the right reason (`CostCalculator.kt:460` reads per-player `PlayerAttackersThisTurnComponent` history, not a battlefield scan).
- **Metadata is exact.** I re-fetched all three from Scryfall: oracle text, mana cost, type line, rarity, collector number, artist (including `Volkan Baǵa`'s combining char), flavor text, and image URI all match character-for-character, as do the BLC row's `oracleId` / `scryfallId` / `releaseDate` / `rarity`.
- **Printing placement is correct.** `check-card-printing` is clean for all three: canonical in WOE (the earliest real expansion; the two PWOE promos are correctly skipped), and BLC gets a presentation-only `Printing(...)` row that matches the repo's `*Reprint.kt` convention exactly (compare `blc/cards/ChartACourseReprint.kt`).
- **Faerie Slumber Party's ordering problem is real and correctly solved**, and the test pins *both* failure directions: tokens created too early would be bounced (0 tokens), a count taken too late would be 0. Test 1 also distinguishes "per opponent" from "per creature" — one opponent with two creatures still yields 2 tokens.
- **Hylda's Crown's discount is proved by mana starvation, not by reading a number back.** That is the right way to test a cost reduction. It also genuinely discriminates: if the discount leaked to an opponent's turn, the cost would be `{0}` + `{T}`, there'd be no mana decision, the `{T}` would be paid immediately and the crown-not-tapped assertion would fail.
- **The token reuse is right.** WOE prints exactly one Faerie token (Scryfall `TWOE` #5: 1/1 blue, flying, "can block only creatures with flying"), so `woeFaerieToken` is the correct helper and needed no change.
- **Owner routing on the mass bounce holds** even though no test covers it: `MoveCollectionExecutor.kt:655` collapses battlefield→hand to `ownerId`, so "their owners' hands" is correct for stolen creatures despite `CardDestination.ToZone(Zone.HAND)` defaulting to `Player.You`.
- **The PR body is honest** about what wasn't checked (no UI playthrough, no e2e, no multiplayer), and the two dropped cards are dropped for correct, specific reasons — I verified `PreventCombatDamageFromGroup` is combat-only and that no `ManaRestriction` covers "MV ≥ 5 or has {X}".

## Issues

### Important

**1. `mtg-sets/.../woe/cards/FaerieSlumberParty.kt:31-33` — the doc comment (and the PR body / LOOP_UNIT) claim a faithful version would need a new primitive. It wouldn't; the primitive exists.**

The card counts *opponents who control a creature when the spell starts resolving*, not *opponents who controlled a creature returned this way*. The comment defends this by asserting the two sets are identical, and the PR body goes further: "there is no primitive today for 'distinct opponents among a gathered collection'". That is not true. `PipelineBuilder.captureControllers` (`PipelineBuilder.kt:538`) writes a parallel list of projected controller ids into a collection slot, and `DynamicAmount.DistinctEntitiesInCollections` (`DynamicAmount.kt:711`) de-duplicates a collection by entity id (`DynamicAmountEvaluator.kt:265-270`). Composed, they *are* "the number of distinct opponents among a gathered collection":

```kotlin
effect = Effects.Pipeline {
    val creatures = gather(CardSource.BattlefieldMatching(GameObjectFilter.Creature))
    val theirs = filterSplit(creatures, GameObjectFilter.Creature.opponentControls()).matching
    val controllers = captureControllers(theirs, name = "faerieSlumberPartyControllers")
    move(creatures, CardDestination.ToZone(Zone.HAND))
    run(woeFaerieToken(count = DynamicAmount.Multiply(
        DynamicAmount.DistinctEntitiesInCollections(listOf(controllers.key)), 2
    )))
}
```

(Capture must stay *before* the move — once the cards are in hand `captureControllers` would read owners, not the pre-bounce controllers.) Either adopt this — it drops the `CountPlayersWith` + `Exists` indirection and models the printed text directly — or keep the current spelling and rewrite lines 31-33 to say plainly *"this approximates 'returned this way' as 'controls a creature at resolution'; they differ only when a creature cannot leave the battlefield"*, and drop the "no primitive exists" claim from the PR body. The wrong claim is the part that must not survive: it will be read as SDK ground truth by the next agent that hits this shape. Behavioural impact of the approximation itself is negligible.

**2. `rules-engine/.../RowdyResearchScenarioTest.kt` — no test where the *opponent's* attackers pay for the spell.**

Both the card doc comment (`RowdyResearch.kt:21`, "unioned over every player's attackers") and the engine (`CostCalculator.kt:460-467` sums `state.turnOrder`) deliberately count attackers controlled by *any* player — and that is the card's actual use case, since it is an instant you hold up while the opponent swings. Every test in the file has player 1 as the attacker, so nothing pins the union. Add a case: player 2 active, player 2 attacks with two creatures, then assert `calculateEffectiveCost(...)` for player 1's Rowdy Research is 4 generic. Without it the "any player" behaviour is asserted only by a comment.

### Minor

**3. `rules-engine/.../RowdyResearchScenarioTest.kt:152` — the discount is never exercised on a real cast.** Tests 1-4 call `CostCalculator.calculateEffectiveCost` directly; the only actual `castSpell` is at full price with 7 Islands. A cast with two attackers and only 5 Islands would prove the reduction reaches the enumerator + handler path, not just the calculator. Note this matches `WitchstalkerFrenzyScenarioTest`'s house style, so it's an improvement rather than a defect.

**4. `mtg-sets/.../woe/cards/RowdyResearch.kt:26-27` — untested rules claim in a comment.** "a Rowdy Research on the stack is always mana value 7" is correct but nothing asserts it. Either add the assertion to the cast test or drop the sentence; a rules claim with no paired test is decoration.

**5. `rules-engine/.../HyldasCrownOfWinterScenarioTest.kt:83, 92` — the off-turn test discards the `execute` result.** It asserts only "nothing got tapped", which reads as "nothing observable happened" rather than "the activation was illegal". Capture the result and assert `result.error shouldNotBe null` (or match the message) so the test states the intended fact. The author already flagged this in LOOP_UNIT; the assertion does discriminate today, so this is about legibility, not coverage.

**6. `rules-engine/.../FaerieSlumberPartyScenarioTest.kt` — no multi-opponent case.** `Multiply(CountPlayersWith(...), 2)` is only ever exercised at 0 and 1 opponents, so nothing distinguishes `CountPlayersWith` from "1 if any opponent has a creature". A three-player scenario (`withPlayers("P1","P2","P3")`, a creature for each opponent → 4 tokens) would pin it. The PR body already lists this as unverified.

**7. `mtg-sets/.../woe/cards/HyldasCrownOfWinter.kt:202-203` — the second ability's `description` override may be redundant.** Ability 1's override is load-bearing (the cost-reduction rider isn't auto-described), but the derived description for `DrawCards(AggregateBattlefield(EachOpponent, Creature.tapped()))` may already read correctly. Worth checking; harmless either way, and Starport Security sets one too.

**8. `backlog/sets/bloomburrow-commander/cards.md:6-15` — the per-colour "Done" table is stale** (columns sum to 12 against 113 implemented). The PR correctly bumped only the `Implemented:` line and the drift predates this change. Do **not** fix it here — separate change.

### Informational (out of scope, do not touch)

`WitchstalkerFrenzyScenarioTest.kt:146` seeds `"Centaur Courser"`, and no `card("Centaur Courser")` exists anywhere in `mtg-sets` — confirmed. `ScenarioTestBase.withCardOnBattlefield` tolerates an unresolvable name (`if (cardDef != null)` guard), so that test's "four attackers" case is probably weaker than it reads. Pre-existing, unrelated to this unit, and correctly left alone by the author — worth a separate issue rather than a fix in this PR.

## Recommendation

Merge-ready once #1 and #2 are addressed. Concretely, for the corrector:

1. Fix `FaerieSlumberParty.kt:31-33` — either adopt the `captureControllers` + `DistinctEntitiesInCollections` composition above, or rewrite the comment to state the approximation honestly and retract the "no primitive exists" claim.
2. Add the opponent-attacked case to `RowdyResearchScenarioTest.kt`.
3. Optional, cheap, in the same pass: #3, #4 and #5.

Leave #8 and the Centaur Courser note alone.

wingedsheep:
Corrections pushed as `ca958b4` — accounting against the review:

**Important**

1. **`FaerieSlumberParty.kt` — "no primitive exists" claim.** Fixed by *adopting* the composition rather than rewriting the comment. The card is now `gather(Creature)` → `filterSplit(opponentControls())` → `captureControllers(theirs)` → `move(all, ToZone(HAND))` → `woeFaerieToken(2 × DistinctEntitiesInCollections(controllers))`, with a `descriptionOverride` so the UI keeps the printed text. Confirmed the pieces before switching: `FilterCollectionExecutor` evaluates `MatchesFilter` through `PredicateEvaluator` with `PredicateContext.fromEffectContext`, so `opponentControls()` is relative to the spell's controller and reads projected control; `MoveCollectionExecutor:653` collapses battlefield→hand to `ownerId`; `CardLinter` already registers `CaptureControllers.storeAs` as a `Space.COLLECTION` write, so the `DistinctEntitiesInCollections` read resolves. The false claim is gone from the doc comment and from the PR body (the body now records that it was wrong and why).

   One correction to the finding itself, for the record: the composition carries the *same* approximation as the old spelling, not a stricter one. Controllers must be captured before the move (afterwards `ControllerComponent` is stripped), so a creature that fails to leave the battlefield still contributes its controller either way. Truly pinning "returned this way" would need controllers captured post-move, which is not expressible — so the doc comment names that residual divergence instead of claiming exactness.

2. **`RowdyResearchScenarioTest.kt` — opponent's attackers.** Added *"an opponent's attackers discount it too — the count is unioned over every player"*: player 2 active, attacks player 1 with two creatures, then `calculateEffectiveCost` for player 1's Rowdy Research asserts 4 generic.

**Minor**

3. **Discount never exercised on a real cast.** Added *"a real cast pays the reduced cost — two attackers make five Islands enough"*. Five Islands cannot pay `{6}{U}`, so the cast only succeeds through the enumerator + payment path if the reduction actually applied.

4. **Untested "always mana value 7" claim.** Folded the assertion into that same cast test: after payment, the spell on the stack reads `manaValue == 7`.

5. **Hylda's off-turn test discarded the `execute` result.** Now captures it and asserts `result.error shouldNotBe null` alongside the untapped assertions. Checked `ActivateAbilityHandler:384` first — the affordability gate rejects the composite cost during `validate`, so the assertion is exact rather than optimistic.

6. **Multi-opponent Faerie case — declined, harness limitation.** `ScenarioTestBase.withPlayers` in `rules-engine` is hard-wired to exactly two players (`player1Id`/`player2Id` fields, `turnOrder = listOf(p1, p2)`), and `game-server`'s `ScenarioTestBase` has no `withPlayers` at all. A three-player case would mean extending the scenario builder — an engine-harness change, not a batch-unit card change. Test 1 (one opponent, two creatures → 2 tokens) still discriminates "per opponent" from "per creature", which is what the de-dup buys.

7. **Ability 2's `description` override — declined, it is load-bearing.** `DrawCardsEffect.description` renders a non-`Fixed` count as `"Draw cards equal to ${count.description}"`, and `AggregateBattlefield`'s COUNT description is `"the number of " + pluralize(filter.description)` with no mention of the `player` scope. The derived text would read *"Draw cards equal to the number of tapped creatures"* — dropping "your opponents control". Kept as-is.

8. **BLC backlog table drift** — left alone as instructed. The Centaur Courser note in `WitchstalkerFrenzyScenarioTest` was likewise not touched.

**Gate:** `just build` — passed. First run failed only on the expected `CardDefinitionSnapshotTest` mismatch for the rewritten card; `just rebless-cards` then re-ran clean, and the `WOE.json` diff is confined to lines 4131-4207, entirely inside the `Faerie Slumber Party` entry (between `Faerie Fencing` at 4021 and `Farsight Ritual` at 4220) — nothing pre-existing moved. `just check-card-printing "Faerie Slumber Party"` still exits 0. Scenario tests: 15 green (Rowdy Research 7, Hylda's Crown 5, Faerie Slumber Party 3).

**Still open:** the multi-opponent count (#6) is unverified above one opponent, for the harness reason above.

PR #1536, opened 2026-07-30
Pr ancestral vision

jonubuu:
# Ancestral Vision (Time Spiral) — Summary

Implements Ancestral Vision (no mana cost, Suspend 4—{U}), which required building printed Suspend as a
new special action mechanic (CR 702.62a/116.2f). Previously the engine only supported Suspend via runtime
grants (Taigam), not a card's own printed keyword.

Testing along the way uncovered and fixed three real bugs:

- A latent CR 202.1b/118.6 violation — cards with no mana cost were being offered as normal free casts.
- A missing card-highlight case in the client for SuspendCardFromHand (and a pre-existing PlotCard gap in
  the same code).
- A silent flash-detection bug that would have broken instant-speed Suspend for any future flash/instant
  card (Ancestral Vision itself never exercises that branch, so it was latent).

Also added a Suspended-cards exile pile in the client (matching Plot/Paradigm) with a live time-counter
countdown badge, and a manual scenario for hand-testing.

10 commits, 41 files changed, +1695/-122. Full regression suite green across rules-engine, mtg-sdk, mtg-
sets, game-server, and web-client (~9000 tests), plus 30 new tests specific to this work.


wingedsheep (changes requested):
I added three focused regression tests locally to `PrintedSuspendTimingTest`. The two existing timing tests still pass, but all three regressions fail against this PR:

```text
PrintedSuspendTimingTest > an instant with printed suspend can be suspended with a non-empty stack (instant speed) PASSED
PrintedSuspendTimingTest > a non-instant card with printed flash and suspend can also be suspended with a non-empty stack PASSED
PrintedSuspendTimingTest > a cast prohibition suppresses the suspend legal action FAILED
PrintedSuspendTimingTest > a direct suspend action is rejected while the player cannot cast spells FAILED
PrintedSuspendTimingTest > explicit payment cannot pay a red suspend cost with a Forest FAILED

5 tests completed, 3 failed
```

Command: `just test-class PrintedSuspendTimingTest`

Please add the following coverage (or equivalent) and make it pass:

```kotlin
import com.wingedsheep.engine.core.PaymentStrategy
import com.wingedsheep.engine.state.components.player.CantCastSpellsComponent

test("a cast prohibition suppresses the suspend legal action") {
    val driver = createDriver()
    val me = driver.activePlayer!!
    driver.passPriorityUntil(Step.PRECOMBAT_MAIN)

    val card = driver.putCardInHand(me, "Test Suspend Bolt")
    driver.giveMana(me, Color.RED, 1)
    val player = driver.state.getEntity(me)!!.with(CantCastSpellsComponent())
    driver.replaceState(driver.state.withEntity(me, player))

    driver.legalActions(me).any {
        val action = it.action
        action is SuspendCardFromHand && action.cardId == card
    } shouldBe false
}

test("a direct suspend action is rejected while the player cannot cast spells") {
    val driver = createDriver()
    val me = driver.activePlayer!!
    driver.passPriorityUntil(Step.PRECOMBAT_MAIN)

    val card = driver.putCardInHand(me, "Test Suspend Bolt")
    driver.giveMana(me, Color.RED, 1)
    val player = driver.state.getEntity(me)!!.with(CantCastSpellsComponent())
    driver.replaceState(driver.state.withEntity(me, player))

    driver.submit(SuspendCardFromHand(me, card)).isSuccess shouldBe false
    driver.getHand(me).contains(card) shouldBe true
}

test("explicit payment cannot pay a red suspend cost with a Forest") {
    val driver = createDriver()
    val me = driver.activePlayer!!
    driver.passPriorityUntil(Step.PRECOMBAT_MAIN)

    val card = driver.putCardInHand(me, "Test Suspend Bolt")
    val forest = driver.putLandOnBattlefield(me, "Forest")

    driver.submit(
        SuspendCardFromHand(
            playerId = me,
            cardId = card,
            paymentStrategy = PaymentStrategy.Explicit(listOf(forest)),
        )
    ).isSuccess shouldBe false
    driver.getHand(me).contains(card) shouldBe true
}
```

There are two fixes needed:

1. Suspend's “could begin to cast” check must use the engine's shared cast-legality path in both `SuspendEnumerator` and `SuspendCardFromHandHandler.validate`. At present it only checks printed Instant/Flash and sorcery timing, so `CantCastSpellsComponent` and the other shared cast prohibitions are bypassed. Please reuse `CastPermissionUtils`/the existing cast-restriction chokepoint rather than duplicating a partial list. The same path should account for granted Flash.

2. `PaymentStrategy.Explicit` must validate that every selected source is controlled by the payer, is a legal available mana source, and collectively produces mana that satisfies the suspend cost. The current handler only checks that each entity exists and is untapped, then taps it, which lets a Forest pay a red suspend cost (and permits arbitrary/opponent entity IDs). Please route this through the established explicit mana-payment machinery used by normal casting.

Once those tests pass, please also add a granted-Flash Suspend case so the permission side of “could begin to cast” is covered, not only the prohibition side.

jonubuu:
added fixes, then added some more tests to check, everything seems to be passing fine

PR #1600, opened 2026-08-01
Send an Aura to the graveyard when its enchant restriction stops holding

wingedsheep:
## The bug

CR 303.4c makes an Aura's "Enchant …" ability a **continuous** restriction, not just a targeting requirement for the Aura spell:

> 303.4c If an Aura is enchanting an illegal object or player as defined by its enchant ability and other applicable effects, … the Aura is put into its owner's graveyard. (This is a state-based action.)

`UnattachedAurasCheck` re-checked three things about a live attachment — the host left the battlefield, the host is gone, and protection gained after the fact — plus the Equipment-specific legality rules of 704.5n. The Aura's *own* enchant restriction was validated at cast time by `TargetValidator` and never again, so an Aura stayed attached — and kept applying its continuous effects — to a host that had stopped qualifying.

Two reproductions, both in the new scenario test:

| Aura | What breaks the restriction | Before |
|---|---|---|
| Cradle of Safety — *Enchant creature you control* | opponent's Act of Treason steals the host | Aura stays, still granting +1/+1 to a creature its controller no longer controls |
| Holy Strength — *Enchant creature* | Imprisoned in the Moon turns the host into a land | Aura stays attached to a land |

## The fix

Re-read the restriction from the card's `auraTarget` and evaluate its filter against the **projected** host, so layer-4 type and control changes are seen. If it fails, 704.5m puts the Aura into its owner's graveyard.

Deliberately narrow in two ways:

- **Filter only, not full targeting legality.** An attached Aura isn't re-targeted, so hexproof/shroud/"can't be the target of" gained after the attachment landed don't dislodge it (CR 702.11b). Protection is the one quality that does, and the existing `hostProtectedFromAttachmentColor` branch right below already handles it.
- **Fails open.** An Aura the check can't judge — printing missing from the registry, an "enchant player" requirement, a filter scoped to a zone other than the battlefield — is left attached rather than destroyed. A wrong verdict here silently removes a card from the game.

The negative half of each test is the real guard against an over-eager implementation: Holy Strength (plain *Enchant creature*) survives the theft, and Imprisoned in the Moon (*Enchant creature, land, or planeswalker*) survives turning its own host into a land.

## A note for card authors

The change makes `auraTarget` load-bearing after resolution, so an Aura that type-changes its own host must list the resulting type in its filter — exactly as the printed cards do (Imprisoned in the Moon enchants "creature, land, or planeswalker"; Sugar Coat enchants "creature or Food"). Documented in `docs/card-sdk-language-reference.md`.

I audited every registered Aura that sets card types or subtypes on its host: all nine either keep the host a creature or already list the resulting type. None self-destruct.

## Verification

- `just test-class AuraEnchantRestrictionStateBasedActionTest` — 2 new tests; both fail without the engine change (confirmed) and pass with it.
- `just test-rules` — 9005 tests, 0 failures.
- `just test` — full suite green. No card-snapshot movement (no SDK behavior change).

Rule numbers checked against the current Comprehensive Rules text (2026-06-19): 303.4c, 704.5m, 704.5n, 702.11b.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -93,6 +93,12 @@ section; do not let SDK additions land without a corresponding doc update.
   `AttachedToComponent` (players are entities too), survives state-based actions while that player is in
   the game, and exposes the player through `Player.EnchantedPlayer` / `EventPattern.LifeGainEvent(EnchantedPlayer)`
   and a `takesDamage(binding = ATTACHED)` "whenever enchanted player is dealt damage" trigger. (Grievous Wound.)
+  The filter is **not** just a cast-time target check: per CR 303.4c it restricts what the Aura may stay
+  attached to, so the engine re-evaluates it against the projected host after every state change and sends
+  the Aura to its owner's graveyard once the host stops matching (CR 704.5m). Write the filter to cover
+  every type the Aura's *own* effects can turn its host into, exactly as the printed card does — Imprisoned
+  in the Moon makes its host a land and so enchants "creature, land, or planeswalker"; an Aura that turns a
+  creature into a land while enchanting only `Targets.Creature` would destroy itself on resolution.
 - `morph: String?` — morph mana cost (cast face-down).
 - `morphCost: PayCost?` — non-mana morph cost.
 - `morphFaceUpEffect: Effect?` — effect that fires when this morph turns face up.
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/mechanics/sba/permanent/UnattachedAurasCheck.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/mechanics/sba/permanent/UnattachedAurasCheck.kt
@@ -1,6 +1,8 @@
 package com.wingedsheep.engine.mechanics.sba.permanent
 
 import com.wingedsheep.engine.core.ExecutionResult
+import com.wingedsheep.engine.handlers.PredicateContext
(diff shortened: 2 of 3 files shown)


wingedsheep:
Validated the "Enchant creature you control" claim against rulings rather than my own reading of 303.4c. It holds, and there's an official ruling that says it almost verbatim — **Cartouche of Solidarity** (2017-04-18):

> Each Cartouche is an Aura with "Enchant creature you control." If another player gains control of either the Cartouche or the enchanted creature (but not both), then the Cartouche will be enchanting an illegal permanent and be put into its owner's graveyard as a state-based action.

Two details in there were worth pinning down, so I've pushed a comments-and-tests-only follow-up (2c7280892c):

**It's symmetric.** Stealing the *Aura* breaks the restriction exactly as stealing the host does — CR 303.4e keeps the two control changes independent ("changing control of the object doesn't change control of the Aura, and vice versa"). Added that direction as a third test, via Blatant Thievery on the Aura. It also asserts the destination, which the first two tests couldn't distinguish: 704.5m says the **owner's** graveyard, and for a stolen Aura that isn't its controller's.

**"But not both."** If one player ends up controlling both the Aura and the host, the attachment is legal again. This is why the check compares the Aura's controller against the host's rather than watching for a control *change* — the naive reading gets it wrong. Noted at the comparison site. I did not add a test for it: with a single control-change effect the state isn't reachable, because the state-based action fires in between the two steals.

Re-ran `just test-rules`: 9006 tests, 0 failures.

PR #1638, opened 2026-08-03
Add Gonti, Night Minister and Cursecloth Wrappings to Aetherdrift

wingedsheep:
The last two Aetherdrift cards that a card definition alone couldn't express. **DFT goes 272/276 → 274/276** (only *Mimeoplasm, Revered One* and *The Aetherspark* remain, both still engine-blocked).

## Gonti, Night Minister — `{2}{B}{B}` Legendary Creature — Aetherborn Rogue

> Whenever a player casts a spell they don't own, that player creates a Treasure token.
> Whenever a creature deals combat damage to one of your opponents, its controller looks at the top card of that opponent's library and exiles it face down. They may play that card for as long as it remains exiled. Mana of any type can be spent to cast a spell this way.

Both halves act on a player who isn't Gonti's controller, which needed two generalizations:

- **`GrantMayPlayFromExile.recipient: EffectTarget`** (default `Controller`) names who gets the play permission. Gonti passes `ControllerOfTriggeringEntity` — the damaging creature's controller, a player the ability otherwise has no handle on. Turn-keyed expiry still follows the *activating* player (the grant's duration is a property of the effect, not of who may use it); per-card `ownerControls` still wins. `TargetResolutionUtils.resolvePlayerTarget` learns to resolve `ControllerOfTriggeringEntity` as a player reference.
- **`SpellCastPredicate.NotOwnedByController`** now compares the card's owner against the **caster** rather than the trigger's controller. The two coincide under "whenever *you* cast a spell you don't own" (Nita, Vaan), where the `Player.You` gate has already pinned caster == controller — so it's behaviour-preserving there, and it's what makes "whenever *a player* casts a spell they don't own" read correctly for every seat. Ownership also falls back to `CardComponent.ownerId`; without that a spell cast out of an opponent's exile read as owner-less and never triggered.

Everything else composes: the theft is the same gather → face-down exile → grant pipeline Black Cat, Cunning Thief and Laughing Jasper Flint use, and the grantee sees the face-down card through the existing permission-visibility path.

## Cursecloth Wrappings — `{2}{B}{B}` Artifact

> Zombies you control get +1/+1.
> `{T}`: Target creature card in your graveyard gains embalm until end of turn. The embalm cost is equal to its mana cost.

Introduces **Embalm (CR 702.128)** in both forms:

- **Printed** — `embalm(cost)` `CardBuilder` helper + `Keyword.EMBALM`, composed like `renew`: `Mana + ExileSelf`, `activateFromZone = GRAVEYARD`, sorcery speed, over a token copy of itself carrying the three printed exceptions (white *instead of* its other colors, Zombie *in addition to* its other types, no mana cost). `CreateTokenCopyOfTarget` gains `noManaCost` for the third.
- **Granted** — `Effects.GrantEmbalm` hands that same ability to a graveyard card at runtime, mirroring `GrantHarmonize` / `GrantFlashback`. The difference in kind: embalm is an ordinary *activated* ability, not an alternative way to cast, so it rides the plain `GrantedActivatedAbility` channel instead of `GrantedKeywordAbility` — which surfaced a real gap, since `ZoneActivatedAbilityEnumerator` only ever read printed card definitions and so could never offer a granted graveyard ability.

## Tests

`GontiNightMinisterScenarioTest` (3) and `CurseclothWrappingsScenarioTest` (3). Gonti's use a **three-seat pod on purpose**: with only two players, any creature dealing combat damage to an opponent of Gonti's controller is itself controlled by Gonti's controller, so a payoff mis-wired to "you" would look correct.

## Verification

`just build` and `just test` both green. `just rebless-cards` moved only the two new cards in `DFT.json`. `just check-card-printing` clean for both.


wingedsheep:
@copilot resolve the merge conflicts in this pull request

PR #1653, opened 2026-08-04
Fix Harmonious Grovestrider having no ward, and gate the attach-scope silent no-op

bolav:
## Summary

Harmonious Grovestrider's printed "Ward {2}" was authored as a `GrantWard` static ability. `GrantWard.filter` defaults to `GroupFilter.attachedCreature()` (`Scope.AttachedTo`) — the Aura/Equipment shape — and the engine reaches attach-scoped grants only by walking the permanents attached to a host. Nothing is ever attached to a Beast, so no ward trigger was generated and no `WARD` keyword was projected: **the card played with no ward whatsoever.**

The fix is the printed-keyword shape every other intrinsic-ward creature already uses, `keywordAbility(KeywordAbility.ward("{2}"))`. `CardBuilder.build()` derives `Keyword.WARD` from the keyword ability, so the display keyword comes back too — the re-blessed EOE snapshot shows both.

Then, rather than fix one card and leave the trap armed, this adds a `CardLinter` rule for the shape — which immediately found a second live instance of the same bug.

## What's in here

**1. Harmonious Grovestrider (EOE) — ward restored.** Plus its scenario test: ward counters an unpayable Bolt (the creature is set to exactly 3/3, so surviving Lightning Bolt can only mean the spell never resolved, and a `SpellCounteredEvent` assertion pins it independently of the graveyard, which can't tell the two outcomes apart), and P/T tracks only your own lands.

**2. New `CardLinter` rule: `AttachedScopeGrantOnNonAttachment`.** A **printed** static ability (`script.staticAbilities` or a `classLevels` entry, any face) whose `GroupFilter` carries `Scope.AttachedTo`, on a card that is not an Aura/Equipment/Fortification and has no `auraTarget`/`equipCost`, is now a build error via the existing `CardLintTest` corpus gate.

This is squarely `CardLinter`'s stated remit — the "silent no-op" class, where something compiles and serializes cleanly while doing nothing. What makes this shape especially invisible is that attach scope is the *default* filter on the `Grant*` static abilities, so it appears in neither the card source nor the snapshot; the check therefore re-encodes each ability with defaults materialized.

Deliberately **not** flagged: attach scope nested inside an *effect*. An effect can change what the card is before the filter is read — The Irencrag is a Legendary Artifact whose trigger turns it into an Equipment and then grants "equipped creature gets +3/+3" — or hand abilities to another object entirely, like an Aura token. Attachability is likewise judged across the whole physical card, so a creature that transforms into an Aura isn't flagged for its other face. Both shapes have unit tests.

**3. Myojin of Night's Reach (CHK) — indestructible restored.** Found by the new rule. "Has indestructible as long as it has a divinity counter on it" was a bare `GrantKeyword(Keyword.INDESTRUCTIBLE)` on the default attach scope, so the Spirit **was never actually indestructible**. Now `Filters.Self`. Its existing scenario test only covered the counter and the discard ability, which is how this survived; added a test asserting the projected `INDESTRUCTIBLE` keyword with the counter present and its absence once the counter is spent.

**4. Docs.** `GrantWard` had no entry in the SDK language reference at all — which is how the wrong shape slipped in. It now sits in the *Composite static abilities* catalog beside its `attachedCreature()`-defaulting siblings, with the attach-scope default and the "use the keyword ability for a card's own ward" warning called out; the `KeywordAbility.Ward` bullet cross-references it. §21 documents the new lint check.

## Verification

| Suite | Result |
|---|---|
| `:mtg-sdk:test` (full module) | 363 tests, 0 failures |
| `:mtg-sets:test` (full module) | 336 tests, 0 failures |
| `CardLintTest` corpus gate | clean across 8,201 cards with the new rule on |
| `CardLinterTest` | 23 tests, incl. 6 new attach-scope cases |
| `HarmoniousGrovestriderScenarioTest` | 2/2 |
| `MyojinOfNightsReachScenarioTest` | 2/2 |
| `WardCounterTest` / `WardLifeCounterTest` | 9/9 — ward mechanic unregressed |

The new Myojin test was falsified against the unfixed card: with the bug reintroduced it fails, and the lint gate fails too, while the *pre-existing* Myojin test still passes — confirming both that the new test is a real guard and why the bug survived this long.

Two blast-radius checks. The new error gates only `CardLintTest`: `CardLoader` hard-fails solely on `MissingCreatureStats`, so card loading is unaffected. And the mtgish emitter can't generate the forbidden shape — `staticHostBlock` is gated on `HostPermanent`, and intrinsic ward already renders as `KeywordAbility.Ward` — so no emitter change was needed.

CR 702.21 (Ward) verified against the Comprehensive Rules; oracle text, mana cost, type line, rarity, and collector number verified against Scryfall; `just check-card-printing "Harmonious Grovestrider"` passes.


bolav:
## Code review — `/review-changes`

Review of the original single-commit fix (`Fix Harmonious Grovestrider having no ward at all`). Everything flagged below was subsequently addressed on this branch; recorded here for the audit trail. Where a finding produced a change, the follow-up is noted inline.

### Verdict

The fix is correct, minimal, and idiomatic. The root-cause diagnosis holds up under inspection, the replacement is the shape every other intrinsic-ward creature uses, and no SDK surface grows. The issues were in the test comment and the docs, not the behavior.

### What's good

- **The diagnosis is right, and the mechanism checks out.** `GrantWard.filter` defaults to `GroupFilter.attachedCreature()` (`mtg-sdk/.../KeywordStaticAbilities.kt:80`), and `TriggerAbilityResolver.kt:560-573` reaches attach-scoped grants only by walking `statics.attachmentsOn(entityId)`. An unattached creature is in nobody's attachment list, so the grant produced no ward trigger — and, with no matching permanent, the layer path projected no `WARD` keyword either. The card genuinely had nothing.
- **The replacement matches corpus convention.** `keywordAbility(KeywordAbility.ward("{2}"))` appears 27 times across `mtg-sets`, including two other EOE cards (`FamishedWorldsire`, `MouthOfTheStorm`); all three cards named in the commit message (Skyward Spider, Guardian Sunmare, Horned Loch-Whale) use exactly this line.
- **The commit's snapshot claim checks out.** `CardBuilder.build():883` derives `Keyword.WARD` from `KeywordAbility.Ward`, so the added `"keywords": ["WARD"]` is exactly what the DSL emits — the display keyword is a genuine second half of the fix, not snapshot noise.
- **No sibling bugs left behind at the time.** Every other default-filter `GrantWard` in the corpus — `HardlightContainment`, `ShelteredByGhosts`, `LavaspurBoots`, `CombatResearch`, `SuperStrength`, and the `RoyalRole` token — is an Aura, Equipment, or Aura token, where the attach-scope default is correct.
- **The emitter isn't the source.** `CardStructure.kt:3390-3395` renders intrinsic ward as `KeywordAbility.Ward`; `StaticAbilities.kt:467` reaches for `GrantWard` only on the enchanted/equipped wording. This was hand-authored.
- **Externals verified.** Scryfall oracle text, mana cost, type line, rarity, and collector number match exactly; `just check-card-printing` exits 0; CR 702.21 is Ward per the local Comprehensive Rules.
- **The test is right-sized.** The full ward payment matrix (can't pay / prompts / declines / pays) is already covered engine-side by `WardCounterTest`, so a card-level test proving "this card has ward at all" is the correct scope.

### Issues

#### Important

**The key comment in the test said the opposite of what makes the test work** — `HarmoniousGrovestriderScenarioTest.kt:45`. It read "Three lands keep the Grovestrider a 3/3 — big enough to survive the Bolt on its own merits." A 3/3 taking Lightning Bolt's 3 damage has lethal damage and dies. The setup was correct — 3/3 vs 3 damage is exactly what makes the survival assertion discriminating — but the comment justified it backwards, and if it were true the assertion would prove nothing. Someone later "fixing" the land count upward on the strength of that comment would have silently gutted the test. *Fixed: the comment now states the 3/3 is precisely lethal, and says not to raise the land count.*

#### Minor

**The graveyard assertion proved nothing** — same file, line 66. `getGraveyardCardNames(caster).contains("Lightning Bolt")` is true whether the Bolt was countered *or* resolved; an instant reaches the graveyard either way. *Fixed: added a `SpellCounteredEvent` assertion that actually separates the two outcomes; the graveyard line stays, with a comment saying why it can't.*

**A lint rule would gate this class where a doc paragraph only warns about it.** This bug is precisely `CardLinter`'s stated bug class — "a name-based reference that doesn't resolve to anything," a silent no-op found at best in playtesting — and `CardLintTest` already turns those into build failures naming the card. *Fixed: added `AttachedScopeGrantOnNonAttachment`. It immediately found a second live instance of the same bug in Myojin of Night's Reach (CHK), which was never actually indestructible, and one instructive false positive (The Irencrag) that narrowed the rule to printed static abilities. Both are covered in this PR.*

**`GrantWard` wasn't in the static-ability catalog** — `docs/card-sdk-language-reference.md`. The commit message said the change documents `GrantWard`, "absent there entirely," but it was described only as an aside inside the `KeywordAbility.Ward` bullet. Its siblings live in the *Composite static abilities* list with their defaults spelled out — `GrantLandwalkOfChosenType(filter = attachedCreature())`, `GrantProtectionFromControlledColors(filter = attachedCreature())`. Someone browsing that list for "grant ward to the enchanted creature" still wouldn't have found it. *Fixed: `GrantWard` now has a catalog entry there; the Ward bullet is a one-line cross-reference.*

**`driver.registerCard(HarmoniousGrovestrider)`** was redundant — `TestCards.all` is `MtgSetCatalog.all.flatMap { it.cards + it.basicLands } + testOnlyCards`, which already includes every EOE card. *Fixed: dropped, with a comment recording why.*

### Recommendation

The behavioral fix was correct as-is; the only pre-merge ask was the backwards comment. The lint rule was filed as the durable fix for the class and turned out to earn its keep immediately.

PR #1654, opened 2026-08-04
Show granted supertypes in the client type line

bolav:
image

# Show granted supertypes in the client type line

A supertype granted by a continuous effect never reached the client. `ClientStateTransformer` built the rendered type line from the **base** `CardComponent`'s supertypes while taking card types and subtypes from the **layer projection** — and granted supertypes live in the projected `types` set alongside both (`StateProjector.extractTypes`). Since `"LEGENDARY"` isn't a `CardType`, the projected-card-types filter dropped it too, so it vanished from the type line entirely.

The engine had it right throughout: `isLegendary()`, `LegendRuleCheck`, and `BlockPhaseManager` all read the projection. Only the view was stale.

Visible as:

- **Origin of Spider-Man** chapter II — "It becomes a legendary Spider Hero in addition to its other types" rendered as `Creature — Cat`.
- **The Ring emblem** — "Your Ring-bearer is legendary" (CR 701.54c) rendered as plain `Creature`, so a player couldn't see that the legend rule had started applying to their bearer.

## What changed

**Engine / view layer**

- `ClientStateTransformer` reads supertypes from the projection for battlefield permanents, falling back to the base type line for the other zones (which have no projection entry).
- Three readers already knew that supertypes are stored by name inside the projected type set, one via a literal `setOf("BASIC", "LEGENDARY", "SNOW", "WORLD")` that a fifth `Supertype` would have silently outgrown. That isolation is now derived once from the enum — `Supertype.fromProjectedTypes` — and used from both `ProjectedState.getSupertypes` and the transformer. Declaration order is printed order, so `Dark Depths` renders "Legendary Snow Land" and `Snow-Covered Plains` renders "Basic Snow Land".
- New `ClientCard.legendaryByEffect`, the mirror of the existing `nonLegendaryCopy`: the printed frame is non-legendary but the permanent is legendary now, so the UI needs an at-a-glance cue that the legend rule applies.
- `nonLegendaryCopy` additionally requires that the permanent isn't *currently* legendary. Both flags previously fired on a de-legendarized copy that an effect made legendary again (an Impostor Syndrome token designated Ring-bearer), which would have badged one card "Legendary" and "Not Legendary" at the same coordinates.

**Client**

- Gold "Legendary" chip mirroring the existing grey "Not Legendary" one, plus a `cardGrouping` key so a granted-legendary permanent doesn't merge into a stack with an identical twin that isn't.
- Both chips now come from one `LegendChip` component. They are meant to read as mirrors of each other, which two copies of ~50 lines of inline style would not have stayed.

**Side effect, deliberate:** this also changes what a player sees looking at *their own* face-down permanent. Opponents already got a hardcoded masked card, but the controller is shown the real name and so falls through the normal transform path — where the base read made their face-down legendary creature render "Legendary Creature" even though its card types and subtypes had already been masked down to "Creature". A face-down permanent has no supertypes (CR 708.2a), and the type line now says so. Covered by a test.

## Tests

`GrantedSupertypeVisibilityTest` (new, `rules-engine`) — a granted `LEGENDARY` renders in the type line; it's visible to the opponent too; a printed legendary still renders; a face-down printed-legendary creature is not legendary even to its controller; granting Legendary back to a non-legendary copy clears `nonLegendaryCopy`. The test cards mirror the real shapes — `AddCardTypeEffect("LEGENDARY", …)` as in Origin of Spider-Man, `CreateTokenCopyOfTarget(removedSupertypes = setOf(LEGENDARY))` as in Impostor Syndrome — rather than pulling SPM into set scope.

`TheRingScenarioTest` gains type-line and flag assertions on the bearer and a non-bearer.

`groupCards.test.ts` gains both legend flags as divergence axes (`nonLegendaryCopy` had none either).

## Verification

- `just test-rules` — full `:rules-engine` suite green.
- `just test-class GrantedSupertypeVisibilityTest` — 5/5.
- `just test-class TheRingScenarioTest` — 9/9.
- `npm run typecheck` — no new errors. Five pre-existing `TS2307` module-not-found errors in `admin/` and `profile/` (`recharts`, `d3-geo`, `topojson-client`, `world-countries`) are unrelated to this change and unfixable in this sandbox, whose network is firewalled.
- `npx vitest run src/store/groupCards.test.ts` could **not** be run here: the `rolldown` native binding is missing from `node_modules` in this sandbox (`MODULE_NOT_FOUND`). The two added `it.each` rows are unexercised on my end and need a CI run.

No screenshot: the app stack couldn't be brought up in this firewalled sandbox. Happy to add one before merge if you'd like it.

## Rules citations

Verified against `MagicCompRules_20260619.txt`:

- **CR 701.54c** — The Ring has "Your Ring-bearer is legendary and can't be blocked by creatures with greater power."
- **CR 708.2a** — a face-down permanent is a 2/2 creature with no text, no name, no subtypes, and no mana cost.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1072,7 +1072,7 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 
 ### Color & type
 
-- `AddCardType(type, target, duration)` — add a card type (e.g. become an artifact).
+- `AddCardType(type, target, duration)` — add a card type (e.g. become an artifact). Also takes a **supertype** — `AddCardType("LEGENDARY", …)` for "it becomes a legendary Spider Hero" (Origin of Spider-Man) — because the layer projection stores supertypes in the same type set; there is no separate `AddSupertype`.
 - `AddSubtype(subtype, target, duration)` — add a subtype temporarily.
 - `SetLandType(landType, target, duration, fromChosenValueKey)` — target land *becomes* the basic land type, **replacing** its existing land subtypes (Rule 305.7); pass `fromChosenValueKey` to read the type from a preceding `ChooseOption(OptionType.BASIC_LAND_TYPE)`. One-shot counterpart to the `SetEnchantedLandType` aura static ability. (Dream Thrush)
 - `ChooseColorForTarget(target)` — target picks a color; stored in context.
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/core/Supertype.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/core/Supertype.kt
@@ -12,5 +12,17 @@ enum class Supertype(val displayName: String) {
     companion object {
         fun fromString(value: String): Supertype? =
             entries.find { it.displayName.equals(value, ignoreCase = true) }
+
+        /**
+         * The supertypes present in a *projected* type set — the engine's layer projection stores
+         * supertypes, card types and subtypes together in one `Set<String>` of uppercase enum names,
+         * so every reader has to isolate them by name. Deriving that from [entries] (rather than
+         * repeating a literal `setOf("BASIC", "LEGENDARY", …)`) keeps the readers correct when a
+         * supertype is added here.
+         *
+         * Returned in declaration order, which is printed order: "Basic Snow Land", "Legendary Snow Land".
+         */
+        fun fromProjectedTypes(types: Set<String>): List<Supertype> =
+            entries.filter { it.name in types }
     }
 }
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/effects/TypeAndColorEffects.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/effects/TypeAndColorEffects.kt
@@ -221,7 +221,10 @@ data class SetLandTypeEffect(
  * Creates a floating effect on Layer.TYPE with AddType modification.
  * Unlike AnimateLandEffect, this does NOT set P/T — it only adds the type.
  *
- * @property cardType The card type to add (e.g., "ARTIFACT", "CREATURE")
+ * @property cardType The card type to add (e.g., "ARTIFACT", "CREATURE"). Also accepts a
+ *   *supertype* — "LEGENDARY" for Origin of Spider-Man's "it becomes a legendary Spider Hero" —
+ *   since the engine projects supertypes into the same type set; there is no separate
+ *   AddSupertypeEffect and none is needed.
  * @property target The permanent to modify
  * @property duration How long the type change lasts (default: Permanent)
  */
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/mechanics/layers/ProjectedState.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/mechanics/layers/ProjectedState.kt
@@ -2,6 +2,7 @@ package com.wingedsheep.engine.mechanics.layers
 
 import com.wingedsheep.sdk.core.Color
 import com.wingedsheep.sdk.core.Keyword
+import com.wingedsheep.sdk.core.Supertype
 import com.wingedsheep.sdk.model.EntityId
 import com.wingedsheep.engine.state.GameState
 import com.wingedsheep.engine.state.components.identity.CardComponent
@@ -171,7 +172,7 @@ class ProjectedState(
      * projected [types] set as card types, so they are isolated by name here for protection checks.
      */
     fun getSupertypes(entityId: EntityId): Set<String> =
-        getTypes(entityId).intersect(setOf("BASIC", "LEGENDARY", "SNOW", "WORLD"))
+        Supertype.fromProjectedTypes(getTypes(entityId)).mapTo(mutableSetOf()) { it.name }
 
     fun hasSubtype(entityId: EntityId, subtype: String): Boolean =
         getSubtypes(entityId).any { it.equals(subtype, ignoreCase = true) }
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/view/ClientDTO.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/view/ClientDTO.kt
@@ -459,12 +459,27 @@ data class ClientCard(
     val copyOf: String? = null,
 
     /**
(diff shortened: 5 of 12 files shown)


bolav:
# Review — granted supertypes in the client type line

Reviewed on the branch with `origin/main` (`0d0e094147`) already merged in — it was an ancestor of HEAD, so no merge commit was needed and no worktree was created.

**Every issue below has been fixed in the branch as pushed.** The findings are recorded here so the reasoning is reviewable rather than buried in a diff.

## Verdict

The behavior is right and the shape is right: a pure anti-corruption-layer fix. The projection was already authoritative — `isLegendary()`, `LegendRuleCheck`, and `BlockPhaseManager` all read it — and only `ClientStateTransformer` was reading base state. No new SDK type was introduced, and none was needed.

## What's good

- **The fix is in exactly the right layer.** The new `displaySupertypes` block mirrors the `displayCardTypes`/`displaySubtypes` pattern immediately above it, with the same battlefield-only/base-fallback shape. Nothing leaked into the SDK, no `Effect`/`Modification`/`StaticAbility` was added, and no card-specific code appeared anywhere.
- **Deriving the supertype list from `Supertype.entries` is the right call** — enum declaration order is printed order, so `Dark Depths` renders "Legendary Snow Land" and `Snow-Covered Plains` renders "Basic Snow Land". Strictly better than the hardcoded literal that already existed in `ProjectedState.getSupertypes`.
- **No projection regression.** Every `values.types` mutation in `EffectApplicator` was checked: `SetCardTypes` removes only card-type strings, `SetAllSubtypes` / `SetBasicLandTypes*` only subtype strings, and `applyTextReplacements` only subtypes. Nothing wipes supertypes wholesale, so switching to the projection cannot drop a printed `Legendary` — Song-of-the-Dryads-style "loses all other card types" effects keep it, as the rules require.
- **Rule citations verified** against `MagicCompRules_20260619.txt`: CR 701.54c and CR 708.2a both check out.
- **Test placement and scope are right.** Engine-level view concern → `rules-engine`; a local `{0}` test card mirroring the real `AddCardTypeEffect("LEGENDARY", …)` shape rather than dragging SPM into set scope.

## Issues found (all now fixed)

### Important

**1. The two legendary chips were not mutually exclusive, despite a code comment asserting they were.** Both flags reduced to `LEGENDARY !in cardComponent.typeLine.supertypes` and then diverged on *different* sources: `nonLegendaryCopy` compares the printed `CardDefinition`, `legendaryByEffect` reads the projection. A non-legendary copy (Impostor Syndrome) later designated Ring-bearer, or targeted by Origin of Spider-Man chapter II, set **both** — and `cardGrouping.ts` pushed both `nonleg` and `grantedleg`, so the data model itself admitted it. The chips render at identical `top`/`left`, so they would have overlapped exactly while the type line read "Legendary Creature — …".

*Fixed* by adding `&& Supertype.LEGENDARY !in displaySupertypes` to `nonLegendaryCopy`. The flag now means "printed legendary, *currently* not legendary" — a permanent an effect has made legendary again simply isn't one, so the grant wins and the comment became true rather than aspirational. New test: `granting Legendary back to a non-legendary copy clears nonLegendaryCopy`, built on a real `CreateTokenCopyOfTarget(removedSupertypes = …)` rather than a hand-poked component.

**2. The diff added a third spelling of "supertypes are stored in the projected `types` set."** `ProjectedState.getSupertypes` already isolated them with a hardcoded `setOf("BASIC", "LEGENDARY", "SNOW", "WORLD")`; `AffectsFilterResolver` and `PredicateEvaluator` each hardcode `"LEGENDARY" in types`; the transformer became a fourth reader. The literal set was a live drift hazard — a fifth supertype added to the enum would silently not appear there.

*Fixed* by putting the derivation on the enum once as `Supertype.fromProjectedTypes(types: Set<String>): List<Supertype>`, now used by both `ProjectedState.getSupertypes` and the transformer. All eight `getSupertypes` call sites consume `Set<String>` of uppercase names and are unaffected.

**3. The face-down claim in the commit message was inaccurate and untested.** It said the change stopped a leak into the "masked type line", but the masked path returns early with a hardcoded `typeLine = "Creature"` for spectators and non-controllers and never leaked. What actually changed is the **controller's own view of their own face-down permanent**, which falls through to the general path: it used to render "Legendary Creature" and now renders "Creature". That is the more correct outcome and consistent with card types and subtypes already coming from the projection — but it was a real user-visible behavior change with no test.

*Fixed*: commit message reworded to describe the controller's own view, and a test added asserting both the controller's and the opponent's view render `"Creature"` for a face-down printed-legendary creature.

### Minor

**4. ~60 lines of near-duplicate chip markup** — the granted chip was a copy of the "Not Legendary" chip with only the gradient, text color, label, icon fill, and strike-through differing, in a file already ~2900 lines. *Fixed*: both now render through one `LegendChip` component, and the shared coronet path is a `CROWN_PATH` constant the commander crown uses too.

**5. `groupCards.test.ts` fixture stopped honoring its own contract** — the `token()` factory documents that it defaults every field `computeCardGroupKey` reads, but `legendaryByEffect` was missing; it only worked because `undefined` is falsy. *Fixed*, plus divergence rows for both legend flags (`nonLegendaryCopy` had none either — pre-existing gap the new axis inherited).

**6. `legendaryByEffect` is a Legendary-specific boolean for a general "supertype granted by effect" fact.** Left as-is — it mirrors `nonLegendaryCopy`, and only the legend rule changes how a permanent must be played around. *Fixed* to the extent that the DTO KDoc now records the narrowing as deliberate, and says to widen to a set rather than add a second boolean if Snow or World ever needs a cue.

**7. Pre-existing, drive-by:** `AddCardTypeEffect` is the canonical way to grant a *supertype* (`AddCardTypeEffect("LEGENDARY", …)`), which its own KDoc contradicted (`@property cardType The card type to add`). *Fixed* in the KDoc and in `docs/card-sdk-language-reference.md`, so the next author doesn't add an `AddSupertypeEffect`.

## Verification

- `just test-rules` — full `:rules-engine` suite green.
- `just test-class GrantedSupertypeVisibilityTest` — 5/5 (one cold-JVM `TimeoutCancellationException` flake on the 120s per-test soft timeout during a run that also recompiled `mtg-sdk`; passes warm).
- `just test-class TheRingScenarioTest` — 9/9.
- `npm run typecheck` — no new errors; five pre-existing `TS2307` module-not-found errors in `admin/` and `profile/` are unrelated and unfixable in this firewalled sandbox.
- `npx vitest run src/store/groupCards.test.ts` — **could not run**: `rolldown`'s native binding is missing from `node_modules` here (`MODULE_NOT_FOUND`). The two added grouping rows are unexercised locally and need a CI run.

PR #1655, opened 2026-08-04
Expire granted abilities whose source-keyed duration has ended

bolav:
image
image

## Problem

Kitesail Larcenist's ETB turns a permanent into a Treasure "for as long as this creature remains on the battlefield". The type change is a floating effect, so `EndedDurationExpiryCheck` reverted it when Kitesail died — but the granted `{T}, Sacrifice this artifact: Add one mana of any color` is stored in `GameState.grantedActivatedAbilities` (that is what lets it survive the Layer 6 "lose all abilities" wipe), and that record carried no source.

`affectedGrantConditionFails` therefore only understood the *affected-object*-keyed gates (`WhileAffectedHasCounter`, `WhileAffectedTapped`) and fell through to `else -> false` for every source-keyed one, while the only other prune sites — in `CleanupPhaseManager` — filter strictly on `EndOfTurn` / `UntilYourNextTurn`. A `WhileSourceOnBattlefield` grant had no expiry path at all: **Bolt the Larcenist and both chosen permanents kept saccing themselves for mana forever.**

## Fix

`GrantedActivatedAbility` and `GrantedStaticAbility` now carry a `sourceId`, mirroring `ActiveFloatingEffect.sourceId` and populated from the same `context.sourceId` by the five executors whose duration comes from the effect. The durable grant and its floating siblings now share one source key, so they expire together by construction rather than by two parallel rules.

`EndedDurationExpiryCheck` grows a `sourceGateHolds` helper — source on the battlefield / tapped / still attached, all decidable without a projection — and both the floating-effect path and grant pruning go through it.

Notes on the shape:

- **Nullable with a default**, so persisted states still deserialize.
- **Fails closed.** `sourceId == null` ⇒ gate closed, so a future executor that forgets to pass it loses the grant loudly instead of leaking it forever.
- **The five untouched construction sites are safe, not just assumed safe.** `CreateTokenExecutor` ×2, `CreateTokenCopyOfTargetExecutor` ×2 and `EachPermanentBecomesCopyOfTargetExecutor` ×1 all pass `Duration.Permanent`, which short-circuits at `else -> true`. That also means no perf regression in the common case: a `Permanent` grant never reaches `getBattlefield()` (an O(n) `List.contains`), exactly as the old `else -> false` early exit did.
- **No SDK change.** `mtg-sdk` is untouched and no card definition changed; this is engine-internal plumbing.
- The three rewritten branches in `activeAffectedEntities` were checked against the deleted code and are behaviourally identical — per-entity filtering degenerates to all-or-nothing for the two gates that ignore the affected object, and the `WhileSourceAttachedToAffected` branch keeps its null/off-battlefield short-circuit.

`WhileSourceTappedAndAffectedPowerAtMostSource` (needs a projection) and the controller-keyed gates (`WhileControlledByController`, `WhileYouControlSource`) stay floating-effect-only: a grant record carries no controller, and no card grants an ability with those durations. That is documented rather than half-implemented.

## Tests

`KitesailLarcenistScenarioTest` covers both clauses of CR 611.2b:

- **"the duration ends later"** — kill Kitesail with a real Lightning Bolt after its ETB has resolved. The pre-existing revert test passed against broken code, because `moveToGraveyard` is a raw zone move that never runs state-based actions and its assertions only read projected types, which re-gate on every projection regardless. The new test asserts on both the grant store and the legal-action enumerator.
- **"the duration ends before the effect would first be applied"** — kill Kitesail while its ETB trigger is still on the stack. The trigger still resolves, so the executor still records the transform and the grant, both keyed to a source that is already gone; nothing may stick.

Both are verified non-decorative: with the source gate stubbed out to reproduce pre-fix behaviour, those two tests fail while the other three pass.

Housekeeping in the same branch: the `pruneEndedGrants` KDoc now names which grant stores are covered and which are not (`grantedTriggeredAbilities`, `grantedReplacementEffects` and `globalGrantedTriggeredAbilities` get only the `EndOfTurn` / `UntilYourNextTurn` filters — fine today, but their executors accept any `Duration`, so the asymmetry is worth a note rather than a silent leak); the class KDoc bullet list is reflowed (the grant paragraph had been inserted mid-list, orphaning two entries) and gains the two it was missing; and the card SDK reference's `BecomeArtifactEffect` entry now states that `grantedAbility` honours source-keyed durations.

Also adds the manual scenario the bug was found with, so the board can be set up in the dev server.

## Verification

| Run | Result |
|---|---|
| `just test-class KitesailLarcenistScenarioTest` | 5/5 PASSED |
| Same, against simulated pre-fix behaviour | 2 FAILED (both grant-expiry tests), 3 passed |
| `just test-rules` (full `rules-engine` suite) | BUILD SUCCESSFUL |

CR 611.2b verified against `MagicCompRules_20260619.txt:2906` — the wording "It doesn't start and immediately stop again, and it doesn't last forever" is what justifies the one-way latch.

https://claude.ai/code/session_0145yakyrz13zRy5cATmB5NT


bolav:
## Review

Reviewed against `main` with `origin/main` already an ancestor of `HEAD` (`be053081da` sits directly on `0d0e094147`), so no merge was needed and no worktree was created.

**Verdict:** the behaviour is right and the shape is right. `WhileSourceOnBattlefield` grants genuinely had no expiry path — the two `CleanupPhaseManager` prune sites filter strictly on `EndOfTurn` / `UntilYourNextTurn`, and `affectedGrantConditionFails` fell through to `else -> false` for every source-keyed duration. The fix closes that at the correct layer with no new SDK type. **No blocking issues.** Four smaller items were raised and all four have been fixed in `ca21c5d275`.

### What's good

- **Root-caused at the right layer.** `sourceId` on the grant records mirrors `ActiveFloatingEffect.sourceId`, and both are populated from the same `context.sourceId` (`FloatingEffectFactory.kt:70` vs. `BecomeArtifactExecutor.kt:127`). The durable grant and its floating siblings now share one source key, so they expire together by construction rather than by two parallel rules. `mtg-sdk` is untouched and no card changed — the SDK-elegance bar is met without argument.
- **`sourceGateHolds` is a real de-duplication, not a shuffle.** All three rewritten branches in `activeAffectedEntities` were checked against the deleted code: per-entity filtering degenerates to all-or-nothing for the two gates that ignore the affected object, and the `WhileSourceAttachedToAffected` branch preserves the null/off-battlefield short-circuit. Behaviourally identical on the floating path.
- **Fails closed.** `sourceId == null` ⇒ gate closed. A future executor that forgets to pass it drops the grant loudly instead of leaking it forever — the right direction for a silent-failure mode.
- **The five untouched construction sites are actually safe**, not just assumed safe: `CreateTokenExecutor` ×2, `CreateTokenCopyOfTargetExecutor` ×2, `EachPermanentBecomesCopyOfTargetExecutor` ×1 all pass `Duration.Permanent`, which short-circuits at `else -> true`. That also means no perf regression in the common case — a `Permanent` grant never touches `getBattlefield()` (an O(n) `List.contains`), where the old code's early `else -> false` had the same property.
- **The test comment at `KitesailLarcenistScenarioTest.kt:142-145` earns its place.** `moveToGraveyard` never runs SBAs and the older assertions only read projected types, which re-gate every frame — so the pre-existing revert test passed against broken code. Documenting that is worth more than the assertion itself.
- **CR 611.2b verified** against `MagicCompRules_20260619.txt:2906`. Correct number, and the quoted "It doesn't start and immediately stop again, and it doesn't last forever" is exactly what justifies the one-way latch.

### Issues raised — all fixed in `ca21c5d275`

**Important · missing test for the second half of the cited rule** — CR 611.2b has two clauses and the change relies on both. The tested one was "the duration ends later". The other — *"if that duration ends before the moment the effect would first be applied … the effect does nothing"* — was cited but untested: kill Kitesail **in response to its own ETB trigger** and the trigger still resolves, so `BecomeArtifactExecutor` still records the transform and the grant, both keyed to a source that is already gone. Before this branch that grant persisted forever, leaving the chosen Courser an ordinary Centaur that could nonetheless sacrifice itself for mana.
*Fixed:* new test `"killing Kitesail in response to its own ETB grants nothing at all"`. Verified non-decorative — with the source gate stubbed out to reproduce pre-fix behaviour, this test and the existing `"goes away when Kitesail dies"` test both FAIL while the other three pass.

**Minor · three of the five grant stores are still outside the prune** — `pruneEndedGrants` covers `grantedActivatedAbilities` and `grantedStaticAbilities`. `grantedTriggeredAbilities`, `grantedReplacementEffects` and `globalGrantedTriggeredAbilities` get only the `EndOfTurn` / `UntilYourNextTurn` filters in `CleanupPhaseManager` — not even the affected-keyed gates (pre-existing for those lists, not introduced here). Nothing leaks today, but `GrantTriggeredAbilityExecutor` accepts any `Duration`, so with two of five stores expiring correctly and three not, the asymmetry is a trap.
*Fixed:* the `pruneEndedGrants` KDoc now names what is covered, what is not, and what the next author has to do.

**Minor · KDoc bullet list broken mid-way** — the new grant paragraph was inserted *between* bullets, orphaning the `WhileControlledByController` and `WhileYouControlSource` entries. `WhileAffectedHasCounter` and `WhileSourceAttachedToAffected` were missing from the list entirely (pre-existing).
*Fixed:* paragraph moved below the list, both missing bullets added.

**Minor · `docs/card-sdk-language-reference.md` understated `BecomeArtifactEffect`** — the `grantedAbility` sentence said only that it is "recorded durably in `grantedActivatedAbilities`" and nothing about expiry, while the neighbouring `grantedStaticAbilities` sentence told authors to "pair it with `Duration.Permanent`". A card author had no signal that `grantedAbility` supports source-keyed durations — precisely the capability this branch adds.
*Fixed:* one clause added naming the behaviour and Kitesail as its first user.

### Verification

| Run | Result |
|---|---|
| `just test-class KitesailLarcenistScenarioTest` | 5/5 PASSED |
| Same, against simulated pre-fix behaviour | 2 FAILED (both grant-expiry tests), 3 passed |
| `just test-rules` (full `rules-engine` suite) | BUILD SUCCESSFUL |

https://claude.ai/code/session_0145yakyrz13zRy5cATmB5NT

PR #1656, opened 2026-08-04
Apply a planeswalker's starting loyalty on every battlefield entry

bolav:
## Summary

A planeswalker only got its loyalty counters when a planeswalker *spell resolved*. CR 306.5b makes
"this permanent enters with a number of loyalty counters on it equal to its printed loyalty number"
an **intrinsic replacement effect** (CR 614.1c), so it applies however the planeswalker enters. Every
other entry — an "exile until this leaves" return (Sheltered by Ghosts), reanimation, a tutor
straight into play, a DFC returning transformed onto a planeswalker back face, a token copy — arrived
at 0 loyalty and was put into its owner's graveyard by state-based actions (CR 704.5i) the instant it
entered. A token that hits the graveyard simply ceases to exist.

## What changed

**`ZoneMovementUtils.applyPlaneswalkerEntryIfNeeded`** is the single entry hook, mirroring the Saga
lore-counter split that already exists. It is called from every battlefield-entry pipeline:

| Pipeline | Call site |
|---|---|
| Zone-change entries (reanimation, linked-exile return, put-onto-battlefield, transform-and-return) | `ZoneTransitionService.moveToZone` battlefield branch |
| Ad-hoc token minting (`BattlefieldEntry.place`, which deliberately skips enters-with setup) | `TokenFromDefinition` + the four `CreateTokenCopyOf*Executor`s + `TokenCreationReplacementHelper.createAttachedPermanentCopies` |
| Cast / resolve | `StackResolver`, while the permanent is still on the stack |

All of them land in the same `EntersWithReplacements.placeEntryCounters` call, so counter-placement
modifiers (Doubling Season, Vorinclex, Pir), the "a counter was placed this turn" tracker, and the
emitted `CountersAddedEvent` behave exactly as they do for a printed "enters with counters" — and no
entry can get the counters twice.

Details worth calling out:

- **Token copies** need this because loyalty is a copiable value (CR 707.2), so the copy has a
  printed loyalty number of its own. Reachable today through Extravagant Replication ("a copy of
  another target nonland permanent you control").
- **The cast path no longer hand-rolls the placement.** It previously applied the counter modifiers
  but emitted no `CountersAddedEvent` and never recorded the placement, so counter watchers and the
  per-turn tracker saw a reanimated planeswalker but not a cast one. Both paths now share one helper.
- **Face-down entries are skipped inside the helper** (a face-down permanent is a nameless 2/2
  creature with no printed loyalty, CR 708.2a), so every caller inherits the check rather than
  repeating it.
- **DFC back faces work** because `returnDfcFace` flips the `CardComponent` before handing the entity
  to `moveToZone`, so the entry step reads the planeswalker face's printed loyalty.

`docs/card-sdk-language-reference.md` is updated in the same change: `startingLoyalty` is placed by
the engine on every entry, and card definitions never place it themselves.

No new SDK types — this is engine wiring over existing primitives.

## Tests

`PlaneswalkerEntryLoyaltyScenarioTest` (8 cases) and `TokenCopyOfPlaneswalkerEntryScenarioTest`:

- returned by an "exile until this leaves" Aura (the reported repro) — comes back at printed loyalty,
  under its owner's control, not in the graveyard
- reanimated from a graveyard (Perennation), with its own keyword counters landing on the same object
- cast — placed **exactly once**, guarding against the two pipelines double-applying
- a DFC returning transformed onto its planeswalker back face — the back face's loyalty, not the
  front's
- a token copy of a planeswalker — copied printed loyalty, and it survives the SBA check
- Doubling Season doubling the entry loyalty on **both** the cast and the reanimation path — the
  modifier parity the shared helper exists for
- negative cases: a face-down entry and a non-planeswalker get no loyalty counters

## Verification

`just test-rules` — **9333 tests passed, 0 failures** across 2952 test classes.

https://claude.ai/code/session_0145yakyrz13zRy5cATmB5NT

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -80,7 +80,10 @@ section; do not let SDK additions land without a corresponding doc update.
 - `dynamicStats(source, powerOffset?, toughnessOffset?)` — the `*`/`*` cycle: composes
   `dynamicPower` + `dynamicToughness` over one shared source, with optional `±` deltas
   (Tarmogoyf's `toughnessOffset = 1`).
-- `startingLoyalty: Int?` — starting loyalty for planeswalkers.
+- `startingLoyalty: Int?` — starting loyalty for planeswalkers. Placed as loyalty counters by the
+  engine's intrinsic enters-with replacement (CR 306.5b) on *every* battlefield entry — resolving from
+  the stack, reanimation, an "exile until this leaves" return, or any other put-onto-the-battlefield
+  effect. Card definitions never place them.
 - `colorIdentity: String?` — override (normally auto-detected). Treated as authoritative in this repo.
 - `colorIndicator: String?` — explicit color indicator (CR 204), e.g. `"B"`. `null` (default) = no
   indicator; the card's color is its mana-cost colors alone. Set it on a face printed with a color
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -82,8 +82,8 @@ section; do not let SDK additions land without a corresponding doc update.
   (Tarmogoyf's `toughnessOffset = 1`).
 - `startingLoyalty: Int?` — starting loyalty for planeswalkers. Placed as loyalty counters by the
   engine's intrinsic enters-with replacement (CR 306.5b) on *every* battlefield entry — resolving from
-  the stack, reanimation, an "exile until this leaves" return, or any other put-onto-the-battlefield
-  effect. Card definitions never place them.
+  the stack, reanimation, an "exile until this leaves" return, any other put-onto-the-battlefield
+  effect, and token copies (loyalty is a copiable value, CR 707.2). Card definitions never place them.
 - `colorIdentity: String?` — override (normally auto-detected). Treated as authoritative in this repo.
 - `colorIndicator: String?` — explicit color indicator (CR 204), e.g. `"B"`. `null` (default) = no
   indicator; the card's color is its mana-cost colors alone. Set it on a face printed with a color
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/effects/ZoneMovementUtils.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/effects/ZoneMovementUtils.kt
@@ -7,6 +7,7 @@ import com.wingedsheep.engine.core.PermanentsSacrificedEvent
 import com.wingedsheep.engine.core.GameEvent as EngineGameEvent
 import com.wingedsheep.engine.handlers.PredicateContext
 import com.wingedsheep.engine.handlers.PredicateEvaluator
+import com.wingedsheep.engine.registry.CardRegistry
 import com.wingedsheep.engine.state.GameState
 import com.wingedsheep.engine.state.ZoneKey
 import com.wingedsheep.engine.state.ComponentContainer
@@ -59,6 +60,7 @@ import com.wingedsheep.sdk.scripting.GameObjectFilter
 import com.wingedsheep.sdk.scripting.RedirectZoneChange
 import com.wingedsheep.sdk.scripting.RedirectZoneChangeWithEffect
 import com.wingedsheep.sdk.scripting.ZoneChangeCause
+import com.wingedsheep.sdk.scripting.events.CounterTypeFilter
 import com.wingedsheep.sdk.scripting.predicates.CardPredicate
 import com.wingedsheep.sdk.scripting.predicates.ControllerPredicate
 import com.wingedsheep.sdk.scripting.predicates.evaluateWith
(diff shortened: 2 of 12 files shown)


bolav:
## Review — CR 306.5b entry loyalty

Reviewed the branch against `main` (already merged in, so this reflects post-merge reality). The
findings below were raised against the first commit (`77f87e8`) and **all of them are fixed in the
follow-up commit** on this branch (`fd541f1`) — recorded here so the reasoning is on the record.

### Verdict

Behavior right, shape right: no new SDK types, just the intrinsic CR 306.5b replacement wired into
the zone-change chokepoint, reusing `EntersWithReplacements.placeEntryCounters` so modifiers, the
counter-placed tracker, and the event all come along. All cited rules verified against the
Comprehensive Rules (306.5b, 614.1c, 704.5i, 708.2/708.2a, 707.2). The gap was coverage, not
correctness.

### What's good

- **Right primitive, zero new types.** Routing through `placeEntryCounters` rather than hand-adding a
  `CountersComponent` is the reuse the SDK principles ask for. `Counters.LOYALTY == "loyalty"` already
  matches what `LoyaltyChangedEvent` sends the client, so no `ClientEvent.kt` work was needed.
- **Right chokepoint, and the double-apply concern handled honestly.** The Saga split is the correct
  precedent, documented at both ends, and the "places its printed loyalty exactly once" test is
  precisely the regression guard it needs.
- **The DFC claim holds up.** `returnDfcFace` flips the `CardComponent` *before* calling `moveToZone`,
  so reading `cardComponent.cardDefinitionId` really does pick up the back face's printed loyalty.
  Reading base state rather than projection is also correct here — printed loyalty is a copiable
  value, and a token/clone's `cardDefinitionId` already points at the copied definition.
- Negative cases (face-down, non-planeswalker) tested, not assumed; docs updated in the same change.

### Findings

**1. Token copies of planeswalkers still entered at 0 loyalty and ceased to exist.** *(Important —
fixed)* `TokenFromDefinition` and the `CreateTokenCopyOf*Executor`s mint through
`BattlefieldEntry.place`, which by design skips enters-with setup — which is exactly why each of them
already called `applySagaEntryIfNeeded` by hand. The planeswalker hook wasn't added alongside, so the
commit's "every battlefield entry" didn't hold for tokens. Confirmed with a throwaway scenario test:
casting "create a token that's a copy of target planeswalker" on an `Ajani, Caller of the Pride` left
one battlefield copy (the original, loyalty 4) and zero graveyard copies — the token entered at 0
loyalty, `PlaneswalkerLoyaltyCheck` (which gates on `projected.isPlaneswalker`) binned it, and the
token vanished. Reachable today via `ncc/ExtravagantReplication`
(`TargetFilter.OtherNonlandPermanent.youControl()`).
→ Fixed: the hook now sits next to every `applySagaEntryIfNeeded` call, plus
`TokenCreationReplacementHelper.createAttachedPermanentCopies` (an Aura can enchant a planeswalker),
with `TokenCopyOfPlaneswalkerEntryScenarioTest` mirroring the Saga test's shape.

**2. The cast path duplicated the step, and the copies had already drifted.** *(Important — fixed)*
The comment said "keep the two in sync", but `StackResolver` applied the counter modifiers while
emitting no `CountersAddedEvent` and never calling `recordCounterPlacement`. So a `CountersPlacedEvent`
watcher and any "a counter was placed this turn" condition saw loyalty counters on a *reanimated*
planeswalker but not on a *cast* one — same card, same counters, different observable behavior.
→ Fixed: the cast path calls `EntersWithReplacements.placeEntryCounters` too, so there is one
implementation and the sync comment reduces to "both call the same helper".

**3. The counter-modifier claim was untested.** *(Important — fixed)* The KDoc cites Vorinclex/Pir
parity as the reason for routing through `placeEntryCounters`, and nothing exercised it.
→ Fixed: Doubling Season doubling asserted on **both** the cast and the reanimation path.

**4. Docs overstated coverage.** *(Minor — fixed)* `card-sdk-language-reference.md` said "every
battlefield entry" while tokens were excluded. Now true, and it names token copies with the CR 707.2
rationale.

**5. The face-down guard sat at the call site.** *(Minor — fixed)* `!options.faceDown` in
`ZoneTransitionService` meant every future caller had to remember it — and finding #1 added six.
`applyBattlefieldEntry` attaches `FaceDownComponent` before that point, so the helper self-guards now.

**6. The DFC-returning-transformed claim was unpinned.** *(Minor — fixed)* Correct by inspection, but
no implemented card reaches `returnDfcFace` with a planeswalker face. Now covered by a synthetic
creature-front / planeswalker-back test card.

One correction to the review itself: I first cited CR 111.4 for token copies carrying printed
loyalty. That's the token name/subtype rule — the right citation is **707.2** ("copiable values …
power, toughness, and/or loyalty"), which is what landed in the code.

### Verification

`just test-rules` on the final tree: **9333 tests passed, 0 failures** across 2952 test classes.

https://claude.ai/code/session_0145yakyrz13zRy5cATmB5NT

PR #1661, opened 2026-08-04
Collapse the *OrPay additional-cost family into one AdditionalCost.OrPay

bolav:
# Collapse the `*OrPay` additional-cost family into one `AdditionalCost.OrPay`

Five near-identical cast-time cost types — `SacrificeOrPay`, `DiscardOrPay`,
`ExileFromGraveyardOrPay`, `BeholdOrPay` and their shared enumerator/handler plumbing — differed
only in *which cost sat on the non-mana leg*. They collapse into one type parameterized by that
cost:

```kotlin
data class OrPay(val cost: AdditionalCost, val alternativeManaCost: String) : AdditionalCost
```

`BlightOrPay` is the one holdout: there is no standalone blight `AdditionalCost` for `OrPay` to
wrap, so collapsing it means introducing one first. Noted in
`backlog/sdk-reusability-consolidation.md`, which proposed this change.

## Why it's smaller

`CastSpellHandler.reduceCostAlternatives` rewrites an `OrPay` to its **leg cost** when the caster
populated that cost's payment field, and drops it otherwise (the pay path — whose only consequence,
the extra mana, is folded in by `applyOrPayManaAdjustments`). Downstream, the leg is just a plain
`Atom` / `Behold` cost, so validation, payment, LKI snapshots (CR 112.7a / 608.2h), behold's
pipeline storage and the turn's discard tracking (CR 701.8) are inherited rather than
re-implemented per or-pay type. `applyOrPayManaAdjustments` is shared by `validate()` and
`execute()` so a cast is priced identically in both.

In the enumerator, every or-pay leg now produces the same `OrPayPath` record, so the three
`LegalAction` emission sites loop over one list instead of repeating a near-identical construction
per cost type.

No card definitions changed: the printed wordings (`Costs.additional.SacrificeOrPay`, `.DiscardOrPay`,
`.ExileFromGraveyardOrPay`, `.BeholdOrPay`) survive as one-line facades over `OrPay`.

## Fixes carried along

- **`costType = "ReturnToHand"` → `"BouncePermanent"`.** `"ReturnToHand"` is in no client cost-type
  list; every other bounce-cost site (activated abilities, Sneak, Web-slinging) emits
  `"BouncePermanent"`. A spell with a `ReturnToHand` additional cost would have surfaced an action
  the client could not prompt for. Pre-existing; fixed on both the plain-atom and or-pay paths.
- **Non-graveyard `ExileFrom` legs are declined rather than mislabelled.** The
  `"ExileFromGraveyard"` costType pins the client's picker to the graveyard, so an `ExileFrom` leg
  on another zone now offers only the pay path instead of showing the wrong zone.
- **`AdditionalCost.Composite` is flattened before alternatives are reduced**, in both
  `reduceCostAlternatives` and `applyOrPayManaAdjustments`, so an alternative nested in a composite
  is reduced *and* priced like a top-level one. Previously a nested or-pay would have been silently
  free.
- **`TapPermanents.excludeSelf`** is now honoured on the or-pay leg, matching the plain-atom path.

## Behaviour note — payment-size tolerance

The deleted `SacrificeOrPay` / `DiscardOrPay` / `ExileFromGraveyardOrPay` validators rejected a
payment whose size was `!= count`. The shared `Atom` validators they now route through check
`size < count`, so an over-sized payment (two permanents for "sacrifice a legendary creature") would
be accepted and both sacrificed. This aligns these four cards with how every other atom cost in the
engine already behaves, and the client always sends exactly `count`. Tightening `<` to `!=` is a
reasonable follow-up but is an engine-wide change to every atom cost, deliberately out of scope
here.

## Verification

- `:rules-engine:test` and `:mtg-sdk:test` — full suites green.
- `:mtg-sets:test --tests "*CardDefinitionSnapshotTest"` — goldens re-blessed; only ECL / FIN / MID /
  SOS / SPM / TDM or-pay cards moved.
- Existing scenario coverage exercises both legs and the leg-unavailable case for each shape:
  `LouisoixsSacrificeScenarioTest`, `PumpkinBombardmentScenarioTest`, `SoaringStonegliderScenarioTest`,
  `CausticExhaleScenarioTest` (behold leg), `SoulsOfTheLostScenarioTest` (the neighbouring `Choice`
  cost-vs-cost shape).

### DIFF
--- a/backlog/sdk-reusability-consolidation.md
+++ b/backlog/sdk-reusability-consolidation.md
@@ -339,8 +339,12 @@ ZoneChange/SpellCast/Attack/DealsDamage variants:
 
 ### Costs (`scripting/AdditionalCost.kt`, `scripting/costs/PayCost.kt`)
 
-- **`AdditionalCost.BlightOrPay`** / **`BeholdOrPay`** → wrap any cost with
-  `OrPayMana(inner: AdditionalCost, manaCost: String)`.
+- ~~**`AdditionalCost.BeholdOrPay`** → wrap any cost with
+  `OrPayMana(inner: AdditionalCost, manaCost: String)`.~~ **Done** — landed as
+  `AdditionalCost.OrPay(cost: AdditionalCost, alternativeManaCost: String)`, which also
+  absorbed the `SacrificeOrPay` / `DiscardOrPay` / `ExileFromGraveyardOrPay` types.
+  `BlightOrPay` is the one holdout: there is no standalone blight `AdditionalCost` for
+  `OrPay` to wrap, so collapsing it means introducing one first.
 - **Strategic:** `AdditionalCost` (cast-time) and `PayCost` (activation-time) share
   concepts but neither references the other. Eventually unify under a single `Cost`
   sealed interface — flag, not an immediate change.
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -463,34 +463,41 @@ definitions construct these through the facade, e.g. `Costs.additional.Sacrifice
   mana cost is waived — `GrantMayCastFromLinkedExile(withoutPayingManaCost = true, additionalCost =
   Costs.additional.PayLifeEqualToManaValueOfSpell)` — so the only cost paid is the life. The amount is
   read from the cast card's mana value, checked at cast time (CR 119.4 — must have at least that much life).
-- `Costs.additional.CostOrPay(atom, alternativeManaCost)` — **cost-vs-mana**: "as an additional cost
-  to cast this spell, \<pay this cost\> **or** pay {mana}". One `AdditionalCost.CostOrPay` covers the
-  whole "do X or pay {N}" family over the shared `CostAtom` vocabulary; the named shapes below are
-  its printed wordings and are one-line facades over it. The enumerator offers up to two cast paths:
-  the **atom path** (base cost + the atom's ordinary selection prompt — the same `costType` a plain
-  atom cost of that kind emits, so no new client UI) and the **pay path** (base cost +
-  `alternativeManaCost` folded in). The atom path is offered only when the board affords the atom's
-  selection, so with nothing to pay it only the pay path is castable — but the cost as a whole is
-  always payable. Which leg was taken is recovered at payment time from **which
-  `additionalCostPayment` field the client populated**, so `atom` must be a selection-carrying atom
-  (`Sacrifice` / `Discard` / `ExileFrom` / `TapPermanents` / `ReturnToHand`) and must not share its
-  field with another additional cost on the same card. `CastSpellHandler.reduceCostAlternatives`
-  then rewrites the cost to a plain `AdditionalCost.Atom` (leg paid) or drops it (pay path), so
-  validation, payment, LKI snapshots and discard tracking (CR 701.8) reuse the atom paths verbatim.
-  `BlightOrPay` / `BeholdOrPay` stay separate types because blight and behold aren't `CostAtom`s.
+- `Costs.additional.OrPay(cost, alternativeManaCost)` — **cost-vs-mana**: "as an additional cost
+  to cast this spell, \<pay this cost\> **or** pay {mana}". One `AdditionalCost.OrPay` covers the
+  whole "do X or pay {N}" family, parameterized by the `AdditionalCost` on the non-mana leg; the
+  named shapes below are its printed wordings and are one-line facades over it. The enumerator
+  offers up to two cast paths: the **leg path** (base cost + the leg cost's ordinary selection
+  prompt — the same `costType` that cost emits standing alone, so no new client UI) and the **pay
+  path** (base cost + `alternativeManaCost` folded in). The leg path is offered only when the board
+  affords the leg's selection, so with nothing to pay it only the pay path is castable — but the
+  cost as a whole is always payable. Which leg was taken is recovered at payment time from **which
+  `additionalCostPayment` field the client populated**, so `cost` must be a selection-carrying cost
+  — a `Behold`, or an atom cost over `Sacrifice` / `Discard` / `ExileFrom` / `TapPermanents` /
+  `ReturnToHand` — and must not share its field with another additional cost on the same card.
+  `CastSpellHandler.reduceCostAlternatives` then rewrites the whole cost to that plain leg cost (leg
+  paid) or drops it (pay path), so validation, payment, LKI snapshots, behold's pipeline storage and
+  discard tracking (CR 701.8) reuse the leg's own paths verbatim. `BlightOrPay` stays a separate
+  type only because there is no standalone blight `AdditionalCost` for `OrPay` to wrap.
+- `Costs.additional.BeholdOrPay(filter = Filters.Any, alternativeManaCost, storeAs = "beheld")` —
+  "behold a [filter] or pay {mana}" (Lys Alana Dignitary). `OrPay(AdditionalCost.Behold(filter,
+  storeAs = storeAs), …)`; the behold path surfaces as a `costType = "Behold"` cost over one
+  candidate pool spanning battlefield *and* hand, and stores the chosen cards under `storeAs` for
+  downstream costs/effects exactly as a plain `Behold` does.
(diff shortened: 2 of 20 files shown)


bolav:
## Code review — SDK elegance pass

Reviewed against `c1d7316683` ("Collapse the atom-backed *OrPay costs into one CostOrPay"), with
`origin/main` (`b080fbaf97`) already an ancestor of the branch. **Every finding below was addressed
in the follow-up commit `ab90b4660e`** — resolutions are inlined so the reasoning stays on the
record.

### Verdict

Right refactor, right shape. Three near-identical `AdditionalCost` types collapse into one
parameterized over the existing `CostAtom` vocabulary, behaviour is preserved, and −430 net lines
fall out. One client-contract bug in a then-unreachable leg, and one open question about whether the
collapse should have gone one parameter further.

### What's good

- **This is the collapse `sdk-design-principles.md` asks for.** `SacrificeOrPay` / `DiscardOrPay` /
  `ExileFromGraveyardOrPay` differed only in which atom sat on the non-mana leg — textbook "the next
  card forces another type". The named facades survive as one-liners, so no card definition changed.
- **`reduceCostAlternatives` reusing the plain `Atom` machinery is the load-bearing insight.**
  Rewriting a paid leg to `AdditionalCost.Atom(atom)` deletes ~200 lines of duplicated validation
  and payment and inherits the LKI snapshots (CR 112.7a / 608.2h) and CR 701.8 discard tracking for
  free, rather than re-implementing them per or-pay type as main did.
- **`applyOrPayManaAdjustments` shared by `validate()` and `execute()`** removes a real drift
  hazard — main kept two hand-maintained copies of the same fold, five costs each.
- **`OrPayPath` + the `orPayPaths` loop** turns 15 near-identical `LegalAction` constructions into
  3, and folds blight/behold into the same shape without changing what they emit.
- **Docs updated in the same change**, and the KDoc that survives is the good kind — *why*
  `BlightOrPay` / `BeholdOrPay` stay separate, not what the code says.
- **The three existing scenario tests are a genuine regression net**: both legs plus the
  atom-unavailable case for each shape.

### Issues

#### Important

**1. `costType = "ReturnToHand"` isn't a costType the client understands.**
`CastSpellEnumerator.kt:2367` emitted `"ReturnToHand"` for a `CostAtom.ReturnToHand` leg, but
`pipelinePhases.ts:178` lists `'BouncePermanent'` — what every other bounce-cost site emits
(`ActivatedAbilityEnumerator.kt:1212`, `SneakCastEnumerator.kt:90`, `WebSlingingCastEnumerator.kt:102`).
An unlisted costType means the client pushes no `costPayment` phase, so an
`OrPay(CostAtom.ReturnToHand(...), "{2}")` would surface a "(Return to hand)" action, send an empty
payment, and the engine would read that as the *pay* path and charge alt mana the action never
advertised. The plain-atom path at `:2248` had the same bug.

> **Fixed** in `ab90b4660e` — `costType = "BouncePermanent"` on both the or-pay leg and the
> plain-atom path.

**2. The collapse stopped one parameter short of `BeholdOrPay`.** `AdditionalCost.Behold` already
existed (`AdditionalCost.kt:268`), so `BeholdOrPay(filter, "{2}", storeAs)` is exactly
`OrPay(Behold(filter, storeAs), "{2}")` if the parameter is typed `AdditionalCost` instead of
`CostAtom`. Leg detection generalizes the same way — `atomPaymentSatisfied(atom)` becomes a
`paymentSatisfied(cost)` that reads `beheldCards` for `Behold` — and the enumerator had *already*
unified all three legs behind `OrPayPath`, so the only place the family was still split was the AST.
Much cheaper to widen the parameter now than after a second card lands on it.

> **Fixed** in `ab90b4660e` — the type is now
> `OrPay(cost: AdditionalCost, alternativeManaCost: String)`, absorbing `BeholdOrPay`; five
> cast-time types became one. The enumerator's `beholdOrPayCost` / `beholdBaseCost` /
> `canAffordBeholdPath` / `beholdPathInfo` locals collapsed into the generic path, and behold's
> pipeline storage + reveal events now come from the existing `Behold` branches. `BlightOrPay` stays
> — there is no standalone blight `AdditionalCost` to wrap — and its KDoc now says so, plus what
> would unblock it. `backlog/sdk-reusability-consolidation.md` had proposed exactly this shape
> (`OrPayMana(inner, manaCost)`); that entry is checked off with `BlightOrPay` noted as the
> remainder.

#### Minor

**3. "Unreachable" was only true for top-level costs.** `reduceCostAlternatives` ran *before*
`Composite` flattening (`CastSpellHandler.kt:1539`, `:2287`), and `applyOrPayManaAdjustments` folded
only top-level costs — so an or-pay nested inside an `AdditionalCost.Composite` would reach the
no-op `CostOrPay` branches *and* never be charged its alternative mana. No card does that today, and
main had the same hole via `filterIsInstance`, but the comment asserted a stronger invariant than
the code enforced.

> **Fixed** — a recursive `flattenComposites` now runs before *and* after reduction, and
> `applyOrPayManaAdjustments` flattens too, so the two always agree on which costs they see. The
> redundant `.flatMap { Composite }` at both call sites is gone.

**4. Payment-size validation loosened from `!=` to `<`.** The deleted validators rejected
`sacrificed.size != count`; the `Atom` validators they now route to check `size < atom.count`
(`CastSpellHandler.kt:1544`, `:1570`, `:1588`). A client sending two permanents for Louisoix's
Sacrifice would sacrifice both. This aligns the or-pay cards with how every other atom cost in the
engine already behaves and the real client always sends exactly `count`, so it reads deliberate.

> **Documented, not changed.** Tightening `<` to `!=` would alter every atom cost in the engine, not
> just these cards — out of scope for this refactor. Called out in the PR body and commit message so
> the tolerance is a recorded decision rather than an accident. Happy to tighten it as a follow-up.

**5. `ExileFrom` label and costType disagreed about the zone.** `orPayAtomCostData` built the label
from `atom.zone` ("Exile from hand") but hardcoded `costType = "ExileFromGraveyard"`, and the client
pins `targetZone = 'Graveyard'` for that costType (`pipelinePhases.ts:756`). Inherited from the
plain-atom path (`:2231`), but the zone-aware label made the mismatch visible.

> **Fixed** — a non-graveyard `ExileFrom` leg is now declined (pay path only) instead of being shown
> against the wrong picker, and the label is graveyard-specific. The constraint is documented on the
> helper and in the SDK reference.

**6. Three test files referenced the deleted types.** Dangling KDoc links in
`PumpkinBombardmentScenarioTest.kt:22,44`, `LouisoixsSacrificeScenarioTest.kt:25,48`,
`SoaringStonegliderScenarioTest.kt:24` — and "Exercises the *new* `…DiscardOrPay` cost" read oddly
for a type that no longer exists.

> **Fixed** — all four scenario tests updated (including `CausticExhaleScenarioTest`, which covers
> the behold leg), plus two more stale references the sweep turned up: a KDoc link in
> `CompositeEffects.kt` and a dead `"BeholdOrPay"` entry in `CardLinter.kt`'s storage-key table.

**7. `orPayAtomCandidates` dropped `TapPermanents.excludeSelf`** (`:2319`) where the plain-atom
branch honoured it (`:271`). A no-op for spells — the card being cast is never the battlefield
permanent — but a trap if reused for abilities.

> **Fixed** — `excludeSelf` is honoured on the leg, matching the plain-atom path.

### Verification

Run after the fixes, all with `--rerun-tasks --no-build-cache` (an earlier cached run reported
BUILD SUCCESSFUL while logging only 28 tests — counts below come from the forced run's result XML):

- `:rules-engine:test` + `:mtg-sdk:test` — **9713 tests across 2991 suites, 0 failures, 0 errors**.
- Both legs and the leg-unavailable case pass for every shape: `LouisoixsSacrificeScenarioTest` (4),
  `PumpkinBombardmentScenarioTest` (3), `SoaringStonegliderScenarioTest` (4),
  `CausticExhaleScenarioTest` (2, behold leg), `SoulsOfTheLostScenarioTest` (4, the neighbouring
  `Choice` cost-vs-cost shape).
- `:mtg-sets:test` card snapshots — 306 tests, 0 failures; goldens re-blessed, only ECL / FIN / MID /
  SOS / SPM / TDM or-pay cards moved.

PR #1663, opened 2026-08-04
Stop the AI going in circles

bolav:
# Stop the AI going in circles

Two bugs from real games, plus the review fixes that followed.

## The loop

The AI activated Aphetto Alchemist (`{T}: Untap target artifact or creature`) eleven times in a row,
aimed at itself, until the game had to be abandoned. A leaf-scoring search has no defence against
this on its own: it compares "take this action" against "pass" by scoring the positions they lead
to, and those two positions are not measured at the same point in the game. Passing carries the game
forward into whatever was about to happen, while a free ability that resolves back onto the same
board stops right where it is. When what is about to happen is bad, doing nothing at all can
*outscore* passing — and then it does so again, from the identical position it just produced.

`StateProgress.digest` reduces a `GameState` to the position a player could point at, and
`Strategist` refuses any candidate whose leaf repeats the position it is acting from or one of the
last 32 it has acted from. That covers both shapes: the inert action (Alchemist untapping itself)
and the cycle (two Alchemists untapping each other, where every step changes the board and the
sequence still goes nowhere).

CR 732.3 makes the same structural claim, and its example turns on exactly the thing
`IGNORED_COMPONENTS` does — the loop repeats a position when "nothing in the game cares how many
times an ability has been activated."

## The crash

An "other target" slot (CR 601.2c) whose every legal target was already spent on an earlier slot
took the AI's whole priority window down with it — in a served game, the AI simply stops playing.
Found by fuzzing self-play with random decks. Mabel's Mettle's "up to one *other* target creature"
with one creature on the board is the minimal case: the enumerator cannot know which target the
earlier slot will take, so it offers that creature for both, and `fillHeuristically`'s
`?: available.first()` fallback could only ever run on an empty list.

Targets are submitted as one flat list sliced back by max counts, so only *trailing* slots may be
left empty. When the rest is optional the prefix is a legal list; otherwise there is no legal list at
all and the action goes back unfilled, for the caller's simulation or the engine to reject.

## Review fixes

The guard as first written could refuse a good play, which is worse than the loop it prevents:

- **Aim an inert ability instead of dropping it.** The target-level inert check lived in
  `chooseCommittedTargets`, which a budget below `BudgetTier.NORMAL` skips entirely — and `ROUTINE`
  is precisely the quiet opponent's-turn window the loop was reported in. There the committed target
  is `TargetSelection.rank`'s, which scores board value and has no notion of whether the ability does
  anything *to* the target, so it can hand back the Alchemist untapping itself while a tapped
  creature sits right there. `Strategist.materialize` buys the simulated pick back for exactly those
  candidates; only the inert path pays for it.
- **Name what the digest excludes, not what it reads.** The first shape was a read-list covering ~20
  of `GameState`'s ~50 fields, and it missed every turn-level rider an ability can set without
  touching a permanent — `turnSpellCostReductions`, `activeCounterPlacementModifiers`,
  `pendingUncounterableSpells`, `damageCantBePreventedThisTurn`, all with live executors. A field
  missing from a read-list makes a real action look inert, and an action that looks inert is refused
  forever; a wrong exclusion costs one wasted activation. `normalized` strips the bookkeeping and
  lets the data class's own `hashCode` supply the rest, so a field added later counts by default.
- **Take the anytime cut on every candidate.** A dropped candidate had already paid for its
  simulation, so `continue`-ing past `budget.expired()` let a board full of inert candidates run
  straight past the budget.

## On `LEGACY_V0`

The guard is **not** behind an `AiProfile` flag, which is the one thing this codebase normally
insists on — `FrozenBaselineTest` exists to stop `LEGACY_V0` drifting, and Phase 4's `fillPartial` is
gated for no other reason. That rule is about *strength*: a change that makes V0 play better silently
rebases every published arena number. A game that has to be abandoned has no strength to compare, so
there is nothing here worth freezing. `FrozenBaselineTest` stays green, but only because its baseline
deck is all-vanilla — no activated abilities, so nothing the guard can fire on. It is not evidence
either way and shouldn't be read as any. This is recorded in `StateProgress`'s KDoc so the next
person doesn't have to rediscover it.

## Tests

- `LoopingActionAiTest` — the inert action, the two-Alchemist cycle, and a sweep over all four
  `BudgetTier`s proving an ability with one inert target and one productive one gets *aimed*, not
  dropped. That sweep fails at `TRIVIAL` without `materialize` (the AI chooses `PassPriority` over
  untapping the tapped Grizzly Bears).
- `OtherTargetFillAiTest` — the optional-slot and mandatory-slot branches, plus a whole priority
  window survived under `AiProfile.PRODUCTION`.
- `StateProgressTest` — both halves of the digest contract: bookkeeping every action advances is not
  a change, a turn-level rider is.

`:ai:test` green, including `FrozenBaselineTest`. Also drops `TargetSelection.bestTarget`, which had
no callers left, and `StateProgress.isInert`, which had no production caller.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/StateProgress.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/StateProgress.kt
@@ -1,0 +1,149 @@
+package com.wingedsheep.ai.engine
+
+import com.wingedsheep.engine.state.GameState
+import com.wingedsheep.engine.state.components.battlefield.AbilityActivatedEverComponent
+import com.wingedsheep.engine.state.components.battlefield.AbilityActivatedThisTurnComponent
+import com.wingedsheep.engine.state.components.battlefield.TargetedByControllerThisTurnComponent
+import com.wingedsheep.engine.state.components.battlefield.TimestampComponent
+import com.wingedsheep.sdk.core.Zone
+import com.wingedsheep.sdk.model.EntityId
+
+/**
+ * "Is this position one I have already been in?" — how the AI tells progress from a treadmill.
+ *
+ * Two shapes of the same bug live here. The first is an **inert** action, one whose resolved
+ * position is the position it started from: Aphetto Alchemist's `{T}: Untap target artifact or
+ * creature` aimed at itself taps for its own cost and then untaps itself again. The second is a
+ * **cycle** — two Alchemists untapping each other — where every single step does change the board
+ * and the sequence still goes nowhere.
+ *
+ * A leaf-scoring search has no defence against either on its own. It compares "take this action"
+ * against "pass" by scoring the two positions they lead to, and those two positions are not
+ * measured at the same point in the game: passing carries the game forward into whatever was about
+ * to happen, while a free ability that resolves back onto the same board stops right where it is.
+ * When what is about to happen is bad, doing nothing at all can therefore *outscore* passing — and
+ * once it does, it does so again from the identical position it just produced, forever. That is the
+ * bug this exists to prevent: the AI activated Aphetto Alchemist until the game had to be
+ * abandoned.
+ *
+ * So this is not a heuristic about value — the evaluator owns that question. It is the structural
+ * claim underneath it, and the rules make the same one: CR 732.3 requires a player whose actions
+ * have "resulted in the same game state being reached multiple times" to make a *different* game
+ * choice, and its example turns on the very thing [IGNORED_COMPONENTS] does — the loop repeats a
+ * position when "nothing in the game cares how many times an ability has been activated."
+ *
+ * [Strategist] is the consumer: it drops any candidate whose leaf repeats a position it has already
+ * acted from.
+ */
+object StateProgress {
+
+    /**
+     * Whether [after] is the same game position as [before], so the action between them
+     * accomplished nothing at all.
+     *
+     * Compares a digest rather than the states themselves. `GameState` equality is unusable here —
+     * a resolved ability leaves behind an orphaned stack entity, a bumped `nextEntityId` and an
+     * advanced `rng`, none of which is a game fact — so [digest] reads the position the way a
+     * player would: zone contents, and everything true of the objects and players in them.
+     */
+    fun isInert(before: GameState, after: GameState): Boolean = digest(before) == digest(after)
+
+    /**
+     * A 64-bit summary of everything about [state] that a player could point at.
+     *
+     * Deliberately blind to bookkeeping that changes on *every* activation regardless of what the
+     * ability did — see [IGNORED_COMPONENTS] — and to whose priority it is, which is what makes an
+     * action's own resolution comparable with the position it started from.
+     *
+     * Turn number and step *are* part of it, which is what keeps the repetition memory honest: the
+     * same board in a later step is a different position, so a digest can only recur inside the
+     * window where recurring means going in circles.
+     *
+     * Per-collection hashes are summed rather than folded, so a `Map` that rehashed between the two
+     * states can't read as a change; order *within* a zone still counts, because library and stack
+     * order are game facts.
+     */
+    fun digest(state: GameState): Long {
+        var h = SEED
+
+        h = h.mix(state.turnNumber)
+        h = h.mix(state.phase.hashCode())
+        h = h.mix(state.step.hashCode())
+        h = h.mix(state.activePlayerId.hashCode())
+        h = h.mix(state.gameOver.hashCode())
+        h = h.mix(state.winnerId.hashCode())
+        h = h.mix(state.stack.hashCode())
+        h = h.mix(state.continuationStack.size)
+        h = h.mix(state.floatingEffects.hashCode())
+        h = h.mix(state.delayedTriggers.hashCode())
+        h = h.mix(state.grantedTriggeredAbilities.hashCode())
+        h = h.mix(state.grantedActivatedAbilities.hashCode())
+        h = h.mix(state.grantedStaticAbilities.hashCode())
+        h = h.mix(state.grantedReplacementEffects.hashCode())
+        h = h.mix(state.grantedKeywordAbilities.hashCode())
+        h = h.mix(state.globalGrantedTriggeredAbilities.hashCode())
+        h = h.mix(state.mayPlayPermissions.hashCode())
+        h = h.mix(state.commanderDamage.hashCode())
+        h = h.mix(state.spellsCastThisTurn)
+        h = h.mix(state.permanentsSacrificedThisTurn)
+
+        var zones = 0L
+        for ((key, contents) in state.zones) {
+            var zone = key.hashCode().toLong()
+            for (entityId in contents) {
+                zone = zone.mix(entityId.hashCode())
+                // A library is hashed by its order alone: 60 cards whose components no game action
+                // touches without also moving them somewhere this digest reads in full.
+                if (key.zoneType != Zone.LIBRARY) zone = zone.mix(objectHash(state, entityId))
+            }
+            zones += zone
+        }
+        h = h.mix(zones)
+
+        var objects = 0L
+        for (entityId in state.stack) objects += objectHash(state, entityId)
+        for (playerId in state.turnOrder) objects += objectHash(state, playerId)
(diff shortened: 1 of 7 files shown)


bolav:
# Review — SDK elegance, correctness, tests

Reviewed at `a639217b39` (before the follow-up commit). `origin/main` was already an ancestor, so no
merge was needed. Every finding below is fixed in `Make the AI loop guard fail safe, and prove it
does`; the resolution is noted inline.

## Verdict

Both bugs are real and the fixes are the right shape: `StateProgress` correctly identifies that a
leaf score can't see a repeated position, and the `TargetSelection` change replaces a
`first()`-on-empty that could only ever throw. No new SDK types are introduced — the whole diff lives
in `ai/` — so the SDK-elegance question doesn't really bite here. Two things needed fixing in the
target-refinement half: on the tiered budget the inert check silently didn't run in exactly the
window the reported bug lives in, and that code path had no test.

## What's good

- **Digesting rather than comparing `GameState` is the right call, and the KDoc says why**: an
  orphaned stack entity, a bumped `nextEntityId` and an advanced `rng` are simulation residue, not
  game facts. That observation is what makes the whole thing work.
- **Summing per-collection hashes instead of folding** so a rehashed `Map` can't read as a change,
  while keeping order *within* a zone because library and stack order are game facts. Same for
  `objectHash` summing over components, since `container.all()` order isn't guaranteed.
- **CR 732.3 verifies exactly**, including the quoted fragment, against
  `MagicCompRules_20260619.txt:6380-6381`. Its example ("nothing in the game cares how many times an
  ability has been activated") is genuinely the justification for `IGNORED_COMPONENTS` — not a
  decorative citation.
- **`leaves` and `leafStates` stay index-aligned** through the new `continue`, and
  `firstCandidate`/`leafScores` indexing still lines up. Easy to break, not broken.
- **`LoopingActionAiTest` test 3 genuinely exercises the cross-decision memory**, not just the
  degenerate self-target: after untapping the tapped Alchemist, untapping it back reaches the
  remembered position, and untapping itself hits `leaf == here`. Both branches from one honest
  scenario.
- **The `TargetSelection` reasoning matches `fillableRequirements`' own contract** — flat list sliced
  by max counts, so only trailing slots may be empty. Both branches tested, and the third test proves
  the priority window survives under `AiProfile.PRODUCTION`.
- `bestTarget` deleted with no callers left anywhere, and oracle text for all three cards matches
  Scryfall exactly.
- The obvious hazard of putting mutable state on `Strategist` is a leak across games. It doesn't
  happen: `AIPlayer.create` is the only construction site, `EngineAiPlayerController` is built per
  session in `AiGameManager`, and `ArenaAgent` builds one per seat per game. Arena reproducibility is
  safe.

## Blocking

**The target-level inert check never ran on the budget tier where the bug lives.** `Strategist`
returned `heuristicTargets(...)` whenever `!budget.allowances.refineTargetsBySimulation`, which
`DecisionBudget.forMillis` sets false below `NORMAL` — i.e. `BudgetTier.ROUTINE`, documented as "an
opponent's-turn window with no immediate threat; upkeep, draw, end step". That is precisely the
Aphetto Alchemist window. On that path the target comes from `TargetSelection.rank`, which ranks own
creatures by `-permanentValue` and has no notion of whether the ability does anything to them, so it
can commit the inert self-target; the ability-level guard then drops the whole candidate and the AI
passes, even though untapping the other creature was available and good. The comment claiming
"`chooseAction` only ever drops the whole ability as inert when *no* target does anything" was
therefore false below `NORMAL`. Not reachable from `PRODUCTION` (it uses `LegacyBudgetPolicy`, whose
`LEGACY` allowances keep refinement on), but reachable from `PHASE4`, `PHASE4_PHASE6` and
`PHASE4_PHASE6_PHASE7` — the last described as "what the plan proposes to ship".

> **Fixed** — `Strategist.materialize` re-commits targets with simulation for exactly the candidates
> the cheap pick made inert, with its own `RESCUE_TARGET_CANDIDATES` floor because a sub-`NORMAL`
> tier can cap candidates at 1. Ordered so a budget that already refines pays nothing at all. The
> comment now matches.

**No test covered the `Double.NEGATIVE_INFINITY` branch.** `LoopingActionAiTest` set up exactly the
right board — Alchemist plus a tapped Grizzly Bears — but only called `StateProgress.isInert` on two
hand-built `ActivateAbility`s; it never routed through `chooseCommittedTargets`. So the one hunk whose
whole purpose is "prefer the productive target over dropping the ability" was unexercised.

> **Fixed** — new test sweeps all four `BudgetTier`s and asserts the AI returns the Alchemist
> activation targeting the bears. Verified it fails without the fix: at `TRIVIAL` it chose
> `PassPriority`.

## Important

**`digest` was an allowlist over `GameState`, and the failure direction was the expensive one.** It
read ~20 of ~50 constructor fields. Unread but action-mutable, each with a live executor:
`turnSpellCostReductions` (`ReduceSpellCostsThisTurnExecutor`), `activeCounterPlacementModifiers`
(`GrantCounterPlacementModifierExecutor`), `pendingUncounterableSpells`
(`MakeNextSpellUncounterableExecutor`), `damageCantBePreventedThisTurn` — plus `dayNight`,
`pendingSpellCopies`, `spellWarpedThisTurn`, `playersWhoCommittedCrimeThisTurn`,
`lastCastSpellColors`. An action changing only one of those reads as inert and becomes permanently
un-takeable. That is the *opposite* direction from the fail-safe argument `IGNORED_COMPONENTS` makes,
which is about components and doesn't transfer to fields. Exposure was bounded in practice — any
ability with a mana or tap cost mutates `ManaPoolComponent` or `TappedComponent` and so can never
read as inert, leaving free ({0}, no-tap) and self-undoing abilities — and no card in the repo trips
it today.

> **Fixed** — `normalized` names the exclusions and lets the data class's own `hashCode` supply the
> rest. `projectedState` is a body property, so it was already out of `hashCode`. `StateProgressTest`
> pins both directions. `objectHash` now mixes the entity id with the component hash rather than
> adding alongside it, so two objects in one zone trading component sets is a change rather than the
> same sum.

**The guard changed `LEGACY_V0` ungated.** `FrozenBaselineTest`'s KDoc is explicit: "If that was
intentional, it belongs behind a *new* profile, not inside `LEGACY_V0`." The precedent is right next
door — `fillPartial` is gated behind `useMeaningfulFilter` for no other reason. The test passes, but
only because its deck is deliberately all-vanilla with no activated abilities, so the guard never
fires; that is not evidence V0 is unchanged.

> **Resolved as documented, not gated** — a game that has to be abandoned has no strength to compare,
> so gating would preserve a hang rather than a baseline. `StateProgress`'s KDoc and
> `docs/ai/architecture.md` now record the decision and the fact that `FrozenBaselineTest` green is
> not evidence either way.

**`continuationStack` contributed only its size**, while every other collection contributed
`hashCode()`, and `pendingDecision` wasn't read at all. Two different pending continuations of equal
depth collide. The `NeedsDecision` exemption covers most of it, but a `Terminal` leaf can still carry
a non-empty continuation stack. The asymmetry read as an oversight.

> **Fixed (documented)** — behaviour unchanged, but the KDoc now says why: a frame holds
> mid-resolution bookkeeping a re-resolution mints afresh, so hashing its contents would make an
> inert action look like progress, and depth alone suffices because the compared states are quiet
> ones.

## Minor

- `remember(here)` was inserted between a comment and the `best.first` that comment describes.
  **Fixed** — moved above the comment block.
- `if (budget.expired()) break` sat after `leaves +=`, so the new `continue` skipped it. Each dropped
  candidate had already paid for a `chooseCommittedTargets` + `simulate`, so the anytime contract
  stopped holding on a board with many inert candidates. **Fixed** — the loop now records
  conditionally and takes the cut unconditionally.
- `StateProgress.isInert` had no production caller. **Fixed** — removed; the test compares digests.
- `chooseCommittedTargets` computed `here` unconditionally, once per affordable candidate, before
  knowing whether the refinement loop would run at all. **Fixed** — now a `lazy` local.

## Verification

`:ai:test` run locally with `--rerun` (an earlier `--rerun-tasks` attempt was killed by the shared
box's OOM mid-compile; the successful run reused compiled classes). Green before and after, including
`FrozenBaselineTest`. The new tier sweep was confirmed to fail with the fix backed out.

PR #1664, opened 2026-08-04
Route each target requirement to a UI that can collect it

bolav:
Route each target requirement of a `ChooseTargetsDecision` to a UI that can actually collect it,
instead of picking one UI for the whole decision.

## The problem

`DecisionUI` decided once per decision: if every legal target in `legalTargets[0]` sat in a
graveyard or exile pile, the whole decision went to `GraveyardTargetingUI`; otherwise all of it went
to `BattlefieldTargetingUI`. Both assumptions break on a decision that mixes zones.

The Spot, Living Portal's ETB is exactly that — "exile up to one target nonland permanent **and** up
to one target nonland permanent card from a graveyard". One slot is a board click, the other needs a
pile picker, and neither UI could serve both:

- routed to the board path, the graveyard card had no clickable target at all;
- routed to the pile path, `GraveyardTargetingUI` hard-coded `submitTargetsDecision({ 0: … })` and
  would have silently dropped every later slot.

## The change

A new `ChooseTargetsUI` owns the walk — `currentReqIndex`, `collectedTargets`, Back — and asks, per
requirement, which UI can collect that slot. `BattlefieldTargetingUI` and `GraveyardTargetingUI`
become per-requirement collectors taking `(requirementIndex, legalTargets, initialSelection,
onComplete, onBack)`; the parent submits every slot together once the last one is answered.

The routing predicate is `getPileTargetCards` in `utils/targeting.ts` — extracted rather than left
inline because the client has no component-test setup, so a pure function is the only thing this
repo can test.

Riding along: `key={pendingDecision.id}` on `ChooseTargetsUI`. Previously two consecutive decisions
reusing the mounted `BattlefieldTargetingUI` kept a stale `currentReqIndex`.

## Review fixes (second commit)

Both were found by reviewing the first commit, and both are visible on The Spot specifically:

- **The pile slot was labeled "Put onto Battlefield" for a card it exiles.** The action wording
  sniffed `effectHint` for `"battlefield"` before `"exile"`, but `ExileUntilLeavesEffect` renders
  `"Exile … until this permanent leaves the battlefield"` and `CompositeEffect` joins The Spot's two
  of them. Extracted the derivation to `derivePileAction` and discounted that boilerplate before the
  sniff. Genuine reanimation always names the battlefield as a *destination* ("onto/to the
  battlefield"), so blink and return-from-exile effects still classify correctly.
- **Back into a pile slot dropped its confirmed picks.** `ChooseTargetsUI` kept `restoredSelection`
  but handed it only to `BattlefieldTargetingUI`. Threaded `initialSelection` into
  `GraveyardTargetingUI` and both its paths, which meant a new optional `initialSelection` prop on
  the shared `ZoneSelectionUI` (defaults to `[]`, so library search and the other callers are
  unaffected).

## Known gap, recorded not fixed

`effectHint` describes the effect as a whole, while this UI now collects one requirement of it. Both
of The Spot's slots exile, so one verb still fits; a composite whose slots take different verbs
("destroy target creature **and** return target card from a graveyard to your hand") would mislabel
the pile slot. The durable fix is an action hint on `TargetRequirementInfo` rather than sniffing
prose — a server-side change, out of scope here. Left as a `TODO(per-requirement action hints)` in
`GraveyardTargetingUI` and noted in `docs/web-client-architecture.md`.

## Scope

Client only. No SDK types, no engine changes — The Spot itself is already implemented
(`mtg-sets/.../spm/cards/TheSpotLivingPortal.kt`) and covered by
`TheSpotLivingPortalScenarioTest`, which passes today; the engine was never the problem, the UI
just couldn't collect what the engine asked for.

## Testing

- `npx tsc --noEmit` — clean (run with `tsconfig.tsbuildinfo` removed, so no incremental caching).
- `targeting.test.ts` covers `getPileTargetCards` (8 cases: graveyard, exile, mixed piles, a
  battlefield target, empty set, missing card, null zone, and the two-slot Spot routing) and
  `derivePileAction` (7 cases, including The Spot's real composed hint string and a blink whose
  text contains both "to the battlefield" and "leaves the battlefield").
- **`npm test` was not run** — the dev container's `node_modules` carries a macOS rolldown binding
  (`@rolldown/binding-darwin-arm64`) so vitest can't start there. The `derivePileAction` cases were
  verified by running the function body directly under node; please let CI confirm the suite.
- **No screenshot** — the firewalled container can't run the stack, and `web-client/AGENTS.md` asks
  for one on user-visible changes. Happy to add a capture of The Spot's two-step flow (board slot →
  pile slot, with the corrected "Exile" wording) before merge if you want it.

## Follow-ups not taken

- An `e2e-scenarios` spec for the two-slot walk (`tests/portal/gravedigger.spec.ts` is the closest
  precedent). Nothing automated exercises the walk itself today — it would have caught the
  mislabeling above. `manual-scenarios/cards/t/the-spot-living-portal.json` is committed as a
  manual aid, not coverage.
- `ZoneSelectionUI` has one secondary-button slot, so Back displaces Cancel on a pile slot while the
  board banner renders both. A cancellable cast can therefore be aborted from a board slot but not
  from a pile slot at the same index.

### DIFF
--- a/docs/web-client-architecture.md
+++ b/docs/web-client-architecture.md
@@ -455,7 +455,15 @@ isn't clickable card-by-card on the board), everything else to `BattlefieldTarge
 highlights board objects through `decisionSelectionState`. The two can mix inside one decision —
 The Spot, Living Portal exiles "up to one target nonland permanent **and** up to one target
 nonland permanent card from a graveyard" — so the choice must not be made once for the whole
-decision. The cast-time equivalent is `TargetingOverlay`'s `ZoneCardTargetingOverlay` routing.
+decision. Both collectors take the restored picks as `initialSelection`, so Back keeps its
+confirmed selection whichever UI owns the slot. The cast-time equivalent is `TargetingOverlay`'s
+`ZoneCardTargetingOverlay` routing.
+
+The wording on a pile slot ("Exile" vs "Put onto Battlefield" vs "Shuffle into Library") comes
+from `derivePileAction` in `utils/targeting.ts`, which sniffs the decision's `effectHint` prose.
+That hint describes the effect as a whole rather than the individual requirement, so a composite
+whose slots take different verbs would mislabel one of them; the durable fix is a per-requirement
+action hint on `TargetRequirementInfo`.
 
 ## Type Mapping
 
--- a/docs/web-client-architecture.md
+++ b/docs/web-client-architecture.md
@@ -446,8 +446,16 @@ Confirming a step snapshots the outgoing `TargetingState` onto
 stack so the player can revise an already-confirmed target before the action is submitted —
 the restored step keeps its confirmed picks selected, and re-confirming recomputes later
 steps' valid-target pools against the revised selection. The resolution-time
-`ChooseTargetsDecision` flow (`BattlefieldTargetingUI`) implements the same back navigation
-over its local requirement index. Cancel still aborts the whole action.
+`ChooseTargetsDecision` flow (`ChooseTargetsUI`) implements the same back navigation over its
+requirement index. Cancel still aborts the whole action.
+
+`ChooseTargetsUI` also decides, **per requirement**, which UI collects that slot: a requirement
+whose legal targets all live in a graveyard or exile pile goes to `GraveyardTargetingUI` (a pile
+isn't clickable card-by-card on the board), everything else to `BattlefieldTargetingUI`, which
+highlights board objects through `decisionSelectionState`. The two can mix inside one decision —
+The Spot, Living Portal exiles "up to one target nonland permanent **and** up to one target
+nonland permanent card from a graveyard" — so the choice must not be made once for the whole
+decision. The cast-time equivalent is `TargetingOverlay`'s `ZoneCardTargetingOverlay` routing.
 
 ## Type Mapping
 
--- a/manual-scenarios/cards/t/the-spot-living-portal.json
+++ b/manual-scenarios/cards/t/the-spot-living-portal.json
@@ -1,0 +1,58 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": [
+      "The Spot, Living Portal",
+      "Murder"
+    ],
+    "battlefield": [
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"}
+    ],
+    "graveyard": [
+      "Savannah Lions",
+      "Lightning Bolt"
+    ],
+    "library": [
+      "Plains",
+      "Swamp",
+      "Plains",
+      "Swamp"
+    ]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Grizzly Bears"},
+      {"name": "Icy Manipulator"},
+      {"name": "Forest"},
+      {"name": "Forest"}
+    ],
+    "graveyard": [
+      "Serra Angel",
+      "Disenchant"
+    ],
+    "library": [
+      "Forest",
+      "Forest",
+      "Forest"
+    ]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "SELF"
+}
--- a/web-client/src/components/decisions/BattlefieldTargetingUI.tsx
+++ b/web-client/src/components/decisions/BattlefieldTargetingUI.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useRef } from 'react'
+import { useState, useEffect } from 'react'
 import { useGameStore } from '@/store/gameStore.ts'
 import type { DecisionSelectionState } from '@/store/slices'
 import type { EntityId, ChooseTargetsDecision } from '@/types'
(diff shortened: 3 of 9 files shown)


bolav:
## Code review — per-requirement target routing

Reviewed on the branch with `origin/main` (`ac4c2e5738`) already the direct parent, so nothing to
merge and no conflicts. Findings 1–3 were fixed in the follow-up commit on this branch and are
recorded below with what changed; 4–6 are open.

## Verdict

The structural move is right and overdue: requirement-walking belongs in a parent that routes
**each** slot to a UI that can collect it, and `BattlefieldTargetingUI` / `GraveyardTargetingUI`
become dumb per-requirement collectors. No SDK types added — a pure client fix, correctly scoped.
The one blocking defect was that the card motivating the change rendered the wrong verb in the new
path; that is now fixed.

## What's good

- **The decomposition itself.** `ChooseTargetsUI` owning `currentReqIndex` / `collectedTargets` /
  Back, with both collectors reduced to `(requirementIndex, legalTargets, onComplete, onBack)`, is
  the correct shape. The old code couldn't express "slot 0 on the board, slot 1 in a pile" at all —
  the graveyard path hard-coded `submitTargetsDecision({ 0: selectedCards })` and would have
  silently dropped every later slot.
- **`key={pendingDecision.id}` on `ChooseTargetsUI`** is a real bug fix riding along: previously two
  consecutive decisions reusing the mounted `BattlefieldTargetingUI` kept a stale `currentReqIndex`.
- **Extracting `getPileTargetCards` into `utils/targeting.ts`.** The repo has no
  jsdom/testing-library — every `*.test.ts` is a pure function — so pulling the routing predicate
  out of the component is exactly how you make it testable here. Eight focused cases including the
  null-zone and missing-card fallbacks.
- Doc comments explain *why* (pile-not-clickable, all-or-nothing on purpose) rather than restating
  the code, and `docs/web-client-architecture.md` was updated in the same change.
- Strict-TS discipline: the `{...(onBack ? … : …)}` spreads are the right idiom under
  `exactOptionalPropertyTypes`.

## Blocking

**1. The Spot's graveyard slot was labeled "Put onto Battlefield" — FIXED.**
`isReanimate = effectHint.includes('battlefield')` was tested before `isExile`, and
`ExileUntilLeavesEffect.description` is `"Exile … until this permanent leaves the battlefield"`
(`mtg-sdk/.../RemovalEffects.kt:263`). `CompositeEffect` joins both halves with `". "`
(`CompositeEffects.kt:42`), so The Spot's `effectHint` was:

> "Exile up to one target nonland permanent until this permanent **leaves the battlefield**. Exile
> up to one target nonland permanent card from a graveyard until this permanent **leaves the
> battlefield**"

→ reanimation won → button "Put onto Battlefield", helper "Optional: choose up to one card to put
onto the battlefield, or decline." The UI promised reanimation and the engine exiled. Pre-PR this
path was unreachable for The Spot, so the branch introduced the visible defect.

A plain reorder would have been wrong — `"Return one of the exiled cards you own to the
battlefield"` (`RemovalEffects.kt:466`) contains both words and really is reanimation. Fixed by
discounting the boilerplate before the sniff, in a `derivePileAction` helper extracted next to
`getPileTargetCards` so it is testable the same way, with the real composed hint string pinned in
`targeting.test.ts`.

## Important

**2. Stepping Back into a pile requirement lost the picks — FIXED.**
The parent maintained `restoredSelection` and threaded it to `BattlefieldTargetingUI` as
`initialSelection`, but the `GraveyardTargetingUI` branch never received it, and that component's
`selectedCards` always initialized to `[]`. A `[pile, board]` decision lost slot 0's cards on Back,
while `[board, pile]` (The Spot's ordering) restored fine — half the mechanism was wired, and
`docs/web-client-architecture.md` claimed parity. Fixed by threading `initialSelection` to both pile
paths, which required a new optional prop of that name on the shared `ZoneSelectionUI`; the doc now
states the parity it actually has.

**3. One whole-effect `effectHint` drives per-requirement wording — RECORDED.**
The verb derivation reads `decision.context.effectHint`, which `TriggerProcessor` builds from
`ability.effect.runtimeDescription` for the *entire* composite, while `targetReq.description` is
per-slot. The Spot survives only because both slots exile. A composite mixing verbs — "destroy
target creature and return target card from a graveyard to your hand" — would label the pile slot
wrong, and this branch is precisely what makes mixed-verb decisions reachable. Not fixable client
side; now a `TODO(per-requirement action hints)` in `GraveyardTargetingUI` plus a note in the
architecture doc. The durable fix is an action hint on `TargetRequirementInfo`.

**4. Nothing automated exercises the walk — OPEN.**
`targeting.test.ts` covers the extracted predicates well, but the actual value of the change — walk
two slots, route each to a different UI, submit both together — has no test, and can't have a
component test in this repo. `e2e-scenarios/` is the right home; `tests/portal/gravedigger.spec.ts`
already drives the pile picker. A `the-spot-living-portal.spec.ts` clicking a board permanent for
slot 0 and a graveyard card for slot 1, asserting both land in exile, would cover it — and would
have caught finding 1. The committed `manual-scenarios/cards/t/the-spot-living-portal.json` matches
the 988-file convention and is a fine manual aid, but it isn't coverage.

## Minor

**5. Cancel is unreachable mid-walk in the pile path — OPEN.** `ZoneSelectionUI` has one secondary
slot, so Back displaces Cancel; `BattlefieldTargetingUI` renders both. The comment acknowledges it,
but the asymmetry means a cancellable cast can be aborted from a board slot and not from a pile slot
at the same index. A third button on `ZoneSelectionUI` would settle it.

**6. Asymmetric `key` — OPEN.** `GraveyardTargetingUI` gets `key={currentReqIndex}`,
`BattlefieldTargetingUI` doesn't (it relies on the `[decision.id, requirementIndex]` effect).
Correct today, but any future local state in the board banner would leak across requirements.
Adding the same key costs nothing and removes the trap.

## Verification

- **`npx tsc --noEmit`: clean**, run with `tsconfig.tsbuildinfo` removed so incremental caching
  couldn't mask anything.
- **`vitest` could not run** in the dev container: `node_modules/@rolldown/binding-darwin-arm64` is
  a macOS binary on Linux, so the runner can't start. The `derivePileAction` cases were instead
  executed against the real function body under node (source copied verbatim, only the aliased
  import stubbed) — all seven pass, The Spot yields `Exile`. `getPileTargetCards` was reviewed by
  inspection. CI should be treated as the authority on the suite.

PR #1665, opened 2026-08-05
Teach the AI what a Room's two halves actually do

bolav:
The AI never cast `Unholy Annex` and never paid to unlock a door.

Both halves of a Room (CR 709.5) carry their rules text on `CardFace.script` and leave the card's
top-level script empty, and `CardIntentAnalyzer` read only that top level — so a repeatable draw
engine came back as `CardIntent.UNKNOWN` and was priced at the flat 0.5 every unreadable permanent
gets. Casting it then cost more in `CardAdvantage` than the board it bought was worth, so passing
won every time.

Unlocking had a second cause stacked on that: `BoardPresence` looked the prior up by card *name*,
and a name does not change when a door opens. Unlocking `Ritual Chamber` did happen — the 6/6 Demon
token it makes is visible on its own — but unlocking `Unholy Annex` moved no evaluated number at
all, and a special action that changes nothing can never beat passing.

## What this does

A multi-face card makes "what can this card do?" and "what is this permanent doing?" different
questions. The analyzer now answers them separately, and callers pick deliberately:

| Reading | Scope | Used for |
|---|---|---|
| `analyze` | every face unioned | a card in hand or on the stack |
| `analyzeSelf` | top-level script only | an ordinary permanent |
| `analyzeFace` | one face | a Room, per unlocked door |

`IntentCatalog.forPermanent(container, cardName)` is the seam that picks between the last two and
the only place that knows a Room from anything else. `BoardPresence.permanentValue` and
`RawBoardFeatures.threatsInPlay` both go through it; `TargetSelection.rank` inherits it by already
routing through `permanentValue`. The hand-side consumers (`HoldPolicy`,
`PlayoutPolicy.priorityScore`, `removalInHand`) keep `forName`, because the union is the right
reading when the decision is whether to *cast* the card.

Two behaviours follow from getting that split right:

- **A Room is worth its open doors.** A locked half's rules text does not exist, so a Room permanent
  is read per unlocked face and opening a door is worth exactly what is behind it. A Room put onto
  the battlefield with both halves locked (CR 709.5d) contributes nothing, correctly.
- **A spent Adventure does not inflate the permanent it left behind.** `Virtue of Loyalty` is an
  enchantment whose Adventure half makes a 2/2 Knight. Unioning faces tagged the enchantment
  `TOKEN_MAKER` + `repeatable` — 3.0 on the `priorValueOf` ladder, where its own end-step trigger is
  worth 1.5. The Adventure resolves to exile and is never text on the battlefield. The same holds
  for Omen and modal-DFC backs: in every layout the SDK has, the face that can sit on the
  battlefield is the top-level one.

Smaller points:

- **`"when you unlock this door"` is not a repeatable engine.** It fires on the unlock transition
  (CR 709.5h), like an ETB. A door *can* be re-locked and unlocked again (CR 709.5g, `LockDoorEffect`),
  so this is a common-case claim rather than an invariant — and it is the right direction to err in:
  under-reading a rare re-unlock costs the AI nothing it had, while reading every open door as
  repeatable value is the bug worth avoiding.
- **The unreadable floor no longer doubles.** Per-face priors combine as a baseline plus what each
  face adds over it, not a raw sum. `UNKNOWN.staticPriorValue` is what an unreadable permanent is
  worth, and a Room with two blank halves is one unreadable permanent, not two.
- **Three readings, three caches**, instead of one map keyed by a synthesized `"<card>//<face>"`
  string — no naming convention has to hold for them to stay apart.

## Tests

`RoomAiTest` pins all three decisions on the reported card — cast the Annex half, unlock
`Ritual Chamber` from the Annex, unlock the Annex from `Ritual Chamber` — and fails on the parent
commit for the cast and the Annex unlock. (The `Ritual Chamber` unlock passed before by accident,
via the Demon token being visible on its own.)

`CardIntentAnalyzerTest` covers the three readings directly: a Room's faces valued independently in
both polarities (Annex repeatable, Chamber not), `analyzeSelf` vs `analyze` on `Virtue of Loyalty`,
single-face equivalence, and `forPermanent` across a plain permanent, one open door, two open doors,
both locked, and the catalog-off position.

## Verification

`:ai:test` — **357 tests, 0 failures, 11 skipped**, run with `--rerun` (a plain run reports
`UP-TO-DATE` and executes nothing).

With card intent off (`AiProfile.LEGACY_V0`, the frozen arena baseline) every path here still
answers 0.5 exactly as before, so no published number rebases — `PuzzleSuiteTest > the AI solves
every puzzle outside KNOWN_FAILURES` and `a zero-weight agent solves strictly fewer puzzles` both
still pass.

Scope is the `ai` module plus a note in `docs/ai/architecture.md` recording the card-vs-permanent
rule. No SDK, engine, server or client contract changes; nothing outside `ai` references
`IntentCatalog` or `CardIntentAnalyzer`.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/evaluation/BoardFeatures.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/evaluation/BoardFeatures.kt
@@ -16,6 +16,7 @@ import com.wingedsheep.engine.state.components.battlefield.SummoningSicknessComp
 import com.wingedsheep.engine.state.components.battlefield.TappedComponent
 import com.wingedsheep.engine.state.components.combat.AttackingComponent
 import com.wingedsheep.engine.state.components.identity.CardComponent
+import com.wingedsheep.engine.state.components.identity.RoomComponent
 import com.wingedsheep.sdk.core.CounterType
 import com.wingedsheep.sdk.core.Keyword
 import com.wingedsheep.sdk.core.Zone
@@ -136,7 +137,7 @@ object BoardPresence : BoardFeature {
         // max keeps both, and guarantees no permanent is ever valued lower than it was before
         // Phase 6 — so a position the AI played correctly on the old numbers still scores at least
         // as well.
-        val prior = intents.forName(card.name)?.let { intentValue(projected, entityId, it) } ?: 0.0
+        val prior = priorValue(projected, entityId, container, card, intents)
 
         // Planeswalkers: the flat 4.0 priced a fresh Jace and a Jace at 1 loyalty identically.
         // Loyalty is both the walker's life total and its remaining activations, so it is the one
@@ -160,6 +161,30 @@ object BoardPresence : BoardFeature {
         // text — a signet, an Oblivion Ring and a Bitterblossom scoring the same number is the
         // blindness Phase 6 exists to remove.
         return maxOf(0.5, prior)
+    }
+
+    /**
+     * The prior for the permanent [entityId] *as it currently stands*.
+     *
+     * For everything except a Room that is the card's own [CardIntent]. A Room (CR 709.5) is one
+     * permanent with a door per half, and a locked half's rules text does not exist — so its value
+     * is the sum of what its **unlocked** faces do, and unlocking a door is what raises it. Reading
+     * the card as a whole would price a Room the same however many doors are open, which is exactly
+     * why the AI would cast a half and then never pay to unlock the other: a special action that
+     * moves no evaluated number can never beat passing.
+     */
+    private fun priorValue(
+        projected: ProjectedState,
+        entityId: EntityId,
+        container: com.wingedsheep.engine.state.ComponentContainer,
+        card: CardComponent,
+        intents: IntentCatalog,
+    ): Double {
+        val room = container.get<RoomComponent>()
+            ?: return intents.forName(card.name)?.let { intentValue(projected, entityId, it) } ?: 0.0
+        return room.unlockedFaces.sumOf { face ->
+            intents.forFace(card.name, face.name)?.let { intentValue(projected, entityId, it) } ?: 0.0
+        }
     }
 
     /**
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/evaluation/BoardFeatures.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/evaluation/BoardFeatures.kt
@@ -8,6 +8,7 @@ import com.wingedsheep.ai.engine.knowledge.CardIntent
 import com.wingedsheep.ai.engine.knowledge.IntentCatalog
 import com.wingedsheep.ai.engine.lifePoolsOf
 import com.wingedsheep.ai.engine.sidesFor
+import com.wingedsheep.engine.state.ComponentContainer
 import com.wingedsheep.engine.state.GameState
 import com.wingedsheep.sdk.scripting.Duration
 import com.wingedsheep.engine.state.components.battlefield.CountersComponent
@@ -16,7 +17,6 @@ import com.wingedsheep.engine.state.components.battlefield.SummoningSicknessComp
 import com.wingedsheep.engine.state.components.battlefield.TappedComponent
 import com.wingedsheep.engine.state.components.combat.AttackingComponent
 import com.wingedsheep.engine.state.components.identity.CardComponent
-import com.wingedsheep.engine.state.components.identity.RoomComponent
 import com.wingedsheep.sdk.core.CounterType
 import com.wingedsheep.sdk.core.Keyword
 import com.wingedsheep.sdk.core.Zone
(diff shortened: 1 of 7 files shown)


bolav:
## Review — `ai-unholy-annex-rooms`

Reviewed `32250bf6c9` ("Teach the AI what a Room's two halves actually do"). All findings below were
addressed in the follow-up commit `48314bbc14` on the same branch, so they are recorded here as the
reasoning behind that commit rather than as open requests.

### Verdict

The diagnosis is right and the fix is the right shape: `analyze` (union of all faces, "does this
card do X *anywhere*") vs `analyzeFace` (one face, "what is functioning on this permanent *now*") is
a genuine semantic split, not a Room special case bolted on. Behaviour is correct and the arena
baseline is untouched. Two follow-ups: the new union reading leaked into a sibling feature extractor
that wasn't updated, and summing per-face priors quietly doubled the "unreadable permanent" floor.

### What's good

- **The rule model is accurate.** Every citation checks out against `MagicCompRules_20260619.txt`:
  709.5 *is* the shared-type-line Room rule (the two "as long as this permanent doesn't have the …
  unlocked designation, it doesn't have the … rules text" statics), 709.5e the unlock special
  action, 709 split, 712 double-faced, 715 adventurer. The claim in `analyzeFace`'s doc — that a
  face is analysed against *the card's* type line "because it is the card that is (or is not) a
  permanent" — is exactly what 709.5a says ("each half … shares the types and subtypes listed on
  that card's shared type line").
- **The seam is in the right place.** `compute(card, scripts)` is a mechanical widening — single-face
  cards get a one-element list and provably identical results. `analyzeFace` deliberately does *not*
  fold `card.script` in (unlike `RoomFaceStatics`), which is what stops a two-door Room
  double-counting the top level. That's the subtle call and it was made correctly.
- **`TargetSelection.rank` gets Room-awareness for free** because it already routes through
  `BoardPresence.permanentValue`. Composition paying rent.
- **Non-Room splits are a real side win.** `Assault // Battery`, `Spite // Malice` et al. carried
  their entire text on faces and read as blank; they now tag correctly for hand-side decisions
  without touching board valuation, since neither half is ever a permanent.
- **The tests earn their place.** Three-way `RoomAiTest` pins cast plus both unlock directions, and
  the *third* case (unlock Annex while Chamber is up) is the one a name-keyed prior could never
  pass — the token from `Ritual Chamber` made the second case pass by accident before. The analyzer
  test asserts both polarities of the door heuristic (Annex repeatable, Chamber *not*), so the
  `DoorUnlockedEvent -> false` branch can't rot silently.
- **The legacy-frozen claim holds.** With `IntentCatalog.NONE`, `forFace` returns null → sum `0.0` →
  `maxOf(0.5, 0.0)`, identical to before.

### Issues

#### Important

**`RawBoardFeatures.kt:163` — `threatsInPlay` still read the whole-card union on battlefield
permanents.** `intents.forName(it.name)?.tags?.any { it in THREAT_TAGS }` saw `TOKEN_MAKER` from a
*locked* `Ritual Chamber`, so a Room with one door shut counted as a threat that doesn't exist — the
precise error the commit fixed in `BoardPresence`, made worse by the union. It feeds
`RawEvaluationWeights.evaluate` and the `ApprenticeArtifact` feature contract, so it reaches the ECL
profiles and any training corpus.
*Fixed:* routed through the new `IntentCatalog.forPermanent`, the same reading `BoardPresence` uses.

**`BoardFeatures.kt:176-188` — the face-aware reading was Room-only, but the union is card-wide.**
`priorValue` special-cased `RoomComponent` and everything else fell back to `forName`, which after
the change included other faces' text. I first read this as latent; it is **live**.
`VirtueOfLoyalty` (WOE) is an *enchantment* whose Adventure half makes a 2/2 Knight, so the union
tagged the enchantment `TOKEN_MAKER` + `repeatable` — 3.0 on the `priorValueOf` ladder, where its own
end-step trigger is worth 1.5. The Adventure resolves to exile and is never on the battlefield.
*Fixed:* added `analyzeSelf` (top-level script only) and made `forPermanent` pick between it and
`analyzeFace`. Checked `CastSpellEnumerator` to confirm the general rule — for Adventure / Omen /
modal DFC the secondary face is always a *spell*, so the permanent is always the top-level face —
which makes this a complete fix rather than a mitigation.

#### Minor

**`BoardFeatures.kt:185-187` — summing faces doubled the unreadable floor.** `staticPriorValue` is
documented as "`0.5` is still the value of a card the analyzer cannot read", but a Room with both
doors open and both faces uninterpretable summed to `1.0`, outranking every other unreadable
permanent for no reason. Separately it mixed calibration philosophies: `priorValueOf` is explicitly
a highest-claim-wins ladder (`candidates.maxOrNull()`).
*Fixed:* the readings now combine as `UNKNOWN + Σ(faceValue − UNKNOWN)`. Single-face cards are
algebraically unchanged.

**`CardIntentAnalyzer.kt:388-396` — the comment overstated "once per door".** `LockDoorEffect` exists
in this codebase (CR 709.5g), so a door can be locked and re-unlocked and the trigger genuinely can
fire again. The heuristic is the right call; the doc asserted an invariant that isn't one.
*Fixed:* reworded as a common-case claim, with the reasoning for why under-reading is the safe
direction, and the citation corrected to 709.5h (abilities that trigger on being given an unlocked
designation).

**`CardIntentAnalyzer.kt:72,86` — face keys shared the whole-card cache map.**
`"${card.name}//${face.name}"` could only avoid collision because real split names use spaced
`" // "` — load-bearing on a printing convention.
*Fixed:* three caches, the face one keyed by a `FaceKey` data class.

**`RoomAiTest.kt:46-49` — `settle` gave up silently.** After 10 rounds it returned with a non-empty
stack and the test would fail somewhere downstream with a confusing message.
*Fixed:* asserts the guard didn't trip.

### Verification notes

- `just test-ai` first reported **everything `UP-TO-DATE` — nothing ran**. Re-run as
  `scripts/gradle-locked :ai:test --rerun` to force execution. Worth knowing if you read a green
  `just test-ai` as proof.
- Post-fix: **357 tests, 0 failures, 11 skipped**; `CardIntentAnalyzerTest` 18/18, `RoomAiTest` 3/3,
  `PuzzleSuiteTest` green so no arena rebase.
- `origin/main` (`2c837352be`) was already an ancestor of `HEAD`, so no merge was needed and none was
  created.
- I did **not** verify the claim that the tests fail on the parent commit — confirming it needs a
  checkout of `main`, and this is a shared working tree. The reasoning is sound for the cast and the
  Annex-unlock cases; the `Ritual Chamber` unlock would have passed before via the Demon token,
  which the commit message already acknowledges.

PR #1666, opened 2026-08-05
Stop "you may discard a card" paying out on an empty hand

bolav:
"You may discard a card. When you do, [payoff]" was paying out on an empty hand.

A discard compiles to a Gather → Select → Move pipeline. With nothing in hand the
gather collects nothing, `SelectFromCollection` auto-selects nothing, and the
pipeline reports **success** — indistinguishable, to the reflexive executor, from a
real discard. So Inti, Seneschal of the Sun handed out its +1/+1 counter and trample
for a discard that never happened, and the attack trigger asked an unanswerable
yes/no every single combat.

## What this does

**Teaches `ReflexiveTriggerEffectExecutor.isActionFeasible` to read the discard
pipeline.** It already recognised `SelectTargetEffect`, `SacrificeEffect`,
`ChooseActionEffect` and `PayFixedCountersEffect`; it now also scores a
`SelectFromCollection` carrying a minimum against the size of the collection the
preceding `GatherCards` will produce. An impossible action never happens, so
CR 603.12's "when you do" never triggers — the whole effect is skipped.

This is deliberately at the engine layer rather than a per-card `feasibility`
declaration, so it fixes every reflexive discard at once: Inti, Shriek Treblemaker,
Awaken the Honored Dead, Gigapede.

**Rebuilds Cool but Rude out of the right primitives.** Level 1 is "you may discard
a card. **If you do**, draw a card" — the draw hangs off the *action outcome*, not
off the yes/no answer, so it becomes `MayEffect(IfYouDo(discard, draw))`. Paired with
the existing `feasibility = HasCardsInZone(Zone.HAND)` so the trigger stops asking
when the hand is empty. No new SDK types; both primitives already existed.

## Conservative in both directions

The walker only ever removes an *impossible* prompt, never a possible one. Three
deliberate fail-open gaps, each with a test:

- **Minimum-less selection modes** (`ChooseUpTo`, `All`, `ChooseAnyNumber`) stay
  feasible — discarding zero cards genuinely performs "discard any number of cards"
  (Miasma Demon), and the reflexive payoff then targets up to zero creatures.
- **A non-empty collection smaller than the requested count** stays feasible.
  `SelectFromCollectionExecutor` clamps the count to the collection size, so with one
  card in hand "discard two cards" discards it and succeeds.
- **Gather sizes are read off pre-action state**, so bookkeeping stops at the first
  step that is neither a gather nor a select — and stays stopped inside nested
  composites. `Effect.then` only flattens when its receiver is already a composite,
  so `DrawCards(1).then(Discard(1))` is `Composite[Draw, Composite[Gather, Select,
  Move]]`; a per-call flag would let the inner walk restart and judge the discard
  against the pre-draw hand.

Unknown gather sources (target-driven, multi-player) and unknown collection names
also fail open.

The gate covers the **mandatory** template too (`optional = false`), not just the
"you may" one — there is no prompt to skip there, but a vacuous action must not pay
out either.

## Tests

- `ReflexiveTriggerFeasibilityTest` — engine-level, named for the mechanic: empty-hand
  suppression, the mandatory path, draw-then-discard staying offered, and
  discard-two-with-one-card paying out.
- `CoolButRudeScenarioTest` — discard-and-draw, decline, and empty hand.
- `IntiSeneschalOfTheSunScenarioTest` — empty hand yields no counter, no trample, no
  impulse exile, and no decision at all.
- `AwakenTheHonoredDeadScenarioTest` — chapter III on an empty hand returns nothing,
  even though chapter II milled legal targets.

Full `:rules-engine:test` + `:mtg-sets:test`: **9850 tests, 0 failures, 0 errors.**

`docs/card-sdk-language-reference.md` is updated with the feasibility contract and
both fail-open gaps.

https://claude.ai/code/session_01At5SgAMrk4ozjg4zhi9ghv

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1962,6 +1962,18 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
   round-trip via `PendingTrigger.carriedPipeline`/`TriggeredAbilityOnStackComponent.carriedPipeline`.
   This differs from the cast-time path (`TargetValidator.effectiveMaxCount`), where the pipeline isn't
   live yet — pipeline-linked caps only work on reflexive/resolution-time targets.
+  With `optional = true`, `ReflexiveTriggerEffectExecutor.isActionFeasible` decides whether the "may
+  [action]?" question is worth asking at all: an action that can't be performed is skipped silently,
+  because answering yes would no-op the action while still firing the reflexive payoff. Alongside
+  `SelectTargetEffect` / `SacrificeEffect` / `ChooseActionEffect` / `PayFixedCountersEffect`, it scores
+  the **Gather → Select → Move pipeline** every discard pattern compiles to — a `SelectFromCollection`
+  whose `SelectionMode` carries a minimum (`ChooseExactly`, `Random`) against the size of the collection
+  the preceding `GatherCards` will produce. So `ReflexiveTriggerEffect(action = Effects.Discard(1),
+  optional = true, …)` on an empty hand never prompts and never pays out (Inti, Seneschal of the Sun).
+  Minimum-less modes (`ChooseUpTo`, `All`, `ChooseAnyNumber`) stay feasible — discarding zero cards is a
+  legal way to perform "discard your hand" (Vaultguard Trooper). Gather sizes are read off the
+  pre-action state, so the bookkeeping stops at the first composite step that is neither a gather nor a
+  select; later selections then fail open rather than being judged against a stale count.
 - **Branching on gathered properties** — "reveal/look, if it's a [type] do X, otherwise Y" needs no
   bespoke effect type; it is the partition + collection-gate composition:
   1. **Partition:** `FilterCollection(from, CollectionFilter.MatchesFilter(filter), storeMatching,
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1962,18 +1962,26 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
   round-trip via `PendingTrigger.carriedPipeline`/`TriggeredAbilityOnStackComponent.carriedPipeline`.
   This differs from the cast-time path (`TargetValidator.effectiveMaxCount`), where the pipeline isn't
   live yet — pipeline-linked caps only work on reflexive/resolution-time targets.
-  With `optional = true`, `ReflexiveTriggerEffectExecutor.isActionFeasible` decides whether the "may
-  [action]?" question is worth asking at all: an action that can't be performed is skipped silently,
-  because answering yes would no-op the action while still firing the reflexive payoff. Alongside
+  `ReflexiveTriggerEffectExecutor.isActionFeasible` decides up front whether the action can happen at
+  all; when it can't, the whole effect is skipped silently. An impossible action never happens, so
+  CR 603.12's "when you do" never triggers — for `optional = true` that means the "may [action]?"
+  question isn't worth asking (answering yes would no-op the action while still firing the payoff),
+  and for `optional = false` it means a vacuous action must not pay out either. Alongside
   `SelectTargetEffect` / `SacrificeEffect` / `ChooseActionEffect` / `PayFixedCountersEffect`, it scores
-  the **Gather → Select → Move pipeline** every discard pattern compiles to — a `SelectFromCollection`
-  whose `SelectionMode` carries a minimum (`ChooseExactly`, `Random`) against the size of the collection
-  the preceding `GatherCards` will produce. So `ReflexiveTriggerEffect(action = Effects.Discard(1),
-  optional = true, …)` on an empty hand never prompts and never pays out (Inti, Seneschal of the Sun).
-  Minimum-less modes (`ChooseUpTo`, `All`, `ChooseAnyNumber`) stay feasible — discarding zero cards is a
-  legal way to perform "discard your hand" (Vaultguard Trooper). Gather sizes are read off the
-  pre-action state, so the bookkeeping stops at the first composite step that is neither a gather nor a
-  select; later selections then fail open rather than being judged against a stale count.
+  the **Gather → Select → Move pipeline** that `Effects.Discard` and the counted `Patterns.Hand`
+  discards compile to — a `SelectFromCollection` whose `SelectionMode` carries a minimum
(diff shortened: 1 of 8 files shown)


bolav:
## Review of `4524882e5a` — "Stop 'you may discard a card' paying out on an empty hand"

Reviewed against `origin/main` (branch was level with it, no merge needed). Everything
below was found on the first commit; the follow-up commit `e29d6613c8` fixes all of it,
so this is the record of what was wrong rather than a list of outstanding work.

### Verdict

Behavior right for both cards, SDK shape right — no new types, and the engine-side fix
sits at the layer where it fixes every reflexive-discard card at once. One latent
correctness bug in the new composite bookkeeping contradicted the code's own documented
counter-example, and three citations (a CR number and a card example repeated in three
files) were wrong.

### What's good

- **Zero new SDK types.** Cool but Rude is rebuilt from primitives that already existed:
  `Effects.IfYouDo` (lowers to `Gate.DoAction`) and the pre-existing
  `MayEffect(feasibility = …)` / `Gate.MayDecide.feasibility`. The `TMT.json` re-bless
  confirms it compiles to `Gated`, not a bespoke node.
- **The engine fix is at the right layer.** Rather than bolting a `feasibility` field onto
  `ReflexiveTriggerEffect` and making every card declare it, `isActionFeasible` learned to
  *read* the Gather → Select → Move shape — fixing Inti, Shriek Treblemaker, Awaken the
  Honored Dead and Gigapede with no card edits.
- **Fail-open discipline throughout.** Unknown gather source, unknown collection name,
  minimum-less selection mode, nested composite — all fall back to "feasible".
- **`gatherableCount` faithfully mirrors `GatherCardsExecutor`** — same
  `predicateEvaluator.matches(state, state.projectedState, …)` call, same
  `PredicateContext.fromEffectContext`. Returning `null` for multi-player refs is correct:
  `TargetResolutionUtils.resolvePlayerRef` genuinely returns null for
  `Player.Each`/`EachOpponent`, so the fan-out case can't be silently under-counted.
- **The Inti test asserts the right things** — all three payoffs absent (counter, trample,
  impulse exile) *and* no decision offered at all.

### Issues

#### Important

**a) `trackable` didn't survive into nested composites, so the KDoc's own counter-example
failed.** `ReflexiveTriggerEffectExecutor.kt:224-244`

`Effect.then()` only flattens when the *receiver* is a composite (`Effect.kt:54-59`), so
`Effects.DrawCards(1).then(Effects.Discard(1))` builds
`Composite[Draw, Composite[Gather, Select, Move]]`. The outer walk set `trackable = false`
on the `Draw` step, but the recursive `isCompositeFeasible` call started a fresh
`var trackable = true`, so the inner `Gather` *was* scored — against the pre-draw hand. On
an empty hand the inner `Select(ChooseExactly(1))` saw `available = 0` and the whole action
was declared infeasible, the prompt never shown, even though the draw would have supplied
the card. That is precisely the case the KDoc claimed to handle ("so 'you may draw a card,
then discard a card' is still offered on an empty hand").

Fixed by making "bookkeeping stopped" part of the value threaded through the recursion —
`gathered: Map<String, Int>?` with `null` meaning stopped — so it propagates inward.

**b) Vaultguard Trooper was the wrong example, in three places.**
`ReflexiveTriggerEffectExecutor.kt:250`, `docs/card-sdk-language-reference.md`,
`CoolButRudeScenarioTest.kt:22-24`

Vaultguard Trooper is a `MayEffect` + `IfYouDoEffect` (`VaultguardTrooper.kt:43-51`), so it
never reaches `isActionFeasible` at all — and `Patterns.Hand.discardHand` is Gather → Move
with **no** `SelectFromCollectionEffect` (`HandPatterns.kt:426-441`), so it isn't an
instance of a minimum-less `SelectionMode` either. Wrong on both axes. The card that
genuinely exercises that branch is **Miasma Demon**: a real `ReflexiveTriggerEffect` whose
action is `Patterns.Hand.discardAnyNumber()` → `SelectionMode.ChooseAnyNumber`. The same doc
sentence's "the Gather → Select → Move pipeline every discard pattern compiles to" was an
overstatement for the same reason.

**c) Wrong CR number.** `CoolButRude.kt:40`

Cited "CR 701.8a — you can't discard a card you don't have". 701.8 is **Destroy**; discard is
**701.9** (701.9a: "To discard a card, move it from its owner's hand to that player's
graveyard"). 701.9 doesn't state the claim either — it follows from there being no card to
move — so the comment was reworded rather than just renumbered.

**d) `minimumSelection` was stricter than the executor it models.**
`ReflexiveTriggerEffectExecutor.kt:252-260`

`SelectFromCollectionExecutor` clamps: `val clamped = minOf(count, eligibleCards.size)`
(`SelectFromCollectionExecutor.kt:170`), so "discard two cards" with one card in hand
discards one and succeeds. It also honours `SelectionRestriction.ReducedMinimumIfMatches`
(line 530), which *lowers* the effective minimum — `minimumSelection` ignored `restrictions`
entirely. Every current card uses n = 1, where the two coincide, so nothing regressed, but
the general rule was wrong. Now `minimum == 0 || available > 0`.

#### Minor

**e) The "stale count" rationale didn't match the code.** The KDoc said later selections
"fail open rather than being judged against a stale count", but `known` was never cleared
when `trackable` flipped. Resolved by the `Map?` rewrite in (a).

**f) The mandatory path had the same hole.** `isActionFeasible` was consulted only when
`effect.optional`. An `optional = false` reflexive discard (The Ancient One, Talion's
Messenger) still emitted `ReflexiveAbilityTriggeredEvent` after a pipeline that auto-selected
nothing. The gate moved into `execute()` to cover both templates.

Worth flagging: this changes behavior for cards outside the original scope. Neither shipping
card can currently reach an empty hand at that point (Talion's draws first, The Ancient One's
payoff would mill 0), so nothing observable changes today — but the gate is now on their path.

**g) `Awaken the Honored Dead` had no scenario test.** Pre-existing, but it's one of the
cards whose behavior this change alters. Added.

### Verification

Full `:rules-engine:test` + `:mtg-sets:test` — **9850 tests, 0 failures, 0 errors**, both
tasks executed (not cached; all 3011 result files freshly written). The two new
`ReflexiveTriggerFeasibilityTest` cases for (a) and (d) fail on the pre-fix code.

Note on this box: two earlier runs died with "daemon disappeared" — the known shared-memory
OOM, environmental rather than a code failure. The green run used
`-Pkotlin.compiler.execution.strategy=in-process -Dorg.gradle.jvmargs="-Xmx3g …"`.

PR #1680, opened 2026-08-07
Add 119 Marvel Super Heroes cards

bolav:
Adds **119 Marvel Super Heroes cards** (MSH 84 → 203 of 276 implemented), the set's seven predefined
tokens, and 25 hand-playtest scenarios covering every new card.

## What's here

- **119 new card definitions** under `mtg-sets/.../definitions/msh/cards/`. Every card's
  `oracleText`, mana cost, type line, P/T, rarity, collector number, artist, and colour identity was
  diffed against the Scryfall MSH bulk data — all 119 match exactly.
- **Two reprints scaffolded at their earliest printing** rather than hosted in MSH: Thirst for
  Knowledge gets its canonical in Mirrodin (plus `Printing(...)` rows for the already-scaffolded JMP,
  CMR and NEO), Blazing Crescendo in Phyrexia: All Will Be One. `just check-card-printing` is clean
  for both.
- **Seven predefined tokens** (The Void, Redwing, Doombot, Zabu, Moloid, Galactus, Sturdy Shield) —
  named tokens carrying their own abilities, so they need registered definitions rather than being
  minted inline.
- **Playtest fixes** found by actually playing the set: the mana-payment UI couldn't cover a hybrid
  pip (Extort was unpayable), sacrifice prompts said "creature" for artifact/land costs, several
  card-state badges were missing or stuck, and The Astonishing Ant-Man's counter cost couldn't be
  paid at all.

## SDK surface

The point of interest for review: **119 cards landed with almost no new SDK vocabulary.** No new
`Effect`, `Trigger`, `StaticAbility`, or `DynamicAmount` type. The whole card-facing delta is two
counter-name constants (`Counters.PLAN`, `Counters.INVASION`) and a handful of parameters on
existing primitives:

| Addition | Why it isn't composable |
|---|---|
| `Costs.RemoveXCounters(self = …)` | The filter-based form asks the player to *distribute* the removal; nothing is ever distributed for a self-scoped cost, so it is unpayable, not merely imprecise. |
| `ManaRestriction.CannotCastSpellsOtherThan` | The family's first **negative** restriction. Every existing variant is a whitelist and blocks non-cast spends (ward costs, "unless you pay {2}") that the printed text allows. |
| `ModifyCounterPlacement(placedByYou = …)` | The placer axis that separates "if counters would be put" from "if **you** would put" — already present on the sibling `DoubleCounterPlacement`. |
| `StatePredicate.IsOnBattlefield` | Cancels the last-known-info fallback the combat predicates carry, so "is he attacking *now*" can't be answered by a corpse. |
| `StatePredicate.IsAttackingAlone` | CR 506.5, on the target filter so CR 608.2b re-checks it at resolution. |
| `SourceReceivedCounterThisTurn(counterType, placedByYou)` | Same two axes, for "you've put one or more +1/+1 counters on ~ this turn". |
| `PlayLandsAndCastFilteredFromTopOfLibrary(spellFilter = null)` | Explicit "lands only", so the rendered ability text stops reading "cast spells matching land". |

`docs/card-sdk-language-reference.md` is updated for every one of these in this change.

## Cards deliberately left out

Three cards were implemented, found to be wrong, and **removed rather than shipped**, with the
blocking defect written up in `backlog/sets/marvel-super-heroes/mechanics.md`:

- **Jennifer Walters** and **Baron Strucker, HYDRA Overlord** — "Do this only once each turn" is an
  effect cap, not a trigger cap, and the engine has no way to express it.
- **Ms. Marvel, Kamala Khan** — a granted power-only CDA that has to track hand size for a duration.

The backlog checklist reflects this: 203 of 276 checked, 73 unchecked, the three above among them.

## Verification

| Gate | Result |
|---|---|
| `just test-rules` | pass |
| `:mtg-sets:test` / `:mtg-sdk:test` | pass |
| `just test-server` | pass |
| `just client-typecheck` | clean |
| `just check-card-printing` | clean for both reprints and sampled new cards |

`web-client` vitest could not run in the dev container (its `node_modules` carries a
`@rolldown/binding-darwin-arm64` native module on a linux/aarch64 box); the new
`manaCoverage.test.ts` is unverified there and should be run on a machine with a matching install.

## Review

This branch was reviewed with `/review-changes` and every finding has been fixed on the branch — see
the review comment on this PR for the full list, including a blocking bug where a `.attacking()`
filter in a zone-change trigger was silently ignored, making Ares, God of War return *every* creature
you controlled that died.


bolav:
# Review: 119 Marvel Super Heroes cards

Reviewed with `/review-changes`. All 119 new card definitions were read in full and checked
clause-by-clause against their `oracleText`; the engine, SDK and client changes were read as a diff
against `main`. **Every finding below has been fixed on the branch** — this comment is the record of
what was found and how it was resolved.

## Verdict

The SDK shape is the strongest thing here: 119 Magic cards for an `mtg-sdk` delta of two counter
constants and a handful of parameters on existing primitives, with no new `Effect`, `Trigger`,
`StaticAbility` or `DynamicAmount` type. Card fidelity is high — 113 of 119 were faithful on first
read, and all 119 `oracleText` strings and every metadata field match Scryfall exactly. The defects
clustered in one place: spots where the engine quietly does not do what a helper's name implies.

## Blocking (fixed)

**`AresGodOfWar.kt` — `.attacking()` in a leaves-battlefield trigger filter is a silent no-op.**
`TriggerMatcher.matchesStatePredicateForZoneChangeTrigger` carries event-side last-known info only
for the counter/power/attachment predicates; everything else falls through to
`matchesStatePredicateForTrigger`, where `StatePredicate.IsAttacking` sits in the explicit "don't
gate" list and returns `true` unconditionally (`TriggerMatcher.kt:1917`). The `youControl()` half
*does* gate, so Ares read as "whenever a creature you control dies, return it to hand" — a chump
blocker, a creature killed by removal on the opponent's turn, anything.

Fixed by moving the attacking test to resolution as
`Conditions.EntityMatches(EffectTarget.TriggeringEntity, GameObjectFilter.Any.attacking())` — the
Garna, Bloodfist of Keld shape, where `PredicateEvaluator` *does* read the exit snapshot (CR 608.2h).
`AresGodOfWarScenarioTest` pins both polarities: an attacker that dies comes back, a creature that
dies without attacking does not. The second assertion fails against the old code.

The SDK reference now documents that `IsAttacking` is unusable as a zone-change trigger filter, so
the next card doesn't rediscover this.

## Important (fixed)

- **`just check-card-printing "Thirst for Knowledge"` failed.** The new Mirrodin canonical left the
  already-scaffolded `jmp`, `cmr` and `neo` without `Printing(...)` rows. All three added; gate clean.
- **`SuperSkrull.kt` — `colorIdentity = "B"`** contradicted Scryfall's `WUBRG` (CR 903.4 counts the
  off-colour activation costs). The field is the authoritative Scryfall identity and feeds
  `DeckValidator` Commander legality, `CommanderDeckGenerator`, `DeckProfiler` and the draft scorers,
  so a mono-black Commander deck would have run him illegally. Now `WUBRG` — the only one of 119 that
  disagreed with Scryfall.
- **`HellcatUndyingVigilante.kt`** placed the +1/+1 counter *after* she re-entered. "With a counter on
  it" is an as-enters replacement (CR 614.1c), so doublers and "enters with an additional counter"
  effects have to see it. Now uses `Effects.Move(addCounterType = …)`, the idiom Grim Reaper already
  uses in this same branch.
- **`BruceBanner.kt`** — `Conditions.SourceIsAttacking` resolves through `PredicateEvaluator`, which
  falls back to `snapshot.wasAttacking` for anything off the battlefield. Enrage routinely fires on
  lethal damage, so a Hulk that died to the very damage that triggered it still read as attacking and
  handed out a free combat phase. Fixed with a new `StatePredicate.IsOnBattlefield` conjunct —
  a live read whose whole job is cancelling that fallback.
- **`DocSamsonSuperPsychiatrist.kt`** — "If **you** would put one or more counters" was modelled with
  `ModifyCounterPlacement`, which had no `placedByYou` field (its sibling `DoubleCounterPlacement`
  does), giving it the Winding Constrictor reading where an opponent's proliferate feeds it too.
  Field added and set.
- **`AvengersAssemble.kt`** — the end-step condition tested *battlefield existence*
  (`attackedThisTurn()` / `enteredThisTurn()` are per-entity markers), so a Hero that attacked and
  then traded in combat gave no draw, though the printed card draws. Trading in combat is the card's
  ordinary line, not a corner. Now both halves are real turn-history reads:
  `Conditions.YouAttackedWithCreaturesThisTurn` over the player's own attacker set (which also kills
  the converse bug — gaining control of an opponent's Hero that attacked no longer qualifies) and
  `DynamicAmounts.subtypeEnteredUnderControlThisTurn`, whose records capture subtypes at entry time.
- **`HydraulicHelper.kt`** — "This mana **can't be spent** to cast a nonartifact spell" is a negative
  restriction, but it was implemented as the positive Guidelight Optimizer whitelist
  (`AnyOf(CardTypeSpellsOrAbilitiesOnly(ARTIFACT), AbilityActivationOnly)`), which is strictly
  narrower: it also rejects paying a ward cost, an "unless that player pays {2}" tax, or a cost
  demanded during resolution. Added `ManaRestriction.CannotCastSpellsOtherThan`, the family's first
  negative variant, backed by a new `SpellPaymentContext.isSpellCast`.
- **Three CR numbers were wrong.** Extort cited as 702.99 (that's Cipher; extort is **702.101**) —
  original to this branch. Connive cited as 702.166 (that's Bargain; connive is a keyword *action*,
  **701.50**) and hybrid mana as 107.4d (that's `{0}`; hybrid is **107.4e**) — both pre-existing repo
  conventions this branch propagated. All fixed repo-wide, including the pre-existing occurrences.
- **`docs/card-sdk-language-reference.md`** documented `Costs.RemoveXCounters` but not its new `self`
  parameter. Now updated, along with entries for every other SDK addition in this branch.

## Minor (fixed)

- `ClientStateTransformer` — `grantedSubtypes` double-reported subtypes from a floating
  `SetCreatureSubtypes` (the `alreadyBadged` filter only excluded `AddSubtype`, so a Turn-to-Frog'd
  creature showed both a `type_changed` badge and "Granted types: Frog"); an empty `else if` branch
  holding only a comment; and `isEveryCreatureType` duplicated `hasAllCreatureTypes` in the same
  file. All three folded into one shared `hasEveryCreatureType` helper that hashes once instead of
  doing ~150 linear scans per card per transform.
- `BeastEruditeAerialist.kt` — `SourceReceivedCounterThisTurn` was a bare `data object` matching any
  counter kind placed by anyone, where the card says "**you've** put one or more **+1/+1** counters".
  Both axes are now parameters, recorded on the permanent at placement time so they survive the
  counters being removed.
- `CrowdOfTrueBelievers.kt` — "attacking alone" was an `ActivationRestriction`, checked once at
  activation. It is a *targeting* restriction, so CR 608.2b must re-check it on resolution. Moved to
  the target filter via a new `attackingAlone()` (CR 506.5).
- `TheSuperHeroCivilWar.kt` — "up to one **other** target creature" didn't exclude the first target,
  so a creature could fight itself. `TargetOther` already enforces exactly this and is validated
  authoritatively by `TargetValidator`; now used.
- `KaZarOfTheSavageLand.kt` — the generated ability description read "cast spells matching land".
  `spellFilter` is now nullable, with `null` meaning lands-only.
- `SpeedYoungAvengerScenarioTest` never registered Speed, Young Avenger — it exercises the bare
  effect, which is right, but the filename promised card coverage it didn't have and
  `just test-class` on it tested nothing about the card. Renamed
  `CantBeBlockedExceptByFilterScenarioTest`, per the AGENTS.md rule that mechanic-level tests are
  named for the mechanic.
- `CommandZoneAbilityEnumerator` didn't get the `sourceId` fix its battlefield sibling received.
- `TheKingpinOfCrime.kt` claimed `canAfford` won't offer the payment at exactly 2 life; it is
  `life >= amount`, so it will, and that is correct (CR 118.3/118.3b, with CR 704.5a ending the game
  afterwards). Comment corrected — the code was already right.

## One thing for the maintainer to sign off on

Two serialized shapes widened from a bare marker to a record, which changes their **wire form**:

- `Conditions.SourceReceivedCounterThisTurn` — `data object` → `data class(counterType, placedByYou)`.
  Serialized `"SourceReceivedCounterThisTurn"` becomes `{"type": "SourceReceivedCounterThisTurn", …}`.
  The SDK facade is now a function with both parameters defaulting to the old behaviour, so the two
  existing call sites (Fractal Tender, the mtgish emitter) just gained `()`.
- `ReceivedCountersThisTurnComponent` — `data object` → `data class(counterTypes, typesFromController)`,
  so the per-turn marker can record *which* kinds landed and who placed them. Presence still means
  what it meant, so the "first counter this turn" reader is unchanged.

Both are additive with defaults and every in-repo reader is updated, but a persisted `GameState`
written before this change would not deserialize. If in-flight saved games matter, say so and I'll
add a migration or keep the old serial names as aliases.

## Deliberately not changed

- **`MisterHydeMonsterWithin.kt`** — "Remove a counter from a creature you control" is mandatory, but
  `RemoveCountersUpTo(1, …)` permits removing zero. Enforcing a real minimum needs a `minTotal`
  threaded through the multi-prompt removal continuation, and a min enforced only for the
  single-counter-kind case would be a silent inconsistency — worse than the honest "up to". The
  divergence is a strictly self-harming line (the `IfYouDo` gate already withholds the draw), and the
  card documents it.
- **`TheThingBenGrimm.kt`** — `detectDamageObserverBatchTriggers` fires once per event batch rather
  than once per damaged player (CR 603.2c). Engine-level, multiplayer-only, correct in two-player;
  out of scope for a card PR.
- **`TheSuperHeroCivilWar.kt`'s `totalManaValueAtMost` cap** is enforced interactively but not by
  `TargetValidator` (which sums only card targets). Pre-existing, shared with Fall of Gil-galad, and
  can only ever be more permissive.

## Verification

`just test-rules`, `:mtg-sets:test`, `:mtg-sdk:test`, `just test-server` all pass; `client-typecheck`
is clean; `just check-card-printing` is clean. The `web-client` vitest suite could not run in the dev
container — its `node_modules` carries a `@rolldown/binding-darwin-arm64` native module on a
linux/aarch64 box — so `manaCoverage.test.ts` is unverified and should be run elsewhere.

PR #1682, opened 2026-08-07
Emerge keyword + the six Eldritch Moon emerge Eldrazi (INR 275 → 281)

wingedsheep:
Every card left in Innistrad Remastered needs engine work first — none is a plain `cardDef`. **Emerge** is the highest-value cluster: one keyword unlocks six of the nine remaining cards.

## Emerge (CR 702.119)

> "Emerge [cost]" means "You may cast this spell by paying [cost] and sacrificing a creature rather than paying its mana cost", plus "if you chose to pay this spell's emerge cost, its total cost is reduced by an amount of **generic** mana equal to the sacrificed creature's mana value."

The interesting part is the coupling: the non-mana choice changes the mana actually charged. Three consequences drove the design.

**Affordability is per-candidate, not per-spell.** `EmergeCastEnumerator` evaluates the reduced cost once for each creature you control and offers only the ones that leave it payable, as an ordinary `SacrificePermanent` selection. Listing an unpayable creature would hand the client — and the AI, which takes the first candidate — an action that errors on submission. Candidates are sorted cheapest-mana-value first so that default pick spends the least valuable body.

**The sacrifice happens after the mana payment.** CR 601.2f–g activate mana abilities before CR 601.2h pays the total cost, so the creature may legally be tapped for mana toward its own emerge cost before it dies. `CastSpellHandler` therefore prices the cast against the chosen creature (still on the battlefield) and sacrifices it once `processPayment` has gone through. There's a test for exactly this line.

**Emerge grants no timing permission.** The spell is cast at its normal timing — which is why Elder Deep-Fiend needs its own flash, and why a Wretched Gryff in hand is not offered during the end step.

The reduction is generic-only: a mana-value-7 creature turns Elder Deep-Fiend's emerge `{5}{U}{U}` into `{U}{U}`, not into `{0}`, and the surplus 2 is wasted.

## Layers touched

| Layer | Change |
|---|---|
| SDK | `KeywordAbility.Emerge(cost)`, `Keyword.EMERGE`, `emerge("{cost}")` CardBuilder helper |
| Engine | `EmergeCasts` (single source of truth), `EmergeCastEnumerator`, `AlternativeCostType.EMERGE`, cost + validation + sacrifice in `CastSpellHandler` |
| Client | Keyword display name; the existing on-battlefield `SacrificePermanent` phase, hoisted ahead of manual mana-source selection (same reason harmonize's creature tap is) |
| mtgish | Emitter renders `emerge("{cost}")`; bridge marks the capability supported. The co-occurring cast trigger stays a SCAFFOLD hole rather than being approximated |
| Docs | `card-sdk-language-reference.md` |

No new decision component was needed — emerge's sacrifice is an ordinary additional-cost selection, so it reuses the pipeline that Torgaar and Casualty already drive.

## Cards

Canonical `card(...)` definitions land in **EMN** (their earliest printing); INR gets `Printing` rows. INR goes 275 → 281 of 290.

| Card | Emerge | Cast trigger |
|---|---|---|
| Abundant Maw | `{6}{B}` | target opponent loses 3 life, you gain 3 life |
| Decimator of the Provinces | `{6}{G}{G}{G}` | creatures you control get +2/+2 and gain trample |
| Distended Mindbender | `{5}{B}{B}` | reveal hand; discard one MV≤3 nonland and one MV≥4 |
| Elder Deep-Fiend | `{5}{U}{U}` | tap up to four target permanents |
| It of the Horrid Swarm | `{6}{G}` | create two 1/1 green Insects |
| Wretched Gryff | `{5}{U}` | draw a card |

All six hang their payload off a **"when you cast this spell"** trigger, which resolves while the Eldrazi is still on the stack. That's load-bearing for Decimator: the Boar is not among "creatures you control" when the pump resolves, so it stays a 7/7 and just brings its own trample and haste. There's an explicit test pinning that.

Abundant Maw's two amounts are independent (`LoseLife(3)` + `GainLife(3)`), not a linked drain — an opponent at 2 goes to −1 and you still gain the full 3. `DrainLife` would have capped the gain; a test caught it.

## Verification

- `EmergeKeywordTest` — 11 tests against the CR: the reduction, generic-only clamping, the surviving colored pip, missing/illegal sacrifice, per-candidate enumeration, no-creature and unaffordable boards, the absent timing permission, tapping the fodder for its own cost, and a normal hard cast.
- One scenario test per card (6 files).
- `just build` (full suite), `just check`, `just rebless-cards`, `just coverage-fixtures --rebless`, `npm run typecheck`, `check-card-printing` clean for all six.

Also adds the missing CMR `Printing` row for Elder Deep-Fiend that `check-card-printing` demands.

## Known limitation

In **manual mana mode** (auto-tap off) the mana-source step still displays the *unreduced* emerge cost — the amount genuinely depends on a choice the player hasn't confirmed yet. The sacrifice pick now runs first so the flow is coherent and the server charges correctly; only the label is pessimistic. Auto-tap (the default) is unaffected.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -5848,7 +5848,7 @@ Flying, Menace, Intimidate, Fear, Shadow, Horsemanship, all basic landwalks (Pla
 (`Keyword.NONBASIC_LANDWALK` — unblockable while the defending player controls any non-basic land;
 `LandwalkRule` checks `typeLine.isLand && !isBasicLand`; Trailblazer's Boots), First Strike, Double
 Strike, Trample, Deathtouch, Lifelink, Vigilance, Reach, Provoke, Defender, Indestructible, Hexproof, Shroud, Haste,
-Flash, Prowess, Flurry, Changeling, Convoke, Delve, Affinity, Storm, Flashback, Harmonize, Mayhem, Disturb, Evoke, Sneak, Ninjutsu, Web-slinging, Impending, Conspire, Casualty, Miracle, Hideaway, Cascade, Plot,
+Flash, Prowess, Flurry, Changeling, Convoke, Delve, Affinity, Emerge, Storm, Flashback, Harmonize, Mayhem, Disturb, Evoke, Sneak, Ninjutsu, Web-slinging, Impending, Conspire, Casualty, Miracle, Hideaway, Cascade, Plot,
 Offspring, Persist, Undying, Enduring, Ascend, Start your engines!, Max speed, Wither, Toxic, Eerie, Vivid, Fateful Bite, Exploit, Daybound, Nightbound, … (display-only — engine effect lives in handlers or
 composite abilities).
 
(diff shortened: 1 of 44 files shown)


wingedsheep:
## Follow-up: the cost label now moves with the choice

The "known limitation" is fixed, and the fix turned out to help everyone rather than just manual-mana players.

**The real problem wasn't the mana step.** With auto-tap on (the default) the player never sees a mana step at all — they see the action menu, then the sacrifice picker. So the picker was where the information was missing: a list of highlighted creatures with no indication that picking the big one makes the spell dramatically cheaper. Emerge's whole decision *is* that trade-off.

**The fix is server-side arithmetic.** `EmergeCastEnumerator` already computes the surviving cost for every candidate — that's how it decides which ones are payable at all. It now sends that map as `AdditionalCostData.costAfterSacrifice`, so:

- The sacrifice overlay shows `You pay {5}{U} → {2}{U}` the moment a creature is clicked (and "pick a creature — its mana value comes off" before that). The pre-reduction cost dims once a pick is made; a reduction that covers the whole cost reads "free".
- Manual mana-source selection prices off the chosen entry, and re-solves its land pre-selection — the server's `autoTapPreview` was solved for the *first* candidate, which may not be the one picked.

The client never re-derives the reduction. The generic-only clamp is a rule, so it stays in the engine and the client renders a string. The overlay block is written against `costAfterSacrifice` generally, so any future cost whose mana depends on which permanent pays it gets the same treatment for free.

**Tests**
- `EmergeKeywordTest` gains a case pinning the map as the player-visible contract: sacrificing a mana-value-3 creature to emerge `{5}{U}` reports `{2}{U}`; a mana-value-7 creature reports `{U}` (generic exhausted, surplus 4 wasted).
- Two `pipelinePhases` cases: the sacrifice step runs before manual mana selection, and exactly once when auto-tap handles the mana.
- `just build`, `just check`, `npm run typecheck`, 415 client unit tests green.

Not visually verified in a running game — ports 8080/5173 were already in use, so I didn't take over a dev server. Happy to boot a scenario and screenshot it if you'd like eyes on the layout.

PR #1688, opened 2026-08-07
Add power-up (CR 702.193) and the 20 Marvel Super Heroes cards it unblocks

bolav:
## Power-up (CR 702.193) and the 20 Marvel Super Heroes cards it unblocks

Implements the set's marquee keyword and the twenty cards that were waiting on it, taking MSH from
203/276 to 223/276 implemented and its blocked list from 67 cards to 47.

### The mechanic

`Power-up — [cost]: [effect]` means *"[cost]: [effect]. If this permanent entered this turn, this
ability's cost is reduced by this permanent's mana cost. Activate this ability only once."*
Authoring is one flag — `isPowerUp = true` in an `activatedAbility { }` block — deliberately shaped
like the neighbouring `isExhaust`, since both mean "activate only once". That flag does three
things:

1. **Renders the `"Power-up — "` prefix** via `ActivatedAbility.describeWithCost`. The legal-action
   enumerator rebuilds the label from the *effective* cost, so the menu shows the discounted cost on
   the turn the permanent entered — which is exactly the turn the player is most likely to use it.
2. **Auto-adds `ActivationRestriction.Once`**, the same per-object lifetime limit `isExhaust` uses.
   Per CR 400.7 a permanent that leaves and re-enters is a new object and may power up again, which
   the existing per-object tracker gives for free.
3. **Switches on the self cost reduction**, which is the genuinely new part.

### The new primitive: `ManaCost.subtract`

Every cost reduction in the engine until now was generic-only (CR 118.7a). Power-up's is not: CR
702.193b subtracts the permanent's *whole printed mana cost*, colored and colorless pips included.
Thanos, the Mad Titan is the card that forces the point — `{C}{W}{U}{B}{R}{G}` − `{R}{W}{B}` =
`{C}{U}{G}`, six mana down to three, none of which a generic-only reduction could have touched.

`ManaCost.subtract(other)` in `mtg-sdk/.../core/ManaCost.kt` implements CR 118.7 pip-wise, in three
tiers: exact symbol match, then a relaxed tier (hybrid pays either half per 118.7e, Phyrexian pays
its color per 118.7f), then spill to generic (118.7b/c/d), floored at zero. `{X}` is inert on both
sides, so Stature's `{X}{U}{U}` − `{U}` = `{X}{U}` still asks for X.

Where CR 118.7e leaves the payer a free choice, the implementation takes whichever choice reduces
the cost most, so the answer never depends on printed order: single-mana pips are assigned by
maximum bipartite matching rather than first-fit (`{W}{U}` − `{W/U}{W/B}` cancels both), and a
monocolored hybrid takes its generic half when that removes more than one mana (`{3}{W}` − `{2/W}` =
`{1}{W}`, not `{3}`).

This is not a power-up helper — **offering** (CR 702.48c) is worded identically and can use it as-is.

### Wiring

The reduction is applied inside `CastPermissionUtils.applyActivatedAbilityCostReduction`, gated on
`EnteredThisTurnComponent` (the same marker `Conditions.SourceEnteredThisTurn` reads). That is the
single choke point all three read sites go through — the legal-action enumerator and both
`ActivateAbilityHandler` sites — so the cost the player is shown and the cost the engine charges
cannot drift apart. `ReduceActivatedAbilityCost.powerUpOnly` ships alongside, mirroring the existing
`exhaustOnly`, which is what Hulk, Gamma Goliath's "power-up abilities of other creatures you
control cost {3} less" needs; the two reductions stack (CR 601.2f).

### One SDK ability replaces two power-gated evasions

Stature, Size Shifter's *"can't be blocked if her power is 1 or less"* is the same shape as Tetsuko
Umezawa, Fugitive's *"creatures you control with power or toughness 1 or less can't be blocked"*,
differing only in scope and stat set. Rather than ship a second near-duplicate type, both are now
one parameterized ability:

```kotlin
Tetsuko  CantBeBlockedWhilePropertyAtMost(1)                                  // defaults
Stature  CantBeBlockedWhilePropertyAtMost(1, setOf(Power), GroupFilter.source())
```

That collapses two SDK types, two components, two serializer rows and two near-identical projector
passes into one of each. The affected set now resolves through `AffectsFilterResolver`, so "you
control" follows the *projected* controller rather than a hand-rolled comparison.

Both variants stay in a **post-layer pass** in `StateProjector`, and that placement is load-bearing:
the gate reads Layer 7 power while a `CantBeBlocked` grant is a Layer 6 ability modification, so a
`ConditionalStaticAbility` over a power comparison would read the *printed* P/T and the evasion would
never switch off however big the creature grew. `docs/card-sdk-language-reference.md` documents that
trap explicitly, since it applies to any static whose gate reads a later layer than the static's own.

### Cards

Brave Brawler · Captain Marvel, Earth's Protector · Aerial Doombot · Bold Biochemist · Stature, Size
Shifter · Ninja of the Hand · Unliving Legionnaire · Human Torch, Johnny Storm · Quicksilver, Brash
Blur · Volcanic Villain · Hercules, Prince of Power · Pet Avengers · Serpent Specialist · She-Hulk,
Jade Defender · White Tiger, Ava Ayala · Abomination, Terrifying Titan · Hulk, Gamma Goliath ·
Thanos, the Mad Titan · Ultron Drone · Viv Vision, Teen Synthezoid.

A few carry their own wrinkles: Abomination is the only hybrid power-up (`{5}{R/G}{R/G}` −
`{3}{R/G}` = `{2}{R/G}`, proving the reduction is symbol-wise rather than color-wise); Quicksilver's
opening-hand clause means a game-start copy never "entered this turn" and so never gets the
discount; White Tiger creates the set's one *named* token, which needed the `name` parameter surfaced
on the `Effects.CreateToken` facade so the legend rule has something to key on; Human Torch's draw
trigger is a true intervening-if while Viv Vision's power check deliberately is not.

Nick Fury, Agent of S.H.I.E.L.D. was implemented and then **removed**: *"If it's a double-faced card,
you may transform it"* has no faithful modelling today — there is no "is a double-faced card"
predicate, so the optional transform could only be offered unconditionally, prompting on
single-faced permanents where the printed card offers no choice. It is written up in
`backlog/sets/marvel-super-heroes/mechanics.md` alongside the other three cards power-up does not
unblock (Wonder Man, Kang the Conqueror, Loki Laufeyson), each with the specific missing piece.

### Tests

- `ManaCostSubtractTest` — every printed MSH power-up cost paired with its source's mana cost as
  ground truth, plus the CR 118.7 subrules the set doesn't happen to exercise.
- `PowerUpKeywordScenarioTest` — once-only enforcement, re-entry reset (CR 400.7), entered-this-turn
  gating, displayed-vs-paid lockstep, and stacking with a `powerUpOnly` static. The fixtures are
  chosen so a regression to a generic-only reduction *fails*: the player is funded with three
  Mountains and one Island, which pays `{3}{U}` but cannot pay the `{1}{U}{U}{U}` a generic-only
  reduction of the same size would leave.
- `ThanosColorlessPipScenarioTest` — the `{C}` pip is colorless (CR 107.4c), not generic, so fifteen
  colored lands still can't pay it; and the odd/even sweep, which is a `ModalEffect` nested inside a
  `Composite` and therefore picks its mode at *resolution* (CR 603.3c).
- `StatureSizeShifterScenarioTest` — the evasion switches off when she grows and comes back when she
  shrinks, read against projected power.
- Manual playtest scenarios for all 20 cards in `manual-scenarios/sets/msh/batch-26..29`.

Full `:mtg-sdk:test`, `:mtg-sets:test` and `:rules-engine:test` are green.


bolav:
## Review — power-up + the 20 MSH cards

Reviewed against `main` with the SDK-elegance lens. Verification actually run, not taken from the
description: `:mtg-sdk:test`, `:mtg-sets:test` and `:rules-engine:test` all green; every cited CR
number checked against `MagicCompRules_20260619.txt` (702.193a/b, 118.7a–g, 702.48c, 400.7, 603.3c —
all correct); all 20 cards' oracle text checked verbatim against Scryfall MSH; `just
check-card-printing` clean for every new card; `just check-backlog` in sync.

**Verdict:** behaviour is right and the SDK shape is right for the marquee mechanic. Findings below
were raised on the first pass and have been **fixed in this branch** — the last three commits are the
fixes, so this comment records what was found and what changed.

### What's good

- **`ManaCost.subtract` is at the right layer and genuinely parameterized.** Not a power-up helper —
  a CR 118.7 operation that offering (CR 702.48c) gets for free. The exact → relaxed → spill tiering
  is the principled reading of 118.7a–f, not a Thanos-shaped special case.
- **One choke point for the reduction.** Putting it inside `applyActivatedAbilityCostReduction` means
  the enumerator and both `ActivateAbilityHandler` sites cannot drift, and the test suite asserts
  displayed-vs-paid lockstep in both the discounted and undiscounted state. That is the failure mode
  that would otherwise reach players.
- **The fixtures are designed to fail on the obvious regression.** Three Mountains and one Island pay
  `{3}{U}` but not `{1}{U}{U}{U}`, so a slip back to `reduceGeneric` produces an opposite answer
  rather than a silent pass. Same trick for the `{C}` pip in the Thanos test. Better than most tests
  in the repo.
- **`isPowerUp` mirrors `isExhaust` exactly** — auto-`Once`, prefix in `describeWithCost`,
  `powerUpOnly` mirroring `exhaustOnly` — instead of inventing a parallel mechanism.
- **`mechanics.md` was rewritten rather than left stale**, with a concrete plan per still-blocked
  card.

### Important — fixed

**a. `CantBeBlockedIfPowerAtMost` duplicated a type that already existed.** It differed from
`GrantCantBeBlockedToSmallCreatures` only in scope (self vs you-control) and stat set (power vs power
or toughness), but cost a second SDK type, component, serializer row, `StaticAbilityHandler` branch
and `StateProjector` pass. The next variant would have forced a third of each.

*Fixed* — both are now `CantBeBlockedWhilePropertyAtMost(maxValue, properties, filter)`:

```kotlin
Tetsuko  CantBeBlockedWhilePropertyAtMost(1)                                  // defaults
Stature  CantBeBlockedWhilePropertyAtMost(1, setOf(Power), GroupFilter.source())
```

One component, one projector pass, one block-evasion rule. The affected set now resolves through
`AffectsFilterResolver`, so "you control" follows the *projected* controller rather than the old
hand-rolled comparison. The post-layer placement — the load-bearing part — is unchanged.

**b. Hulk's KDoc example was arithmetically wrong.** Serpent Specialist's mana cost is `{G}`, so
`{3}{G}` − `{G}` = `{3}` and Hulk's −`{3}` floors it to `{0}` — the result is `{0}`, and no colored
pip survives. *Fixed*: the example is now She-Hulk (`{4}{G}{G}` − `{3}{G}` = `{1}{G}`, then −`{3}` =
`{G}`), which actually demonstrates the claim, with Serpent Specialist kept as the bottoms-out case.

**c. Captain Marvel's KDoc stated a rule that is wrong, and that the engine disagrees with.** It said
that after flashing her in on the opponent's end step "the discount is still live when you untap".
CR 702.193a is scoped to the turn she entered, and `BeginningPhaseManager` strips
`EnteredThisTurnComponent` from *every* permanent at *every* untap step — so the engine already gets
this right and only the comment was wrong. *Fixed*: the line that does work is powering up during
that same opponent's turn.

**d. Three stale "ConditionalStaticAbility" claims survived the Stature fix** (`e9b498f382`) — the
card KDoc, the MSH mechanics note, and the scenario test's clue strings all still described the
clause as a `ConditionalStaticAbility`, directly contradicting the reference-doc entry added in the
same branch which says in bold *not* to spell it that way. *Fixed* in all three.

**e. `ManaCost.subtract`'s determinism claim was false.** The KDoc promised "the total amount reduced
is the same either way", but a monocolored hybrid's halves are worth different amounts (`{3}{W}` −
`{2/W}` gave `{3}`, where the generic half gives `{1}{W}`), and greedy first-fit can strand a pip
another could have paid (`{W}{U}` − `{W/U}{W/B}` cancelled only one). *Fixed* by taking the larger
reduction: single-mana pips are now assigned by maximum bipartite matching, and a monocolored hybrid
compares its generic half against the generic actually left to remove, tie-breaking to the colored
half because that also clears a color requirement. No printed MSH cost hits either case, but the
invariant was documented and wrong.

**f. Thanos's odd/even sweep had no test at all.** The `{C}`-pip test only covered affordability. The
sweep is a `ModalEffect` nested inside the ability's `Composite`, so it takes the resolution-time
mode-picking path of CR 603.3c and must survive resumption from a composite continuation — the
least-travelled path in the branch — and the KDoc's claim that `notSourceItself()` is load-bearing
(Thanos is mana value 3, so "odd" would kill him) was unverified. *Fixed*: new tests assert the
prompt appears at resolution rather than activation, that each parity destroys the right creatures,
that zero counts as even, and that Thanos survives "odd".

### Minor — fixed

- **g.** `applyPowerUpSelfReduction` claimed a face-down creature "has no mana cost". Turning face
  down doesn't touch `CardComponent`; face-down permanents simply have no abilities to activate
  (CR 708.2). The *copy* half of the sentence was correct and is kept.
- **h.** `UltronDrone` mixed `"Robot"` with `Subtype.VILLAIN.value`, and `WhiteTigerAvaAyala` used
  `"Cat"`/`"God"` — all three constants exist, now used.
- **i.** Import ordering in `StateProjector` and `StaticAbilityHandler`.
- **j.** Restored the blank line after the MSH card-count header.
- **k.** The `.firstOrNull()` on the evasion static now carries a note about the one-component-per-
  permanent limit instead of dropping a second instance silently.

### Noted, not changed

Only 11 of 213 pre-existing MSH cards have a scenario test naming them, so the 20 cards here landing
with mechanic-level coverage plus two card-level tests is consistent with the set, not a regression
against it. Worth a separate pass if per-card coverage for MSH is ever wanted — it is not this
branch's debt to pay.

PR #1701, opened 2026-08-07
Add shield, haste, and menace counters, plus Captain America and Super-Adaptoid

bolav:
Adds **shield counters** (CR 122.1c) and the last two **keyword counters** (CR 122.1b), fixes a
silently-inert controller-hexproof grant, and implements the two MSH cards that need them:
Captain America, Super-Soldier (#9) and Super-Adaptoid (#250).

## Shield counters (CR 122.1c)

One or more shield counters on a permanent create a *single* replacement effect and a *single*
prevention effect — "if this permanent would be destroyed as the result of an effect, instead remove
a shield counter from it" and "if damage would be dealt to this permanent, prevent that damage and
remove a shield counter from it". Both consume exactly **one** counter per event however many are on
the permanent, so `consumeShieldCounter` is a fixed-at-one atom shared by four chokepoints:

- `DamageUtils.dealDamageToTarget` — noncombat/effect damage.
- `CombatDamageManager` — combat damage marks itself rather than routing through
  `dealDamageToTarget`, so it needs its own call site. Applied to the whole simultaneous batch
  (CR 510.2) rather than per assignment: a creature blocked by two attackers is dealt damage once
  and spends one counter, not two. The first-strike and regular damage steps *are* separate events,
  so a double striker costs its target one counter in each.
- `ZoneMovementUtils.destroyPermanent` and `MoveCollectionExecutor`'s destroy branch — destruction
  by an effect, single-target and board-wipe alike.

Deliberately **not** the lethal-damage state-based action: 122.1c replaces destruction "as the result
of an *effect*", and a shielded creature still dies to the SBA. Per the same rulings, unpreventable
damage is still dealt but still removes a counter, so consumption sits outside the `cantBePrevented`
guard; sacrifice is unaffected; and an indestructible permanent never "would be destroyed", so its
counter is not spent. It is not regeneration — no tap, no removal from combat.

The two damage paths order the shield counter differently against other replacement effects
(`DamageUtils` puts redirection and Anti-Venom first; the combat path spends the counter first,
because it has to run over the whole batch before per-assignment replacements are reachable). Both
are legal CR 616.1 orderings; the difference is documented on `applyShieldCounterToDamage` rather
than papered over.

## Keyword counters

Haste and menace were the last two of the eleven CR 122.1b keyword counters missing from
`StateProjector.KEYWORD_COUNTER_MAP`.

## Gated controller hexproof

`StaticAbilityHandler` stamped the controller-hexproof marker only for a bare
`GrantHexproofToController`, so wrapping it in a `ConditionalStaticAbility` ("as long as ~ has a
shield counter on him, you … have hexproof") switched the grant off entirely rather than gating it.
The marker cannot resolve the gate at stamp time — it is stamped once as the permanent enters, while
the condition flips later — so the condition now rides on the marker and is re-evaluated on every
read. The four duplicated reader scans (`TargetValidator`, `TargetEnumerationUtils`, `TargetFinder`,
`ClientStateTransformer`) now share one `ControllerHexproof` helper, and the bare-vs-gated lookup is
a reusable `controllerGrant<A>()` extractor.

Note the sibling controller-grants (`GrantShroudToController`, `GrantProtectionToController`,
`OpponentsCantMakeYouSacrifice`, `GrantCantLoseGame`) still match the bare type only and would
reproduce the same silent no-op if gated. That is called out in the extractor's KDoc with the fix to
apply when a card first needs it; converting all five now would mean changing four more marker
components with no card driving it.

## Cards

**Captain America, Super-Soldier** — `{1}{W}{W}` Legendary Creature — Human Soldier Hero 3/2. First
strike; enters with a shield counter; as long as he has one, you and other Heroes you control have
hexproof. The counter carries its own replacement/prevention pair, so the reminder text needs nothing
card-side. He does not protect himself — "you and other Heroes" excludes him, so the keyword grant is
a `GroupFilter` with `excludeSelf = true`. Two statics because they address different kinds of
object, both under the same `ConditionalStaticAbility`.

**Super-Adaptoid** — `{2}` Legendary Artifact Creature — Robot Villain \*/2. Power equal to the
number of legendary creatures you control; on enter or attack, copy each of eleven keywords the
target has that he lacks, as keyword counters. The eleven clauses are generated from a list rather
than written out eleven times; the counters do the granting through `KEYWORD_COUNTER_MAP`.

## Client

Shield / haste / menace counter badges and icons. No new `GameEvent` — the existing
`CountersRemovedEvent` covers consumption, so no `ClientEvent.kt` branch was needed.

## Verification

- `:rules-engine:test` — full suite green, run with `--rerun-tasks --no-build-cache` (a plain run was
  entirely cache-served and proved nothing).
- New scenario tests: `ShieldCounterScenarioTest` (9), `CaptainAmericaSuperSoldierScenarioTest` (4),
  `HasteMenaceCounterScenarioTest` (4), `SuperAdaptoidScenarioTest` (3).
- `:mtg-sets:test` — `CardDefinitionSnapshotTest`, `CardLintTest`, `FacadeBoundaryTest` green.
- `just check-card-printing` clean for both cards (canonical in `msh`, their earliest real printing).
- Every CR number cited in code, comments and commit messages checked against the Comprehensive
  Rules text: 122.1b, 122.1c, 510.2, 510.4, 613.1f, 616.1, 702.12b, 702.111b.


bolav:
## Review — `msh-shield-counters`

Reviewed against `main` with the SDK-elegance lens. Every finding below was then fixed on the branch
and the suites re-run, except the one noted at the end as deliberately deferred.

### Verdict

Behavior is right and the SDK shape is right. Shield/haste/menace land as **engine-realized counters
with zero new SDK types** — both cards are pure composition (`EntersWithCounters`, `AddCounters`,
`ConditionalStaticAbility`, `Composite`). The one genuinely new piece of state, the `condition` on
`GrantsControllerHexproofComponent`, generalizes an existing marker instead of special-casing
Captain America.

### What's good

- **`consumeShieldCounter` is the right atom** — fixed at 1, with CR 122.1c's "one *or more* counters
  create a *single* replacement/prevention effect" spelled out as the reason.
- **Combat damage applied per *batch*, not per assignment.** A creature blocked by two attackers is
  dealt damage once (CR 510.2) and spends one counter. That's the subtle part of this mechanic.
- **The excluded chokepoint is the right one.** `LethalDamageCheck` deliberately doesn't consult
  shield counters — 122.1c replaces destruction "as the result of an *effect*", and an SBA isn't one.
  All three destroy paths traced; coverage matches the doc.
- **`ControllerHexproof` collapses four duplicated battlefield scans** and is the only way a gate
  that flips mid-game could work, given the marker is stamped once at ETB.
- **Super-Adaptoid's eleven clauses generated from a list**, with the counters doing the granting via
  `KEYWORD_COUNTER_MAP` — all 11 keywords are in the map, so nothing dead-ends.
- Oracle text matches Scryfall exactly for both cards; `check-card-printing` clean for both.

### Important — fixed

1. **Wrong CR number for indestructible.** `ShieldCounterScenarioTest` cited **CR 701.7b** twice.
   701.7 is *Create* (tokens); indestructible is **CR 702.12b** ("A permanent with indestructible
   can't be destroyed… they ignore the state-based action that checks for lethal damage").
2. **Wrong CR number for menace.** `HasteMenaceCounterScenarioTest` cited **702.111a**, which only
   says menace is an evasion ability. The two-blocker requirement is **702.111b**.
3. **`SuperAdaptoidScenarioTest` claimed coverage it didn't have.** Its KDoc advertised the
   enters/attacks absorption and the "and Super-Adaptoid doesn't" clause; neither was tested. The
   doesn't-clause is the load-bearing claim from the commit message ("he never stacks two counters
   for one keyword") and regresses silently if `SourceHasKeyword` stops reading projected state.
   Added a test that casts him, absorbs flying/trample/haste off one donor, uses the absorbed haste
   to attack the same turn, and points the attacks trigger at a second donor sharing flying but
   adding menace — asserting menace goes to 1 while flying *stays* at 1.
4. **First-strike double-consumption cited with no test.** `CombatDamageManager` documents that the
   two combat damage steps are separate events costing a counter each — the one place the
   batch-scoped implementation could fail in the opposite direction. Added a double-striker blocker
   (CR 510.4) asserting counters go 2 → 1 after the first-strike step and 1 → 0 after the regular
   step, with nothing marked.
5. **Captain America's two halves only half-tested.** Both off-state tests asserted only the *player*
   half. The halves ride different machinery — the player half through the marker +
   `ControllerHexproof`, the Heroes half through a projected continuous effect — so the sidekick now
   gets its own assertion in both.

### Minor — fixed

6. `hasShieldCounter` was dead code (no call sites). Removed.
7. Unused imports left by the `ControllerHexproof` extraction — `GrantsControllerHexproofComponent`
   and `PlayerHexproofComponent` were import-only in `TargetFinder`, `TargetEnumerationUtils` and
   `TargetValidator`. Removed.
8. `consumeShieldCounter`'s "the two chokepoints that call this" list named `DamageUtils` (which
   actually calls `applyShieldCounterToDamage`) and **omitted `CombatDamageManager`** — the one call
   site a reader wouldn't guess. Same omission in `CounterType.kt`. Both corrected to four, split by
   half.
9. **Replacement-ordering asymmetry** between the combat and noncombat damage paths: `DamageUtils`
   consults Glarecaster-style redirection and the Anti-Venom self-replacement *before* the shield
   counter; the combat path spends the counter first, because it must run over the whole batch before
   per-assignment replacements are reachable. So a permanent with both a shield counter and a
   Glarecaster shield keeps the counter against a Shock but spends it against a blocker. Both are
   legal CR 616.1 orderings, and unifying them would mean giving up either the batch scoping or the
   redirect-first preference — so it is now documented at both call sites rather than left implicit.
10. Fully-qualified inline call sites (`com.wingedsheep.engine.core.applyShieldCounterToDamage(…)`,
    `…targeting.ControllerHexproof.…`) in five places. Replaced with imports;
    `MoveCollectionExecutor` and `CombatDamageManager` already wildcard-import `engine.core`.
11. **The gate fix covered hexproof only.** `GrantShroudToController`, `GrantProtectionToController`,
    `OpponentsCantMakeYouSacrifice` and `GrantCantLoseGame` still `any { it is X }` on the bare type,
    so wrapping any of them in a `ConditionalStaticAbility` reproduces exactly the silently-inert bug
    this branch fixes. Extracted the bare-vs-gated lookup into a reusable
    `controllerGrant<A>(): ControllerGrant?` and documented the remaining exposure there. Converting
    the other four now would mean changing four more marker components with no card driving it, so
    that is left as a signposted follow-up rather than speculative work.
12. `TapLayerStateHandlers` still said "Deathtouch counter (CR 122.1c)" two lines below the
    `FlyingCounter` comment this branch correctly changed to 122.1b. Fixed.

### Not fixed (deliberate)

13. **No client badge for trample / indestructible / hexproof counters**, which Super-Adaptoid can now
    create. The right-edge badge stack runs firstStrike 32 → doubleStrike 172 → haste 200 →
    menace 228, so there is room; those three simply never had badges. Left out at the author's
    direction — pre-existing gap, cosmetic only.

### Verification

Both gates run and confirmed green after the fixes, not taken on trust:

- `:rules-engine:test` — full suite, forced with `--rerun-tasks --no-build-cache` after a plain run
  came back entirely `UP-TO-DATE` (cache-served, proving nothing). 9757 tests, 0 failures, 0 skipped.
- The four scenario classes re-run after the fixes: 21 tests, all passing.
- `:mtg-sets:test` — `CardDefinitionSnapshotTest` 308 / 0 failures, plus `CardLintTest`,
  `FacadeBoundaryTest`, `CardDiscoveryTest`.
- Every CR number in the diff checked against the Comprehensive Rules text: 122.1b, 122.1c, 510.2,
  510.4, 613.1f, 616.1, 702.12b, 702.111b.

bolav:
The four sibling controller-grants are still ungated. GrantShroudToController, GrantProtectionToController,
OpponentsCantMakeYouSacrifice and GrantCantLoseGame still match the bare type only, so wrapping one in a ConditionalStaticAbility
reproduces the exact silently-inert bug this branch fixes for hexproof. I extracted the shared controllerGrant<A>() and documented
the exposure plus the fix, but didn't convert them — that means changing four more marker components to data classes and threading
conditions through their readers, with no card driving it. Say the word if you'd rather have it done now.

PR #1703, opened 2026-08-08
Add a per-turn effect budget for triggered abilities, plus Jennifer Walters and Baron Strucker

bolav:
# Add "Do this only once each turn" (CR 603.2h) for triggered abilities, and two MSH cards that need it

Marvel Super Heroes prints two different "only once each turn" riders, and the engine only had one of
them. This adds the missing one and ships the two cards it was blocking.

## The new SDK primitive

`TriggeredAbility.effectOncePerTurn: Boolean` — the printed rider **"Do this only once each turn"**.

The engine already had `oncePerTurn`, for the *other* printed wording, "This ability triggers only once
each turn" (Crossbones, Moon Girl, Knight of Wundagore, Ant-Man). The two are not interchangeable —
what differs is **what spends the cap**:

| Printed wording | Flag | What spends the cap |
|---|---|---|
| "**This ability triggers** only once each turn" | `oncePerTurn` | the *first trigger*, whether or not anything came of it |
| "**Do this** only once each turn" | `effectOncePerTurn` | the *action*, when it is actually taken |

The governing rule is **CR 603.2h**: *"A triggered ability may have an instruction followed by 'Do this
only once each turn.' This ability triggers only if its source's controller has not yet taken the
indicated action that turn."* So the rider is a stateful trigger condition keyed to the action:

- While the action is untaken, **every** matching event triggers. In a multi-block, *every* damaged
  creature puts its own instance of The Sensational She-Hulk's mirror on the stack.
- The "you may" is answered **as an instance resolves** (Legolas, Counter of Kills ruling), so the
  controller declines down the line to the one carrying the number they want.
- Once the action is taken, the ability **stops triggering** for the rest of the turn and instances
  already on the stack **do nothing as they resolve** (Nykthos Paragon / Riveteers Ascendancy rulings).

Modelling it as `oncePerTurn` was tried and reverted on an earlier branch: the trigger cap burns the
turn's only fire on the first trigger — even a declined one — so the biggest damage number later in the
same combat is unreachable, which is the entire point of the card. Same defect for Baron Strucker: with
a trigger cap, two Villains entering together means only the first triggers, so you can't choose which
one connives. **Existing `oncePerTurn` cards are untouched**, and there are regression tests pinning
that.

How it works:

- `TriggerProcessor` lowers the flag into new `Gate.OnceEachTurn` gates around the effect. For an
  optional ability that is a sandwich: a `spend = false` check **outside** the consent gate
  (`MayDecide` / `MayPay` / `MayPayX`) and the spending gate **inside** it. The inner one is why
  declining costs nothing — you can decline instance after instance and still take the action on the
  one you want. The outer one is why an instance whose turn is already used up resolves silently
  instead of raising a yes/no that cannot matter.
- With a gate outside the consent gate, a *targeted* capped trigger no longer matches `asMayDecide()`
  at the top of its effect, so it takes the ordinary targeted path: targets are chosen on announcement
  (CR 603.3d) and the "you may" is asked at resolution, one instance at a time. That is where CR puts
  both choices anyway, and it is what removes the "accept three prompts, get one mirror" trap.
- `GatedEffectExecutor.executeOnceEachTurn` checks and spends in one atomic step against a
  `TriggeredAbilityEffectAppliedThisTurnComponent` on the source permanent, keyed by ability id (two
  capped abilities on one permanent, or the same ability on two permanents, never share a budget).
  Registered for serialization; cleared at end of turn by `CleanupPhaseManager`.
- Once the action is taken, later matching events that turn are dropped silently at trigger-processing
  time — per CR 603.2h they never trigger, so no event is emitted.
- Capped abilities are excluded from the batched may-question — one shared yes/no would answer for
  every instance and take away the choice of *which* one to use.
- Cards never author `Gate.OnceEachTurn` directly; they set `effectOncePerTurn = true`.
- The lowering recognises a consent gate only at the **top** of the effect tree; an `optional` ability
  whose "you may" is buried under a wrapper would get the budget gate outside it. Documented in the
  flag's KDoc and the SDK reference.

`docs/card-sdk-language-reference.md` is updated in the same change: §8 gains a trigger-cap vs
effect-cap table and the mechanics, and the `Gate` catalogue gains `Gate.OnceEachTurn`.

## Cards

- **Jennifer Walters // The Sensational She-Hulk** (MSH #18) — transforming DFC
  (`CardDefinition.doubleFacedCreature`, the Bruce Banner shape). Front: the lock is the reusable
  `PlayersCantCastSpells(EachOpponent, IsYourTurn)` (Voice of Victory precedent) plus a sorcery-speed
  `TransformEffect` activated ability. Back: reach/trample, the same lock, and the damage mirror —
  a `DealsDamageEvent(recipient = CreatureYouControl)` observer (`TriggerBinding.ANY`, the Kazarov
  shape) whose payoff is `MayEffect(Effects.DealDamage(ContextProperty(TRIGGER_DAMAGE_AMOUNT),
  <any target>))` with `effectOncePerTurn = true`.
- **Baron Strucker, HYDRA Overlord** (MSH #88) — the discount is `ModifySpellCost(YouCast(Villain),
  ReduceGeneric(1))` (the Tombstone, Career Criminal / Undead Warchief shape); the connive trigger is
  `Triggers.entersBattlefield(Villain-you-control, TriggerBinding.OTHER)` with
  `MayEffect(Effects.Connive(TriggeringEntity))` and `effectOncePerTurn = true`.

## Tests

One test file per card, plus one named for the mechanic:

- `EffectOncePerTurnTest` — the primitive, using two inline enchantments that differ only in which cap
  they carry. Covers: declining down a batch keeps offering the next instance; taking the action makes
  the rest resolve silently with no prompt; a later matching event that turn doesn't trigger; the turn
  boundary resets it; two sources keep separate budgets; and three regression tests proving
  `oncePerTurn` behaves exactly as before.
- `TriggeredAbilityDescriptionTest` (mtg-sdk) — the auto-generated " Do this only once each turn."
  suffix, and that `oncePerTurn` does not get it.
- `JenniferWaltersScenarioTest` — a real multi-block with three blockers dealt **1 / 2 / 5** damage in
  one combat-damage event: all three instances offered when declined, only one mirror lands; accepting
  the first silences the rest; and, taking a different instance in each of three runs, the three runs
  between them produce all three numbers — which pins `TRIGGER_DAMAGE_AMOUNT` per instance and proves
  "decline down to the 5" is a reachable line. Plus a declined combat trigger leaving the turn's use
  for a bigger later hit, and the front-face lock.
- `BaronStruckerScenarioTest` — the second Villain still triggers after the first is declined; only one
  connive per turn; his own entry doesn't trigger; the Villain discount.

## Verification

- `just test` — passed.
- `just rebless-cards` — the only snapshot change is `mtg-sets/src/test/resources/snapshots/cards/MSH.json`,
  +222 lines, containing only the two new cards. No existing card moved.
- `just check-card-printing` clean for both cards (canonical in MSH, their earliest printing).
- Backlog updated: both cards ticked, counts resynced with `just fix-backlog` (227/276), and the
  mechanics blocker section retitled SHIPPED.

## What was *not* done

This PR came out of an autonomous loop. There was **no manual playthrough in the web client, no UX
pass from either seat, and no e2e coverage** — everything above is engine-level test evidence only.

One known usability gap is called out rather than fixed, and it is pre-existing engine behaviour: the
yes/no prompt text is the ability's static description, so three simultaneous She-Hulk prompts look
identical and don't tell the player which damage number each carries. Answering them one at a time at
resolution means no answer is wasted, but choosing the *biggest* number still needs triggering context
in the decision payload. Tracked, not fixed here.

Two statements in an earlier draft of this description were wrong and are corrected above, for the
record: dropping a matching event once the action has been taken is **the rule** (CR 603.2h, and the
Riveteers Ascendancy ruling in as many words), not a shortcut or a simplification; and the instance
that spends the turn's use is the **first to resolve**, not the last — the stack is LIFO.

Seven cards outside MSH (`AncientCornucopia`, `EarthKingdomGeneral`, `IrreverentGremlin`,
`LegolasCounterOfKills`, `PlanetariumOfWanShiTong`, `SpiderVerse`, `Terrasymbiosis`) printed the same
rider but predated this flag and were still on `oncePerTurn`, so declining burned their turn's fire.
All seven are **migrated here**, each with a scenario test that declines one instance and takes a
later one; Gatherer rulings on five of them say "once you *choose* to …, that ability won't trigger
again that turn" in as many words, and the Legolas ruling spells out both halves.

Planetarium of Wan Shi Tong forced one generalization of the lowering. Its effect is
`Composite(look at the top card, May(cast it))` and its ruling ties the turn's use to the *cast*, not
the look, so `withEffectBudgetGate` now searches the **tail** of a `CompositeEffect` for the consent
gate — the payoff of a "do X, then you may Y" instruction is its last step. That also closes the
review's M3: the caveat is now "outermost or last" rather than "top only".

---

**Independent review is still pending.** A reviewer agent will record its findings on this PR and a
corrector agent may push follow-up commits, so this is not final.


bolav:
**Verdict: approve with changes — 0 blocking, 5 important, 8 minor.** The shipped behaviour is correct (verified against CR 603.2h and the official Nykthos Paragon / Riveteers Ascendancy / Legolas rulings); it does *not* degenerate to the trigger cap and existing `oncePerTurn` cards are untouched. The findings are a wrong/missing rule citation carried into the canonical SDK reference, two self-disclosed "concerns" that are factually backwards, seven existing repo cards left on the flag this PR declares wrong, and a prompt raised for instances whose budget is already gone.

Full review below.

---

# Review — unit u01 (`loop-msh-u01`), MSH

Reviewed in place on `loop-msh-u01`, diffed against its intended base `msh-shield-counters`
(`git diff msh-shield-counters...HEAD`). No merge of `origin/main` was possible or needed — the branch
is already based on its intended parent and the network is firewalled. Tier: **full** (mtg-sdk +
rules-engine source, `docs/card-sdk-language-reference.md`), confirmed against the 19-file list.
Test suite not re-run: Stage A gated `just test` green and there is nothing to resolve.

## Verdict

**The behaviour is right — and it is right for reasons the PR does not know.** I verified the
implementation against CR 603.2h and against the official Gatherer rulings on Nykthos Paragon and
Riveteers Ascendancy: every instance triggers while the action is untaken, only one applies, other
instances "do nothing as they resolve", declining does not spend it, and the budget is per-source-not-
per-controller. All four match the code exactly. This does **not** degenerate to the trigger cap, and
existing `oncePerTurn` cards are untouched and pinned by regression tests.

The problems are (a) the governing rule is **CR 603.2h**, which the unit never found and whose text
contradicts the "effect cap, not a trigger cap / the triggering is left alone" framing repeated in
~12 places including the canonical SDK reference; (b) two of the author's own disclosed "concerns" are
wrong — one apologises for behaviour that is actually the rule, the other misstates which instance wins;
(c) seven cards already in this repo print the same rider and are still on the flag this PR's new docs
declare wrong, with no trace left anywhere; and (d) the consent prompt is asked even when the budget is
already gone, which is what makes the card's headline play unreachable at a real table.

Nothing here is Blocking. 5 Important, 8 Minor.

## What's good

- **The core design is correct and I could not break it.** Gate-inside-consent-gate is the right
  placement, and check-and-spend in one executor step (`GatedEffectExecutor.executeOnceEachTurn`,
  `:420-448`) really is atomic — there is no window for two instances to both pass.
- **Per-(source, ability) budget keying is exactly right**, and it is not obvious. The Nykthos Paragon
  ruling is explicit: *"if you control multiple Nykthos Paragons, you will be able to do this once for
  each of them."* `EffectOncePerTurnTest`'s "two sources each keep their own budget" pins the correct
  semantics against a plausible-but-wrong controller-scoped reading.
- **`batchKeyOf` exclusion** (`TriggerProcessor.kt:171-175`) is a genuinely sharp catch — one shared
  yes/no would have silently destroyed the "which instance" choice, and nothing in the type system
  would have caught it.
- **`EffectOncePerTurnTest`'s two-warden design** — two inline enchantments differing only in which
  flag they carry, with prompts-asked and life-gained as independently observable channels — is the
  best shape available for this and makes the regression guard airtight.
- Reusing `PlayersCantCastSpells(EachOpponent, IsYourTurn)`, `ModifySpellCost(YouCast(Villain),
  ReduceGeneric(1))`, `Effects.Connive(TriggeringEntity)` and the existing `doubleFacedCreature` shape
  means neither card added a card-specific type. Both cards are oracle-exact against Scryfall (cost,
  type line, P/T, every ability, both faces).
- Snapshots carry `effectOncePerTurn: true` for both cards (`MSH.json:1705`, `:6866`), so a future
  regression that drops the flag shows up in the snapshot diff.

## Issues

### Important

**I1 — The governing rule is CR 603.2h; it is never cited, and the "effect cap, not a trigger cap"
framing is a rules mis-statement.**
Verified locally in both rules files (`/workspace/MagicCompRules_20260619.txt:2580`, and identically
`MagicCompRules_20260417.txt:2576`):

> **603.2h** A triggered ability may have an instruction followed by "Do this only once each turn."
> This ability triggers only if its source's controller has not yet taken the indicated action that turn.

So the CR expresses this rider as a *stateful trigger condition*, not as a cap on the effect. The unit
cites only CR 603.2 and 603.2c and asserts the opposite — "leaves the triggering alone and only limits
application", "the ability still triggers once per matching event" — in:
`mtg-sdk/.../scripting/TriggeredAbility.kt:59-73`; `mtg-sdk/.../effects/GatedEffects.kt:243-249`;
`rules-engine/.../event/TriggerProcessor.kt:290-295`; `docs/card-sdk-language-reference.md:3520` (the
table row) and `:3522-3529`; `mtg-sets/.../msh/cards/JenniferWalters.kt:48-57`;
`mtg-sets/.../msh/cards/BaronStruckerHydraOverlord.kt:27-34`;
`rules-engine/src/test/.../EffectOncePerTurnTest.kt:25-26`; `JenniferWaltersScenarioTest.kt:21-25`;
`BaronStruckerScenarioTest.kt:20-23`; `backlog/sets/marvel-super-heroes/mechanics.md:412-425`;
`LOOP_UNIT.md`; `build/pr/u01-pr-body.md:19-25`.

The *observable* behaviour shipped is correct, so this is a documentation/citation defect — but it
lands in `card-sdk-language-reference.md`, which AGENTS.md calls the canonical catalog, and it would
teach the next author to expect post-spend triggering that the engine (correctly) does not do.
Restate as: **the ability triggers on every matching event as long as the controller has not yet taken
the action this turn; once taken it stops triggering for the rest of the turn, and instances already on
the stack do nothing as they resolve** (CR 603.2h; rulings on Nykthos Paragon / Riveteers Ascendancy).
The contrast with `oncePerTurn` survives intact and is still the point: `oncePerTurn` is spent by the
*first trigger* regardless of whether the action happened, `effectOncePerTurn` by the *action*.

**I2 — Two disclosed "concerns" are wrong, in opposite directions; left as-is they will misdirect the
corrector.**
- `LOOP_UNIT.md` ("Dropping a spent-budget trigger … Strictly, the ability should still go on the stack
  and resolve doing nothing … an observable simplification"), echoed by the comment at
  `TriggerProcessor.kt:290-295` ("the same shortcut the engine already takes for a trigger with no legal
  targets") and `build/pr/u01-pr-body.md:103-105`. **Backwards.** Per CR 603.2h and the Riveteers
  Ascendancy ruling — *"Once you have chosen to return a creature to the battlefield, further instances
  of sacrificing creatures the same turn will not cause the ability to trigger"* — dropping is the rule.
  Putting it on the stack would be the bug. Rewrite the comment to state the rule, and delete the
  caveat from the PR body; do not "fix" this.
- `build/pr/u01-pr-body.md:100-101` and the matching `LOOP_UNIT.md` bullet: "the budget then applies only
  the **last-resolving** instance". The stack is LIFO, so the instance that spends the budget is the
  **first to resolve** — the last one put on the stack. Factual error.

**I3 — Seven cards already in this repo print "Do this only once each turn" and are still on
`oncePerTurn`, and the PR leaves no trace of them.**
`big/cards/AncientCornucopia.kt:33-34` · `tla/cards/EarthKingdomGeneral.kt:40,69` ·
`dsk/cards/IrreverentGremlin.kt:27,57` · `ltr/cards/LegolasCounterOfKills.kt:43` ·
`tla/cards/PlanetariumOfWanShiTong.kt:24,47` · `spm/cards/SpiderVerse.kt:40,46` ·
`eoe/cards/Terrasymbiosis.kt:24,36`.
All seven are "you may …" triggers, so all seven have precisely the defect this unit exists to fix:
**declining burns the turn's only fire.** The official Legolas ruling is unambiguous — *"'Do this only
once each turn' lets you choose whether or not to untap Legolas as the triggered ability resolves. If
you don't untap it, the ability will trigger again the next time the condition is met."* The repo's
Legolas is wrong today, and `card-sdk-language-reference.md:3514-3529` now tells authors that flag is
the wrong one, with seven shipped counter-examples and no acknowledgement. Migrating them is a one-line
flag swap plus a test each; at minimum, name them in a follow-up backlog entry so the new guidance is
not silently self-contradicting.

**I4 — The consent prompt is raised even when the budget is already spent, and for a targeted trigger
it is raised before any instance resolves — which makes the card's headline play unreachable.**
`TriggerProcessor.withEffectBudgetGate` (`:1493-1518`) places the budget gate *strictly inside* the
consent gate, so the executor asks the yes/no first and only afterwards discovers the budget is gone.
For She-Hulk the yes/no is asked at put-on-stack time (`processMayThenTargetTrigger`, dispatched at
`:344-346`), so in a multi-block the player answers three identical prompts and picks three targets
before any of them resolves; at least two are guaranteed no-ops. There is no CR 603.3b
trigger-ordering decision in the engine (I found no `OrderTriggers*` anywhere in `rules-engine/src/main`)
and the prompt text is the ability's static description carrying no damage number, so the player
cannot tell which prompt is the big hit. `JenniferWaltersScenarioTest.kt:53-57` admits this in a
comment — uniform 2 damage "keep[s] the assertion independent of the order the three triggers are
offered in".

Concrete fix, no new SDK type: add `val spend: Boolean = true` to `Gate.OnceEachTurn`
(`GatedEffects.kt:260-265`) and have `withEffectBudgetGate` emit a check-only gate *outside* the consent
gate plus the spending gate inside it:

```kotlin
GatedEffect(
    gate = Gate.OnceEachTurn(ability.id, spend = false),
    then = consentGated.copy(then = GatedEffect(Gate.OnceEachTurn(ability.id), then = consentGated.then))
)
```

An already-spent instance then silently does nothing with no prompt — exactly the Nykthos ruling's
"other instances will do nothing as they resolve" — the decline path is unchanged, and the "say yes
twice" trap disappears. (`FeasibilityCheck` on the outer `Gate.MayDecide` would also work — the
executor already honours it, `GatedEffects.kt:131-136` — but that needs a third `FeasibilityCheck`
variant where a Boolean does not.)

Note on the author's characterisation (`u01-pr-body.md:95-101`): only the first half is pre-existing
(identical prompt text for identical triggers — and note this change deliberately switches off the
batching that mitigates it, `TriggerProcessor.kt:175`, correctly, but that means it is a *new instance*
of the old gap). The second half — accepting several while only one applies — is **created by this
change**; no prior ability had a resolution-time budget that silently no-ops after a "yes".

**I5 — No test uses distinct damage amounts inside one batch.**
The whole point of the card is taking the *biggest* number, and every multi-block case uses 2/2/2
(`JenniferWaltersScenarioTest.kt:56-85`). Nothing proves that `ContextProperty(TRIGGER_DAMAGE_AMOUNT)`
resolves per-instance within a batch, nor that declining instances 1–2 and accepting instance 3 mirrors
*that* instance's number rather than some other instance's. Add: three blockers dealt 1 / 2 / 5, decline
down to the 5, assert the opponent lost exactly 5. If that test cannot be written deterministically,
that is itself the confirmation of I4.

### Minor

**M1** — `TriggerProcessor.kt:296-308` emits `AbilityFizzledEvent(…, "Its effect has already been
applied this turn")` for an ability that, per CR 603.2h, never triggered. Nothing fizzled; the game log
and any client animation will show a phantom ability. Prefer emitting no event once I1's framing is
corrected.

**M2** — The spent-budget early return at `TriggerProcessor.kt:296` sits *before* the
`consumesDelayedTriggerId` removal (`:321-323`) and before `markTriggerFiredEver` (`:330-333`). No
shipped card combines `effectOncePerTurn` with `triggersOnce` or a consuming delayed trigger, but a
future one would leak a delayed trigger or lose its lifetime mark. Move the drop below those two steps.

**M3** — `withEffectBudgetGate` (`TriggerProcessor.kt:1509-1516`) only recognises a consent gate at the
*top* of the effect tree (`effect as? GatedEffect` with a `May*` gate). An ability whose effect is a
`CompositeEffect` wrapping a `MayEffect`, or a `MayEffect` one level down, silently gets the budget gate
on the outside — so declining spends the budget, the exact defect this design exists to prevent, with no
warning. Neither shipped card hits it. Add the caveat to the `effectOncePerTurn` KDoc
(`TriggeredAbility.kt:59-74`), or make the lowering log/fail loudly when an `optional` ability has no
recognisable consent gate.

**M4** — `GatedEffects.kt:263` uses the fully-qualified `com.wingedsheep.sdk.scripting.AbilityId` in the
property declaration. There is no clashing `AbilityId` in the file; add a normal import.

**M5** — `Gate.OnceEachTurn` is the only `Gate` that carries an ability identifier, which the author
correctly flags as slightly off for a "what must succeed" vocabulary. Acceptable as shipped — the
alternative genuinely costs two new types. For a future cleanup: `EffectContext` already carries
`abilityIdentity` (`EffectContext.kt:55`), though it is definition-scoped and nullable for synthesized
sources, so the explicit id is the safer choice today. Worth one sentence in the KDoc saying why.

**M6** — `TriggeredAbility.kt:107`'s auto-generated " Do this only once each turn." suffix has no test;
both shipped cards set `descriptionOverride`. One assertion in `mtg-sdk` covers it.

**M7** — Informational, not a defect of this unit: Scryfall reports Jennifer Walters as `layout:
modal_dfc` with the back face carrying `{3}{G}{W}{W}` and **no** color indicator, whereas
`JenniferWalters.kt:95-96` models it as a transform back face with `manaCost = ""` and
`colorIndicator = "GW"`. That is exactly what the four existing MSH DFCs do (`BruceBanner.kt:94`,
`KingTChalla.kt:94`, `TonyStark.kt:94`, `MonicaRambeau.kt:77`), so the card is consistent with the set's
established convention and should move with them if the convention is ever revisited — not separately.

**M8** — `mechanics.md:412` retitles the section "SHIPPED ✅" but keeps the "correctly `oncePerTurn =
true` today" line about Crossbones / Moon Girl / Knight of Wundagore / Ant-Man. That line is accurate
(those four genuinely print "**This ability triggers** only once each turn" — verified), but the same
paragraph is the natural home for the I3 follow-up list.

## Recommendation

1. Re-cite **CR 603.2h** and restate the rider's semantics everywhere listed in I1, in the same words
   in code, docs, cards, tests, `mechanics.md` and the PR body.
2. Delete the two wrong disclosures (I2) — the drop is the rule, not a shortcut; and it is the
   first-resolving instance that spends the budget.
3. Apply the `spend: Boolean` split from I4 so an already-spent instance never prompts, and add the
   distinct-amounts multi-block test from I5.
4. Record the seven mis-flagged existing cards (I3) as a follow-up, or migrate them.
5. Sweep the Minors; M2 and M3 are the two with future teeth.

Everything else in the diff is sound and should be kept as written.

bolav:
# u01 — correction pass against the review

Branch `loop-msh-u01`, based on `msh-shield-counters`. One new commit on top of the author's work; no
history rewritten. Verdict was 0 Blocking / 5 Important / 8 Minor.

## Important

**I1 — CR 603.2h, and the framing that contradicted it. Fixed.**
Verified locally first: `MagicCompRules_20260619.txt:2580` reads *"603.2h A triggered ability may have
an instruction followed by 'Do this only once each turn.' This ability triggers only if its source's
controller has not yet taken the indicated action that turn."* The reviewer is right, so every
citation and the "effect cap, not a trigger cap / the triggering is left alone" framing was rewritten
to the actual rule — the rider is a stateful trigger condition keyed to the *action*. Sites changed:
`TriggeredAbility.kt` (flag KDoc), `CardBuilder.kt`, `GatedEffects.kt` (`Gate.OnceEachTurn` KDoc),
`TriggerProcessor.kt` (drop comment, `batchKeyOf`, `withEffectBudgetGate`, `effectBudgetSpent`),
`GatedEffectExecutor.kt`, `docs/card-sdk-language-reference.md` (§8 section + table + the `Gate`
catalogue entry), `JenniferWalters.kt`, `BaronStruckerHydraOverlord.kt`, `EffectOncePerTurnTest.kt`,
`JenniferWaltersScenarioTest.kt`, `BaronStruckerScenarioTest.kt`,
`backlog/sets/marvel-super-heroes/mechanics.md`, `LOOP_UNIT.md`, `build/pr/u01-pr-body.md`. The
`oncePerTurn` contrast survives and is now stated as *what spends the cap*: the first trigger there,
the action here. The Nykthos Paragon / Riveteers Ascendancy / Legolas rulings were re-fetched from
Scryfall and quoted where they carry the weight.

**I2 — Two wrong disclosures. Prose corrected, code untouched (correct as written).**
The reviewer is right on both. Dropping a matching event once the action has been taken *is* CR
603.2h plus the Riveteers ruling, not a shortcut — the caveat is deleted from `LOOP_UNIT.md` and the
PR body and the code comment now states the rule. And the instance that spends the turn's use is the
**first to resolve** (LIFO), not the last — corrected in both. No behaviour change was made here.

**I3 — Seven mis-flagged existing cards. Declined the conversion, recorded the debt.**
Verified all seven do print the rider and do carry `oncePerTurn` (`AncientCornucopia`,
`EarthKingdomGeneral`, `IrreverentGremlin`, `LegolasCounterOfKills`, `PlanetariumOfWanShiTong`,
`SpiderVerse`, `Terrasymbiosis`). Converting them is a separate unit — out of scope here. But the new
guidance no longer sits next to seven silent counter-examples: `docs/card-sdk-language-reference.md`
now carries a **Known drift** callout naming all seven as a queued follow-up, and
`mechanics.md` has a matching follow-up entry (which also gives M8's paragraph its home).

**I4 — `spend: Boolean` on `Gate.OnceEachTurn`. Taken, with a caveat the review did not state.**
`Gate.OnceEachTurn` gained `spend: Boolean = true`; `executeOnceEachTurn` skips the stamp when it is
false; `withEffectBudgetGate` now emits the sandwich
`OnceEachTurn(spend = false) → May… → OnceEachTurn() → effect` for an optional ability (a single
spending gate for a mandatory one). An instance whose turn is already used up resolves silently.

The consequence the review didn't trace, and the reason I checked it against CR rather than just
applying the patch: with a gate outside the consent gate, `asMayDecide()` no longer matches at the top
of the effect, so a *targeted* capped trigger routes through `processTargetedTrigger` instead of
`processMayThenTargetTrigger`. Consent therefore moves from put-on-stack time to resolution time. That
is not a side effect to regret — it is where CR puts both choices (targets on announcement, CR 603.3d;
the "you may" as the ability resolves, which the Legolas ruling states outright), and without it the
outer gate would have nothing to suppress, since all three prompts of a multi-block precede any
resolution. Cost: targets are now chosen for instances the player later declines. Blast radius is
limited to `effectOncePerTurn` abilities, i.e. the two new cards.

Also hardened while in there: the `optional` double-wrap guard (`ownsConsentGate`) now looks through a
`Gate.OnceEachTurn`, so a capped `optional` ability can't be given a second "you may".

**I5 — Distinct damage amounts in one batch. Fixed.**
`JenniferWaltersScenarioTest` now blocks three attackers of power **1 / 2 / 5** (inline vanilla
fixture creatures, so the amounts don't rest on a remembered catalog P/T) and the new test
*"each instance mirrors its own damage number, so the biggest is reachable"* runs that combat three
times, taking a different instance each time, and asserts the three runs between them produce
`{1, 2, 5}`. That is order-independent — the prompts carry no damage number, so an
accept-the-nth-and-assert-n test would have been guessing — and it pins `TRIGGER_DAMAGE_AMOUNT` to the
instance rather than the batch. It passes, which incidentally also confirms I4's fix works end to end.

## Minor

- **M1 — phantom `AbilityFizzledEvent`. Fixed.** The drop now returns no event at all; per CR 603.2h
  nothing triggered, so nothing can fizzle.
- **M2 — move the drop below the delayed-trigger removal / `triggersOnce` mark. Declined.** It
  contradicts I1's own framing: an ability that never triggers must not consume its one-shot delayed
  trigger nor burn its lifetime fire. The current ordering is the correct one; a comment now says so
  explicitly, along with the note that no shipped card combines the flags.
- **M3 — consent gate recognised only at the top of the effect tree. Fixed (documented).** Caveat
  added to the `effectOncePerTurn` KDoc and to the SDK reference: keep the consent gate outermost.
- **M4 — fully-qualified `AbilityId`. Fixed** — normal import.
- **M5 — why `Gate.OnceEachTurn` carries an ability id. Fixed** — the KDoc now says it:
  `EffectContext.abilityIdentity` is definition-scoped and null for synthesized sources, which would
  silently merge or lose budgets.
- **M6 — untested description suffix. Fixed** — new `mtg-sdk` test
  `scripting/TriggeredAbilityDescriptionTest.kt` covers the appended rider, its absence for
  `oncePerTurn`, and `descriptionOverride` winning.
- **M7 — Jennifer Walters DFC modelling. Declined (informational).** The reviewer already concluded
  the card follows the set's established convention and should move with the other four MSH DFCs if
  it ever changes. No action.
- **M8 — mechanics.md paragraph. Fixed** as part of I3: the four genuine trigger-cap cards keep their
  accurate line, and the follow-up list now lives in the same section.

## Gate

`just test` — **passed**: `BUILD SUCCESSFUL`, 10,994 tests reported PASSED, zero FAILED.

The first attempt died mid-run with *"Gradle build daemon disappeared unexpectedly"* during
`:mtg-sets:test` — the box's documented OOM kill, not a test failure (no test had failed at that
point). It left five orphaned test JVMs from my own worktree reparented to PID 1, holding ~5 GB; I
reaped those and re-ran clean.

Also run: `just rebless-cards` → **no snapshot file changed at all**, which is the expected result —
these corrections are SDK/engine and prose, and don't touch any card definition, so nothing moved in
`MSH.json` or any other set. `just check-card-printing` was not re-run: no printing placement was
touched in this pass.

Narrower runs used while iterating (all green): `EffectOncePerTurnTest` 9/9,
`JenniferWaltersScenarioTest` 5/5.

Commit: `df7904ca39 Address review findings for u01`, on top of the author's `c01be9da9b`. No history
rewritten, nothing pushed (no network here).

## Still open

- The prompt text is the ability's static description, so three simultaneous She-Hulk prompts still
  don't tell the player *which* damage number each carries. Answering them one at a time at resolution
  means no answer is wasted now, but picking the biggest number needs triggering context in the
  decision payload — a separate feature, not built here.
- The seven mis-flagged pre-existing cards (I3) are recorded, not fixed.
- There is still no CR 603.3b trigger-ordering decision in the engine; the controller cannot order
  simultaneous triggers. Pre-existing, unchanged by this unit.

bolav:
## Review round 2 — three rules bugs and the minors

Reviewed the branch against `main` and fixed everything found. Two commits on top of the seven already here.

### Blocking

**A modal back-face cast reported the front face's mana value (CR 712.8f).** `SpellCastEvent` and `CastSpellRecord` both took theirs from the pre-swap `CardComponent`. CR 712.8c does say a *nonmodal* transformed spell computes its mana value from the front, but CR 712.8f gives a modal double-faced spell "only the characteristics of the face that's up" with no such exception. Casting The Sensational She-Hulk for `{3}{G}{W}{W}` therefore reported mana value **2** to `TRIGGERING_SPELL_MANA_VALUE` and to the turn's cast history, so "a spell with mana value N" payoffs (Kellan, the Kid) misfired. `StackResolver` now resolves one `spellManaValue` alongside the face swap; `CastSpellHandler` mirrors it for the cast record.

### Important

**Printed flash was read off the wrong face (CR 712.11c).** `CastSpellHandler`'s timing check read `cardDef.keywords` — the whole card — while the type line beside it already used the face being cast. A modal DFC with a flash *front* therefore let its sorcery-speed *back* be cast at instant speed. Found by the new CR 712.11c test, which needs an inline card whose two faces disagree about flash: nothing in the MSH cycle splits that way, since King T'Challa prints flash on both, so a front-reading gate would otherwise still have passed.

**The consent-gate placement rule is now enforced rather than documented.** The lowering only finds a "you may" at the top of the effect or a composite's tail; anywhere else the budget gate lands outside it and declining spends the turn's use — the exact defect `effectOncePerTurn` exists to fix, and invisible at the table because the prompt looks identical either way. `loweredEffectBudget` now throws on that shape, and `EffectOncePerTurnLoweringTest` sweeps every card in the pool for it. Since the condition depends only on the card definition and never on game state, the sweep is what guarantees the throw can't fire mid-game.

### Then one more, found while fixing the above

**A nonmodal DFC lost its front-face mana value across a transform (CR 712.8e).** Pre-existing, not introduced by this branch, but adjacent enough that leaving it would have made the new comments misleading. `buildCardComponentForDfcFace` took mana cost off the destination face like every other characteristic — right for a modal DFC, but CR 712.8e carves mana value out for a nonmodal one, and a transform back prints no mana cost at all. A transformed Delver of Secrets read as mana value 0 instead of 1, and every reader of `CardComponent.manaValue` saw it: the mana-value predicates, `EntityNumericProperty.ManaValue`, emerge's reduction.

Mana value is now decoupled from mana cost by `CardComponent.manaValueOverride`, with `dfcBackFaceManaValue` as the single place the modal/nonmodal split is decided. All three flip routes feed through it — the battlefield transform, the return-from-exile/graveyard transform, and `StackResolver`'s cast-transformed swap — and last commit's `spellManaValue` now falls out of the same helper instead of repeating the layout check. Not modelled: CR 712.8e's other clause (a permanent *copying* a nonmodal back face has mana value 0, CR 202.3b), which is a property of the copy rather than of the flip.

### Minors

- Dropped the unreachable `cantCastSpell` branch and the duplicate `castRestrictions` check in `enumerateModalBackFace` — the caller already gates both.
- `modalBackCastFace` was queried three times per `validate()`; resolved once.
- Reattached two KDocs that a new function had been inserted between in `GatedEffectExecutor` (one orphan predates this branch).
- `CastProvenance` says `"back face"` rather than `"its back face"`, so the capitalized badge reads "Back face" like its neighbours.
- The playtest scenario is committed as `manual-scenarios/sets/msh/loop-msh-u01-new-cards-castable.json`. Note this breaks the `batch-NN-...` convention the other 30 files in that directory follow — happy to renumber it to `batch-31` if you'd rather.
- `LOOP_UNIT.md`: the "which prompt is which — still open" entry was stale (`DynamicHint` shipped in this branch); corrected, and a review-corrections section added.

### New tests

- `DfcManaValueTest` — CR 712.8e both directions, plus the modal boundary next to it.
- `DisturbKeywordTest` — the resolved permanent's mana value (the file already covered the spell's).
- `ModalDfcPermanentBackTest` — back-face-spell mana value, front-face regression guard, and three CR 712.11c timing cases.
- `EffectOncePerTurnLoweringTest` — the card-pool sweep plus the guard's own accept/reject cases.

### Verification

- `just test` — green on the run before the last commit. On the final run one test, `ConniveTargetingTest`, hit the 120s Kotest soft timeout under load; it passes standalone and `just test-rules` re-ran the whole module clean afterwards (0 failures), so it was the box, not a regression. Worth knowing rather than glossing over.
- `just test-rules` — BUILD SUCCESSFUL, 0 failures, task actually executed.
- `just check-card-printing` clean for both new cards; `just check-backlog` in sync.
- CR 603.2h, 712.3, 712.8c, 712.8e, 712.8f, 712.11b, 712.11c and 712.13 checked against the 2026-06-19 Comprehensive Rules; every quotation in the branch is verbatim and every number correct.
- Scryfall: all five heroes are `modal_dfc` with the printed back costs and no colour indicators; the four rulings quoted in the migrated cards (Legolas, Spider-Verse, Ancient Cornucopia, Planetarium) match word for word.

### One thing I did not change

`AlternativeCostType.MODAL_BACK_FACE` rides the alternative-cost enum for plumbing it doesn't really belong to, as its own KDoc admits. It's harmless today — the only consumer is `CastProvenance`, which is display-only — but it does mean a legacy `CastSpell(useAlternativeCost = true)` with no `alternativeCostType` now silently picks the back face for these cards, via the priority-chain fallthrough. Nothing in the tree constructs that shape, so I left it rather than widen the change.

PR #1731, opened 2026-08-09
Add Teamwork N (CR 702.194), plus Helicarrier Strike, Repulsor Blast and Team Tactics

bolav:
# Teamwork N (CR 702.194) + Helicarrier Strike, Repulsor Blast, Team Tactics

Marvel Super Heroes' second new keyword mechanic, plus the three simplest cards that use it.

> **Stacked PR — base is `main`.** Bottom of the stack; merge first.
>
> | # | Branch | What it adds | |
> |---|---|---|---|
> | 1 | `loop-msh-u02` | **Teamwork N** (CR 702.194) + 3 plain-rider cards | ⬅ **this PR** |
> | 2 | `loop-msh-u03` | the 5 modal "choose both instead" cards + 2 enumerator fixes |  |
> | 3 | `loop-msh-u04` | the 4 remaining non-modal cards |  |
> | 4 | `loop-msh-u05` | `TapReason` on the tap event + Agent Maria Hill — 13 of 13 |  |
>
> Its prerequisite [#1703](https://github.com/wingedsheep/argentum-engine/pull/1703) — `loop-msh-u01`, the per-turn effect budget
> (CR 603.2h), the MSH modal-DFC hero cycle and two cards — is **already merged into
> `main`**, so it no longer appears in this stack.
>
> Review this unit alone with `git diff main...loop-msh-u02`.
>
> ✅ **Re-gated on the current `main`.** Rebased onto `main` at `26f5e1c8ca` (46 new commits) and
> re-run on `loop-msh-u05`, the top of the stack, so a green run covers this branch. Flags:
> `--no-build-cache --rerun-tasks`, mandatory after a rebase here or Gradle serves a stale `:mtg-sdk`
> compile from cache and passes falsely. **`BUILD SUCCESSFUL` in 18m 38s, 12229 passed, 0 failures,
> zero `FROM-CACHE`**, all 12 module test tasks in one run.
>
> This unit needed six conflict resolutions, all additive collisions with **collect evidence**
> (CR 701.59, Murders at Karlov Manor), which rides the same optional-additional-cost rail as
> teamwork — so both mechanics add to the same `ChoiceSlot` enum, the same `Conditions` facade block,
> the same `KeywordAbility` display `when`, both cast enumerators, and the same SDK-reference section.
> `CollectEvidenceScenarioTest` (21) passes alongside `TeamworkMechanicScenarioTest` (18), so neither
> mechanic reads as the other. Log: `build/pr/loop-msh-u05-gate4.log`.

## The primitive

**Teamwork N** — *"As an additional cost to cast this spell, you may tap any number of creatures you
control with total power N or more"* (CR 702.194a). A spell whose controller declared that intention
was cast **using teamwork** (CR 702.194b), and the card's own riders branch on that fact.

Authoring is one line in the card DSL, deliberately the same shape as `bargain()`:

```kotlin
teamwork(2)
```

### What it reuses rather than rebuilds

- **The kicker / bargain rail.** `KeywordAbility.OptionalAdditionalCost` with
  `declaredSlot = ChoiceSlot.TEAMWORK`. No new keyword-ability type, no new `GameAction`, no new
  `actionType` — the cast variant is the existing `CastWithKicker`, labelled "(Teamwork 2)". The slot
  (not a boolean) is what keeps "cast using teamwork" a *different* fact from "kicked" and
  "bargained", so a teamwork spell never satisfies a "whenever you cast a kicked spell" payoff.
- **`Conditions.TeamworkWasPaid`** is a one-line facade over `CastChoiceMade(ChoiceSlot.TEAMWORK)`,
  mirroring `WasBargained` / `WebSlungCostWasPaid` — no new `Condition` subtype.
- **The crew selection.** "Tap any number of creatures you control with total power N or greater" is
  precisely the crew payment (CR 702.122b). The work here was re-exposing it on the cast rail, not
  building it: a new `mechanics/cost/VariablePermanentsCost` is the single shared answer to *which*
  permanents can pay and *what* the chosen ones measure, mirroring `CrewEnumerator` — untapped only
  (CR 701.26a), controlled by the payer, matched **and summed through projected state** so a lord
  bonus or a +1/+1 counter counts, and with no summoning-sickness check (CR 302.6 governs the `{T}`
  symbol in an activation cost, not a tap paid as a cost — same as crew). Both cast enumerators, the
  cast validator, `CostHandler`, `ActivateAbilityHandler` and the built-in AI read it, so no reader
  of `CostAtom.VariablePermanents` can disagree with another about affordability.
  **Crew and saddle were not migrated onto it** and keep their own copy of the eligibility rule in
  `CrewEnumerator` / `CrewVehicleHandler`: they are a different activation shape (`CrewVehicle`, not
  a cost atom), and their *measure* must stay separate, since crew sums through
  `CrewSaddleContributionEvaluator` so a "crews Vehicles as though its power were 2 greater" static
  raises a crew total and must not raise a teamwork one. So this is a mirror of crew's selection, not
  an extraction of it; folding crew's candidate selection in is recorded as a follow-up (pointer
  comments at both sites).
- **The crew/saddle client payload.** `TapForPowerCreatureData` / `TapForPowerCreatureInfo`, now also
  carried on the additional-cost DTO under `costType = "TapForTotalPower"`.
- **The existing modal machinery.** The ten queued teamwork cards include the "choose one; if cast
  using teamwork, choose both instead" shape (CR 702.194c); that is
  `dynamicChooseCount = DynamicAmount.Conditional(Conditions.TeamworkWasPaid, 2, 1)` on the existing
  `modal { }` block — no new modal *type*, but it did need two wiring fixes to actually fire (see
  *Post-review corrections* below).

### What genuinely had to be added

- **SDK:** `Keyword.TEAMWORK`, `ChoiceSlot.TEAMWORK`, `Conditions.TeamworkWasPaid`,
  `Costs.additional.TapForTotalPower(n)`, and the `teamwork(n)` `CardBuilder` helper.
- **The cost atom — the real gap.** `CostAtom.VariablePermanents` gained
  `PermanentCostAction.TAP`, `VariableCostMeasure.TOTAL_POWER`, and a `minMeasure` field: a floor on
  the *measure* rather than on the count, which is what "any number … with total power N or more"
  actually says. Existing atom descriptions are unchanged.
- **The atom became a spell additional cost**, not only an activated-ability cost: validation and
  payment branches in `CastSpellHandler`, `paymentSatisfied`, `CostHandler.canPayAdditionalCost`, and
  a `TAP` payment branch in `CostHandler.payVariablePermanentsList`. Three branches of the resulting
  `when`s are unreachable today and deliberately kept, because the `when`s are exhaustive over
  `PermanentCostAction` and each delegates to an existing helper rather than reimplementing
  anything: `SACRIFICE` and `EXILE` on the cast path, and `TAP` in
  `CostHandler.payVariablePermanentsList` (no printed card pays a TAP `VariablePermanents` cost from
  an activated ability yet).
- **Enumerators:** a `VariablePermanents` branch in `CastSpellEnumerator.enumerateKicker` and its
  mirror in `CastFromZoneEnumerator`. The cost info is published even when the threshold is out of
  reach, so the greyed-out variant still tells the player what teamwork would ask for.
- **AI:** a `"TapForTotalPower"` branch in `Strategist.withAutomaticPayments` — the cheapest single
  creature that clears the threshold on its own if there is one, otherwise biggest power first, so
  the AI turns as few (and as small) bodies sideways as possible.
- **Client:** a `TapForTotalPower` case in the `costPayment` pipeline phase (writing
  `additionalCostPayment.variableCostPermanents`), `requiredTotalPower` / `powerByEntityId` on
  `TargetingState`, and a total-power confirm gate plus a "power 2/2" readout in `TargetingOverlay`.
  Selection happens on the battlefield, not in an overlay. Powers come from the server — the client
  derives no rules.
- **Docs:** `docs/card-sdk-language-reference.md` gains a Teamwork N section (both the plain-rider and
  the modal "choose both instead" authoring shapes, since ten later cards will be written from it),
  the new cost-atom axes, the `Costs.additional.TapForTotalPower` bullet, and the `TeamworkWasPaid`
  condition entry.

CR numbers cited here (702.194 / 702.194a-c, 701.26a, 302.6, 702.122b, 601.2b) were checked against
the local Comprehensive Rules text, not recalled.

## The cards

| Card | Text | Composed from |
|---|---|---|
| **Helicarrier Strike** [15] `{W}` Instant | Teamwork 2. Deals 2 damage to target attacking or blocking creature; 4 instead if cast using teamwork. | `teamwork(2)` + `Effects.DealDamage` fed a `DynamicAmount.Conditional(TeamworkWasPaid, 4, 2)` — one damage event, not 2+2, so prevention shields and damage triggers see what the printed card does |
| **Repulsor Blast** [150] `{3}{R}` Sorcery | Teamwork 2. Deals 5 damage to target creature; if cast using teamwork, also 2 to that creature's controller. | `teamwork(2)` + `Effects.DealDamage` + `ConditionalEffect(TeamworkWasPaid, DealDamage(2, EffectTarget.TargetController))` |
| **Team Tactics** [155] `{1}{R}` Instant | Teamwork 1. Target creature gains double strike; also trample if cast using teamwork. | `teamwork(1)` + two `Effects.GrantKeyword` grants, the second under a `ConditionalEffect` |

All three are MSH-only printings per Scryfall, so MSH is canonical for each;
`just check-card-printing` passes for all three.

## Tests

- `TeamworkMechanicScenarioTest` — the mechanic itself: paying with several creatures, paying with
  one, declining the cost entirely, a selection short of the threshold, an already-tapped creature,
  an opponent's creature, a lord's bonus counting toward the threshold (projected power), summoning
  sickness not mattering, the tapped creatures actually being tapped, the "cast using teamwork" fact
  surviving to resolution and onto the resolving permanent, the intervening-if not firing without it,
  teamwork-vs-kicked separation in both directions, and the advertised legal action (candidates,
  threshold, affordability, label).
- One scenario test per card — `HelicarrierStrikeScenarioTest`, `RepulsorBlastScenarioTest`,
  `TeamTacticsScenarioTest` — each asserting both branches.
- New fixture: `ScenarioTestBase.castSpellWithTeamwork(...)`.

## Verification

`just test` — passed. `just rebless-cards` moved only the three new cards in
`mtg-sets/src/test/resources/snapshots/cards/MSH.json`. `just check-card-printing` clean for all
three. Backlog ticked and `just fix-backlog` run.

## What was **not** done

This PR came out of an autonomous loop, and the following were not done:

- **No manual playthrough in the web client.** Nobody has clicked a teamwork cast.
- **No UX pass from both seats.** The "power 2/2" progress readout, the greyed-out unaffordable
  variant, and the on-battlefield creature selection are untested visually.
- **No e2e test.**
- **The web client was not type-checked.** `web-client/node_modules` is absent in this container and
  there is no network, so `just client-typecheck` could not run. The client changes are small and
  reviewed by eye, but they are unverified by a compiler.
- **No mtgish bridge entry.** The corpus has no `Teamwork` keyword IR tag (only
  `WhenAPermanentBecomesTappedToPayATeamworkCost`, which belongs to Agent Maria Hill's later unit), so
  there is nothing to map. Declined deliberately, not overlooked.

## Follow-up

Ten further teamwork cards are queued as later units on top of this branch: Murdock's Crusade,
Atlantis Attacks, We Say Thee Nay!, Cruel Alliance, Too Evil to Stay Dead, Widow's Bite, HULK SMASH!,
Earth's Mightiest Heroes, Go Nuts! — all ordinary card work now — plus **Agent Maria Hill**, which
still needs one engine addition: "whenever she becomes tapped **to pay a teamwork cost**" needs a
tap-reason on `TappedEvent` and a matching `EventPattern.TapEvent` predicate.

Independent review is still pending: a reviewer agent will record findings on this PR and a corrector
agent may push follow-up commits, so this is not final.

## Post-review corrections

A second commit on this branch addresses the review:

- **The modal "choose both instead" recipe never fired (blocking).**
  `CastSpellHandler.effectiveModalChooseCounts` built its evaluation context without
  `declaredCostSlot`, so `Conditions.TeamworkWasPaid` inside a `dynamicChooseCount` always read
  false and a teamwork cast silently picked one mode. Fixed by passing the declaration into the
  context. Alongside it, `CastSpellEnumerator.enumerateKicker` now emits a `CastSpellModal` variant
  when the declared cast's spell effect is modal (a modal card has no card-level target
  requirements, so the declared variant was previously advertised as a plain no-mode cast that
  always failed validation with "Too few modes chosen"), and the client's modal phase collects a
  `TapForTotalPower` payment the same way it already collects a blight one. Covered end to end by
  three new `TeamworkMechanicScenarioTest` cases on a modal teamwork card.
- **Affordability no longer sums negative power.** `VariablePermanentsCost.canPay` measured the
  whole candidate list, so a 3/3 next to a debuffed −2-power creature reported teamwork 2 as
  unpayable; the ceiling is now the sum of non-negative contributions, which is what the payer can
  actually reach — and what the AI already assumed. Tested.
- **Prose corrections.** CR 702.194a cites rules **601.2b and 601.2f–h** (the earlier note's
  "601.2b–f" was not a range the rule uses), and the engine's declare-then-target ordering is
  rules-correct: the variant carrying the declaration is chosen before any pipeline phase, and
  CR 702.194c explicitly wants the declaration before targets. Only the client's cost-collection
  order differs, with no rules consequence — the earlier note describing that as a deviation was
  wrong. The "single shared answer" claim and the unreachable-branch disclosure above were corrected
  in the same pass, and the doc's `Filters.Creature` default was really `GameObjectFilter.Creature`.
- **Left for later units, deliberately.** Crew/saddle migration onto `VariablePermanentsCost`
  (a second mechanic, out of this PR's scope) and the `TappedEvent` tap-reason for Agent Maria Hill
  (its own queued unit) — both now carry pointer comments at every site the future change touches.


bolav:
# Review (round 2) — Teamwork N + Helicarrier Strike / Repulsor Blast / Team Tactics

Reviewed in place on `loop-msh-u02` (clean tree). `origin/main` (`26f5e1c`) is already an ancestor of
`HEAD`, so no merge was needed and this reflects post-merge reality; u01's work has landed in `main`,
so `main...HEAD` is exactly the u02 diff.

**Verdict: behavior is right and the SDK shape is right.** No blocking issues. Everything found below
was citation/documentation accuracy plus one missing rules test, and all of it is fixed in
`d3284b8334` on this branch.

## What's good

- **`teamwork(n)` is genuinely one line.** `TeamworkDsl.kt` adds
  `OptionalAdditionalCost(TapForTotalPower(n), declaredSlot = ChoiceSlot.TEAMWORK)` and nothing else;
  `Conditions.TeamworkWasPaid` is a facade over `CastChoiceMade`. Zero new `Effect`, `Condition`,
  `KeywordAbility` or cast action. The slot — not a shared boolean — carries the mechanic's identity,
  and that separation is real: `ConditionEvaluator.evaluateWasKicked` and `TriggerMatcher.kt:1689`
  both compare against `ChoiceSlot.KICKED` specifically.
- **The cost-atom extension is parameterized for the next card, not this one.** `action` / `xMeasure`
  / `minMeasure` are three independent axes on the existing `CostAtom.VariablePermanents`, and
  `minMeasure` — a floor on the *measure* rather than the count — is the genuinely reusable
  generalization. `Costs.additional.TapForTotalPower` is the named facade over it.
- **`VariablePermanentsCost` is the right consolidation.** One object answers candidates / measure /
  canPay for both cast enumerators, the cast validator, `CostHandler`,
  `ActivateAbilityHandler.variableCostX` and the AI, so they cannot disagree about affordability.
- **The negative-power clamp in `canPay` is a subtle catch**, correctly reasoned (the payer picks a
  subset, so a −2/2 is left out rather than dragging the sum down) and pinned by a test with a real
  debuff static.
- **Projection is used everywhere it must be** — `projected.getBattlefieldControlledBy`,
  `getController`, `getPower`, and `predicateEvaluator.matches(state, projected, …)`.
- **Helicarrier Strike models "4 instead" as one `DynamicAmount.Conditional`, not 2 + 2**, and the
  card comment says why: splitting it would show two damage events to prevention shields and
  damage triggers.
- **Tests earn their keep** — one file per card asserting both branches, plus a mechanic file
  covering CR 701.26a (already tapped), CR 302.6 (summoning sickness irrelevant to a tap cost),
  projected power via a lord, the CR 601.2h rewind, the durable flag on a resolved permanent, and
  teamwork-vs-kicked separation.

Verified independently: `just check-card-printing` passes for all three cards (MSH is the earliest
real printing for each), and oracle text, mana cost, type line, rarity, collector number, artist and
flavor text match Scryfall MSH exactly.

## Findings — all fixed in `d3284b8334`

### Important

**1. CR 702.194c was miscited in four places for the modal "choose both instead" shape.** Checked
against the current Comprehensive Rules: 702.194c is about **targets** — *"If part of a spell's
ability has its effect only if teamwork was used to cast it, and that part of the ability includes
any targets, the spell's controller chooses those targets only if teamwork was used…"* It says
nothing about how many modes get chosen; that is CR 700.2, branching on the CR 601.2b declaration.
Was wrong in `TeamworkDsl.kt`, `Conditions.TeamworkWasPaid`, `docs/card-sdk-language-reference.md`
and `backlog/sets/marvel-super-heroes/mechanics.md`. `TeamworkMechanicScenarioTest` already had it
right ("CR 702.194b + 700.2"). *Fixed: all four now cite 700.2 / 601.2b, and each explicitly notes
that 702.194c is the targets rule so it doesn't drift back.*

**2. CR 702.194c's actual rule was documented as supported but never exercised.** The language
reference claimed a teamwork-only clause with its own target is declared via `kickerTarget(...)` /
`kickerEffect` — plausible, since the enumerator prices the plain cast against `targetRequirements`
and the declared cast against `kickerTargetRequirements` — but no shipped card and no test touched
it, and it is exactly the corner that will bite when the nine unblocked cards land. *Fixed: added a
"Teamwork Rally" inline card (the Goblin Barrage shape on the teamwork slot) and four cases — only
the teamwork variant announces the clause's extra target (the plain cast's `targetRequirements` is
null with `targetCount == 1`, the declared cast's has 2); the plain cast resolves against the base
target and leaves the player untouched; the declared cast damages both; and declaring teamwork
without the extra target is rewound.*

### Minor

**3. CR 702.122b cited for the crew *selection*.** 702.122a is the selection rule ("Tap any number of
other untapped creatures you control with total power N or greater"); 702.122b only defines when a
creature "crews a Vehicle". *Fixed in `TeamworkDsl.kt` and the language reference.*

**4. A comment described a parameter that doesn't exist.** `CostHandler`'s `canPayAdditionalCost`
branch said "An ability-scoped caller must pass one [`sourceId`]" — the function has no such
parameter, so no caller can. *Fixed: now says reaching this from an ability means adding the
parameter first.*

**5. Forward-TODO comments in engine code.** The two "TAP-REASON HOOK (one of two)/(two of two)"
blocks were a cross-referenced note for an unimplemented card (Agent Maria Hill), which
`mechanics.md` already records in full. *Fixed: one line each, pointing at mechanics.md.*

**6. `castSpellWithTeamwork` read base controller.** The new fixture used `ControllerComponent`
rather than projected control — harmless in these scenarios, but the repo's rule is projection for
battlefield reads and the fixture is now shared. *Fixed: resolves payers through
`projectedState.getBattlefieldControlledBy`.*

**7. A stray blank line orphaned a KDoc.** An empty line was inserted between `*/` and
`fun castSpellBargained` in `ScenarioTestBase`, detaching that function's documentation — unrelated
to teamwork. *Fixed.*

**8. Doc blockquote ran on.** The `> **Teamwork N**` block followed the collect-evidence blockquote
with no blank separator, so markdown rendered the whole Teamwork section as a lazy continuation of
that blockquote's last bullet. *Fixed.*

**9. `mechanics.md` header contradicted its own body.** "3 of 13 implemented, the other 10 now
unblocked" versus a body listing 3 implemented + 9 buildable + Agent Maria Hill still blocked. The
45 → 33 blocked arithmetic matches the body. *Fixed, and the header now names the still-blocked card.*

**10. Two unnecessary `!!` assertions** in `TeamworkMechanicScenarioTest` produced compiler warnings.
*Fixed by binding the already-null-checked `additionalCostInfo` once.*

**11. `CastFromZoneEnumerator` has no modal path.** Not a regression — that file has never emitted
`CastSpellModal` — but it means the modal-teamwork fix in `CastSpellEnumerator` doesn't carry to a
flashback or graveyard-cast grant. *Recorded as a follow-up in `mechanics.md` next to the
crew-selection consolidation; no card reaches it today.*

## Gate

`just test-rules`: **10151 passed, 0 failed** (was 10147 before the four new CR 702.194c cases),
including all 22 `TeamworkMechanicScenarioTest` tests and the three per-card scenario tests.

The rest of the tree was green in an earlier `just test` run — `mtg-sdk`, `mtg-sets` (including
`CardDefinitionSnapshotTest > MSH`, `CardLintTest`, `FacadeBoundaryTest`), `ai` and `game-server`,
1021 passed / 0 failed — but that run ended when the Gradle daemon was OOM-killed before
`rules-engine` finished, which is why the two suites are reported separately. The round-2 commit
touches only KDoc in `mtg-sdk` and comments in the two engine handlers, so nothing in those modules
changed behaviorally; a single clean end-to-end `just test` on a quiet box would still be worth
having before merge.

## Recommendation

Merge. No blocking issues remain.

PR #1750, opened 2026-08-10
Add the five modal teamwork cards, and make "choose both instead" mandatory

bolav:
# The five modal teamwork cards (MSH), and a mandatory floor for "choose both instead"

Marvel Super Heroes' five `teamwork(n)` modal cards — Murdock's Crusade [24], Atlantis Attacks [46],
Widow's Bite [122], HULK SMASH! [135], Go Nuts! [168] — plus the engine work they exposed.

> **No longer stacked.** Its prerequisite `loop-msh-u02` (Teamwork N itself) merged as
> [#1731](https://github.com/wingedsheep/argentum-engine/pull/1731), so this branch was rebased off
> the stack onto `main` and is an ordinary PR against `main`. Five commits, 23 files.
>
> ✅ **Gated on `main` at `b9d89050eb`, after the rebase.** `--no-build-cache`, mandatory here or
> Gradle serves a stale `:mtg-sdk` compile and passes falsely.
>
> | Modules | Result |
> |---|---|
> | `:rules-engine:test` + `:mtg-sets:test` | **10601 passed, 0 failed** |
> | `:mtg-sdk:test` + `:game-server:test` | **897 passed, 0 failed** |

## The cards

All five print the same shape:

> Teamwork N *(As an additional cost to cast this spell, you may tap any number of creatures you
> control with total power N or more.)*
> **Choose one. If this spell was cast using teamwork, choose both instead.**

| Card | Cost | Modes |
|---|---|---|
| Murdock's Crusade [24] | `{1}{W}` Sorcery, teamwork 4 | exile creature toughness ≥ 4 / exile enchantment mana value ≥ 4 |
| Atlantis Attacks [46] | `{5}{U}{U}` Sorcery, teamwork 4 | target player creates a 6/5 hexproof Leviathan / bounce one or two nonland permanents |
| Widow's Bite [122] | `{1}{B}` Instant, teamwork 3 | grant deathtouch / −2/−2 |
| HULK SMASH! [135] | `{1}{R}` Instant, teamwork 4 | destroy noncreature artifact / Rabid Bite one-sided damage |
| Go Nuts! [168] | `{G}` Sorcery, teamwork 3 | +1/+1 counter / fight |

No new card-specific engine types — every mode composes existing vocabulary (`ForEachTargetEffect`,
`TargetObject(count = 2, minCount = 1)`, `TargetFilter.Enchantment.manaValueAtLeast`, the Rabid Bite
`damageSource` shape). Oracle text, mana costs, rarities, collector numbers and artists verified
against Scryfall; `just check-card-printing` clean on all five (canonical in MSH, the earliest real
printing).

## "Choose both instead" is mandatory — and wasn't

`ModalEffect.dynamicChooseCount` raises only the **ceiling**. It models Flame of Anor's "you *may*
choose two instead", so the floor stayed at the printed 1 — a player could declare teamwork, tap
their team, and then take a single mode. Not theoretical: the enumerator advertised
`minChooseCount = 1` on the teamwork variant and `ModalModeSelector` unlocks confirm at
`totalChosen >= minChooseCount`, so it was reachable through normal play.

- **`ModalEffect.dynamicMinChooseCount`** — the mandatory sibling. Same evaluation, same clamping,
  lower bound. The two exist because the printed wording splits: "may choose two instead" leaves the
  floor at one, "choose both **instead**" does not.
- **`teamworkModal { }`** (`TeamworkDsl.kt`) wires a modal block for that wording, setting both
  bounds from one `Conditional` and deriving "both" from the modes declared — a three-mode printing
  needs no new recipe. The five cards each shed six lines of copy-pasted `DynamicAmount.Conditional`.

## One authority for the mode count

`CastSpellHandler` (validating a submitted cast) and `CastSpellEnumerator` (advertising one) had
separate implementations of "how many modes can this cast choose", and they had already drifted —
the enumerator's knew nothing about the blight path and nothing about the floor. Both now call
**`ModalChooseCounts.forCast`**, which returns the `min..max` range.

The enumerator also now drops a modal variant when fewer than the **minimum** modes are available,
not the maximum. Gating on the maximum happened to be right for these cards only because the floor
was wrong.

### That fix reaches beyond teamwork

It failed an existing test — and the test was wrong. `CastSpellEnumeratorTest` asserted that
**Brigid's Command** ("Choose two —") is still offered with only one mode available. It isn't
castable: a mode that would be illegal can't be chosen (CR 700.2a), so there is no legal pair to
announce and the cast can't be proposed (CR 601.2). Wizards' own Cryptic Command ruling says the
same — *"You must choose two different modes."* The test's name already read "dropped entirely"
while its body asserted the opposite; it now asserts the drop.

## Tests

One file per card, `<CardName>ScenarioTest.kt`, four cases each plus extras — 25 in total:

1. plain cast → exactly the one chosen mode resolves, the other mode's victim untouched, nothing
   tapped;
2. teamwork cast → both modes resolve, payer tapped;
3. plain cast asking for both modes → rejected, card stays in hand;
4. **teamwork cast asking for one mode → rejected and rewound**, card in hand *and* payer untapped.

Rule corners worth naming:

- **Go Nuts!** points both modes at the same 2/2 so the counter lands before the fight and it trades
  with a 3/3 — pinning mode ordering within one resolution (CR 608.2c).
- **HULK SMASH!** bites a 0/8 Wall of Stone, so the assertion is `damage == 6` rather than "the 2/2
  died" — that pins the *amount*, and thereby that `EntityReference.Target(0)` is scoped to the
  mode's own target list rather than the spell's first target overall.
- **Murdock's Crusade** and **Widow's Bite** assert against `getLegalActions`, covering the half the
  handler-driven tests structurally can't: the advertised range at both ends, per-mode target
  candidates with negative controls on both filters, and the teamwork variant being absent when only
  one mode has a legal target.

## Also in here

- CR **702.194c → CR 700.2 / 601.2b** in five card KDocs and one test KDoc. 702.194c is about
  *targets*; the mode count is CR 700.2 branching on the CR 601.2b declaration —
  `docs/card-sdk-language-reference.md` already said exactly that, and the KDocs contradicted it.
- `docs/card-sdk-language-reference.md` updated in the same change: the mandatory variant, the
  `teamworkModal` recipe, and the one-authority rule.
- A manual playtest scenario (`manual-scenarios/sets/msh/loop-msh-u03-new-cards-castable.json`)
  whose board is sized for the teamwork casts, not just the mana — disjoint payers totalling 18
  power, and every mode's target restriction represented with a negative control beside it.
- Backlog ticked: MSH 235 / 276, teamwork 8 of 13 implemented.

## Not done

No manual playthrough in the web client, no UX pass, no e2e run. The web client was not
type-checked (no `node_modules` in this environment). No client changes are in this diff.

### DIFF
--- a/LOOP_UNIT.md
+++ b/LOOP_UNIT.md
@@ -1,12 +1,12 @@
 # Unit u03 — the five modal teamwork cards (MSH)
 
-Branch `loop-msh-u03`, stacked on `loop-msh-u02`. Batch unit, ordinary card work: every card is
-`teamwork(n)` plus the documented modal recipe. **No new SDK vocabulary** — the card diff is five card
-files, five test files, the MSH snapshot, and two backlog files.
+Branch `loop-msh-u03`, an independent PR against `main` — `loop-msh-u02` merged upstream as
+PR #1731, so this branch was rebased off the stack onto `origin/main` and carries only its own
+commits. Batch unit, ordinary card work: every card is `teamwork(n)` plus the modal recipe.
 
-> **Amended after review.** The branch also carries two fixes to `CastSpellEnumerator.kt`, which is
-> `loop-msh-u02`'s code rather than this unit's, because both defects are only reachable through
-> these five cards and the two branches merge in order. See *Review corrections* at the bottom.
+> **Amended twice after review.** The branch also carries fixes to `CastSpellEnumerator.kt` /
+> `CastSpellHandler.kt` and one new SDK field, because the defects are only reachable through these
+> five cards. See *Review corrections* and *Round-2 corrections* at the bottom.
 
 ## Cards
 
@@ -27,9 +27,9 @@ files, five test files, the MSH snapshot, and two backlog files.
   `TargetObject(count = 2, minCount = 1, filter = NonlandPermanent)` bounced through
   `ForEachTargetEffect(Effects.ReturnToHand(ContextTarget(0)))`.
 
-All five: `modal(chooseCount = 2, minChooseCount = 1, dynamicChooseCount =
-DynamicAmount.Conditional(Conditions.TeamworkWasPaid, Fixed(2), Fixed(1)))` — the recipe documented
-in `docs/card-sdk-language-reference.md` → *Teamwork N*. Thresholds verified against Scryfall
+All five: `teamworkModal { }` — the recipe documented in
+`docs/card-sdk-language-reference.md` → *Teamwork N*, which pins the mode count to exactly one on a
+plain cast and exactly all of them on a declared one. Thresholds verified against Scryfall
 (3, 4, 3, 4, 4 respectively).
 
 **Nothing was dropped.** The two shapes flagged as possible drops both turned out to be covered by
@@ -57,12 +57,14 @@ share a memory window. The final `just build` is green over everything.
 
 ## Tests
 
-Five files, `<CardName>ScenarioTest.kt`, three cases each (15 total, all passing):
+Five files, `<CardName>ScenarioTest.kt`, four cases each plus extras (25 total, all passing):
 
 1. plain cast → exactly the one chosen mode resolves, the other mode's victim is untouched, nothing
    tapped;
 2. teamwork cast → both modes resolve and the payer is tapped;
-3. plain cast declaring both modes → rejected, card stays in hand.
+3. plain cast declaring both modes → rejected, card stays in hand;
+4. teamwork cast declaring one mode → rejected and rewound, card in hand *and* payer untapped
+   (added in round 2 — "choose both instead" is mandatory in both directions).
 
 Go Nuts!'s teamwork case points both modes at the same 2/2 so the +1/+1 counter lands before the
 fight and it trades with a 3/3 — that pins mode ordering within the single resolution. HULK SMASH!'s
@@ -80,17 +82,17 @@ enumeration for a declared cast) reappeared — the teamwork branch is reached a
   `targetId` and cannot express per-mode target lists. That bypasses the legal-action enumerator, so
   these tests prove the *handler* accepts the shape but not that the client is offered the
   `CastSpellModal` "(Teamwork N)" variant for these particular cards. `TeamworkMechanicScenarioTest`
-  covers the advertised action generically; I did not add a per-card enumerator assertion.
+  covers the advertised action generically. Murdock's Crusade and Widow's Bite now carry per-card
+  `getLegalActions` assertions; the other three still rely on the generic coverage.
 - `Effects.Exile` is `MoveToZoneEffect(destination = EXILE)`; I did not check whether anything in
   MSH cares about the distinction between "exile" and "exile face down"/linked exile here. The
   printed text is a plain exile, so I believe this is right.
 - Atlantis Attacks' token is named `"Leviathan Token"` by the executor's default
   (`creatureTypes.joinToString(" ") + " Token"`) since I passed no explicit `name`. That matches
   what Brigid's Command and the other token cards do, but the printed token is just "Leviathan".
-- Each card's KDoc cites CR 702.194c for the "choose both instead" shape. 702.194c is strictly
-  about *targets* being skipped when teamwork wasn't used; that is exactly what
-  `dynamicChooseCount` produces here, and it matches the wording already in
-  `backlog/.../mechanics.md`, but it is a slightly loose citation.
+- ~~Each card's KDoc cites CR 702.194c for the "choose both instead" shape.~~ Corrected in round 2
+  to CR 700.2 / 601.2b; the loose citation was in fact wrong, and contradicted this repo's own SDK
+  reference.
 - Not done: no manual playthrough in the web client, no UX pass from either seat, no e2e run, no
   `/generate-scenario` JSON. The web client was not type-checked (no node_modules, no network).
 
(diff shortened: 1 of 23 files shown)


bolav:
# Review — the five modal teamwork cards

Reviewed against `main` with the branch rebased onto it, using the repo's SDK-elegance bar
(`docs/sdk-design-principles.md`). Every CR number below was checked against
`MagicCompRules_20260619.txt`, and every card against Scryfall Oracle.

**All findings below were fixed on this branch** in `Make teamwork's "choose both instead"
mandatory`. They are recorded here because the fix changed the SDK surface and one existing test's
expectation, and both deserve a reviewer's eye.

## Verdict at the time of review

The card modelling was right and genuinely composed existing vocabulary — no new SDK types, which is
the central bar. But "choose both instead" was not actually mandatory: the engine accepted a
teamwork-declared cast that picked one mode, and the client offered exactly that. The branch's own
enumerator gate and the doc it updated both asserted the opposite, so it was internally inconsistent
on its headline mechanic.

## What's good — kept through the fix

- **Zero new SDK vocabulary in the cards.** Five `cardDef`s built from existing primitives.
- **Card data is exact.** Mana cost, type line, oracle text, rarity, collector number and artist
  match Scryfall character-for-character on all five. `just check-card-printing` clean;
  the MSH snapshot was a pure insertion with nothing else moved.
- **`MurdocksCrusadeScenarioTest`'s `getLegalActions` cases** pin the enumerator half that
  handler-driven tests structurally cannot, with negative controls on *both* filters.
- **`HulkSmashScenarioTest`** uses a 0/8 Wall of Stone so the assertion is `damage == 6` rather than
  "the 2/2 died". A weaker test would have passed on the wrong target binding.
- **`GoNutsScenarioTest`** points both modes at the same 2/2 so the counter lands before the fight —
  a real CR 608.2c corner, not a restatement of the happy path.
- The playtest scenario is carefully built: disjoint teamwork payers summing to 18, and Orcish
  Oriflamme chosen over Castle specifically because Castle's +0/+2 would perturb Murdock's toughness
  threshold.

## Blocking — fixed

### `ModalEffect` raised the ceiling but not the floor

`effectiveModalChooseCounts` returned `modalEffect.minChooseCount to max`, so the floor stayed at
the printed `1` even with `TEAMWORK` declared. `validateChosenModeShape` therefore accepted
`chosenModes = [0]` on a teamwork cast, and the interactive path let "Done" fire after one mode.

Reachable through normal UI, not just a hand-crafted action: the enumerator advertised
`minChooseCount = minOf(1, 2) = 1`, and `ModalModeSelector.tsx` renders "Choose 1–2" with confirm
enabled at `totalChosen >= minChooseCount`. **Repro:** tap Craw Wurm for Teamwork 4 on Murdock's
Crusade, pick only "Street Justice", confirm — accepted. The printed card says *choose both*; CR
601.2 makes an unsatisfiable announcement illegal, not optional.

It also made the branch self-contradictory — the enumerator dropped the declared variant on the
grounds that *"'choose both instead' is not optional"*, and the SDK reference said *"the teamwork
cast picks both"*, while the handler disagreed with both.

**Fix as applied.** `chooseAllIfBlightPaid` was already this exact mandatory shape hardcoded to one
cost, so the generalisation rather than a second special case: `ModalEffect.dynamicMinChooseCount`
alongside `dynamicChooseCount`, plus a `teamworkModal { }` recipe that sets both from one
`Conditional` and derives "both" from the modes declared. Each card now carries the inverse of the
test it already had — a teamwork cast choosing one mode is rejected *and rewound*, creature
untapped. `chooseAllIfBlightPaid` can collapse into the same field later.

## Important — all fixed

### The enumerator gated on the maximum where the minimum is the constraint

`requiredModeCount = kickerChooseCount` was only correct because the floor *should* have been 2 and
wasn't — the right answer via the wrong quantity. Now gates on the effective minimum.

**This turned out to be live, not latent.** It failed `CastSpellEnumeratorTest`'s case asserting
that Brigid's Command ("Choose two —") is still offered with only one mode available. That test was
wrong: a mode that would be illegal can't be chosen (CR 700.2a), so there is no legal pair to
announce and the cast can't be proposed (CR 601.2) — Wizards' own Cryptic Command ruling is
explicit, *"You must choose two different modes."* The test's name already said "dropped entirely"
while its body asserted the opposite. It now asserts the drop. **Worth a second opinion, since it
changes behaviour for every mandatory-count modal, not just teamwork.**

### Two copies of "what mode count can this cast reach"

The enumerator's private `effectiveModalMaxChooseCount` duplicated the handler's dynamic branch —
same `EffectContext`, same evaluator, same `coerceIn` — and had already drifted twice in one diff:
it skipped the blight path and skipped the floor entirely. Now one `ModalChooseCounts.forCast`
returning `min..max`, called by both. "The handler stays the authority" is only true if there is one
implementation.

### CR 702.194c cited for the mode count in six places

`AtlantisAttacks.kt`, `GoNuts.kt`, `HulkSmash.kt`, `MurdocksCrusade.kt`, `WidowsBite.kt` and
`WidowsBiteScenarioTest.kt` all said *"The modal shape of teamwork (CR 702.194c)"*. Verified against
the Comprehensive Rules: 702.194c is about **targets** in a teamwork-only clause. The mode count is
CR 700.2 branching on the CR 601.2b declaration, with 702.194b defining "cast using teamwork" — and
`docs/card-sdk-language-reference.md` said so verbatim, *"CR 700.2 governs the mode count … **not**
CR 702.194c, which is about targets"*, from this stack's own earlier correction. All six fixed; the
702.194c citations that remain in the tree are the genuine ones (teamwork-only clause with its own
target), each checked.

## Minor — author's discretion, not addressed

- `WidowsBite.kt` cites CR 601.2c for "both modes may point at the same creature" with no test for
  the same-creature case. Go Nuts! tests its analogue.
- `HulkSmash.kt` claims "deathtouch/lifelink on it apply" — untested. A deathtouch attacker vs the
  0/8 Wall would be two lines and is the interesting half of the Rabid Bite shape.
- `AtlantisAttacksScenarioTest` only ever targets `player1` with the token mode; it's printed
  "target *player*".
- The Leviathan token is named `"Leviathan Token"` vs the printed `"Leviathan"` — repo-wide default,
  not this card's problem.

## Verification

Re-run on `main` at `b9d89050eb` after the rebase, `--no-build-cache` (mandatory here or Gradle
serves a stale `:mtg-sdk` compile and passes falsely):

| Modules | Result |
|---|---|
| `:rules-engine:test` + `:mtg-sets:test` | **10601 passed, 0 failed** |
| `:mtg-sdk:test` + `:game-server:test` | **897 passed, 0 failed** |

All new tests confirmed executed by name, not merely absent from the failure list. Snapshot
re-blessed: `MSH.json` +75 lines, 0 removed — the five `dynamicMinChooseCount` blocks and nothing
else.

PR #1751, opened 2026-08-10
Add the four remaining non-modal teamwork cards to Marvel Super Heroes

bolav:
# Add the four remaining non-modal teamwork cards to Marvel Super Heroes

> **Independent PR against `main`.** Every teamwork unit ahead of this one has landed upstream —
> `loop-msh-u01` as [#1703](https://github.com/wingedsheep/argentum-engine/pull/1703),
> `loop-msh-u02`, and `loop-msh-u03` as
> [#1750](https://github.com/wingedsheep/argentum-engine/pull/1750). This branch was rebased off the
> stack onto `origin/main` (`b11a8d70fe`) and carries only its own three commits; there is no
> merge order to respect.
>
> Review with `git diff origin/main...loop-msh-u04`.
>
> ✅ **Gated on the current `main` (`b11a8d70fe`).** `just build` with `--no-build-cache
> --rerun-tasks` — see the *Verification* section below for the result. The previously reported
> figure (12229 passed) was measured on `main` at `26f5e1c8ca`, 50 commits behind and before
> `loop-msh-u03` merged, so it no longer stands for this base.

Adds the four remaining non-modal **teamwork** cards to Marvel Super Heroes. Ordinary card work —
the teamwork rail (`teamwork(n)`, `Conditions.TeamworkWasPaid`, `ChoiceSlot.TEAMWORK`) shipped in
earlier units and is unchanged here. No engine or SDK code is touched.

## Cards

| Card | What it does | Primitives it composes |
|---|---|---|
| **Cruel Alliance** [92] — {2}{B} Sorcery, teamwork 2 | Exile target creature with mana value 3 or less; if cast using teamwork, instead exile target creature (no restriction) and gain 3 life. | The shared optional-additional-cost rail's `kickerTarget` / `kickerEffect` slots (the Fight with Fire / Brave the Wilds shape) + `Effects.Exile` + `Effects.GainLife` + `Effects.Composite`. |
| **Too Evil to Stay Dead** [118] — {2}{B} Sorcery, teamwork 4 | Choose target creature card in your graveyard with mana value 4 or less; if teamwork, instead any creature card in your graveyard. Return it to the battlefield. | Same `kickerTarget` / `kickerEffect` pair, over `TargetFilter.CreatureInYourGraveyard` / `Targets.CreatureCardInYourGraveyard` + `Effects.PutOntoBattlefield`. |
| **We Say Thee Nay!** [82] — {1}{U} Instant — Arcane, teamwork 2 | Counter target spell unless its controller pays {2}; {4} instead if cast using teamwork. | One `Effects.CounterUnlessDynamicPays` whose generic amount is `DynamicAmount.Conditional(Conditions.TeamworkWasPaid, 4, 2)`. |
| **Earth's Mightiest Heroes** [165] — {4}{G}{G} Sorcery, teamwork 5 | Reveal the top eight; put a creature card onto the battlefield, or any number if cast using teamwork; rest to the graveyard. | Gather → Select → Move pipeline, with a `ConditionalEffect(TeamworkWasPaid, …)` swapping only the selection mode (`SelectionMode.ChooseAnyNumber` vs `ChooseUpTo(1)`) — the same split KTK's See the Unwritten uses for ferocious. |

Cruel Alliance and Too Evil to Stay Dead are a shape the rail already covered but the docs hadn't
named: the teamwork "instead" replaces the **target requirement**, not an amount, so each branch is
announced under its own filter. `CardScript.kickerTargetRequirements` already *replace*
`targetRequirements` on a declared cast, so a teamwork cast can legally target something the plain
cast could not — and the plain cast drops out of the enumeration entirely when only the wider filter
has a legal target. The rules basis is **CR 601.2c** ("a spell may require *alternative targets*
only if an alternative or additional cost was chosen for it"), with **CR 702.194c** covering the
other direction (the plain cast never announces the teamwork clause's target). One bullet was added
to the Teamwork entry in `docs/card-sdk-language-reference.md` to record that; no SDK change
accompanies it. While in that entry, the pre-existing **modal "choose both instead"** bullet was
corrected from CR 702.194c to **CR 702.194b** — 702.194c governs *targets*, not mode counts, and
702.194b ("abilities may refer to a spell cast using teamwork") is what `CastSpellEnumerator` and
`CastSpellHandler` already cite for that same shape.

## Tests

One scenario test file per card, 17 tests, every card asserting **both** branches.

Because the three teamwork bugs found in the earlier units all lived in `CastSpellEnumerator`'s
legal-action advertising path — invisible to tests that drive `CastSpell` directly — two of these
files assert against `getLegalActions`:

- `CruelAllianceScenarioTest` — with only a mana-value-6 creature on the board the plain `CastSpell`
  is **not** advertised while the `CastWithKicker` teamwork variant is, carrying that creature as a
  valid target; and with no creature available to tap, the teamwork variant is advertised
  **unaffordable** (matching the enumerator's deliberate "greyed out, not hidden" behaviour for an
  unpayable teamwork cost).
- `TooEvilToStayDeadScenarioTest` — the same contrast over the graveyard, in both directions: the
  plain cast absent when only an expensive creature card is available, and offered with the
  *narrow* `validTargets` list when a cheap one is.

We Say Thee Nay!'s teamwork branch pins the tax at the actual `{4}` (a board where the controller
can pay it, asserting the prompt names `{4}` and that paying saves the spell), not merely "more
than the {2} on the table".

## Verification

- `just build` — **passed**.
- `just rebless-cards` — only the four new cards moved in
  `mtg-sets/src/test/resources/snapshots/cards/MSH.json` (pure insertions, no existing card touched).
- `just check-card-printing` for all four — each is canonical in MSH, its only real printing.
- `just fix-backlog` — MSH now 239/276; the teamwork section of
  `backlog/sets/marvel-super-heroes/mechanics.md` updated to 12 of 13 implemented, with **Agent
  Maria Hill** the only one still blocked (it needs the tap-reason addition on `TappedEvent`).

## What was *not* done

This PR came out of an autonomous loop. Specifically:

- **No manual playthrough in the web client.** Nothing here was clicked through by a human.
- **No UX pass from both seats.** In particular, Earth's Mightiest Heroes' teamwork branch shows an
  eight-card selection overlay with the noncreature cards greyed out; the decision payload is
  asserted in the test (`maxSelections` 1 vs 3) but the overlay itself was never looked at. Likewise
  the "pay {2} / pay {4}" prompt on We Say Thee Nay! was only exercised through the engine.
- **No e2e test.** No new visual or UX mechanic was introduced, so none was added.
- No card was dropped from this unit. The conditional target filter — the shape most likely to have
  needed a new primitive — turned out to be expressible on the existing rail.

## Review

Two independent reviews of this branch are complete and their findings were addressed in follow-up
commits — see `build/pr/loop-msh-u04-correction.md`.

- **Round 1** (0 blocking, 1 important, 7 minor). The important one was the unpinned `{4}` teamwork
  tax on We Say Thee Nay!, now fixed, along with the missing enumerator and edge-case assertions.
- **Round 2** (0 blocking, 2 important, 3 minor), all in the paperwork rather than the code: this
  PR body still described a stacked PR against an already-merged `loop-msh-u03` and quoted a gate
  measured on a 50-commit-stale base (both corrected above), the new
  `docs/card-sdk-language-reference.md` bullet was missing the CR 601.2c citation its card KDocs
  carry, `LOOP_UNIT.md` named the wrong base, and `CruelAlliance.kt` spelled out
  `TargetCreature(filter = TargetFilter.Creature)` where the `Targets.Creature` facade says the same
  thing.

### DIFF
--- a/LOOP_UNIT.md
+++ b/LOOP_UNIT.md
@@ -1,8 +1,10 @@
 # Unit u04 — MSH teamwork, the four non-modal payoffs
 
-Branch `loop-msh-u04`, stacked on `loop-msh-u03`. Ordinary card work: the teamwork rail
-(`teamwork(n)`, `Conditions.TeamworkWasPaid`, `ChoiceSlot.TEAMWORK`) already shipped in u02/u03 and
-is unchanged here. No engine or SDK code was touched.
+Branch `loop-msh-u04`, an independent PR against `main` — `loop-msh-u03` merged upstream as
+PR #1750, so this branch was rebased off the stack onto `origin/main` (`b11a8d70fe`) and carries
+only its own commits. Ordinary card work: the teamwork rail (`teamwork(n)`,
+`Conditions.TeamworkWasPaid`, `ChoiceSlot.TEAMWORK`) already shipped in u02/u03 and is unchanged
+here. No engine or SDK code was touched.
 
 ## Cards
 
--- a/LOOP_UNIT.md
+++ b/LOOP_UNIT.md
@@ -27,16 +27,21 @@ is unchanged here. No engine or SDK code was touched.
 
 ## Tests
 
-One file per card, all under `rules-engine/src/test/kotlin/.../scenarios/`. 13 tests, every card
+One file per card, all under `rules-engine/src/test/kotlin/.../scenarios/`. 17 tests, every card
 covering both branches.
 
-Three of the four assert against `getLegalActions` directly, not just `execute(CastSpell(...))`,
-because the teamwork bugs found in u02/u03 all lived in `CastSpellEnumerator`'s advertising path:
+Two of the four assert against `getLegalActions` directly, not just `execute(CastSpell(...))`,
+because the teamwork bugs found in u02/u03 all lived in `CastSpellEnumerator`'s advertising path.
+The other two (We Say Thee Nay!, Earth's Mightiest Heroes) drive `castSpell`/`CastSpell` only —
+neither has a branch-dependent target or mode, so their enumeration path is the generic one
+`TeamworkMechanicScenarioTest` already covers.
 
 - `CruelAllianceScenarioTest` — with only a mana-value-6 creature on the board the plain `CastSpell`
   is **not** advertised while the `CastWithKicker` teamwork variant is, carrying that creature as a
-  valid target; and with no creature to tap, the teamwork variant is advertised **unaffordable**.
-- `TooEvilToStayDeadScenarioTest` — same shape over the graveyard.
+  valid target; with a cheap creature present both are advertised and the plain one's
+  `validTargets` is the narrow list; and with no creature to tap, the teamwork variant is
+  advertised **unaffordable**.
+- `TooEvilToStayDeadScenarioTest` — same shape over the graveyard, both directions.
 
 ## Gate
 
--- a/LOOP_UNIT.md
+++ b/LOOP_UNIT.md
@@ -1,156 +1,72 @@
-# Unit u03 — the five modal teamwork cards (MSH)
+# Unit u04 — MSH teamwork, the four non-modal payoffs
 
-Branch `loop-msh-u03`, an independent PR against `main` — `loop-msh-u02` merged upstream as
-PR #1731, so this branch was rebased off the stack onto `origin/main` and carries only its own
-commits. Batch unit, ordinary card work: every card is `teamwork(n)` plus the modal recipe.
-
-> **Amended twice after review.** The branch also carries fixes to `CastSpellEnumerator.kt` /
-> `CastSpellHandler.kt` and one new SDK field, because the defects are only reachable through these
-> five cards. See *Review corrections* and *Round-2 corrections* at the bottom.
+Branch `loop-msh-u04`, stacked on `loop-msh-u03`. Ordinary card work: the teamwork rail
+(`teamwork(n)`, `Conditions.TeamworkWasPaid`, `ChoiceSlot.TEAMWORK`) already shipped in u02/u03 and
+is unchanged here. No engine or SDK code was touched.
 
 ## Cards
 
-- **Widow's Bite** [122] `{1}{B}` Instant, teamwork 3 — `Effects.GrantKeyword(DEATHTOUCH)` /
-  `Effects.ModifyStats(-2, -2)`, one `Targets.Creature` per mode.
-- **HULK SMASH!** [135] `{1}{R}` Instant, teamwork 4 — `Effects.Destroy` on
-  `TargetFilter(GameObjectFilter.Artifact.notCreature())` / the Rabid Bite one-sided
-  `Effects.DealDamage` with `amount = EntityProperty(Target(0), Power)` and
-  `damageSource = <your creature>`. Mode 2 carries two targets of its own.
-- **Go Nuts!** [168] `{G}` Sorcery, teamwork 3 — `Effects.AddCounters("+1/+1", 1)` /
-  `Effects.Fight`, the Epic Fight shape.
-- **Murdock's Crusade** [24] `{1}{W}` Sorcery, teamwork 4 — `Effects.Exile` behind
-  `TargetFilter.Creature.toughnessAtLeast(4)` / `TargetFilter.Enchantment.manaValueAtLeast(4)`.
-  The printed mode names ("Street Justice —", "Legal Justice —") are part of the mode
-  descriptions, which become the client's button text.
-- **Atlantis Attacks** [46] `{5}{U}{U}` Sorcery, teamwork 4 — `Effects.CreateToken(6, 5, BLUE,
-  "Leviathan", HEXPROOF, controller = <target player>)` / a single
-  `TargetObject(count = 2, minCount = 1, filter = NonlandPermanent)` bounced through
-  `ForEachTargetEffect(Effects.ReturnToHand(ContextTarget(0)))`.
-
-All five: `teamworkModal { }` — the recipe documented in
-`docs/card-sdk-language-reference.md` → *Teamwork N*, which pins the mode count to exactly one on a
-plain cast and exactly all of them on a declared one. Thresholds verified against Scryfall
-(3, 4, 3, 4, 4 respectively).
-
-**Nothing was dropped.** The two shapes flagged as possible drops both turned out to be covered by
-existing target vocabulary: "one or two target nonland permanents" is `count = 2, minCount = 1`
-(Succumb to the Cold / Amazing Acrobatics precedent), and "enchantment with mana value 4 or
-greater" is `TargetFilter.Enchantment.manaValueAtLeast(4)`.
-
-## Gate
-
-`just build` — **passed** (`BUILD SUCCESSFUL`, all modules).
-
-Getting there took several runs because the Gradle daemon was OOM-killed three times on this shared
-box (`daemon disappeared`), never with a compile or test error. The runs that stuck used
-`scripts/gradle-locked … -Pkotlin.compiler.execution.strategy=in-process --max-workers=1`, and
-`:rules-engine:test` had to be run on its own once so the test compile and the test run did not
-share a memory window. The final `just build` is green over everything.
-
-- `just rebless-cards` — `mtg-sets/src/test/resources/snapshots/cards/MSH.json` only, and a **pure
-  insertion** (571 added lines, 0 removed): Atlantis Attacks, Go Nuts!, HULK SMASH!, Murdock's
-  Crusade, Widow's Bite. No unrelated card moved.
-- `just check-card-printing` — ok for all five (canonical in MSH, the earliest real printing).
-- Backlog ticked; `just fix-backlog` resynced MSH to 235 / 276. The teamwork section of
-  `mechanics.md` now reads 8 of 13 implemented, with the remaining four (We Say Thee Nay!, Cruel
-  Alliance, Too Evil to Stay Dead, Earth's Mightiest Heroes) plus the still-blocked Agent Maria Hill.
+- **Cruel Alliance** [92] — {2}{B} Sorcery, teamwork 2. Plain: exile target creature with mana value
+  3 or less. Teamwork: exile target creature (no restriction) and gain 3 life. Composes the shared
+  optional-additional-cost rail's `kickerTarget` / `kickerEffect` slots (the Fight with Fire / Brave
+  the Wilds shape) so the declared cast announces its *own* target requirement, plus
+  `Effects.Exile` + `Effects.GainLife` + `Effects.Composite`.
+- **Too Evil to Stay Dead** [118] — {2}{B} Sorcery, teamwork 4. Same `kickerTarget` / `kickerEffect`
+  shape: plain targets a creature card in your graveyard with mana value 4 or less, teamwork targets
+  any creature card in your graveyard. Both branches end in `Effects.PutOntoBattlefield`
+  (`TargetFilter.CreatureInYourGraveyard`, `Targets.CreatureCardInYourGraveyard`).
+- **We Say Thee Nay!** [82] — {1}{U} Instant — Arcane, teamwork 2. One
+  `Effects.CounterUnlessDynamicPays` whose generic amount is a
+  `DynamicAmount.Conditional(Conditions.TeamworkWasPaid, 4, 2)` — a single "counter unless they pay"
+  event whose price changes, not two counter effects.
+- **Earth's Mightiest Heroes** [165] — {4}{G}{G} Sorcery, teamwork 5. Gather → Select → Move
+  pipeline (`GatherCardsEffect(TopOfLibrary(8), revealed)` → `SelectFromCollectionEffect` →
+  two `MoveCollectionEffect`s), with a `ConditionalEffect(TeamworkWasPaid, …)` swapping only the
+  selection mode: `SelectionMode.ChooseAnyNumber` vs `SelectionMode.ChooseUpTo(1)`. Structurally the
+  same split KTK's See the Unwritten uses for its ferocious "two instead of one".
 
 ## Tests
 
-Five files, `<CardName>ScenarioTest.kt`, four cases each plus extras (25 total, all passing):
+One file per card, all under `rules-engine/src/test/kotlin/.../scenarios/`. 13 tests, every card
+covering both branches.
 
-1. plain cast → exactly the one chosen mode resolves, the other mode's victim is untouched, nothing
-   tapped;
-2. teamwork cast → both modes resolve and the payer is tapped;
-3. plain cast declaring both modes → rejected, card stays in hand;
-4. teamwork cast declaring one mode → rejected and rewound, card in hand *and* payer untapped
-   (added in round 2 — "choose both instead" is mandatory in both directions).
+Three of the four assert against `getLegalActions` directly, not just `execute(CastSpell(...))`,
+because the teamwork bugs found in u02/u03 all lived in `CastSpellEnumerator`'s advertising path:
 
-Go Nuts!'s teamwork case points both modes at the same 2/2 so the +1/+1 counter lands before the
-fight and it trades with a 3/3 — that pins mode ordering within the single resolution. HULK SMASH!'s
-teamwork case pins that the bite amount reads *that mode's* first target (the 6/4 Craw Wurm) rather
-than the spell's first target overall (the artifact) — the snapshot confirms
-`targetRequirements[0]` is "target creature you control".
+- `CruelAllianceScenarioTest` — with only a mana-value-6 creature on the board the plain `CastSpell`
+  is **not** advertised while the `CastWithKicker` teamwork variant is, carrying that creature as a
+  valid target; and with no creature to tap, the teamwork variant is advertised **unaffordable**.
+- `TooEvilToStayDeadScenarioTest` — same shape over the graveyard.
 
-Neither of the two u02 regressions (missing `declaredCostSlot` in the evaluation context; no modal
-enumeration for a declared cast) reappeared — the teamwork branch is reached and both modes resolve.
+## Gate
 
-## Uncertain / worth a reviewer's eye
+`just build` (see the verdict block for the result). The four scenario tests were also run on their
+own first and are green. `just rebless-cards`, `just check-card-printing` ×4 (all ok, all canonical
+in MSH), `just fix-backlog` (239/276).
 
-- The tests drive `game.execute(CastSpell(...))` directly with `chosenModes` +
-  `modeTargetsOrdered`, because `ScenarioTestBase.castSpellWithTeamwork` takes only a single
-  `targetId` and cannot express per-mode target lists. That bypasses the legal-action enumerator, so
-  these tests prove the *handler* accepts the shape but not that the client is offered the
-  `CastSpellModal` "(Teamwork N)" variant for these particular cards. `TeamworkMechanicScenarioTest`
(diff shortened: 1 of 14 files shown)


bolav:
**Verdict: approve with changes — no blocking findings.** Behaviour is right, SDK shape is right, oracle fidelity is exact on all four cards.

**Findings: 0 blocking · 1 important · 7 minor.** Nothing here should hold the merge; fix the Important one first.

---

# Review — unit u04 (`loop-msh-u04`), MSH batch: the four non-modal teamwork cards

Reviewed in place on `loop-msh-u04`, diffed against its stack base `loop-msh-u03`
(`git diff loop-msh-u03...HEAD`). `origin/main` is unreachable in this container, so no merge was
performed; the four branches beneath this one are the intended base and were not reviewed.

**Tier: full.** Confirmed from `--stat`: 13 files, and the author's "no engine/SDK code touched"
claim holds — nothing under `rules-engine/src/main`, `mtg-sdk/src/main`, `game-server`, `ai`, or
`web-client` changed. The only non-card, non-test file is `docs/card-sdk-language-reference.md`
(+11 lines) plus the two backlog files and the regenerated `MSH.json` snapshot.

## Verdict

The behaviour is right and the SDK shape is right. Four cards, zero new SDK types, every branch
built from primitives that already shipped. Oracle text is byte-exact against Scryfall on all four,
the printing gate is green ×4, and the two cards with a branch-dependent target filter are tested
where the previous units' bugs actually lived — `getLegalActions`, not `execute(CastSpell)`. The
central claim under scrutiny (that `kickerTargetRequirements` *replace* `targetRequirements`) is
genuine documented behaviour of the shared optional-additional-cost rail, not an accident. One
Important test gap: We Say Thee Nay!'s teamwork branch is never actually pinned to `{4}`.

## What's good

- **No new SDK surface, and each card picks the right existing primitive.** `DynamicAmount.Conditional`
  feeding one `CounterUnlessDynamicPays` (rather than two counter effects) on We Say Thee Nay! is
  the correct call — the printed card is a single "counter unless they pay" event whose price
  changes. Earth's Mightiest Heroes putting the `ConditionalEffect` around only the select-and-move
  half, so the reveal and both `MoveCollectionEffect`s stay shared, is likewise the minimal split.
- **The enumerator-level tests are the right instinct and the assertions are real.**
  `CruelAllianceScenarioTest.kt:106-132` asserts the *negative* (`casts.none { actionType ==
  "CastSpell" }`) on a board where only a mana-value-6 creature exists, and then asserts the
  teamwork variant carries that creature in `validTargets`. `:134-160` asserts the mirror image —
  plain offered with `validTargets shouldBe listOf(bears)`, i.e. the narrow filter really does
  exclude Craw Wurm. Together those two prove both directions of the filter swap at the advertising
  layer. `TooEvilToStayDeadScenarioTest.kt:132-156` proves the negative over the graveyard.
- **The `kickerTargetRequirements` claim checks out.** `CardScript.kt:212-219` documents exactly
  this ("When non-empty and the cast declared *any* slot on that rail, these replace
  `targetRequirements`"), naming Fight with Fire and Brave the Wilds. `CastSpellEnumerator.kt:1942`
  and `CastSpellHandler.kt:649,2994` implement it. Crucially, `enumerateKicker` is a standalone
  private function (`CastSpellEnumerator.kt:1792-2126`) whose target gate at `:2048` runs
  `allRequirementsSatisfied` over the *kicker* requirements, independent of the plain-cast path at
  `:1030` — so the plain cast being unofferable cannot suppress the teamwork variant, and vice
  versa. The rail is not being leaned on by accident.
- **The disclosed `isAffordable` divergence is the established contract, not a deviation.**
  `CastSpellEnumerator.kt:1889-1892` says so in its own comment ("The cost info is published even
  when the threshold is out of reach, so the greyed-out variant still tells the player what teamwork
  would ask for; affordability is the separate flag"), and the u02/u03 mechanic test already asserts
  it: `TeamworkMechanicScenarioTest.kt:520-521` (`teamworkCast.isAffordable shouldBe false` when
  total available power is short). `CruelAllianceScenarioTest.kt:158` matches it exactly. The unit
  brief's "not advertised" expectation was the thing that was wrong; the author was right to flag it
  rather than bend the code. No inconsistency across the thirteen teamwork cards.
- **A real correction to u03's notes, verified.** `mechanics.md` now says the modal "choose both
  instead" five are Murdock's Crusade, Atlantis Attacks, Widow's Bite, HULK SMASH! and Go Nuts! —
  u03's text had named Repulsor Blast and Team Tactics instead. Grepping `dynamicChooseCount` across
  the eight teamwork cards confirms the new list is the correct one and the old one was wrong.
- **Metadata fidelity is exact.** All four verified against Scryfall (`set=msh`): name, mana cost,
  type line (including `Instant — Arcane`, which round-trips through the snapshot at
  `MSH.json:17769`), oracle text verbatim including the teamwork reminder, rarity, collector number,
  artist, Cruel Alliance's flavor text, and all four `imageUri` values character-for-character.
  Teamwork thresholds checked individually: 2 / 4 / 2 / 5, all correct.
- **`just check-card-printing` green ×4**, all canonical in MSH (MSH is each card's only printing).
  The `MSH.json` snapshot diff is purely additive — no removed lines.
- **One test file per card**, no batching, all in `rules-engine` `scenarios/`. 13 tests, count
  verified (5 / 4 / 2 / 2).

## Issues

### Blocking

None.

### Important

**1. We Say Thee Nay!'s teamwork branch is never pinned to `{4}`.**
`rules-engine/.../scenarios/WeSayTheeNayScenarioTest.kt:109-115`. The plain branch is pinned
properly — `:64` asserts `decision.prompt shouldContain "{2}"`. The teamwork branch is not: the
board gives Player 2 four Forests, two of which pay for Grizzly Bears, leaving exactly two untapped,
and the test then asserts only that *no decision is offered at all* (`getPendingDecision()
.shouldBeNull()`). That assertion passes for any tax greater than 2 — 3, 4, 5, or 100 — so the
`DynamicAmount.Conditional`'s `ifTrue = 4` is unverified. A regression changing 4 to 3 or 5 would go
undetected, and that number is the entire point of the card. Fix: either raise Player 2's untapped
mana (e.g. `withLandsOnBattlefield(2, "Forest", 6)`) and assert `decision.prompt shouldContain "{4}"`
before asserting they can't/won't pay, or add a third test on a board where Player 2 *can* pay {4},
asserting the prompt names `{4}` and that paying it saves Grizzly Bears. The existing "cannot pay"
test is worth keeping as well — it just isn't sufficient on its own.

### Minor

**2. The CR citation on the two filter-swapping cards is half the story.**
`mtg-sets/.../msh/cards/CruelAlliance.kt:19` and `TooEvilToStayDead.kt:21` both cite CR 702.194c.
Verified against `/workspace/MagicCompRules_20260619.txt:5433` — 702.194c is real and does cover one
direction ("If part of a spell's ability has its effect only if teamwork was used ... and that part
includes any targets, the spell's controller chooses those targets only if teamwork was used"), i.e.
the plain cast not announcing the teamwork target. But the *load-bearing* half for these two cards
is the reverse: on a declared cast the plain target requirement is replaced, not added to. That is
**CR 601.2c** (`:2461`): "a spell may require **alternative targets** only if an alternative or
additional cost was chosen for it." Suggest citing 601.2c alongside 702.194c in both KDoc blocks
(and in the new doc bullet), since 601.2c is precisely what `kickerTargetRequirements`-replaces-
`targetRequirements` implements.

**3. `docs/card-sdk-language-reference.md` — the new bullet is accurate and correctly placed, but
the neighbouring modal bullet it sits beside carries a wrong rule number.**
The added bullet (`:6059-6069`) is correct on every checkable claim: the `kickerTarget` /
`kickerEffect` pairing, "replace" semantics, "the plain cast is dropped from the enumeration
entirely when only the wider filter has a legal target" (confirmed by
`CastSpellEnumerator.kt:1030` vs `:2048`), and the restated-trailing-sentence caveat. It is placed
with the rest of the Teamwork shape list, before the closing `Conditions.TeamworkWasPaid` note, and
contradicts nothing the earlier units wrote. **However**, the pre-existing bullet three entries above
it — `**Modal "choose both instead"** (CR 702.194c)` at `:6038` — cites 702.194c for a mode-count
rule; 702.194c is about targets and has nothing to do with modal choose-counts. The engine's own
comment for that same shape (`CastSpellEnumerator.kt:1967`) cites **702.194b**, so the doc and the
code currently disagree. That line is u03's, not this unit's, but this unit is the one editing the
entry: worth changing to 702.194b (or dropping the number) while here.

**4. `LOOP_UNIT.md` overstates the enumerator test coverage.** The Tests section claims "Three of the
four assert against `getLegalActions` directly" and then lists only two. Two is correct — Cruel
Alliance and Too Evil to Stay Dead. We Say Thee Nay! and Earth's Mightiest Heroes drive
`castSpell`/`CastSpell` only. (That's the right call for those two — neither has a branch-dependent
target or mode, so their enumeration path is the generic one `TeamworkMechanicScenarioTest` already
covers — but the notes should say two.)

**5. `TooEvilToStayDeadScenarioTest` has no positive enumerator assertion.** It asserts the plain
cast is *absent* when the graveyard holds no cheap creature (`:147-149`) but never asserts the
converse at the advertising layer — that with both a Hill Giant (MV 4) and a Shivan Dragon (MV 6) in
the graveyard, the plain variant is offered with `validTargets` containing only the Giant. Cruel
Alliance covers that shape at `:150-153`, so this is redundancy rather than a hole, but it's two
lines and it's the assertion that would catch a filter that silently widened.

**6. Earth's Mightiest Heroes' min-0 path is untested.** `EarthsMightiestHeroesScenarioTest.kt`
asserts `minSelections shouldBe 0` on both branches (`:181`, `:219`) but never exercises selecting
zero cards — the "you *may* put" / "any number" reading where all eight revealed cards go to the
graveyard. Also untested: a library holding fewer than eight cards (`GatherCardsEffect` with
`TopOfLibrary(8)` must take as many as possible). Both are one short test each.

**7. `CruelAlliance.kt:28-29` states a rules corner it doesn't test.** The KDoc asserts "an animated
land (mana value 0) is a legal plain-cast target and a token (also 0) is too." That's correct
(`PredicateEvaluator.kt:384-387` reads `card.manaValue`, and a permanent with no mana cost has mana
value 0), but per the project's "every cited rule gets a test" rule the claim is currently
decorative. Either add a token or animated-land case to the plain-cast test, or drop the sentence.

**8. Cosmetic inconsistency between the two branches of Too Evil to Stay Dead.**
`TooEvilToStayDead.kt:47` builds the plain requirement as a raw
`TargetObject(filter = TargetFilter.CreatureInYourGraveyard.manaValueAtMost(4))` while `:53` uses the
`Targets.CreatureCardInYourGraveyard` facade. They are the same shape
(`Targets.kt:260` is exactly `TargetObject(TargetFilter.CreatureInYourGraveyard)`), so this is
correct — just visually asymmetric for two lines that are meant to be read as a pair. Author's
discretion. Not a facade-boundary violation: 46 other raw `TargetCreature(`/`TargetObject(` uses
exist in the MSH card package and `FacadeBoundaryTest` does not police `TargetRequirement`
constructors.

## Notes on things I checked and found clean

- Projected vs base state: nothing in this unit reads battlefield characteristics directly. The one
  battlefield filter (Cruel Alliance's `manaValueAtMost(3)`) goes through
  `targetUtils.buildTargetInfos` → `PredicateEvaluator`, which handles the face-down projection case
  at `PredicateEvaluator.kt:385`. Both graveyard filters read a non-battlefield zone, where base
  state is correct.
- No new `GameEvent`, no continuous effect, no layer interaction, no dies/leaves trigger — the
  613.8, `ClientEvent.kt`, and last-known-information classes don't apply here.
- Named targets are bound by name (`CardBuilder.kt:1029-1032`, `:1135-1138` both call
  `requirement.withId(name)` and return `EffectTarget.BoundVariable(name)`), so the two branches'
  targets cannot collide by index. `CardLinter.kt:637-646` puts both branches' ids in one scope with
  `targetCount = maxOf(...)`, which is correct for a 1-and-1 split.
- Backlog arithmetic: 235 → 239, four boxes ticked, matches.
- I did **not** re-run the test suite (stage A gated green and each `just` run takes the
  machine-global 30-minute lock). `just check-card-printing` was run and is green ×4. Every finding
  above is from reading, from Scryfall, or from the local Comp Rules — none depends on a fresh test
  run.

## Recommendation

Not blocking. Fix **#1** (pin We Say Thee Nay!'s teamwork tax at `{4}` with a board that can
actually pay it) before merge — it's the only finding that leaves real behaviour unverified. **#2**
and **#3** are cheap citation fixes worth folding into the same pass, and **#3** also cleans up an
inconsistency this entry already had. **#4** through **#8** are the author's discretion; **#5** and
**#6** are the ones most worth the two lines each if the corrector is in the files anyway.

bolav:
## Review — round 2 (`loop-msh-u04`)

Reviewed in place on `loop-msh-u04` with a clean tree. `origin/main` (`b11a8d70fe`) is already an
ancestor of HEAD, so no merge was needed and no merge commit was produced.

### Verdict

The behavior is right and the SDK shape is right. Four cards, **zero new SDK types** — every shape
composes the existing optional-additional-cost rail. Oracle text, mana costs, type lines, rarities,
collector numbers, artists, flavor text and image URIs all match Scryfall exactly for all four; the
CR citations verify against the local Comprehensive Rules; all four scenario suites plus
`TeamworkMechanicScenarioTest` and the full `:mtg-sets:test` gate are green on this base. Every
issue found was in the paperwork wrapped around the diff, not the code.

**Verification run independently of the PR description:**

- `just test-class CruelAllianceScenarioTest` — `BUILD SUCCESSFUL`
- `:rules-engine:test` for `TooEvilToStayDeadScenarioTest`, `WeSayTheeNayScenarioTest`,
  `EarthsMightiestHeroesScenarioTest`, `TeamworkMechanicScenarioTest` — `BUILD SUCCESSFUL`
- `:mtg-sets:test` (snapshot + `FacadeBoundaryTest` + `CardLinter`) — `BUILD SUCCESSFUL`
- `just check-card-printing` ×4 — all `ok`, canonical in MSH, its only real printing
- Snapshot re-bless is a pure insertion: 378 added, 0 removed, exactly the four cards

### What's good

- **The central call is correct.** Cruel Alliance / Too Evil to Stay Dead look like they need a new
  "teamwork changes the target filter" primitive; they don't. `kickerTargetRequirements` already
  *replace* `targetRequirements` on any declared cast (`CastSpellHandler.kt:692`,
  `CastSpellEnumerator.kt:2093` — both gated on `declaredCostSlot != null`, not on kicker
  specifically), so the rail carried it unchanged. Recognising that instead of inventing a type is
  the whole point of the bar, and adding a doc bullet naming the shape is the right follow-through.
- **We Say Thee Nay! is one `CounterUnlessDynamicPays` with a `DynamicAmount.Conditional`, not two
  counter effects.** That is the faithful model of a single "counter unless they pay" event whose
  price changes, and the KDoc says why.
- **Earth's Mightiest Heroes wraps only the select-and-move half** in the `ConditionalEffect`,
  sharing the reveal and both `MoveCollectionEffect`s. The branch is as small as the printed
  difference.
- **Two of four suites assert against `getLegalActions`**, and in both directions: the plain cast
  *dropped* when only the wide filter has a legal target, and the plain cast advertising the
  *narrow* `validTargets` when both are offered (`TooEvilToStayDeadScenarioTest.kt:162`). That
  second assertion is the one that catches a silently-widened filter, and it is easy to omit.
- **`WeSayTheeNayScenarioTest.kt:125`** pins the tax at exactly `{4}` (`shouldContain "{4}"` +
  `shouldNotContain "{2}"`) on a six-Forest board where the payment can actually be made — the
  difference between proving "more than 2" and proving the number.
- `LOOP_UNIT.md`'s "Things I'm unsure about" section flags the `isAffordable == false` divergence
  from the unit brief rather than hiding it.

### Issues

#### Important

**1. The PR body opened as a stacked PR against an already-merged base.** The header read
"**Stacked PR — base is `loop-msh-u03`, not `main`.** Merge the stack in order," with a stack
table. `loop-msh-u03` merged upstream as #1750 and is an ancestor of this HEAD, and
`CLAUDE.local.md` rules out stacked PRs across the fork boundary regardless. Posting it would have
told the maintainer to merge a base that is already in `main`.

**2. The gate claim in the PR body was 50 commits stale.** It reported "Re-gated on the current
`main` … rebased onto `main` at `26f5e1c8ca` … 12229 passed". `origin/main` is now `b11a8d70fe`;
`git rev-list --count 26f5e1c8ca..origin/main` is 50, and the branch has been rebased onto that
newer base since. A rebase invalidates the gate.

#### Minor

**3. `f25fc938c1`'s commit message describes two doc edits the commit does not contain.** It lists
"`card-sdk-language-reference.md`: same 601.2c note on the new bullet, and fix the pre-existing
modal … bullet from CR 702.194c to CR 702.194b", but `git show f25fc938c1 --stat` touches no docs
file. The consequence is real: the new bullet carried no CR citation while both card KDocs it
describes cite CR 601.2c + 702.194c.

**4. The second half of that message was also moot.** The modal bullet already reads "CR 700.2 …
CR 601.2b — *not* CR 702.194c", which is correct and came from u03 — as do all five u03 modal card
KDocs. The correction file's carry-forward about five cards still citing 702.194c described text
that no longer exists.

**5. `LOOP_UNIT.md:3`** — "Branch `loop-msh-u04`, stacked on `loop-msh-u03`". Same staleness as #1.

**6. `CruelAlliance.kt:52`** — `TargetCreature(filter = TargetFilter.Creature)` spells out the
factory's own default (`TargetRequirement.kt:156`), so it is exactly `Targets.Creature`. The
sibling card uses the `Targets.*` facade for its kicker target.

**7. `manual-scenarios/sets/msh/loop-msh-u04-new-cards-castable.json`** — Alice's creatures total 9
power (Bears 2, Hill Giant 3, Serra Angel 4) against teamwork thresholds summing to 11 (2 + 4 + 5),
so at most two of the three sorceries can be teamworked per main phase. Workable across a turn
cycle since both mains are stop steps.

### Things checked that are fine

- **Projection** — no battlefield type/P/T reads in the diff. Mana value is a card characteristic
  and the graveyard reads are correctly base-state (the Too Evil KDoc says so explicitly).
- **Filter composition** — `TargetFilter.manaValueAtMost` is
  `copy(baseFilter = baseFilter.manaValueAtMost(n))`, purely additive, so
  `CreatureInYourGraveyard`'s `ownedByYou()` + `zone = GRAVEYARD` cannot be dropped by the
  `.manaValueAtMost(4)` chain.
- **Zero-creature reveal on Earth's Mightiest Heroes** does not stall:
  `SelectFromCollectionExecutor.kt:193` and `:222` both handle `eligibleCards.isEmpty()`, and with
  `showAllCards = true` it emits a 0/0 decision rather than hanging.
- **Raw `GatherCardsEffect` / `SelectFromCollectionEffect` / `MoveCollectionEffect` constructors**
  in `EarthsMightiestHeroes.kt` are the established convention (~20 other card files do it) and are
  not in `FacadeBoundaryTest`'s forbidden list — the module gate passes.
- **`target(...)` vs `kickerTarget(...)` binding** — both produce `EffectTarget.BoundVariable(name)`
  with distinct names per branch, so there is no index collision between the two requirement lists.

### Resolution

All findings except #7 are fixed on this branch. #7 was declined by the author; the scenario stays
as-is, and the reasoning is recorded in `build/pr/loop-msh-u04-correction.md`. #4 was retracted
rather than actioned — the edit it asked for was neither made nor correct.

Fixes landed in commit `ba07d02088` ("Address round-2 review findings for u04") plus in-place
updates to the PR body, which is not tracked by git.

PR #1754, opened 2026-08-10
Add a tap-reason to the tap event, and Agent Maria Hill

bolav:
# Add a tap-reason to the tap event, and Agent Maria Hill

> **Depends on `loop-msh-u04`, which is not merged yet.** u01–u03 have all landed upstream (u03 as
> [#1750](https://github.com/wingedsheep/argentum-engine/pull/1750)); this branch is built locally on
> `loop-msh-u04` so the two can be developed together, but it is **not** a stacked PR — once u04
> lands, this branch is rebased onto `origin/main` and opened on its own against `main`.
> **Do not open this PR until u04 has merged**, or its diff will carry u04's four commits.
>
> Review this unit alone with `git diff loop-msh-u04...loop-msh-u05`.
>
> ⚠️ **The previously reported gate is void.** It read "`BUILD SUCCESSFUL` in 18m 38s — 12229 passed"
> on `main` at `26f5e1c8ca`. That figure does not stand, for a reason worth stating plainly: this
> branch was based on `5abb4d3055`, the **u03 side branch**, which is the *second parent* of the
> merge commit `b11a8d70fe` — not on merged `main`. It was therefore missing 7 mainline commits (the
> MKM card batch, the ISD/INR double-faced reprint-art work, and engine changes to `TokenArtRegistry`,
> `TransformEffectExecutor`, `CardEntityFactory` and `TriggerProcessor`). Since this unit threads
> `TapReason` through `TriggerMatcher`, `TriggerDetector` and `CostHandler`, that code had never been
> compiled against the current engine.
>
> ✅ **Re-gated after rebasing onto the current `loop-msh-u04` / `origin/main` (`b11a8d70fe`).** See
> *Verification* below.

The last teamwork unit for Marvel Super Heroes. It adds the one thing the mechanic was still missing
— a *cause* on the tap event — and ships the card that reads it. With this, **all thirteen MSH
teamwork cards are implemented**.

## The primitive: `TapReason`

`TappedEvent` previously carried only `tappedById` — *who* tapped the permanent. That is not enough
for "whenever this becomes tapped **to pay a teamwork cost**": a teamwork tap, an attack tap, a crew
tap and a mana tap are all performed by the permanent's own controller, so attribution is identical
across all four and cannot separate them.

`TapReason` (`mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/TapReason.kt`) is a serializable
enum — deliberately an enum rather than a boolean, so the next card that wants to name crew, saddle,
convoke or the attack declaration adds a constant instead of a second flag. It threads through:

- `TappedEvent.reason`, defaulted to `UNSPECIFIED` so an event serialized before the field existed
  still decodes;
- the `tap()` atom (`rules-engine/.../core/TapHelpers.kt`), which is where nearly every tap runs and
  the place a cause is attached (two mana-payment sites synthesise a `TappedEvent` by hand for a
  `{T}, Sacrifice this` source and never call the atom — they report `UNSPECIFIED`, and `TapReason`'s
  KDoc names all three so the next cause is stamped everywhere);
- `EventPattern.TapEvent.reason: TapReason?` — **null means "any cause"**, so every existing
  "becomes tapped" trigger is unchanged;
- matching on all three tap paths: `TriggerMatcher` (per event), `TriggerDetector.detectTapBatchTriggers`
  (the "one or more become tapped" batch, narrowed by reason exactly as it is already narrowed by
  tapper), and `AttachmentTriggerDetector` (ATTACHED binding);
- the DSL: `Triggers.becomesTapped(binding, filter, reason)`, `Triggers.OneOrMoreBecomeTapped(filter,
  reason)` for the batch wording, and the SELF facade `Triggers.BecomesTappedForTeamwork`. `reason`
  defaults to `null` — "any cause" — everywhere, so no existing trigger changes meaning.

### Which tap sites are classified, and which are not

**Classified: teamwork only.** `CastSpellHandler` stamps the reason when it pays a
`CostAtom.VariablePermanents(TAP)` additional cost, deriving it from the *declared cast-choice slot*
(`TapReason.forChoiceSlot`) rather than from the atom. That matters: `VariablePermanents(TAP)` is a
generic atom any mechanic may reuse, so the atom alone cannot say a tap was a teamwork tap — the
declaration the caster made can (CR 601.2b; teamwork itself is CR 702.194a, and "cast using teamwork"
is CR 702.194b — all verified against the local rules text, not from memory). It is stamped only on
the additional cost that the declared optional ability actually contributed, so a card's own printed
tap cost can't be relabelled by an unrelated declaration.

**Everything else reports `UNSPECIFIED`, on purpose** — attacking, crew, saddle, convoke, mana
abilities, a `{T}` activation cost, "tap target permanent" effects, and the activated-ability
`VariablePermanents(TAP)` payer in `CostHandler` (an ability cost has no declared slot, so naming a
cause there would be a guess). Attack and crew each do have a single chokepoint and could have been
classified cheaply, but nothing reads them yet, and the asymmetry is the point: an under-claimed
reason only makes a trigger stay silent, while a wrong one makes it fire wrongly. The enum's KDoc and
the SDK reference both carry the recipe for naming the next cause.

### One chokepoint instead of two

The two previous units left `TAP-REASON HOOK (one of two)` / `(two of two)` markers on the two places
that tap for this atom, and suggested folding them together. I took the suggestion: both now call
`VariablePermanentsCost.tapAll(state, chosen, reason)`, and both markers are gone.

Reviewer note on a side effect of that fold: `CostHandler.payVariablePermanentsList`'s TAP branch
used to interleave per-permanent validation with tapping, which rejected a duplicated id incidentally
(the second pass saw it already tapped). Validation now runs as a pass over the pre-payment state, so
there is an explicit `distinct()` guard, mirroring the one `CastSpellHandler.validate` already has. It
sits directly after the chosen list is resolved — ahead of both the count and measure floors it
protects, and covering the SACRIFICE and EXILE branches too, which had only rejected duplicates by
accident. That branch is unreached by any printed card today.

### How the payment loop recognises the declared cost

The stamp is applied to exactly one of the costs being paid — the one the declared optional ability
contributed — so the loop has to be able to pick that cost out of the flattened list. Both sides go
through the *same* `reduceCostAlternatives` call and the match is a membership test on the reduced
list. That is load-bearing rather than stylistic: the reduction flattens an `AdditionalCost.Composite`
into its parts and resolves an `AdditionalCost.Choice` to the leg the caster picked, so comparing a
reduced atom against an unreduced wrapper can never match — and a failed match doesn't fail loudly, it
silently stamps `UNSPECIFIED` and the trigger stays silent. `Costs.additional.TapForTotalPower` emits
a bare atom today, so the shapes are equal either way; reducing both sides is what keeps that true for
the first teamwork cost that is authored as a composite. (Round-2 review finding — see the review
comment.)

## The card

**Agent Maria Hill** (MSH #2) — {W} Legendary Creature — Human Spy Hero, 2/1. "Whenever Agent Maria
Hill becomes tapped to pay a teamwork cost, put a +1/+1 counter on her and draw a card." Verified
field by field against Scryfall; MSH is her first printing, so the canonical definition lives there.
No new vocabulary beyond the primitive above — she composes `Triggers.BecomesTappedForTeamwork` with
`Effects.Composite(Effects.AddCounters(+1/+1, 1, Self), Effects.DrawCards(1))`.

She is the set's only *payer-side* teamwork payoff: `Conditions.TeamworkWasPaid` answers "was this
spell cast using teamwork", but her trigger's subject is a creature on the battlefield that may or
may not have been among the payers.

## Tests

`TapReasonScenarioTest` covers the primitive itself (named for the mechanic, per the one-card-one-file
rule's engine-test exception): a teamwork tap carries `TEAMWORK` while the same cast's mana taps carry
`UNSPECIFIED`; a plain cast claims teamwork nowhere; an attack tap and a crew tap are `UNSPECIFIED`; a
cause-agnostic "becomes tapped" trigger (Interface Ace) still fires on a teamwork tap; the pattern's
description and its cause-agnostic default; and `TappedEvent` serialization, including legacy JSON
with no `reason` field.

It also covers all **three** sites where a tap trigger is matched, because the cause has to be applied
at each of them: the per-event one (`TriggerMatcher`, which every SELF case above goes through), the
ATTACHED one (`AttachmentTriggerDetector`) and the batch one
(`TriggerDetector.detectTapBatchTriggers`). The latter two have no printed card that reads a cause, so
they use two test-only cards — an Aura ("whenever enchanted creature becomes tapped to pay a teamwork
cost, draw a card") and a batch reader ("whenever one or more creatures you control become tapped to
pay a teamwork cost…") — each proved in both directions, teamwork tap fires / attack tap silent. The
batch case also pins that two creatures tapped in one payment draw **one** card, not two (CR 603.2c).

`AgentMariaHillScenarioTest` proves both directions for the card: she triggers when tapped to pay
Repulsor Blast's Teamwork 2 (her printed power 2 reaches the threshold alone), and stays silent when
tapped by attacking, by crewing a Vehicle, by an opponent's tap effect (Crippling Chill — the one
direction where `tappedById` is not her controller, so the place cause and attribution could be
confused), and when a teamwork cost is paid by a different creature. One assertion goes against
`getLegalActions` rather than `execute` — three enumeration bugs in this mechanic have all been in the
advertising path — checking the teamwork cast variant is still offered and prices her at power 2 as an
eligible payer.

## Docs and bookkeeping

`docs/card-sdk-language-reference.md` updated in the same change: the `becomesTapped` entry gained
`reason`, a new `BecomesTappedForTeamwork` entry documents `TapReason` and the under-claiming policy,
and the Teamwork N mechanic block gained the payer-side paragraph. The mtgish bridge gained one
comment noting that `WhenAPermanentBecomesTapped` renders the cause-agnostic trigger only and must not
be used to draft the teamwork wording — no new capability is registered, because the IR has no
teamwork tag and this is a predicate on an existing trigger, not a new emittable primitive.
`backlog/sets/marvel-super-heroes/cards.md` ticks the card (header resynced with `just fix-backlog`);
`mechanics.md`'s Teamwork N section is marked fully shipped, 13 of 13.

## Verification

Gated in two passes: a full per-module sweep after the rebase onto the current `loop-msh-u04` /
`origin/main` (`b11a8d70fe`), then the modules the round-2 fix touches re-run on the final commit.
**12348 passed, 0 failed across every module in the build.**

| Module | Tests | Result | Run on |
|---|---:|---|---|
| `:rules-engine:test` | 10314 | ✅ 0 failed | final commit (round-2 fix) |
| `:mtg-sdk:test` | 427 | ✅ 0 failed | final commit (round-2 fix) |
| `:mtg-sets:test` | 341 | ✅ 0 failed | final commit (round-2 fix) |
| `:game-server:test` | 470 | ✅ 0 failed | post-rebase sweep |
| `:ai:test` | 437 | ✅ 0 failed | post-rebase sweep |
| `:mtg-search:test` | 163 | ✅ 0 failed | post-rebase sweep |
| `:mtgish-tooling:test` | 139 | ✅ 0 failed | post-rebase sweep |
| `:gym:test` | 43 | ✅ 0 failed | post-rebase sweep |
| `:gym-server:test` | 10 | ✅ 0 failed | post-rebase sweep |
| `:gym-trainer:test` | 4 | ✅ 0 failed | post-rebase sweep |

The four modules not re-run on the final commit don't compile against anything it changes: the fix
touches `CastSpellHandler`, `VariablePermanentsCost` and two `rules-engine` tests, plus one
default-valued parameter added to `Triggers.OneOrMoreBecomeTapped` in the SDK — additive, so no
existing caller in any module changes meaning. `:mtg-sdk` and `:mtg-sets` were re-run anyway because
they are what would break if that were wrong.

The arithmetic: `loop-msh-u04`'s 12331 plus this unit's 17 net new tests — `AgentMariaHillScenarioTest`
6, `TapReasonScenarioTest` 11 (8 originally, 4 added by the round-2 fix for the ATTACHED and batch
match sites, 1 removed with `requireAtMostOnePerCast`).

**What this run proves that the previous one could not.** The old base (`5abb4d3055`) was the u03
side branch, so this unit's `TapReason` threading through `TriggerMatcher`,
`TriggerDetector.detectTapBatchTriggers`, `AttachmentTriggerDetector` and `CostHandler` had never
been compiled against the engine changes that merge `b11a8d70fe` brought in — `TriggerProcessor`,
`CardEntityFactory`, `TransformEffectExecutor` and the token executors. They now build and pass
together.

Gate reported per module rather than as a single `just build` figure because a whole-build
invocation cannot complete on this machine — attempts are killed with `Gradle build daemon
disappeared unexpectedly`, each time with zero failing tests. Every module above ran to
`BUILD SUCCESSFUL`. Logs: `build/pr/loop-msh-u05-gate-*-postrebase.log` for the sweep,
`build/pr/loop-msh-u05-gate-reviewfix-*.log` for the final-commit re-runs.

- `just rebless-cards` — only Agent Maria Hill moved in
  `mtg-sets/src/test/resources/snapshots/cards/MSH.json`.
- `just check-card-printing "Agent Maria Hill"` — clean.

## What was **not** done

- No manual playthrough in the web client, and no UX pass from either seat.
- No e2e test. `ClientEvent.PermanentTapped` does **not** carry the new reason — nothing in the UI
  needs it today (the trigger is detected server-side and appears as an ordinary stack object), but a
  future "why did this tap?" affordance would need it plumbed.
- The mtgish emitter cannot draft this wording; that is stated in the bridge rather than worked around.

## Review

Reviewed twice; both rounds' findings are fixed on this branch.

### Round 2 — **1 blocking, 2 important, 2 minor** (full text posted as a PR comment)

- **Blocking:** the tap cause was dropped for any declared teamwork cost that isn't already a bare
  atom — the payment loop compared a *reduced* cost against the *unreduced* declaration, so a
  composite or choice wrapper never matched and fell through to `UNSPECIFIED`. Latent today
  (`TapForTotalPower` emits a bare atom) but silent when it bites. Both sides now go through the same
  reduction — see *How the payment loop recognises the declared cost* above.
- `VariablePermanentsCost.requireAtMostOnePerCast` (added in round 1) is **deleted**: it was propping
  up the broken comparison rather than fixing it, and with the reduction fix it is dead defensive code
  asserting a shape no facade can produce. Its reasoning survives as a comment where the comparison
  lives.
- Two of the three tap-trigger match sites — ATTACHED and batch — had **no test**, and the batch one
  wasn't even reachable from the SDK (`OneOrMoreBecomeTapped` took no `reason`). Parameter added,
  both sites now covered in both directions by two test-only cards.
- Minors: a KDoc that claimed `TapReason` names more constants than it does, and a missing assertion
  that her trigger sits on top of the spell it paid for (CR 601.2i, CR 603.3).

### Round 1 — **0 blocking, 2 important, 7 minor**

Both important findings and five minors
are addressed in a follow-up commit on this branch — the shared-selection invariant was stated and
asserted (round 2 then removed the assertion as the wrong fix and kept the reasoning as a comment;
see above), `TapReason`'s KDoc no longer claims `tap()` is the only place a cause can be
attached, the `distinct()` guard is hoisted, `EventPattern.TapEvent.reason` documents that patterns
must use `null` rather than `UNSPECIFIED`, the card's KDoc drops a misapplied CR 603.2c gloss, and the
opponent-taps-her negative is now tested. The reviewer's drive-by was also taken: three stale **CR
603.2f** citations for "tapping is a transition" (`TapHelpers.kt` ×2, `EventPattern.kt`, and the
`YouTap` entry in the SDK reference) are corrected to **CR 701.26a**, the rule this unit's new code
already cited correctly.


bolav:
**Verdict: approve with changes — no blocking issues.** Findings: 0 blocking, 2 important, 7 minor. Safe to merge once the two Important items are addressed; neither is a wrong-behaviour bug today.

# Review — u05: tap reason (`TapReason`) + Agent Maria Hill

Branch `loop-msh-u05`, reviewed in place against its stacked base: `git diff loop-msh-u04...HEAD`
(20 files, +859/−127). No `origin/main` merge attempted — the container is firewalled and
`loop-msh-u04` is the intended base. Tier: FULL.

## Verdict

The behaviour is right and the SDK shape is right. `TapReason` is a genuinely new axis on the tap
event that no existing primitive could express, the card composes only existing facades, and all
three trigger-matching paths are wired. **No blocking issues.** One latent ordering hazard in how
the reason is derived and one doc invariant that is factually false are worth fixing before merge;
the rest is polish.

## What's good — verified, don't throw away

- **The refactor is behaviourally equivalent, and I checked it rather than took the author's word.**
  In `CostHandler.payVariablePermanentsList` the only line that changed semantics is
  `newState.getEntity(id)` → `state.getEntity(id)` for the already-tapped guard. `projected` was
  already captured from the pre-payment `state` in the old code, and `predicateEvaluator.matches`
  already read `state`, so the *only* case where the interleaved loop and the validate-then-tap
  split differ is a repeated id — which the new `distinct()` guard now rejects explicitly. Failure
  mid-loop discarded the partially-tapped state in both versions. Equivalent.
- **"Unreached by any printed card" holds.** `CostAtom.VariablePermanents` is constructed in exactly
  three places in the SDK (`Costs.kt:270` EXILE, `:289` SACRIFICE, `:471` the teamwork TAP builder),
  and only the teamwork keyword produces the TAP variant. Teamwork is a *spell* additional cost, so
  all 13 teamwork cards (including the 12 on the branches beneath) pay through `CastSpellHandler`,
  never through `CostHandler`. The activated-ability TAP branch is genuinely dead today.
- **Deriving the cause from the declared cast-choice slot is the right call, and the "two optional
  costs" worry does not apply.** `CastSpell.declaredCostSlot` is a single nullable `ChoiceSlot`
  (`GameAction.kt:82`) and `declaredOptionalCosts` filters `cardDef.keywordAbilities` by that one
  slot — the engine has no representation for "kicked *and* teamwork" on one cast, so there is no
  ambiguity to resolve. `VariablePermanents(TAP)` really is generic vocabulary and the atom alone
  could not name the cause.
- **Only two payment sites exist and both now go through `VariablePermanentsCost.tapAll`.**
  `CostPaymentContinuationResumer:80` explicitly errors on `VariablePermanents`, and
  `CostPaymentService` treats it as unaffordable/non-executable. The chokepoint claim is true.
- **All three matching paths wired, and there is no fourth.** `grep` for `EventPattern.TapEvent ->`
  across `rules-engine`, `game-server`, `ai`, `mtg-sdk`, `mtgish-tooling` returns exactly
  `TriggerMatcher.kt:460`, `AttachmentTriggerDetector.kt:128`, `TriggerDetector.kt:2364`. The batch
  rewrite is equivalent when `reason == null`, so no existing tap trigger changed behaviour.
- **Extensibility/serialization done properly.** Enum not boolean; `reason` appended *last* in
  `TapEvent`'s parameter list so every positional caller is untouched; `TappedEvent.reason`
  defaulted so legacy JSON decodes (pinned by a test); `applyTextReplacement` uses `copy(...)` so
  the reason survives text replacement.
- **Every CR number in the diff checks out** against `MagicCompRules_20260619.txt`: 702.194a
  (teamwork additional cost — correct, and correctly *not* 702.194b/c, which are the "using
  teamwork" rider and the target rule u04 dealt with), 701.26a (only untapped permanents can be
  tapped), 601.2b (announcing optional additional costs), 603.2c, 508.1f (attacker tap is not a
  cost).
- **Oracle fidelity is exact.** Name, `{W}`, `Legendary Creature — Human Spy Hero`, 2/1, uncommon,
  #2, artist, flavour text and oracle text all match Scryfall character-for-character.
  `just check-card-printing "Agent Maria Hill"` → `ok: canonical is in the card's earliest real
  printing`. Snapshot delta is the one card.
- **The tests prove the discrimination in both directions**, which is the whole point of this card:
  positive (counter + draw + projected 3/2), and negatives for attacking, crewing, and a teamwork
  cost paid by someone else — plus a `getLegalActions` assertion, which is the right instinct given
  the enumerator's history in this mechanic. `TapReasonScenarioTest` separately pins the axis
  itself, including that a cause-agnostic trigger (Interface Ace) still fires on a teamwork tap.

## Issues

### Blocking

None.

### Important

**1. The reason derivation silently depends on `allAdditionalCosts` ordering plus a shared payment
list.** `CastSpellHandler.kt:2521-2532`. Every `VariablePermanents` atom in `allAdditionalCosts`
taps the *same* `action.additionalCostPayment.variableCostPermanents` list, and card-level costs
from `resolveAdditionalCostsForMode` are added to that list **before** `declaredSlotAdditionalCost`
(`:2353-2357`). So if a card ever carries a printed `VariablePermanents(TAP)` additional cost
*alongside* teamwork, the printed atom taps all of `chosen` first with `UNSPECIFIED`, and the
declared teamwork atom then finds them already tapped — `tap()` returns no event, and the teamwork
cause is lost entirely. Not reachable today (only the teamwork builder emits a TAP atom), but this
is precisely the "surfaces two sets later" shape: the reason's correctness now rests on an
invariant nothing states or asserts. Expected: either state the invariant explicitly at
`:2521` ("at most one `VariablePermanents` atom per cast; the shared `chosen` list assumes it"),
or make it enforced — e.g. compute the reason once before the `allAdditionalCosts` loop and let the
first TAP atom consume it, or reject a second TAP atom outright rather than silently double-paying
one selection. The author's own flag about structural equality (`additionalCost ==
declaredSlotAdditionalCost`) is the *milder* half of this; the ordering hazard is the sharper one
and was not flagged.

**2. `TapReason`'s KDoc states a single-chokepoint invariant that is false, in the paragraph that is
the recipe for adding the next cause.** `TapReason.kt:10-14` — "the engine funnels every one of them
through the same tap atom (`com.wingedsheep.engine.core.tap`). That atom is therefore the only place
a cause can be attached." Two sites construct `TappedEvent(...)` directly, bypassing `tap()`:
`ManaPaymentWindow.kt:191` and `ManaPaymentContinuationResumer.kt:1269`, both synthesising a tap
event for a `{T}, Sacrifice this` mana source that never actually gets `TappedComponent`. Behaviour
today is correct (both default to `UNSPECIFIED`, which is the honest answer for a mana tap), but a
future author classifying, say, a mana cause will follow this doc to one site and miss two. Expected:
soften to "nearly every tap routes through the tap atom" and name the two synthetic-event exceptions,
so the recipe sends the next author to all three.

### Minor

**3. The `distinct()` guard sits below the check it claims to protect, and only covers TAP.**
`CostHandler.kt:955-963`. The comment says "a repeated id would double-count toward the measure
floor above" — but the `minCount` (`:929`) and `minMeasure` (`:934-943`) checks have already run by
then, on the duplicated list. Net behaviour is still rejection, so this is not a bug; but the guard
would be both truthful and more useful hoisted to right after `toPay` is resolved (`:927`), where it
also covers the SACRIFICE and EXILE branches, which today reject duplicates only incidentally
(`paySacrificeList`'s count arithmetic; the EXILE loop's "not found" on the second pass).

**4. `EventPattern.TapEvent(reason = TapReason.UNSPECIFIED)` is constructible and means something no
card wants.** `EventPattern.kt:1543`. It would match only taps the engine has *not* classified — a
predicate whose meaning silently changes the day someone classifies a new cause, and one that renders
with no clause at all (`:1558` special-cases it out of the description). The author clearly noticed
the shape but didn't close it. Expected: document on the `reason` property that patterns must use
`null` or a classified cause, or normalise `UNSPECIFIED` → `null` inside
`Triggers.becomesTapped` (`Triggers.kt:1492`).

**5. Test gap — the mana and `{T}`-activation negatives are only proven on lands, not on a
creature.** `TapReasonScenarioTest.kt:68-73` asserts the Mountains tapped for the same cast carry
`UNSPECIFIED`, which is a fair proxy, but neither file taps *Agent Maria Hill herself* for mana or
for a `{T}` activation cost — the two causes `TapReason`'s own KDoc lists. Structurally these cannot
regress (only one site ever passes a non-default reason), so this is belt-and-braces, but the card's
test file is where a reader looks for "and not for mana".

**6. Test gap — the "not by an opponent's effect" claim is untested.**
`Triggers.kt:1446-1449` says the trigger never fires "for the same creature tapped to attack, to
crew, for mana, **or by an opponent's effect**". No test taps her with a "tap target creature"
effect. Cheap to add via `TapUntapExecutor`, and it is the one direction where `tappedById` and
`reason` could plausibly be confused.

**7. Card KDoc misapplies CR 603.2c.** `AgentMariaHill.kt:26-28` — "one cast that taps her and two
others fires this once (CR 603.2c — one trigger event, one occurrence per tapped permanent)". 603.2c
is about one *event* containing multiple *occurrences*, which is the ANY-binding/batch case; her
trigger fires once because she is one permanent with a SELF binding, not because of 603.2c. The
gloss "one occurrence per tapped permanent" actually argues the opposite of the sentence it
supports. Reword or drop the citation.

**8. Pre-existing CR misattribution in the immediate neighbourhood (out of scope, drive-by).**
`TapHelpers.kt:72`, `EventPattern.kt:1531` and the `YouTap` entry in
`docs/card-sdk-language-reference.md` all cite **CR 603.2f** for "tapping is a transition / an
already-tapped permanent emits no tap event". 603.2f is *"if a triggered ability's trigger condition
is met but the object is at no time visible to all players, the ability does not trigger"* — nothing
to do with tapping. The correct citation is **CR 701.26a**, which this diff's new
`VariablePermanentsCost.tapAll` KDoc gets right. Not introduced by this unit, so not its
responsibility, but the new code now sits directly beside three copies of the wrong number.

**9. `ClientEvent.PermanentTapped` does not carry the reason** (`ClientEvent.kt:1039`). Correctly
scoped out — no new event type, so the anti-corruption layer needs no new branch, and nothing in the
UI reads a tap cause today. Noting only so the corrector does not treat it as an oversight.

## Recommendation

Merge-able after (1) and (2). Concretely:

1. `CastSpellHandler.kt:2521` — state or enforce the "at most one `VariablePermanents` atom per
   cast" invariant that the shared `chosen` list and the ordering of `allAdditionalCosts` currently
   rely on for the reason to survive.
2. `TapReason.kt:10-14` — correct the single-chokepoint claim and name
   `ManaPaymentWindow.tapOrSacrifice` / `ManaPaymentContinuationResumer` as the two synthetic
   `TappedEvent` sites.
3. Optional but cheap, in descending value: hoist the `distinct()` guard (3), add the
   opponent-taps-her negative (6), fix the 603.2c gloss (7), and — as a separate drive-by if the
   corrector wants it — replace the three stale 603.2f citations with 701.26a (8).

No `just test` re-run was performed at this stage (Stage A already gated it green, and Gradle
daemons are being OOM-killed in this container). `just check-card-printing "Agent Maria Hill"` was
run and is clean.

bolav:
## Review round 2 — one real bug, one deleted invariant, and the two untested match sites

Re-reviewed the branch after round 1's fixes landed. **1 blocking, 2 important, 2 minor** — all
addressed in the commit on top.

### Blocking

**The teamwork tap cause is dropped for any declared cost that isn't already a bare atom.**
`CastSpellHandler` decides a tap's `TapReason` by asking whether the additional cost it is currently
paying *is* the one the caster declared:

```kotlin
val reason = if (additionalCost == declaredSlotAdditionalCost) { … }
```

The left side comes out of `flattenedAllCosts` — the list after `reduceCostAlternatives` — while the
right side is the raw `AdditionalCost` read straight off the declared optional ability. That
reduction is not identity-preserving: it flattens an `AdditionalCost.Composite` into its parts and
resolves an `AdditionalCost.Choice` down to the leg the caster picked. So the moment a teamwork cost
is authored as anything but a bare `Atom` — a composite ("tap creatures with total power N **and**
discard a card"), or a choice — the reduced atom being paid is never `==` the unreduced wrapper, the
comparison falls through to `TapReason.UNSPECIFIED`, and Agent Maria Hill silently does not trigger
on a cost she was tapped to pay.

It is latent rather than live today: `Costs.additional.TapForTotalPower` emits a bare atom, so every
one of the thirteen shipped teamwork cards takes the matching branch. But the failure mode is a
trigger that quietly stays silent, with nothing in the logs, and the first card that composes a
teamwork cost with anything else walks straight into it.

**Fixed** by putting the declared cost through the *same* reduction and testing membership:

```kotlin
val declaredSlotCosts: List<AdditionalCost> = reduceCostAlternatives(
    listOfNotNull(declaredSlotAdditionalCost), currentState, action.playerId, action.additionalCostPayment
)
…
val reason = if (additionalCost in declaredSlotCosts) { … }
```

Both sides now live in the same representation, which is the only way this comparison can be stated
without depending on the shape a card author happened to choose.

### Important

**`VariablePermanentsCost.requireAtMostOnePerCast` is deleted.** Round 1 added it to assert the
"one `VariablePermanents` selection per cast" invariant that the *old* identity check leaned on: with
two such atoms, the un-reduced comparison could match the wrong one. That was propping up the
comparison rather than fixing it. With the reduction fix above, the branch no longer depends on the
atom being unique, and what remains is an `IllegalStateException` thrown from the cast path for a
shape no facade can produce — dead defensive code plus a test that only ever called the helper
directly, never through a cast, so it proved nothing about the engine. The reasoning it captured is
worth keeping, and it is kept: as a comment on the declared-slot reduction, where the next reader of
that comparison will actually be standing. 39 lines of code and a test go away.

**Two of the three match sites had no test at all.** The cause has to be applied wherever a tap
trigger is matched, and there are three such places — `TriggerMatcher` (per event),
`AttachmentTriggerDetector` (ATTACHED binding) and `TriggerDetector.detectTapBatchTriggers` (the
"one or more" batch). All three narrow by `reason` in the source, but every test on the branch went
through the first one, because Agent Maria Hill is a SELF trigger and she is the only card that
reads a cause. A regression at either of the other two was invisible.

They have no printed card, so `TapReasonScenarioTest` now defines two test-only cards and covers
each in both directions:

- **Teamwork Tether** — an Aura, "whenever enchanted creature becomes tapped to pay a teamwork cost,
  draw a card". Fires when the enchanted creature pays Repulsor Blast's teamwork cost; silent when
  the same creature is tapped by attacking.
- **Teamwork Roll Call** — "whenever one or more creatures you control become tapped to pay a
  teamwork cost, draw a card". Two creatures tapped in one payment draw **one** card, not two
  (CR 603.2c); the same two tapped by declaring attackers draw none — which also pins the narrowing
  behaviour: a batch that mixes causes is filtered down to the matching taps and still fires once,
  rather than being discarded wholesale, exactly as `YouTap`'s batch narrows by tapper.

**`Triggers.OneOrMoreBecomeTapped` had no `reason` parameter**, which is why the batch site was
untestable in the first place: the detector reads `TapEvent.reason`, but no card could set it
through the batch facade. Added, defaulted to `null` (cause-agnostic, so every existing caller is
unchanged), with the narrowing semantics documented on the function and in the SDK reference entry.

### Minor

- **A KDoc claim that outran the enum.** `AgentMariaHillScenarioTest`'s header said "the remaining
  two causes `TapReason` names — a mana ability and a `{T}` activation cost". `TapReason` names
  exactly two members, `UNSPECIFIED` and `TEAMWORK`; a mana ability and a `{T}` cost are two of the
  *unclassified* causes listed under `UNSPECIFIED`, not constants. Reworded — the deliberate
  under-claiming is the point of the design and the doc shouldn't read as though the enum is larger
  than it is.
- **Stack order was asserted nowhere.** The teamwork cost is paid *during* casting, so the trigger
  can only reach the stack once the spell has become cast (CR 601.2i), and CR 603.3 then puts it on
  top — it resolves before the spell it paid for. Nothing tested that; the existing test resolved
  the whole stack and checked the end state, which passes either way. `AgentMariaHillScenarioTest`
  now asserts the stack is 2 deep with her trigger on top before resolving.

### Not changed

- The under-claiming policy itself (attack, crew, saddle, convoke and mana taps all report
  `UNSPECIFIED`) stands as designed and as documented — a missing cause makes a reader stay silent,
  a guessed one makes it fire wrongly.
- `ClientEvent.PermanentTapped` still doesn't carry the reason. Nothing in the UI reads it today.

PR #1776, opened 2026-08-11
Converge the copy paths on one CopyExceptions vocabulary (+ 3 MSH cards)

bolav:
# Converge the copy paths on one `CopyExceptions` vocabulary (+ 3 MSH cards)

Unit `u06` of the MSH missing-mechanics run — the "Copy-with-exceptions: name, added types, longer
durations" section of `backlog/sets/marvel-super-heroes/mechanics.md`.

## The gap

Three MSH cards are copy effects with long "except …" clauses, and none of them could be modelled:

- `EachPermanentBecomesCopyOfTargetEffect` exposed only `addedKeywords`, `powerOverride`,
  `toughnessOverride` and `retainActivatingAbility` as copy exceptions. No name override, no added
  types, and — the one that actually kills a card — no way to *remove* a supertype.
- Its token sibling `CreateTokenCopyOfTargetEffect` already had a dozen exception riders and its own
  hand-rolled copy of the type-line arithmetic. Two implementations of CR 707.9, drifting.
- The permanent executor honoured only `Duration.Permanent` / `EndOfTurn` / `UntilNextEndStep`,
  silently degrading anything else to permanent.

## The design — convergence, not more riders

Rather than bolting five more fields onto one effect, the "except …" half of a copy effect became a
value type in its own right:

- **`CopyExceptions`** (`mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/effects/CopyExceptions.kt`)
  — `nameOverride`, `addedKeywords`, `addedSupertypes` / `removedSupertypes`,
  `addedCardTypes` / `overrideCardTypes`, `addedSubtypes` / `overrideSubtypes`,
  `addedColors` / `overrideColors`, `powerOverride` / `toughnessOverride`, `noManaCost`.
  Everything on it is a *copiable* value (CR 707.2), so a later copy of the copy sees it. The
  add-vs-override split is not an invention: **CR 205.1a** makes a stated card type or subtype
  *replace* the existing one by default, and **CR 205.1b** is the exception that retains the prior
  types, keyed on the phrase "in addition to its other types" / "still a [type]". This set contains
  one card of each kind.
- **`CopyExceptionApplier`** (`rules-engine/.../handlers/effects/copy/CopyExceptionApplier.kt`) — the
  single place that turns a `CopyExceptions` into a modified `CardComponent`. Four copy paths call
  it: the permanent-becomes-a-copy executor, both token-copy executors (targeted and self), and the
  `EntersAsCopy` clone path in `ModalAndCloneContinuationResumer`. The token executor's Aura-host
  type-line probe (which used to compute the post-exception type line separately from the token it
  then built) shares `CopyExceptionApplier.typeLine` too. Deliberately still outside it: the two
  paths whose only exception is the boolean `removeLegendary`, and which therefore have no arithmetic
  to share — `CreateTokenCopyOfEquippedCreatureExecutor` (Helm of the Host) and spell copies
  (`StormCopyEffectExecutor`, CR 707.10, where the copy lives on the stack). The KDoc names exactly
  that set rather than claiming universality.
- **All three copy effects take an `exceptions: CopyExceptions` field.**
  `EachPermanentBecomesCopyOfTargetEffect.exceptions` replaces the three flat riders it had;
  `CreateTokenCopyOfTargetEffect` and `CreateTokenCopyOfSourceEffect` gained the same field
  (surfaced on the `Effects.CreateTokenCopyOfTarget` / `CreateTokenCopyOfSelf` facades). That
  matters more than it looks: a read-only view over the old riders would have made the convergence
  one-directional — a new exception added to `CopyExceptions` would be expressible on the permanent
  path and *unreachable* on the token ones without also adding a rider, which is the drift the type
  exists to end.
  Their ~20 existing flat riders stay (card definitions author them; their serialized shape
  shouldn't churn) and are folded in by one shared rule, `CopyExceptions.over(base)`: additive axes
  union, replacing and single-valued axes take the modern field and fall through to the riders.
  `CopyExceptions.None.over(base) == base`, so a card using only riders is bit-for-bit unchanged —
  confirmed by `:mtg-sets:test` passing with no snapshot re-bless. The riders are frozen; new
  exceptions go on `CopyExceptions`. `retainActivatingAbility` deliberately stayed on the effect: it
  isn't a characteristic.
- **`Duration.UntilYourNextTurn`** in the copy-revert path — `RevertCopyAtYourNextTurnComponent(playerId)`,
  a sibling to the two existing revert markers, expired in
  `CleanupPhaseManager.expireUntilYourNextTurnEffects` alongside every other "until your next turn"
  effect (post-untap of that player's next turn). Registered for serialization.

A latent bug fell out of the merge: the permanent path applied a P/T override with
`baseStats?.let { … }`, so it dropped the override entirely when the copy source had no base stats.
That is exactly Absorbing Man copying a land. The shared applier conjures base P/T instead.

## Cards

- **Shuri, Wakandan Inventor** [75] — `ModifySpellCost(YouCast(Artifact), ReduceGeneric(1))` for the
  discount; the ability is the Fleeting Reflection two-target copy shape (`affected` = the artifact
  that changes identity, `target` = the copy source, wrapped in `TargetOther` so "a **second**
  target" can't repeat the first) with `CopyExceptions(removedSupertypes = {LEGENDARY})`. Without the
  removal direction the copy is a second legendary permanent with the same name under one controller
  and the legend rule (**CR 704.5j**) bins one of them before it does anything.
- **Absorbing Man** [199] — `Triggers.FirstMainPhase`, an optional target across artifact / non-Aura
  enchantment / land, `Duration.UntilYourNextTurn`, and the additive exception set: name, legendary,
  creature, Human Villain, 4/4, vigilance. The long duration is load-bearing — copying replaces his
  card component wholesale, so his own trigger is gone while the copy is up and back in time to fire
  again once the copy reverts at his next untap step.
- **Taskmaster, Mercenary Mimic** [232] — the same shape with a cross-zone target (battlefield **or**
  graveyard, via `TargetFilter.or`) plus `sourceFromAnyZone`, and the *replacing* reading of the type
  clause (`overrideCardTypes` / `overrideSubtypes`) — see the open question below.

## Rules cited (checked against `MagicCompRules_20260619.txt`)

- **CR 707.2** — copiable values; what a copy takes.
- **CR 707.9b** — "some copy effects modify a characteristic as part of the copying process"; the
  modified value becomes part of the copy's copiable values.
- **CR 205.1a** — a stated card type or subtype *replaces* the object's existing ones.
- **CR 205.1b** — the retention exception: "in addition to its other types" / "still a [type]" keeps
  the prior card types, supertypes and subtypes. Together with 205.1a this is the add-vs-override
  switch. (CR 707.9d, cited in the first revision of this PR, governs only whether a
  *characteristic-defining ability* is copied — it is not that switch. Corrected everywhere.)
- **CR 704.5j** — the legend rule, i.e. what "except it isn't legendary" is protecting against.

## Open question for review

**Taskmaster's type clause.** His oracle text reads "he's a legendary Human Mercenary Villain
creature" with **no** "in addition to his other types", while Absorbing Man in the same set has the
phrase. Given CR 205.1a/205.1b I read Taskmaster as *replacing* card types and subtypes: copying a Goblin
Wizard makes a Human Mercenary Villain, and copying an artifact creature drops the artifact type.
Scryfall has no rulings for the card yet. If a reviewer reads it the other way it is a two-field
change in `TaskmasterMercenaryMimic.kt`. Related, smaller call: legendary is `addedSupertypes` rather
than an override (`CopyExceptions` has no supertype-override field by design), so a copied Snow
creature would keep Snow.

## Tests

- `CopyExceptionsTest` (`rules-engine/.../handlers/effects/copy/`) — the mechanic: add vs override on
  each axis (card types included), an addition on an axis with no override of its own, the
  add-then-remove supertype order, P/T conjured onto a copy source that had none, the half-specified
  override, and pins that both token effects' flat riders map onto the same vocabulary (the
  convergence itself) — including that `None.over(base)` is `base` untouched, that a modern field
  and a legacy rider combine rather than dropping each other, and that a stated type line renders
  its article while a bare supertype clause doesn't.
- `ShuriWakandanInventorScenarioTest` — the copy is not legendary, both same-named permanents survive
  the legend-rule state-based action, the copy wears off at cleanup, and the artifact discount lets a
  `{1}{U}` artifact be cast off one Island.
- `AbsorbingManScenarioTest` — additive typing and 4/4 conjured onto a noncreature source, vigilance,
  the copy surviving the opponent's turn and reverting at his own untap, the trigger firing again
  afterwards and *resolving onto a second target* (a land — the other branch of his filter, and the
  case with no base P/T to copy), and the declined optional target being a clean no-op. Plus three
  assertions driven through `getLegalActions` rather than through `execute`, one per enumerator the
  resolve-by-definition-id fix touches that a printed MSH card can reach: the copied Futurist Forge's
  activated ability, The Mind Stone's mana ability, and Dependable Quinjet's Crew 4. Each fails if
  the enumerator resolves by `CardComponent.name`, because the printed Absorbing Man has none of the
  three.
- `TaskmasterMercenaryMimicScenarioTest` — the replacing type line (including an artifact creature
  copy, where the artifact card type is dropped), the graveyard clause of the cross-zone target (with
  the copied card staying put), and the until-your-next-turn window.

## Verification

All run through `just` / `scripts/gradle-locked` (machine-global lock), per module — a whole-build
run gets OOM-killed on this box.

```
just test-rules                                 # BUILD SUCCESSFUL — 10,340 tests, 0 failed
scripts/gradle-locked :mtg-sets:test            # BUILD SUCCESSFUL (snapshots + FacadeBoundaryTest)
just rebless-cards                              # BUILD SUCCESSFUL — diff reviewed, see below
scripts/gradle-locked :ai:compileKotlin :game-server:compileKotlin \
    :gym:compileKotlin :mtg-search:compileKotlin    # BUILD SUCCESSFUL (downstream of the SDK change)
just check-card-printing "Shuri, Wakandan Inventor"      # ok — canonical is msh
just check-card-printing "Absorbing Man"                 # ok — canonical is msh
just check-card-printing "Taskmaster, Mercenary Mimic"   # ok — canonical is msh
just fix-backlog                                # MSH 239 → 242 / 276
```

After the review-correction commit, both gates were re-run on the corrected tree:

```
just test-rules                                 # BUILD SUCCESSFUL — 10,344 tests, 0 failed
scripts/gradle-locked :mtg-sets:test            # BUILD SUCCESSFUL — snapshots unmoved
```

**Final gate, after the second review round** (commit *Finish the copy-exception convergence…*),
run as two halves — because `just test` is Gradle-only and **no `just` recipe reaches the web
client**, which this branch changes by ~1000 lines:

```
just test                                       # BUILD SUCCESSFUL — rules-engine 10,563 tests,
                                                #   mtg-sets (snapshots + FacadeBoundaryTest) and
                                                #   game-server green; 0 failed
cd web-client && npm run typecheck              # clean
cd web-client && npm test                       # 39 files, 553 tests, 0 failed
```

`:mtg-sets:test` passing **without a re-bless** is the evidence for the "bit-for-bit unchanged"
claim about the new `exceptions` field on the two token effects: a defaulted field that no existing
card sets changes no serialized shape.

One run of `just test` in between failed on `ConniveTargetingTest` with a
`TimeoutCancellationException`. That class contains no reference to anything in this diff and passes
in isolation (`just test-class ConniveTargetingTest`); it is this box's memory pressure, not a
regression. The full gate was re-run rather than assumed, and is green.

Logs, in the worktree: `build/pr/loop-msh-u06-gate.log` (full rules suite),
`loop-msh-u06-gate-correction.log` (both re-runs after the review corrections),
`loop-msh-u06-sets.log`, `loop-msh-u06-diag.log`. The accounting against the review is
`build/pr/loop-msh-u06-correction.md`.

An earlier run of the full suite had 3 failures, all in the two new scenario tests and all
test-harness bugs, not engine behaviour: `passUntilPhase` returns immediately when the game is
already at the requested phase/step (so two upkeep hops in a row were one hop), and
`ScenarioBuilder` starts both libraries empty, so the opponent decked out on their draw step and
silently stalled the turn advance. Both fixed in the tests; the final run above is clean.

**Expected snapshot movement beyond the three new cards.** `Likeness Looter` (WOE) and
`Mimeoplasm, Revered One` (DFT) are the only existing cards that set the three fields that moved into
`exceptions`, so their serialized ability JSON changes shape. Nothing else in any set snapshot moves —
that was checked, not assumed.

## Playtest fixes (third commit)

The cards were then played manually in the running app, which turned up two bugs neither the scenario
tests nor the review could see — both live on the *player-facing* path (legal-action enumeration and
the targeting UI) that a test driving `GameAction`s directly never touches. Full accounting:
`build/pr/loop-msh-u06-playtest-fix.md`.

**1. Taskmaster's battlefield ∪ graveyard target could not be chosen in the UI.** Server-side
targeting was correct (`TargetFinder.findObjectTargets` recurses over the union's clauses, and
`TaskmasterMercenaryMimicScenarioTest` passes). The client's `TargetingOverlay` reduced "where do
these targets live" to one all-or-nothing boolean with exactly two render branches — a full-screen
pile picker when *every* valid target sits in a graveyard/exile, or board clicking otherwise. A mixed
union fell to board clicking, where a graveyard pile isn't individually clickable. Sorceress's
Schemes works only because graveyard ∪ exile is pile-only; Taskmaster is the repo's first *mixed*
union.

The boolean was not loosened — the restriction it encodes is real. A third case was added: the
targeting banner stays up (board clickable) *and* offers a `Graveyard (n)` button that opens the same
`ZoneCardTargetingOverlay`, whose existing "View Battlefield" button hands control back. Both routes
call the same `addTarget`/`removeTarget`, so picks from either side satisfy one requirement.

No payload change was needed and none was made: the client already has per-target zones from
server-sent card state (`gameState.cards[id].zone`), which is what the existing pile detection reads.
`TargetEnumerationUtils.getTargetZone` returns `null` for a union deliberately, and says so. The
dumb-terminal rule holds — the client only picks the widget that can reach a target the server
already marked legal.

The split moved into a pure `partitionTargetsByZone` helper in `web-client/src/utils/targeting.ts`,
next to the all-or-nothing `getPileTargetCards` it complements, because the web-client has no
jsdom/testing-library and pure `utils/` helpers are the only unit-testable seam.

**2. A copy with a `nameOverride` lost the copied card's activated abilities.** Absorbing Man copying
Futurist Forge was never offered its `{3}{U}, Sacrifice this artifact: Draw two cards`.
`ActivatedAbilityEnumerator` resolved the permanent's definition by `CardComponent.name`, and this
unit's own `nameOverride` is precisely what makes the name disagree with `cardDefinitionId`. The name
resolved to the printed Absorbing Man, who has no activated abilities.

`ActivateAbilityHandler` had *always* resolved by `cardDefinitionId`, so the ability was **executable
but not enumerable** — which is why every scenario test passed while the player saw no button.

Six enumerator sites moved to `cardDefinitionId`, chosen by triage rather than a sweep: each reads a
battlefield permanent (so a renamed copy reaches it) *and* each had an authoritative counterpart
already resolving by id, i.e. each was a live enumerate-vs-execute asymmetry —
`ActivatedAbilityEnumerator` (own + "any player may activate"), `ManaAbilityEnumerator`,
`CrewEnumerator` (vs `CrewVehicleHandler`), `SaddleEnumerator` (vs `SaddleMountHandler`), and
`CombatEnumerator`'s `CanBlockAnyNumber` blocker cap (vs `BlockPhaseManager`). The other ~25
`getCard(x.name)` sites read hand/library/graveyard/exile/command cards or stack objects, where no
copy effect can rename anything today; they are listed individually with reasons in the accounting
file, along with the one follow-up (the seven stack sites become wrong the day spell copies route
through `CopyExceptions`).

Regression tests go through the player-facing path on purpose: `AbsorbingManScenarioTest` asserts the
copied ability is **enumerated** by `getLegalActions` (then executes it), and
`web-client/src/utils/targeting.test.ts` covers the mixed union reporting both routes.

Gates: `just test-rules` — 10,345 tests, 0 failed — plus `npm run typecheck` and `npm test` in
`web-client` (37 files / 525 tests). `:mtg-sets:test` was not re-run: no card definition, SDK type or
snapshot moved.

One caveat, disclosed rather than papered over: the `just test-rules` *recipe* exited non-zero even
though the build printed `BUILD SUCCESSFUL` and every test passed. The error is a
`CorruptedCacheException` on `/workspace/.gradle/9.6.1/fileHashes/fileHashes.bin` raised while
writing test *reports* after the test task finished. That cache directory is shared with sibling
containers, so this is environmental; per the repo's guidance it was reported and left alone rather
than cleaned. `just test-class AbsorbingManScenarioTest` on the same tree finished `BUILD SUCCESSFUL`.
Logs: `build/pr/loop-msh-u06-gate-rules.log`, `build/pr/loop-msh-u06-gate-absorbingman.log`,
`build/pr/loop-msh-u06-gate-playtest-fix.log`.

## Playtest fix, round two (fourth commit) — the *other* targeting path

Re-playtesting Taskmaster showed the graveyard half was **still** unselectable. The fix above was
correct but landed in the wrong place: the client has two targeting paths, and only one of them was
taught the mixed case.

- Casting a spell / activating an ability goes through `targetingState` → `TargetingOverlay`. That is
  what the third commit fixed.
- A **triggered** ability's targets arrive as a `PendingDecision` → `DecisionUI` → `ChooseTargetsUI`,
  a separate component that was still calling the all-or-nothing `getPileTargetCards`. Taskmaster's
  ability is a trigger, so it never touched the fixed path.

`ChooseTargetsUI` now gets the same three-way routing, and the two paths were converged onto one
shared decision so they can't drift again:

- `routeTargetsByZone` (+ `describePileZones`) in `web-client/src/utils/targeting.ts` — `board` /
  `pile` / `mixed` plus the pile label — built on the existing `partitionTargetsByZone`.
  `TargetingOverlay` was rewritten onto it; its local label IIFE is gone.
- `getPileTargetCards` was **deleted**: `ChooseTargetsUI` was its only caller, and leaving two helpers
  that answer the same question differently is what caused this.
- The decision path's per-requirement walk moved into a pure reducer,
  `web-client/src/components/decisions/chooseTargetsWalk.ts` (`chooseTargetsWalkReducer` +
  `chooseTargetsView`), so the mixed flow is unit-testable — the web-client has no
  jsdom/testing-library, so a pure seam is the only way to test it. Behaviour of the walk is
  unchanged: advance, submit-all-at-once, Back with restored picks, dedup across requirements.
- Mixed rendering mirrors the action path's idiom: `BattlefieldTargetingUI` keeps the banner up
  (permanents stay clickable) and gains a `Graveyard (n)` button; `GraveyardTargetingUI` /
  `ZoneSelectionUI` gain an `onViewBattlefield` handoff. The picks travel with the player in both
  directions, so either half fills the same requirement and either side's Confirm submits them.
  Single-select pile toggles now replace at the cap, the rule `toggleDecisionSelection` already used
  on the board — without it a mixed "up to one" slot deadlocks.

**Audit for a third copy of the bug.** Mixed-zone `SelectCardsDecision`s were already safe
(`zones.size > 1` → `MultiZoneSelectionUI`, which renders every zone together). One latent gap was
found and closed: a `useTargetingUI` selection whose options all sit in a graveyard/exile pile would
have gone to the board-click `BattlefieldSelectionUI`; it now falls through to the modal picker. No
shipped card reaches it today — engine callers guard it themselves — but
`ChainSpellContinuationResumer` hardcodes `useTargetingUI = true` for a copied spell's targets and is
safe only because the Chain cycle targets permanents and players. Everything else
(`CardSelectionDecisionUI`, `LibrarySearchUI`, sacrifice / discard / behold flows) is single-zone by
construction.

Gates (client-only change, no gradle): `npm run typecheck`, `npm test` (38 files / 537 tests),
`npm run build` — all green. Log: `build/pr/loop-msh-u06-gate-decision-targeting.log`; accounting:
`build/pr/loop-msh-u06-decision-targeting-fix.md`.

## Playtest fix, round three (fifth commit) — the pile picker's verb was a false statement

With the graveyard half finally selectable, playtesting Taskmaster surfaced the copy itself: the
picker said `Optional: choose up to one card to return to your hand, or decline.` with a
`Return to Hand` button. Nothing was returned to anyone's hand — the card is copied and stays in the
graveyard. Client-only, no gradle. Full accounting: `build/pr/loop-msh-u06-pile-verb-fix.md`.

`derivePileAction` (`web-client/src/utils/targeting.ts`) infers the verb by sniffing the decision's
`effectHint` prose for `battlefield`, `shuffle` + `library`, or `exile`, and **fell through to a
`Return to Hand` default** when none matched. `EachPermanentBecomesCopyOfTargetEffect`'s description
("… becomes a copy of up to one target creature on the battlefield or creature card in a graveyard
…") matches none of the three, so Taskmaster silently inherited reanimation wording.

The bug fixed here is the **default**, not the missing branch: any pile effect outside those three
keywords told the player the wrong thing would happen. A wrong verb is worse than a vague one — it is
a false statement about what the game is about to do.

- **The fallback is neutral now** — `Confirm Target` / "choose … to target", which is the exact label
  `GraveyardTargetingUI` already uses on its mandatory path, so the neutral case borrows the
  component's existing voice rather than inventing a register.
- **Return-to-hand became an explicit branch** (word-bounded `\bhands?\b`, so "beforehand" can't trip
  it). `MoveToZoneEffect(destination = HAND)` renders "Return … to its owner's hand", so nothing that
  read "Return to Hand" *correctly* lost it.
- **A copy branch was added ahead of everything else** — `becomes a copy of` / `token that's a copy
  of` → `Copy`. It must precede the battlefield branch: Taskmaster's hint names the battlefield as a
  place to *target*, not a destination. The token-copy wording is included because it is the same
  class of falsehood about the picked card (the token arrives on the battlefield; the card does not).

Branch order is first-match-wins and is now documented in the KDoc and pinned by a test for every
multi-keyword edge: copy > battlefield (new, load-bearing), battlefield > exile (blink), exile > hand
(`RemovalEffects.kt:518` renders "cards from your hand or graveyard" — the hand as a *source*), and
the `leaves the battlefield` boilerplate stripped first.

**Deliberately not fixed here:** the durable fix is a per-requirement **action hint on
`TargetRequirementInfo`** so the client stops sniffing prose — a server change, out of scope for a
client-only round. The `TODO(per-requirement action hints)` in `GraveyardTargetingUI.tsx` stays and
was sharpened with what this bug proved: prose sniffing doesn't merely go vague, it produces an
actively misleading verb, and lengthening the keyword list is not convergence.

Gates: `npm run typecheck`, `npm test` (38 files / 543 tests), `npm run build` — all green. No gradle
was run. Log: `build/pr/loop-msh-u06-gate-pile-verb.log`.

## Not done

- No UX pass beyond the four playtest bugs above. The mixed-union picker has not been screenshotted
  for this PR, and the two triggered mimics still raise an optional target prompt every one of their
  controller's turns; nobody has judged how that cadence reads over a long game.
- No e2e / Playwright coverage.
- No AI advisor work — the built-in AI will pick copy targets with generic heuristics.
- This PR came out of an autonomous loop and has had two independent review rounds. Round one found
  0 blocking / 3 important / 8 minor; that follow-up commit fixed 10 and declined 2 with reasons
  (`build/pr/loop-msh-u06-correction.md`, review preserved as
  `build/pr/loop-msh-u06-review-round1.md`). Round two found 0 blocking / 2 important / 5 minor and
  ran the gate independently; six of those are fixed in the final commit and the seventh was a
  documentation-only note folded into this body (posted as a PR comment — see
  `build/pr/loop-msh-u06-review.md`). Still open by choice: the copy-revert path emits no
  `GameEvent` for the client to animate — a pre-existing gap shared by all three revert markers,
  worth one follow-up for all of them.


bolav:
# Review — `loop-msh-u06` (copy-with-exceptions + 3 MSH cards)

Second independent review round, `review-changes` (SDK-elegance lens) over the full diff.
Reviewed in place on `loop-msh-u06`; `origin/main` is already fully contained
(`git rev-list --count HEAD..origin/main` = 0), so no merge was needed and the diff is post-merge
reality as-is.

**Verdict:** behaviour is right and the SDK shape is the right one. This is convergence, not a new
one-off type: `CopyExceptions` + `CopyExceptionApplier` delete four hand-rolled copies of the same
type-line arithmetic and the three cards fall out of it as pure data. **0 blocking, 2 important,
5 minor.**

All findings below have since been fixed on this branch (commit *"Finish the copy-exception
convergence and cover the enumerator fixes"*), except **Minor #5**, which was a documentation-only
note and is folded into the PR body. Each entry records what was done.

---

## Verified independently

Not taken on trust from the PR body:

- **Full gate re-run from scratch.** `just test` → BUILD SUCCESSFUL, 0 failed. `:rules-engine:test`
  10,563 tests; `:mtg-sets:test` (snapshots + `FacadeBoundaryTest`) and `:game-server:test` green.
- **Web-client gate, which no `just` recipe reaches.** `npm run typecheck` clean; `npm test`
  39 files / 553 tests. See Minor #5.
- **Oracle text** for all three cards fetched from Scryfall — exact match, including the presence of
  "in addition to his other types" on Absorbing Man and its **absence** on Taskmaster.
- **Every CR number** checked against `MagicCompRules_20260619.txt`: 205.1a, 205.1b, 400.7, 601.2c,
  702.128a, 704.5b, 704.5j, 707.2, 707.9 / 707.9b. All correct. In particular 205.1a's text really
  does say a stated subtype "replaces any existing subtypes from the appropriate set", and 205.1b
  really is keyed on the phrase "in addition to its other types" — so the central design call is
  resting on the right rule.
- **Printing placement** — `just check-card-printing` clean for all three (MSH canonical, earliest
  real printing; Shuri's later `pmei` printing correctly unscaffolded).
- **The one intentional behaviour change** (see Minor #3) traced across every call site in
  `mtg-sets`.

**On the question the unit flagged for a second opinion — Taskmaster's type clause reads as
*replacing*, and I agree.** CR 205.1b keys retention on exactly the phrase Taskmaster's oracle
lacks and Absorbing Man's has, in the same set. A contrast that deliberate is a strong signal.
`overrideCardTypes` / `overrideSubtypes` with `addedSupertypes` for legendary is the correct
encoding, and keeping supertypes additive (so a copied Snow creature stays snow) is right rather
than merely conservative — neither 205.1a nor 205.1b provides a "set the supertype line" operation.

---

## What's good

- **`CopyExceptions` is parameterized for the next card, not this one.** The rare change that
  *removes* SDK surface: three flat riders come off `EachPermanentBecomesCopyOfTargetEffect` and
  land in a shared value type. The add/override split is derived from the rules rather than from
  what these three cards happened to need, and `retainActivatingAbility` correctly stayed on the
  effect because it isn't a characteristic — that's the right line to draw.
- **`CopyExceptionApplier.typeLine` is shared with `auraTypeLineOf`**, so the Aura-host probe and
  the token actually built can't disagree about what the copy is. Reuse that prevents the *next*
  bug rather than fixing the last one.
- **The latent P/T bug is a real find.** The old permanent path used `targetCard.baseStats?.let { … }`,
  silently dropping an override when the copy source had no stats — exactly Absorbing Man copying a
  land. Directly tested at both unit and scenario level.
- **`AbsorbingManScenarioTest`'s enumeration test has the right instinct**: it asserts through
  `getLegalActions` rather than `execute(ActivateAbility(...))`, with a comment explaining that the
  ability was *executable but not enumerable*. That asymmetry is what a player experiences as "the
  button isn't there".
- **`derivePileAction`'s neutral fallback is a correctness win well beyond this unit.** Replacing
  "Return to Hand" with "Confirm Target" means an unrecognised effect now goes vague instead of
  making a false statement about what the game is about to do.
- **`chooseTargetsWalk.ts` as a pure reducer** is the right answer to "the web client has no DOM
  test environment" — the walk gets real tests instead of none.
- Docs updated in the same change, thoroughly; backlog arithmetic (32 → 29) checks out.

---

## Important

### 1. Four of the five enumerator name→id fixes had no test — **fixed**

`ActivatedAbilityEnumerator.kt:496` was covered; `ManaAbilityEnumerator`, `CrewEnumerator`,
`SaddleEnumerator`, `CombatEnumerator` and the second `ActivatedAbilityEnumerator` site shipped on
assertion alone. Not hypothetical — `ManaAbilityEnumerator`'s own comment names "Absorbing Man
copying a mana rock", and MSH ships exactly that.

Fixed by two more assertions in `AbsorbingManScenarioTest`, both through `getLegalActions`:

- copies **The Mind Stone** (`{T}: Add {W}`) → asserts the mana ability is offered off Absorbing Man;
- copies **Dependable Quinjet** (Crew 4) → asserts Crew is offered on the copy and is affordable off
  two Hill Giants.

Both are non-vacuous: the printed Absorbing Man definition has neither a mana ability nor Crew, so a
name lookup fails them. `board()` gained an `extraPermanents` parameter.

### 2. The convergence was one-directional on the authoring surface — **fixed**

`CopyExceptions` was the shared *value*, but the token effects only exposed a **read-only**
`copyExceptions` view over their flat riders. So a new exception added to `CopyExceptions` — say
`removedKeywords` — was expressible on `EachPermanentBecomesCopyOfTarget` and `EntersAsCopy` and
**unreachable from the token paths** until someone also added a flat rider and mapped it. That is
precisely the drift the change set out to end, and the doc's "New copy exceptions belong here, not
on an individual effect" therefore overstated what the code supported.

Fixed the way the finding suggested: both token effects now take a real
`exceptions: CopyExceptions = None` field, surfaced on the `Effects.CreateTokenCopyOfTarget` /
`CreateTokenCopyOfSelf` facades. The merge rule lives once, as `CopyExceptions.over(base)` —
additive axes union, replacing and single-valued axes take the modern field and fall through to the
riders. `None.over(base) == base`, so the ~20 rider-only cards are bit-for-bit unchanged, which is
confirmed empirically: `:mtg-sets:test` passes with **no snapshot re-bless**. Four new tests pin the
merge, including that a `nameOverride` (not expressible as a rider at all) now combines with
rider-borne types instead of one silently dropping the other.

---

## Minor

1. **Doc parameter order** — `card-sdk-language-reference.md` listed
   `(… excludeTarget, sourceFromAnyZone, affected, exceptions …)`; the data class is
   `excludeTarget, affected, sourceFromAnyZone, exceptions`. Named arguments make it harmless in
   practice, but it's the canonical catalog. **Fixed**, along with the convergence paragraph.

2. **`CopyExceptions.clauses()` dropped the article** — Taskmaster rendered as "except it's
   legendary Human Mercenary Villain creature". **Fixed** with a `withArticle` helper: "it's **a**
   legendary Human creature", "it's **an** artifact in addition to its other types", while a
   supertype-only clause correctly stays bare ("it's legendary"). Tested.

3. **One silent semantic change rides along** — the old token executor unioned `addCardTypes` on top
   of `overrideCardTypes`; `CopyExceptionApplier.typeLine` now lets the override win outright.
   Checked every call site: only Shelob uses `overrideCardTypes`, and it pairs it with
   `addedSubtypes`, not `addCardTypes` — so no printed card changes behaviour. It is the right
   reading of CR 205.1a, is documented on the field and pinned by a test. **No change needed**;
   worth stating explicitly rather than letting a reviewer discover it.

4. **`ChooseTargetsUI`'s submit guard latched forever** — `submittedRef` was a bare boolean, so if
   the server ever re-sent the *same* decision id after rejecting a submission, the walk could never
   answer it: `walk.submission` is already set, the effect's deps don't change, and the ref stays
   true. **Fixed** by holding the sent payload instead of a boolean — StrictMode re-runs the effect
   with the *same* object (suppressed by identity) while a re-confirm builds a fresh one (sent). A
   reducer test now pins that `confirm` never returns the same object twice, since the fix depends
   on it.

5. **The gate description understated what was run** — `LOOP_UNIT.md` listed only Gradle recipes
   though this unit changes ~1000 lines of TypeScript, and **`just test` is Gradle-only**: the
   web-client suite is not reachable from any `just` recipe. **Fixed** in `LOOP_UNIT.md` and the PR
   body, both of which now record `npm run typecheck` and `npm test` as a separate half of the gate.

---

## Note on the gate

One `just test` run during this review failed on `ConniveTargetingTest` with a
`TimeoutCancellationException`. That class contains no reference to anything in this diff and passes
in isolation (`just test-class ConniveTargetingTest`) — it is this box's memory pressure, not a
regression. The full gate was re-run rather than assumed, and is green.

PR #1777, opened 2026-08-11
Add ward costs with a non-listed cost: WardCost.PlayerCounters and WardCost.Choice (+2 MSH cards)

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1776

# Add ward costs with a non-listed cost: `WardCost.PlayerCounters` and `WardCost.Choice`

Unblocks the last two Marvel Super Heroes cards whose ward is not one of the six shapes the SDK
knew: **The Serpent Society** [226] and **Titania, Rugged Rumbler** [235].

## The gap

`WardCost` (`mtg-sdk/.../scripting/effects/StackEffects.kt`) had exactly six variants — `Mana`,
`Life`, `DynamicLife`, `Discard`, `Sacrifice`, `Composite` — and `Composite` is **AND** (Gisa, the
Hellraiser's "Ward—{2}, Pay 2 life"). There was no disjunction, and no way to pay a ward cost in
counters placed on the paying player. `WardCounterEffectExecutor`'s `when` over the cost was
exhaustive, as were the three description renderers in `KeywordAbility.kt`,
`KeywordStaticAbilities.kt` and `StackEffects.kt`.

### `WardCost.PlayerCounters(counterType, amount)` — "Ward—Get five poison counters"

`KeywordAbility.wardPlayerCounters(Counters.POISON, 5)`. The counters go on the **paying** player
(CR 122.1 — a counter is a marker placed on an object *or player*), placed through the ordinary
`AddCountersEffect` executor with the payer as the effect's controller. Nothing bespoke follows from
that: counter-placement replacement effects apply, `CountersAddedEvent` is emitted, and the
ten-poison state-based action (CR 122.1f) falls out on its own.

It belongs in `WardCost` rather than being a one-off because it is a *cost shape*, not a card: the
same variant covers energy, rad, or any other player-scoped counter, and it composes inside
`Composite` and `Choice` like every other atomic ward cost. It is also the only ward cost with **no
affordability precondition** — a player can always get counters — which is a property of the cost,
not of the card, and the executor now encodes it in one place rather than special-casing Serpent
Society at a call site.

### `WardCost.Choice(options)` — "Ward—Discard a card or pay {2}"

`KeywordAbility.wardChoice(...)`, the OR sibling of `Composite`'s AND, modelled on the two
cost-vocabularies that already have this shape: `AdditionalCost.Choice` and `PayCost.Choice`. It
follows the shape of `CostPaymentService.choicePrompt` — a `ChooseOptionDecision` listing only the
options the payer can actually pay, plus a trailing decline; the reduced list is stored on the
continuation so the chosen index maps directly. One deliberate divergence: the decline is labelled
"Counter spell" rather than `choicePrompt`'s "Don't pay", naming the consequence and matching the
`noText` the existing yes/no ward prompts already use. Picking an option charges it through *that
cost's own* ordinary ward flow, so the disjunction contributes a picker and no payment logic. If no
option is payable the spell is countered with no prompt, as an unpayable ward cost always has.

Making it a `WardCost` variant rather than a one-off keeps "or" out of `Composite`, where it would
otherwise have arrived as an ad-hoc third state, and it is immediately general: a `Composite` option
works, options are not restricted to two, and a mana option is just a `Mana` in the list.

### Two supporting pieces

- **`WardCost.clause`** — the self-contained verb phrase for a cost ("discard a card", "pay {2}",
  "get five poison counters"), as opposed to `description`'s object phrase ("a card"). The three ward
  renderers each supply the verb themselves ("Ward—Discard " + description), which works only while a
  cost is rendered alone; a disjunction must render each option *with its own verb*. Only `Choice`
  reads `clause`, so no existing card's rendered text changes.
- **`WardCounterEffectExecutor.canPayWardCost`** — now the single source of truth for "unpayable ward
  cost → counter without a prompt" (CR 702.21a). Every per-cost handler consults it (the mana, life,
  discard and sacrifice can-pay checks were inline copies before), and `Choice` uses it to filter the
  options it offers, so the picker can never advertise a leg the payer would then fail to pay.

### One shape, two rails (Titania)

Titania prints "discard a card or pay {2}" twice — once as an additional cost, once as a ward cost.
They are deliberately **the same shape reused, not two implementations**, and the facades are named
to match: `Costs.additional.DiscardOrPay("{2}")` and `KeywordAbility.wardDiscardOrPay("{2}")`, the
latter a one-liner over `wardChoice`.

They stay separate *types* because the rails genuinely differ, and collapsing them would be wrong:
an additional cost's mana leg folds into the spell's own mana cost at cast time
(`AdditionalCost.OrPay` — the enumerator offers a discard cast and a `{2}`-more cast), whereas a ward
cost is paid standing alone as the ward trigger resolves, so on the ward side the mana leg is an
ordinary option with no special status. That asymmetry is documented on both facades and in the SDK
reference, so the next card with this wording lands on the right one.

## Cards

- **The Serpent Society** [226] — {1}{B}{G} 3/4. Deathtouch; Ward—Get five poison counters; "whenever
  another creature you control with deathtouch dies, each opponent sacrifices a nontoken creature of
  their choice". The death trigger's "with deathtouch" is matched against **last-known information**
  (`TriggerMatcher`'s `HasKeyword` LKI branch, the Jackdaw Savior path) — the creature is already in
  the graveyard when the trigger is detected. The edict is the ordinary
  `Effects.Sacrifice(Creature.nontoken(), target = EachOpponent)`.
- **Titania, Rugged Rumbler** [235] — {2}{B/G} 5/5, the two `DiscardOrPay` rails above and nothing
  else.

## Rule citations (verified against `MagicCompRules_20260619.txt`)

- **CR 702.21a** — "Ward [cost] means 'Whenever this permanent becomes the target of a spell or
  ability an opponent controls, counter that spell or ability unless that player pays [cost].'"
- **CR 702.21b** — a ward cost's X is determined when the ability resolves (pre-existing;
  `DynamicLife` relies on it).
- **CR 122.1** — "A counter is a marker placed on an object **or player** …".
- **CR 122.1f** — "If a player has ten or more poison counters, that player loses the game as a
  state-based action."
- **CR 119.4** — a player can pay only life they have (the `Life` can-pay check).
- **CR 609.3** (backed by **CR 101.3**) — an effect that attempts something impossible does only as
  much as possible, which is why an opponent with no nontoken creature sacrifices nothing.

## mtgish generator

`wardKeywordLine` learned the disjunction: `_Cost: "Or"` → `KeywordAbility.wardChoice(...)`, via a new
`wardCostExpr` leg renderer. Every leg must render to a faithful `WardCost` or the whole ward line
still declines to SCAFFOLD rather than emit a ward missing an option. The existing single-cost
branches keep their facade rendering byte-for-byte, and no ward in the corpus uses `Or` today, so no
golden moves. The IR has **no player-counter cost tag**, so `WardCost.PlayerCounters` has nothing to
map to and is deliberately not taught.

## Verification

| Command | Result |
|---|---|
| `just test` | **BUILD SUCCESSFUL**, exit 0 — `:rules-engine:test` ran in full (10,340 assertions passing); `:mtg-sets:test`, `:game-server:test`, `:ai:test`, `:gym:test`, `:mtg-search:test` up-to-date from the passing runs below |
| `just rebless-cards` | exit 0 — **only** `The Serpent Society` and `Titania, Rugged Rumbler` were added to `mtg-sets/src/test/resources/snapshots/cards/MSH.json` (+112 lines, 0 removals). No unrelated card moved, so no shared SDK behaviour changed |
| `just check-card-printing "The Serpent Society"` | ok — canonical is in the card's earliest real printing |
| `just check-card-printing "Titania, Rugged Rumbler"` | ok — canonical is in the card's earliest real printing |
| `just check-backlog` | all 14 card-count headers in sync |

Full logs: `build/pr/loop-msh-u07-gate.log` (final `just test`),
`build/pr/loop-msh-u07-gate2.log` (targeted re-run), `build/pr/loop-msh-u07-rebless.log`.

Two failures on the first `just test` pass, both since resolved and both re-run green:

- `CardDefinitionSnapshotTest > MSH` — expected for two new cards; fixed by `just rebless-cards`.
- `ConniveTargetingTest > nonland discard` — a `TimeoutCancellationException` (the hang watchdog),
  in code this change does not touch. Re-ran in isolation and **passed**; the box was under memory
  pressure at the time. Not caused by this change.

One test of mine failed on that pass and was fixed: `TheSerpentSocietyScenarioTest`'s OTHER-binding
case killed the 3/4 Society with two 3-damage burn spells, and the priority passes between them
crossed a cleanup step that wiped the first three damage. It now uses a single 5-damage spell, so the
death happens inside one priority window. A test-harness bug, not an engine one.

### What was **not** done

- No manual playthrough in the web client, from either seat. The two new prompts reuse existing
  decision types (`YesNoDecision` for the counter cost, `ChooseOptionDecision` for the disjunction),
  both already rendered by the client, so no new UI was written — but that was reasoned about, not
  observed in a running client.
- No e2e / Playwright run, and no UX pass on the option labels beyond asserting their text in the
  engine tests.
- No AI-advisor work: the built-in AI answers both decisions through the generic decision paths.
- The multiplayer / Two-Headed Giant interaction of a player-counter ward cost was not exercised.

### Tests added

- `WardPlayerCountersTest` — engine-level, per mechanic: counters land on the payer; declining
  counters the spell; the cost is always payable (empty hand, empty board, 1 life still prompt); a
  payer already on five poison loses to the SBA; it composes inside `Composite`; it works when
  granted by `GrantWard`; oracle-text rendering.
- `WardCostChoiceTest` — engine-level, per mechanic: both legs offered and each paid; an unpayable
  leg is not offered; the trailing decline counters; declining *inside* the chosen leg counters; no
  payable option counters with no prompt; static `GrantWard` grant; oracle-text rendering.
- `TheSerpentSocietyScenarioTest`, `TitaniaRuggedRumblerScenarioTest` — one file per card.

## Review corrections

An independent reviewer looked at this branch (0 blocking, 2 important, 10 minor) and a corrector
pass followed. **10 findings fixed, 2 declined**; the full accounting is in
`build/pr/loop-msh-u07-correction.md`. The load-bearing ones:

- **CR 608.2c → CR 609.3 (backed by CR 101.3).** The wrong rule was cited for "does as much as it
  can", in the Serpent Society KDoc and in this body; both re-verified against
  `MagicCompRules_20260619.txt` and fixed. Every other CR number this change adds was re-checked and
  is correct.
- **`spellOutCount` deleted** — it was a byte-for-byte copy of `numberToWord`
  (`mtg-sdk/.../scripting/util/TextUtils.kt`), which `WardCost.PlayerCounters.description` now uses.
- **`WardCounterEffect.description` conjugates the `Choice` branch.** It read "…unless its controller
  discard a card or pay {2}"; a `wardPaymentVerbPhrase` helper now supplies the third-person form per
  variant and `Choice` joins those. Every other branch renders the identical string as before.
- **The counter-on-decline duplication is finished, not half-done.** All nine copies of the
  exile-or-plain counter block in `ManaPaymentContinuationResumer.kt` now route through one
  `counterForUnpaidCost` helper (renamed from `counterWard`, since it also serves the generic
  non-ward "counter unless its controller pays").
- **The paused counter-placement branch no longer drops a composite's remaining parts** — it errors
  instead of silently dropping them, with a comment recording why the branch is unreachable today.
- **A granted-deathtouch test** was added for the Serpent Society's last-known-information claim, and
  the Titania test KDoc's cast-rail coverage claim was corrected to what it actually asserts.

Re-gated after the corrections: `just test` → **BUILD SUCCESSFUL**, exit 0, 12,073 PASSED, **zero**
FAILED (`build/pr/loop-msh-u07-gate-correction.log`).

## Review corrections (round 2)

A second review pass (`build/pr/loop-msh-u07-review-round2.md`, posted as a PR comment) found **0
blocking, 0 important, 6 minor**. Five were fixed; one was declined as not worth the churn.

- **Dead `exileOnCounter` dropped from both new continuations.** The only producer of
  `exileOnCounter = true` is `CounterEffectExecutor`, which builds the mana / life / discard family —
  neither `CounterUnlessPlayerCountersContinuation` nor `WardCostChoiceContinuation` is ever
  constructed with it, so it was a serialized field on a persisted frame that could only ever be
  `false`. Both call sites now pass `exileOnCounter = false` explicitly, and each continuation's KDoc
  records *why* it differs from its siblings.
- **"Don't pay" → "Counter spell" in two KDocs.** The `WardCounterEffectExecutor` class KDoc and the
  `WardCostChoiceContinuation` KDoc described a decline label the code does not use.
- **The other two text renderings are now asserted.** The `WardCost` taxonomy is rendered three ways
  — the keyword line, `WardCounterEffect.description`'s third-person "Counter it unless its
  controller ~", and `GrantWard.description` — and only the first was covered. Both mechanic tests
  gained a case pinning the other two for the new variant, including that `Choice` joins *conjugated*
  phrases ("discards a card or pays {2}"), not clauses.
- **`WardCost.Choice(options)` added to the variant enumeration** in
  `docs/card-sdk-language-reference.md`; it was described in the prose below but missing from the
  grep-able list.
- **The "always payable" claim is now labelled an assumption.** It holds only because nothing grants
  a *player* `CANT_RECEIVE_COUNTERS` today — that flag is permanent-scoped and no card in the repo
  sets it. A Melira / Solemnity-shaped card would make the placement a silent no-op and the ward
  would be paid for free. The `PlayerCounters` KDoc now says so and points at the can-pay branch to
  revisit.

**Declined:** the duplicate `sacrificeCandidates` scan in `handleSacrificeCost` (once inside
`canPayWardCost`, once for the prompt's option list). It runs once per ward trigger; threading the
candidate list back out of the can-pay predicate would cost more clarity than the scan costs time.

Re-gated after round 2 (`build/pr/loop-msh-u07-gate-review2.log`): `just test` reported **BUILD
FAILED** on **one** test, `ConniveTargetingTest > nonland discard`, with a
`TimeoutCancellationException` from the 120s hang watchdog. Same environmental flake as the first
pass — the box had 84 MB of memory available at that moment. Re-run in isolation with 7.2 GB
available it **passes** (`build/pr/loop-msh-u07-gate-review2-connive.log`, 2/2). It is in connive
targeting, which this change does not touch, and it does not exercise ward, continuations or
counters.

Everything else in that run passed, including the round-2 additions:
`WardCostChoiceTest` **9/9** and `WardPlayerCountersTest` **8/8**
(`build/pr/loop-msh-u07-gate-review2-wardchoice.log`,
`build/pr/loop-msh-u07-gate-review2-wardcounters.log`).

**A clean full-suite `just test` has not been observed since the round-2 commit** — the one run
attempted hit the flake. Worth one more pass on a quiet box before merge.


bolav:
## Review (round 2) — `WardCost.PlayerCounters` + `WardCost.Choice`

Reviewed the branch in place against `loop-msh-u06`. `origin/main` (`498bccc131`) is already an
ancestor of HEAD, so there was nothing to merge. Primary lens: SDK elegance
(`docs/sdk-design-principles.md`).

### Verdict

Behavior is right and the SDK shape holds up. Both new `WardCost` variants earn their place:
`PlayerCounters` is a payment shape no existing primitive expresses (the payer *receives* a marker
rather than spending a resource, which is exactly why it has no affordability gate), and `Choice` is
the OR combinator sibling to `Composite`'s AND — it adds a picker and delegates every payment to the
handler that already existed. Both are parameterized for the next card, not this one
(`counterType`/`amount`, arbitrary option list). No blocking issues.

### What's good

- **`Choice` adds a picker and nothing else.** `resumeWardCostChoice` routes the chosen option back
  through `chargeNextWardPartOrNull` rather than re-implementing payment, so the spell-left-the-stack
  guard and the composite chaining stay single-sourced. `Composite`-inside-`Choice` and
  `Choice`-inside-`Composite` both fall out for free.
- **`canPayWardCost` consolidation** (`WardCounterEffectExecutor.kt`) is the change that makes "only
  offer payable legs" correct by construction — four inline can-pay checks folded into one exhaustive
  `when`, so `Choice`'s filter and the per-handler gate can't drift.
- **`counterForUnpaidCost`** collapses eight copy-pasted 10-line decline blocks into one. Checked
  branch by branch: behavior-preserving, including the `precedingEvents` ordering at the waterbend
  sub-cost site.
- **`PlayerCounters` goes through the ordinary `AddCountersEffect` executor**, so replacement effects,
  `CountersAddedEvent` and the ten-poison SBA come for free — and the test actually proves the SBA
  rather than asserting it in a comment.
- **Test coverage is real, not decorative.** One engine test per mechanic plus one scenario test per
  card (per the one-card-one-file rule), covering the corners that matter: LKI-granted deathtouch on
  the dies trigger, the OTHER binding excluding the Society's own death, the unpayable-leg filter,
  declining *inside* the chosen leg, composite chaining, and static `GrantWard` grants for both
  variants.
- **Every CR citation checks out** against `MagicCompRules_20260619.txt`: 122.1, 122.1f, 702.21a,
  702.21b, 119.4, 701.59b, 609.3, 101.3, 601.2f, 810.9a. Oracle text matches Scryfall byte-for-byte
  on both cards; `just check-card-printing` passes for both (canonical in `msh`, the earliest real
  printing).
- **The mtgish `"Or"` tag is verified, not guessed.** The corpus has `_Cost: "Or"` 3× with exactly the
  `args: [cost, cost]` array shape the emitter assumes (none of them a ward, as the notes say).
- **The duplication was written down instead of hidden.** The `backlog/sdk-reusability-consolidation.md`
  note gives the picker pattern an explicit extraction trigger ("a third cost vocabulary"). Right
  call — `CostPaymentService` and `WardCounterEffectExecutor` can't call each other today.

### Issues — Blocking

None.

### Issues — Important

None.

### Issues — Minor

1. **Dead `exileOnCounter` on both new continuations.** The only producer of `exileOnCounter = true`
   is `CounterEffectExecutor`, which builds the mana/life/discard family; neither
   `CounterUnlessPlayerCountersContinuation` nor `WardCostChoiceContinuation` is ever constructed
   with it. It is a serialized field on a persisted frame that can only ever be `false`, which
   AGENTS.md's "no unused fields" rule forbids. Drop the parameter from both and pass
   `exileOnCounter = false` at the two `counterForUnpaidCost` call sites.
2. **KDoc says "Don't pay", the code says "Counter spell".** The `WardCounterEffectExecutor` class
   KDoc and the `WardCostChoiceContinuation` KDoc both describe the trailing option as "Don't pay",
   while `handleChoiceCost` labels it `"Counter spell"` and its own KDoc explains why.
3. **The two new text renderings have no assertions.** The tests pin
   `KeywordAbility.Ward.description` for both variants, but nothing in the repo ever constructs
   `WardCounterEffect(cost).description` — so the new `wardPaymentVerbPhrase` branches
   (`"gets five poison counters"` and the `" or "` join) and the new `GrantWard.description` branches
   are untested. The refactor of the pre-existing branches is behavior-preserving, so this is not a
   regression risk, but four one-line assertions would cover all three renderings instead of one.
4. **`WardCost.Choice` missing from the variant list** in `docs/card-sdk-language-reference.md`. The
   enumeration line gained `WardCost.PlayerCounters(counterType, amount)` but not
   `WardCost.Choice(options)`; `Choice` appears only in the prose below.
5. **Double battlefield scan on ward—sacrifice.** `canPayWardCost` calls `sacrificeCandidates`, then
   `handleSacrificeCost` immediately calls it again for `validPermanents`. Once per ward trigger, so
   the cost is nil — noted only because the fold is otherwise clean.
6. **The "always payable" claim is load-bearing in four places** (SDK KDoc, executor KDoc,
   continuation KDoc, language reference) and holds today only because nothing grants
   `CANT_RECEIVE_COUNTERS` to a *player* — that flag is permanent-scoped and no card in the repo sets
   it. If a Melira/Solemnity-shaped card ever lands, `AddCountersExecutor` will silently no-op and the
   ward gets paid for nothing. Whether CR makes such a cost unpayable is genuinely ambiguous, so no
   code change now — but one sentence naming the assumption would keep the next author from treating
   it as a proved invariant.

### Recommendation

Merge-ready. The cleanup pass (1–4, plus the KDoc sentence for 6) is mechanical with no behavior
change. Item 5 is not worth the churn.

The `clause`-vs-`description` split the author flagged for a second opinion is the right call:
`Composite` renders Gisa as "Ward—{2}, pay 2 life" off the *object* phrase, so collapsing to a single
clause property would change existing oracle text. Leave it.

### Verified by the reviewer, not taken on trust

- `just test` on the reviewed tip → **BUILD SUCCESSFUL**. `WardCostChoiceTest` 8/8,
  `WardPlayerCountersTest` 7/7, `TheSerpentSocietyScenarioTest` 6/6,
  `TitaniaRuggedRumblerScenarioTest` 4/4, zero failures.
- `just check-card-printing` for both cards → ok. `just check-backlog` → 14 headers in sync.
- Oracle text for both cards re-fetched from Scryfall and compared character by character.
- Every CR number re-grepped in `MagicCompRules_20260619.txt`.

### Follow-up applied

All of 1–4 and the KDoc sentence for 6 were fixed on this branch after the review (item 5 declined);
see the "Review corrections (round 2)" section of the PR body.

Post-fix gating: the two mechanic suites pass with the new assertions (`WardCostChoiceTest` 9/9,
`WardPlayerCountersTest` 8/8). The full `just test` run after the fixes hit one unrelated flake —
`ConniveTargetingTest > nonland discard`, a 120s watchdog timeout with 84 MB of memory available —
which passes in isolation on a quiet box. A clean full-suite run has not been observed since the
round-2 commit; worth one more pass before merge.

PR #1778, opened 2026-08-11
Refresh the vendored Draftsim tables, adding HOB, MBC and MSH

bolav:
Re-extracted every ratings/removal table from a current draft-app bundle (index-DmeA26Yz.js) instead of the June snapshot. MSH now has a real 0-5 pick rating for 344 names plus archetype columns, so MSH pools take the scorer's archetype-aware `jm` path rather than the color-bias fallback; its curated removal list grew from 31 to 61 entries. HOB and MBC arrive the same way, and vintage_cube picks up the ratings drift since the last pull (140 values moved, +7/-2 names). EOE's table rewrote in a different row order with not one rating changed, so `--check` now reports that as "reordered only" rather than letting a real change hide inside reorder noise on the next pull.

MBC's table carries a Fixing column but an empty Archetype one, so its archetypes file is mana-fixing data only, with no synergy tags.

Three extraction fixes come with it, all verified against the bundle's own `KM`/`qM`/`$M`:

- `splashable` was never read (column 10), so every splashable card in TMT was flagged false and splashable-only rows were dropped entirely.
- The marker in that column is per-table: TMT and SOS write "yes", MSH writes "X". Matching only "yes" left Thanos, the Mad Titan un-splashable.
- Archetype names were lowercased on the way in; the bundle preserves case and only lowercases the role. (TMT's own table is lowercase upstream, so its names stay that way.)

HOB's upstream removal list has only 10 rows -- the set is too fresh to be tagged -- so `removal/_overrides.json` gains 16 hand-checked entries for it. Those are the one part of `draftai/` not reproducible from the bundle.

The extraction is now scripted and reproducible rather than an ad-hoc node script that never landed: `bolav/ai-rating-score/tools/extract_from_bundle.py` pulls the tables out of a bundle or a saved HAR, and `tools/import_draftsim_sets.py` regenerates `ai/src/main/resources/draftai/` from them, with `--check` to diff instead of write. Every other set's generated JSON is byte-identical to what was committed, which is what pins the port -- with one gap: `--check` only compares files the current tables still generate, and `archetypes/FDN.json` predates upstream dropping FDN's archetype columns, so it is no longer regenerated from anything.


bolav:
## Review — vendored Draftsim table refresh

Data-only refresh; no SDK surface is added, so the elegance bar isn't in play. The join is solid and the two advertised extraction fixes are real — but one of them is only half-fixed (MSH marks `Splashable` with `X`, not `yes`, so MSH's one splashable card is still dropped), and two files ship data that doesn't match what the code and docs say about them (MBC, FDN).

**Status:** issues 1, 4, 5, 6 and 7 below are already fixed on `draftsim-update` locally (folded into the same commit, so this PR's head needs a force-push to show them). Issues 2 and 3 are deliberately left open — see the note at the end.

Verified locally with `scripts/gradle-locked :ai:cleanTest :ai:test` — BUILD SUCCESSFUL, `DraftsimDataTest` green.

### What's good

- **The join actually works.** Independently verified: all 156 HOB and 239 MSH canonical `card("…")` names resolve through `nameKey` into the new ratings tables — 100%, not just past the test's 0.9 floor.
- **Removal coverage is complete.** Pulled all 180 HOB and 256 MSH nonland cards from Scryfall and pattern-matched oracle text for removal shapes: every one is tagged. The only flag was `Front Porch Sentries` (a -1/-1 dies trigger), correctly untagged. The 8 HOB overrides spot-checked are right, including `Thorin's Last Stand` tagged artifact/enchantment-only.
- **The splashable fix is genuine where the marker is `yes`:** upstream TMT has exactly 9 splashable rows and `TMT.json` now has 9, and the three splashable-only rows the old extractor dropped entirely (`leader's talent`, `madame null, power broker`, `the last ronin`) are back.
- **The casing question checks out against upstream** — TMT really does ship `[alliance enabler]` while MSH/HOB ship `[Artifacts Enabler]`, so the mixed casing across committed files is faithful, not a regression.
- Data hygiene is clean: every archetype and removal key resolves into its own set's ratings table, ratings keys are already `nameKey`-normalized, and MSH removal grew 31→61 with nothing dropped.

### Issues

#### Important

**1. `archetypes/MSH.json` — Thanos, the Mad Titan should be `splashable: true`.** *(fixed)* Upstream marks MSH's `Splashable` column with `X`, while TMT/SOS use `yes`; the importer tests `cell(10).lower() == "yes"`, so the exact fix this commit advertises still drops MSH's only splashable row. Evidence: `MSH.txt` column `Splashable` has distinct values `{'X': 1}`, on `Thanos,_the_Mad_Titan` (RWB mythic, 3.1); HOB/MBC are genuinely empty. `splashable` gates `DraftsimDeckBuilder.kt:329`, so a RWB bomb upstream wants splashable into a two-colour build silently isn't. Fixed by matching a marker set (`yes`/`x`/`true`/`1`) and regenerating — one file moved, `archetypes/MSH.json`.

**2. MBC is a fixing-only table, but the code treats any non-empty map as "tagged".** *(open)* `archetypes/MBC.json` has 238 records with zero archetype tags — upstream's `Archetype` column is empty for MBC; only `Fixing` is populated. `DraftsimScorer.kt:402` (`hasArch`) and `:501` (`scoreBoosterAuto`) branch on `tables.archetypes.isNotEmpty()`, so merely shipping the file routes MBC pools onto the `jm` path with no archetype signal — `archColorMap` → empty, `archetypeScores` → empty, `d` collapses to `h < G2`. Dormant today (no `MtgSet` with code MBC, and `ConstructedRatings` strips archetypes), which also means nothing will catch it later. Suggested fix: gate on tags, not map size — add `hasArchetypeTags = archetypes.values.any { it.archetypes.isNotEmpty() }` to `DraftsimSetTables` and use it at `DraftsimScorer.kt:402/501/542` and `DraftsimDeckScorer.kt:162`, so fixing data stays usable for the manabase logic.

**3. `archetypes/FDN.json` is stale and no longer reproducible — and `--check` can't see it.** *(open)* The current bundle's FDN table has 8 columns (no `Archetype`/`Fixing`/`Splashable`), so the importer emits 0 archetype rows for FDN, yet the 174-record file from the June snapshot stays committed and `--check` still prints "matches byte for byte". That's a blind spot in the claim "every other set's generated JSON is byte-identical … which is what pins the port": the check only compares files it still generates. Meanwhile FDN drafts keep taking the `jm` path off archetype data upstream has withdrawn. Either drop `FDN.json`, or keep it with an explicit comment that it's pinned to the 2026-06 snapshot — and make `--check` report orphaned outputs either way.

#### Minor

**4. Stale tagged-set enumerations.** *(fixed)* The list was spelled out in eight places and this PR updated four; `DraftsimDraftAdvisor.kt:11` ("FDN/SOS/TMT"), `DraftsimScorer.kt:384` ("the 4 tagged sets"), `SPEC_draft_entrypoints.md:152`, `SPEC_scoring.md:282` and `:363` were still stale. Since the list changes on every data refresh, the fix drops the enumerations entirely — `DraftsimData` carries the one canonical statement, pointing at `draftai/archetypes/`, and the rest just say "tagged sets".

**5. `ratings/EOE.json` is a whole-file rewrite with zero content change** *(fixed)* — same 322 keys, same 322 values, only the order moved (upstream row order churn). The commit message said EOE "picks up the ratings drift"; it didn't. `--check` now classifies `reordered only, no values changed` separately from `differs`, so the next refresh can't hide a real change inside 7 KB of noise, and the commit message is corrected. (`vintage_cube` is genuine drift: 140 values changed, +7/−2 keys, `tarmogoyf` and `multiversal passage` dropped.)

**6. The 16 hand-authored HOB `_overrides.json` entries weren't mentioned in the commit.** *(fixed)* They're the one part of this diff that can't be regenerated from the bundle, so they now get a line in the message. `_overrides.json`'s `_comment` also called the per-set files "Scryfall-generated" — they're bundle-extracted; corrected at the source file so the next import doesn't revert it, along with the same confusion in the test comment.

**7. Test shape.** *(fixed)* The new MSH join test duplicated the LTR one, leaving two `withClue` idioms in the file (the private helper hardcoded "LTR spells"; the new test inlined `io.kotest.assertions.withClue`). Now a loop over `listOf("LTR", "MSH", "HOB")` — HOB ships in this PR and had no coverage — with the helper deleted. Also added the assertions that pin the extraction fixes the commit is built around: TMT has splashable records and the splashable-only row `The Last Ronin` survives, and MSH's Thanos is splashable (that one fails against the old importer).

**8. Provenance.** *(open, low priority)* The pipeline lives in a gitignored dir outside the repo, and `SPEC_scoring.md:33` points at a `data/README.md` that doesn't exist. Worth vendoring the two scripts (e.g. `tools/draftsim/`) or adding that README with the bundle → JSON flow, the column layout, and the parser symbols.

### Recommendation

Force-push the amended branch to pick up 1, 4–7. Issues 2 and 3 are the remaining judgement calls: the `jm` path is still gated on map size rather than actual tags, and `archetypes/FDN.json` is still committed although nothing regenerates it — the commit message now discloses the latter as a `--check` blind spot instead of leaving the byte-identical claim unqualified. Everything else in the diff is sound and well-verified.

PR #1779, opened 2026-08-11
Fix all Kotlin compiler warnings across every module

bolav:
Clears every Kotlin compiler warning in the repo. Two commits: the first covers `rules-engine` and `ai` main sources, the second the remaining 43 across all ten modules and all three source sets.

**No behavior change is intended anywhere**, and for most of the diff that is guaranteed by construction: each removed `!!`, `?.`, and `?:` fallback sits behind a smart cast the compiler now proves, so if it compiles the semantics are identical. The handful of hunks that are *not* mechanical are called out below.

## Warnings cleared

| Module / source set | Count |
|---|---|
| `rules-engine` main | 15 |
| `rules-engine` test | 27 |
| `rules-engine` testFixtures | 6 |
| `ai` main | 1 |
| `ai` test | 1 |
| `game-server` main | 5 |
| `game-server` test | 1 |
| `mtgish-tooling` main | 3 |

## The hunks that aren't mechanical

**`EffectApplicator.kt` — deleted a duplicate `Modification.SetName` branch.** `git log -L` shows the branch was added at line 99 by 648e0d9c (Witness Protection) and then *independently re-added verbatim* at line 294 by 6bad4cb2 (Honest Work), which didn't notice the existing one. Both live in the same `when` opened at line 54, so the second was unreachable and Kotlin flagged it. The bodies are byte-identical, so removing it is behavior-preserving — and it closes a trap where a future edit to one copy would silently do nothing.

**`MaralenFaeAscendantScenarioTest.kt` — deleted a shadowed extension.** The file declared a private `TestGame.isInExile` shadowed by the identical `ScenarioTestBase` member, so both call sites (`:130`, `:222`) already resolved to the member. Dead code.

**`BlightPaymentAiTest.kt` — fixed a vacuous filter.** It selected Evershrike's Gift's ability with `first { it.activateFromZone != null }`, always true: `ActivatedAbility.activateFromZone` is `Zone`, not `Zone?`, defaulting to `BATTLEFIELD`. Now `== Zone.GRAVEYARD`, which is what the test meant and what the card declares.

**`UserDtoSerializationTest.kt` — the test was pinning the wrong mapper.** It built its `ObjectMapper` from the deprecated `Jackson2ObjectMapperBuilder`, and its own comment claimed that meant it saw "exactly the modules Spring MVC uses". Under Spring Boot 4.1 / Framework 7 that is no longer true — Jackson 2 (`com.fasterxml`, 2.21.4) and Jackson 3 (`tools.jackson`, 3.1.4) are both on the classpath, and MVC serializes through `JacksonJsonHttpMessageConverter`, which is Jackson 3. So the test guarding the `isAdmin` wire shape was validating a mapper the server does not use. It now takes the mapper straight off that converter. All three assertions still pass, so the `isAdmin` shape does hold in production.

**Redis TTL — deprecated overload.** Spring Data Redis deprecates `set(key, value, Long, TimeUnit)`; the five call sites move to the equivalent `Duration` overload.

## Verification

- **Compile:** `compileKotlin compileTestKotlin compileTestFixturesKotlin --rerun-tasks` across all ten modules — 21 compile tasks, BUILD SUCCESSFUL, **0 warnings, 0 errors**. `--rerun-tasks` mattered: a plain `just test-rules` came back all-`UP-TO-DATE` in 8s, which is exactly the false-green this repo is prone to.
- **Tests:** every module's suite ran green.

| Module | Tests | Failures | Errors |
|---|---|---|---|
| `rules-engine` | 10,527 | 0 | 0 |
| `game-server` | 483 | 0 | 0 |
| `ai` | 448 | 0 | 0 |
| `mtg-sdk` | 427 | 0 | 0 |
| `mtg-sets` | 342 | 0 | 0 |
| `mtg-search` | 163 | 0 | 0 |
| `mtgish-tooling` | 142 | 0 | 0 |
| `gym` / `gym-server` / `gym-trainer` | 57 | 0 | 0 |
| **Total** | **12,589** | **0** | **0** |

One caveat worth stating rather than burying: the first full `just test` run reported `ConniveTargetingTest > nonland discard` FAILED with a `TimeoutCancellationException` — a coroutine timeout, not an assertion, in a file this PR does not touch. It passes on its own and on a full `rules-engine` re-run (10,527 tests, 0 failures). That is the load-related flake this box is known for, not a regression.

## Notes for the reviewer

- No new SDK types (`Effect` / `Modification` / `Trigger` / `Condition`), so the SDK-elegance bar doesn't bind here.
- No card definitions or `Printing(...)` rows moved, so no `check-card-printing` run applies.
- No CR rule numbers added or changed. The `702.80a` / `702.2b` / `704.5h` / `122.6a` citations visible in the `DamageUtils` hunk are untouched context.
- No `@Suppress` anywhere — warnings are fixed, not silenced.
- No new tests: behavior is provably unchanged by the compiler, so the existing suites passing on a forced rerun is the appropriate evidence.
- The first commit's subject says "in rules-engine and ai", which was accurate for its own scope (main sources only); the second commit completes the claim across the repo.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/Strategist.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/Strategist.kt
@@ -282,7 +282,7 @@ class Strategist(
             // Fill in targets on the returned action so the processor can execute it.
             // The committed target is chosen by simulation (not just the heuristic) so the
             // AI sees the real resolved board, including effects already on the stack.
-            best!!.first
+            best.first
         } else {
             pass ?: legalActions.first()
         }
@@ -292,7 +292,7 @@ class Strategist(
                 state, evaluationState, playerId, startNanos,
                 pass = pass, passScore = passScore, adjustedPassScore = adjustedPassScore,
                 adjusted = adjusted, dropped = dropped.orEmpty(),
-                chosenAction = if (takeAction) best!!.first else null,
+                chosenAction = if (takeAction) best.first else null,
             )
         }
         return chosen
--- a/ai/src/test/kotlin/com/wingedsheep/ai/engine/BlightPaymentAiTest.kt
+++ b/ai/src/test/kotlin/com/wingedsheep/ai/engine/BlightPaymentAiTest.kt
@@ -8,6 +8,7 @@ import com.wingedsheep.engine.support.TestCards
 import com.wingedsheep.mtg.sets.definitions.ecl.cards.CinderStrike
 import com.wingedsheep.mtg.sets.definitions.ecl.cards.EvershrikesGift
 import com.wingedsheep.sdk.core.Step
+import com.wingedsheep.sdk.core.Zone
 import com.wingedsheep.sdk.model.Deck
 import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.shouldBe
@@ -25,7 +26,7 @@ class BlightPaymentAiTest : FunSpec({
         val gift = driver.putCardInGraveyard(player, "Evershrike's Gift")
         val creature = driver.putCreatureOnBattlefield(player, "Force of Nature")
         repeat(2) { driver.putLandOnBattlefield(player, "Plains") }
-        val abilityId = EvershrikesGift.activatedAbilities.first { it.activateFromZone != null }.id
+        val abilityId = EvershrikesGift.activatedAbilities.first { it.activateFromZone == Zone.GRAVEYARD }.id
         val legal = LegalActionEnumerator.create(driver.cardRegistry).enumerate(driver.state, player)
             .single { (it.action as? ActivateAbility)?.abilityId == abilityId }
 
--- a/game-server/src/main/kotlin/com/wingedsheep/gameserver/repository/RedisGameRepository.kt
+++ b/game-server/src/main/kotlin/com/wingedsheep/gameserver/repository/RedisGameRepository.kt
@@ -14,8 +14,8 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.context.annotation.Primary
 import org.springframework.data.redis.core.RedisTemplate
 import org.springframework.stereotype.Component
+import java.time.Duration
 import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.TimeUnit
 
 /**
  * Redis-backed implementation of GameRepository.
@@ -63,8 +63,7 @@ class RedisGameRepository(
             redisTemplate.opsForValue().set(
                 gameKey(gameSession.sessionId),
                 json,
-                redisProperties.ttlMinutes,
-                TimeUnit.MINUTES
+                Duration.ofMinutes(redisProperties.ttlMinutes)
             )
             logger.debug("Persisted game session ${gameSession.sessionId} to Redis")
         } catch (e: Exception) {
@@ -124,8 +123,7 @@ class RedisGameRepository(
             redisTemplate.opsForValue().set(
                 lobbyLinkKey(gameSessionId),
                 lobbyId,
-                redisProperties.ttlMinutes,
-                TimeUnit.MINUTES
+                Duration.ofMinutes(redisProperties.ttlMinutes)
             )
         } catch (e: Exception) {
             logger.error("Failed to persist lobby link for game $gameSessionId", e)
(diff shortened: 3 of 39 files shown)


bolav:
## Review — `compiler-warning-cleanup`

Reviewed with `origin/main` merged in (the branch was already 0 commits behind, so no merge commit was needed). Both commits reviewed together.

### Verdict

Sound. A pure compiler-warning cleanup with **zero behavior change**, and for most of the diff that is guaranteed by construction: each removed `!!`, `?.`, and `?:` fallback sits behind a smart cast the compiler now proves, so if it compiles the semantics are identical. The hunks that *aren't* mechanical each turn out to be a genuine latent-bug find rather than a regression. Compiles clean repo-wide and every suite is green.

**Gates run for this review** (not taken from the description):

| Gate | Result |
|---|---|
| `compileKotlin compileTestKotlin compileTestFixturesKotlin --rerun-tasks`, all ten modules | BUILD SUCCESSFUL, 21 compile tasks, **0 warnings, 0 errors** |
| Full test suite, all modules | **12,589 tests, 0 failures, 0 errors** |

`--rerun-tasks` mattered: a plain `just test-rules` came back all-`UP-TO-DATE` in 8s, which is exactly the false-green this repo is prone to after a rebase or a cached compile.

### What's good

- **The `EffectApplicator` deletion is the real prize.** `git log -L` shows `is Modification.SetName -> { values.name = mod.name }` was added at line 99 by 648e0d9c (Witness Protection) and then *independently re-added verbatim* at line 294 by 6bad4cb2 (Honest Work), which didn't notice the existing branch. Both live in the same `when` opened at line 54, so the second was unreachable. Bodies are byte-identical, so deleting it is behavior-preserving — and it removes a trap where a future edit to one copy would silently do nothing.
- **The author knew where to stop.** `DamageUtils.kt:388` still reads `sourceId != null && projected.hasKeyword(...)` in the `else` branch, and that's correct: `hasWither` being *false* does not imply `sourceId != null`, so only the `if (hasWither)` branch could drop the guard. That asymmetry is the easy thing to botch in this kind of sweep.
- **No `@Suppress` anywhere in the diff.** Warnings are fixed, not silenced — the whole point of the exercise.
- `MoveToZoneEffectExecutor.kt:274` correctly *keeps* `effect.positionFromTop!!` (a cross-module `val` on an `mtg-sdk` type can't be smart-cast) while dropping `faceDownMode!!` at line 287 (bound to a local `val` first). Subtle distinction, handled correctly.
- Three of the test-source warnings were treated as signal rather than noise — the shadowed `isInExile` extension, the vacuous `activateFromZone != null` filter, and the Jackson 2 vs Jackson 3 mapper mismatch. Each is a real (if latent) defect that the warning was pointing at, and each was fixed at the cause rather than by adding a cast.

### Issues

**Minor — none blocking.**

1. **`ReplacementEffectUtils.kt:82` now leans on K2 data-flow exhaustiveness.** `ReplacementEffect` is a sealed interface with many subtypes; the second `when` only type-checks as exhaustive because the *earlier* `when (effect)` at line 77 exits via `else -> continue`. It compiles, so it's sound, and it's arguably better than the old `else -> false`. No change needed — but the robustness property someone might assume is there isn't: adding a third counter-placement `ReplacementEffect` subtype still gets silently skipped by line 77's `else -> continue` rather than producing a compile error. Pre-existing, not introduced here.

2. **`UserDtoSerializationTest` now asserts against a different mapper than before.** Correctly so — Spring Boot 4.1 serializes MVC responses with Jackson 3, not the Jackson 2 `ObjectMapper` the old builder produced, so the test was previously validating a mapper the server doesn't use. Flagging it only because it is the one place where the *meaning* of a test changed rather than just its syntax; the three assertions still pass, which is the evidence that the `isAdmin` wire shape genuinely holds in production.

### Axes checked and not applicable

Worth stating explicitly since they're the usual review axes here: no new SDK types, so the SDK-elegance bar doesn't bind. No card definitions or `Printing(...)` rows moved, so `check-card-printing` doesn't apply. No CR rule numbers added or changed — the `702.80a` / `702.2b` / `704.5h` / `122.6a` citations visible in the `DamageUtils` hunk are untouched context. And no new tests are warranted: behavior is provably unchanged by the compiler, so the existing suites passing on a forced rerun is the appropriate evidence.

### Note on the run

The first full `just test` reported `ConniveTargetingTest > nonland discard` FAILED with a `TimeoutCancellationException` — a coroutine timeout, not an assertion, in a file this branch does not touch. It passes on its own and on a full `rules-engine` re-run (10,527 tests, 0 failures). Load-related flake, not a regression, recorded here so the green totals above aren't read as hiding it.

### Recommendation

Merge as-is.

PR #1782, opened 2026-08-11
Fix Emeritus of Truce entering prepared unconditionally

bolav:
## Emeritus of Truce entered prepared unconditionally

`Emeritus of Truce // Swords to Plowshares` (SOS #13) carried `keywords(Keyword.PREPARED)`. It
shouldn't: the creature face has no printed "This creature enters prepared." line — it only ever
becomes prepared from the conditional tail of its ETB trigger ("Then if an opponent controls more
creatures than you, this creature becomes prepared").

Scryfall reports `keywords: ["Prepared"]` on **every** `layout: "prepare"` card, whether or not the
card actually enters prepared, and that list was copied verbatim when the card was authored.

In this engine the keyword is load-bearing, not cosmetic. `StackResolver` enters a permanent
prepared only when the `PREPARE` layout *and* `Keyword.PREPARED` are both present, so the card
entered prepared the moment it resolved — handing its controller a free castable Swords to
Plowshares copy — and the `ConditionalEffect` on the ETB trigger was a dead no-op, since
`PreparationLogic.makePrepared` bails when the permanent is already prepared.

Dropping the keyword restores the conditional path. No SDK change: the card already composed
`CreateToken then ConditionalEffect(OpponentControlsMoreCreatures, BecomePrepared(Self))`, and the
keyword was short-circuiting it.

### Set-wide check

Swept all 36 SOS prepare-layout cards against Scryfall oracle text. After this change,
`keywords(… Keyword.PREPARED …)` and the printed "enters prepared" line agree on every card, in both
directions. Emeritus of Truce was the only mismatch, and `PREPARE` layout exists only in SOS.

The mtgish emitter is not the source: its golden declines this card to a STRUCTURE stub and emits no
keyword (it reads the IR's `AsPermanentEnters[EntersPrepared]` rule, not Scryfall's keyword list), so
no emitter fix is owed.

`backlog/sos-engine-gaps.md` already specified the behavior correctly — this was authoring drift
against a written spec, not a spec gap.

## Second commit — review follow-up

`Keyword.PREPARED`'s KDoc claimed the keyword was *"display-only — the behavior is driven by the
PREPARE layout."* That is false, and it is precisely the trap that produced this bug: the next author
reads "display-only", concludes the keyword is cosmetic, copies Scryfall's list, and ships another
silently-always-prepared card. It now states the iff the resolver enforces, names
`Effects.BecomePrepared` as the alternative for cards that only *become* prepared, and warns about
Scryfall's blanket tag.

The "Then if" ordering was cited as **CR 603.4**, which is the intervening-"if" clause rule and says
the opposite of what was claimed — its closing parenthetical is what establishes that a mid-text
"if" is *not* an intervening if. The asserted behavior comes from **CR 608.2c** (instructions
followed in the order written) and **CR 608.2h** (information determined as the effect is applied).
Both the card KDoc and the test name now cite those.

Also in this commit:

- The KDoc said the "rest of the Emeritus cycle" carries the keyword. It's Abundance, Ideation, and
  Woe — Conflict lacks it too.
- `docs/card-sdk-language-reference.md` documented a `MakePreparedEffect` / `PreparedService`
  that does not exist; removed, with its content folded into the correct `BecomePreparedEffect` entry
  seven lines above.

## Tests

`EmeritusCycleScenarioTest` gains four Emeritus of Truce cases over a shared `playTruce(opponentBears,
tokenTo)` helper:

| Board | Expected |
|---|---|
| 0 Bears, token to self — you 2, them 0 | not prepared, no exile copy |
| 3 Bears, token to self — you 2, them 3 | prepared, castable Swords copy in exile |
| 2 Bears, token to self — you 2, them 2 | not prepared (tied) |
| 2 Bears, token to **opponent** — you 1, them 3 | prepared |

The last two are the interesting pair. They share an identical starting board and differ only in who
receives the Inkling, so together they pin that the condition reads **both** battlefields **after**
the token is created. Checked before the token existed, case 3 would prepare (2 > 1); counting only
your own side, case 4 would not.

## Verification

Run with `--rerun-tasks` throughout — a plain `just test-class` reports `:rules-engine:test
UP-TO-DATE` and passes without executing anything.

- `EmeritusCycleScenarioTest` — all 4 Truce cases PASSED, whole class green
- `SosPrepareCardsScenarioTest` — PASSED
- `:mtg-sets:test CardDefinitionSnapshotTest` — BUILD SUCCESSFUL, including `SOS: card trees match
  golden` (the snapshot edit is exactly what a re-bless produces)
- `just check-card-printing "Emeritus of Truce"` — ok, exit 0
- 0 failures across every run

No downstream coupling: the web client reads `isPrepared` off `PreparedComponent`, not the keyword;
`ai/` has no references to `PREPARED`; `CastFromZoneEnumerator`, `CastSpellHandler`, and
`BecomePreparedExecutor` all gate on `CardLayout.PREPARE`.

## Known follow-ups, deliberately not in this PR

- **A `CardLinter` guard for this bug class.** A `PreparedKeywordMismatch` finding — for every
  `CardLayout.PREPARE` card, assert `keywords.contains(PREPARED)` matches whether the oracle text
  contains "enters prepared" — would make the whole class a build failure across all sets. It's a
  pure local check on the `CardDefinition`, and `CardLintTest` + `lint-allowlist.txt` already provide
  the gate and the escape hatch. Same precedent as `AttachedScopeGrantOnNonAttachment`, which was
  added after Harmonious Grovestrider shipped with no ward. Kept out because it's an SDK addition
  with its own language-reference update and full `mtg-sets` run.
- **`Conditions.OpponentControlsMoreCreatures` in multiplayer.** `evaluateBattlefieldAggregate`
  flatMaps across all opponent ids, so it compares the *combined* opponent creature count to yours.
  Correct in two-player, wrong for "an opponent" with three or more. Also affects
  `OpponentControlsMoreLands` and Beza, the Bounding Spring, so it needs its own branch and tests.


bolav:
## Review — SDK elegance, correctness, rules accuracy, tests

Reviewed on the branch with `origin/main` merged in (`git merge origin/main` → **Already up to date**;
no merge commit, no conflicts). All findings below were raised against the first commit
(`cc379db4`) and are **fixed in the second commit** (`0e81fdf8`) — recorded here for the trail.

### Verdict

The behavior fix is correct and complete. Verified the premise directly against Scryfall:
`Emeritus of Truce // Swords to Plowshares` is `layout: "prepare"` with `keywords: ["Prepared"]`, but
the creature face's oracle text carries **no** "This creature enters prepared." line — Scryfall tags
`Prepared` on every prepare-layout card regardless. `StackResolver` gates entering-prepared on
`layout == PREPARE && cardDef.keywords.contains(Keyword.PREPARED)`, so the keyword was load-bearing:
the creature entered prepared unconditionally, and the ETB `ConditionalEffect` was a dead no-op
(`PreparationLogic.makePrepared` bails when already prepared). Removing the keyword restores the
conditional path, which the new tests exercise for the first time.

### What's good

- **The fix is the minimal correct one.** No new SDK type, no new effect — the card already composed
  `CreateToken then ConditionalEffect(OpponentControlsMoreCreatures, BecomePrepared(Self))`. Nothing
  to say about SDK elegance; the keyword was simply short-circuiting an already-correct composition.
- **The `opponentBears = 2` test earns its place.** It's the discriminating case: 2 > 1 (before the
  token) would prepare, 2 > 2 (after) does not. It pins the resolution ordering rather than just the
  happy path, and the 0/2/3-bear triple covers false / boundary / true.
- **`playTruce(opponentBears)` factoring** collapses three near-identical setups into one helper
  without hiding the varying input.
- **The negative assertion is complete** — `findExileCopy(...) shouldBe null` checks the *consequence*
  of not being prepared, not just the flag, matching how the rest of the file asserts.
- **The KDoc paragraph is the right kind of comment**: it records why a keyword is *absent*, which is
  exactly what a future Scryfall re-sync would silently undo.

### Completeness checks run during review

- **Swept all 36 SOS prepare-layout cards** against Scryfall oracle text. After the fix,
  `keywords(… Keyword.PREPARED …)` and the printed "enters prepared" line agree on every card, in
  both directions. Emeritus of Truce was the only mismatch in the set; `PREPARE` layout exists only
  in SOS.
- **The mtgish emitter is not the source.** `sos.emitted.golden.txt` declines this card to a
  STRUCTURE stub and emits no keyword — it reads the IR's `AsPermanentEnters[EntersPrepared]` rule,
  not Scryfall's keyword list. No emitter fix is owed (AGENTS.md: fix the emitter, not the generated
  card — not applicable here).
- `just check-card-printing "Emeritus of Truce"` → **ok**, exit 0 (canonical in `sos`, its earliest
  real printing).
- No downstream coupling: the web client reads `isPrepared` off `PreparedComponent`, not the keyword;
  `ai/` has zero references to `PREPARED`; `CastFromZoneEnumerator`, `CastSpellHandler`, and
  `BecomePreparedExecutor` all gate on `CardLayout.PREPARE`.
- **The backlog already specified this correctly** — `backlog/sos-engine-gaps.md` says "A few cards
  (Emeritus of Truce, …) *conditionally* become prepared via an ETB clause". Authoring drift against
  a written spec, not a spec gap.

### Findings

> ⚠️ `just test-class EmeritusCycleScenarioTest` reported `BUILD SUCCESSFUL` with
> `:rules-engine:test UP-TO-DATE` — **nothing ran**. Every result below is from a `--rerun-tasks`
> re-run.

**Important — `mtg-sdk/.../core/Keyword.kt`: the SDK doc that caused this bug was left in place.**
`PREPARED`'s KDoc read *"Display-only on the keyword — the behavior is driven by the PREPARE
layout."* That is false, and `StackResolver` is the proof: a PREPARE-layout permanent enters prepared
**only** if the keyword is present. The failure path is exactly the bug being fixed — the next author
adding a prepare card reads "display-only", concludes the keyword is cosmetic, copies Scryfall's
`keywords: ["Prepared"]` verbatim, and ships another silently-always-prepared card. This belongs in
the same change as the fix: it's the root cause of the class, not a drive-by.
✅ Fixed — the KDoc now states the iff, names `Effects.BecomePrepared` for cards that only *become*
prepared, and warns about Scryfall's blanket tag.

**Important — `EmeritusCycleScenarioTest.kt`: wrong CR number in the new test name.** The test was
titled `"the Inkling is counted — CR 603.4 'Then if' is checked after the token is made"`. CR 603.4 is
the *intervening-"if" clause* rule and says nothing about mid-resolution ordering; its closing
parenthetical is what establishes that a mid-text "if" is **not** an intervening if. The asserted
behavior comes from **CR 608.2c** ("The controller of the spell or ability follows its instructions
in the order written") plus **CR 608.2h** ("if an effect requires information from the game … the
answer is determined only once, when the effect is applied"). AGENTS.md makes verifying rule numbers
a hard rule, and a test name is what the next prepare-card author copies.
✅ Fixed — test name and card KDoc both cite 608.2c/608.2h; the KDoc now also explains *why* 603.4
does not apply rather than mis-citing it.

**Minor — `EmeritusOfTruce.kt`: "the rest of the Emeritus cycle carries" is inaccurate.** Emeritus of
Conflict has no `Keyword.PREPARED` either (Scryfall-confirmed) — it's 3 of 5, not 4 of 5.
✅ Fixed — names Abundance, Ideation, and Woe, and notes Conflict is in the same position as Truce.

**Minor — the same explanation lived in two places** (the card KDoc and the test-side comment). The
card KDoc is the one that gets read during a re-sync.
✅ Fixed — the test comment is now a two-line pointer at the KDoc.

**Minor — one untested axis.** All three cases targeted *yourself* with the Inkling. Targeting the
opponent grows *their* count, so `opponentBears = 2` + target-opponent flips the condition to true
(3 > 1). That's the case where the choice of target changes the outcome, and the only one that
distinguishes "count both battlefields after the token" from "count only yours".
✅ Fixed — added as a fourth test; `playTruce` gained a `tokenTo` parameter.

### Test results (`--rerun-tasks`, on the final tree)

- `EmeritusCycleScenarioTest` — all 4 Emeritus of Truce cases **PASSED**, whole class green
- `SosPrepareCardsScenarioTest` — **PASSED**
- `:mtg-sets:test CardDefinitionSnapshotTest` — **BUILD SUCCESSFUL**, including `SOS: card trees
  match golden`, so the snapshot edit is exactly what a re-bless produces
- **0 failures** across every run

### Pre-existing, outside this diff

- `docs/card-sdk-language-reference.md` documented a `MakePreparedEffect` /
  `PreparedService.makePrepared`. Neither symbol exists — it's `BecomePreparedEffect` /
  `PreparationLogic.makePrepared`, correctly described seven lines above. Stale duplicate entry.
  ✅ Removed in `0e81fdf8` (safe, and the diff already touches this area).
- `Conditions.OpponentControlsMoreCreatures` uses `AggregateBattlefield(Player.EachOpponent, …)`, and
  `DynamicAmountEvaluator.evaluateBattlefieldAggregate` flatMaps across all opponent ids — so it
  compares the **combined** opponent creature count to yours. Correct in two-player, wrong in
  multiplayer, where "an opponent controls more creatures than you" means any single opponent. Also
  affects `OpponentControlsMoreLands` and Beza, the Bounding Spring. **Not fixed** — changing a
  shared primitive needs its own branch and tests.

### Recommended follow-up

This bug class has a natural home in `CardLinter`. `AttachedScopeGrantOnNonAttachment` was added for
precisely this shape of failure (Harmonious Grovestrider shipping with no ward because a default
filter made a printed static a silent no-op). A `PreparedKeywordMismatch` finding — asserting, for
every `CardLayout.PREPARE` card, that `keywords.contains(PREPARED)` matches whether the oracle text
contains "enters prepared" — would make the whole class a build failure across all sets. It's a pure
local check on the `CardDefinition`, no network, and `CardLintTest` + `lint-allowlist.txt` already
provide the gate and the escape hatch. Kept out of this PR as an SDK addition with its own
language-reference update.

PR #1794, opened 2026-08-11
Bind {X} at cast time; make Leatherhead's counter removal a real cost

bolav:
Two playtest bugs, plus the fixes from reviewing them.

## Day of Black Sun destroyed every creature

A caster that announced no X — the AI's `CastSpell` carries none, and a synthesized free cast never picks one — left `SpellOnStackComponent.xValue` null all the way to resolution, where `CardPredicate.ManaValueAtMostX` fails *open*. That fail-open is deliberate: legal-action enumeration runs before X is chosen, so an X spell must still be offered, which is why the resolution-only siblings `ToughnessAtMostX` / `PowerAtLeastX` fail closed instead. So the fix binds X rather than touching the predicate.

Per CR 601.2b a spell with `{X}` in its cost has X announced as it is cast, and a caster who paid nothing for X cast it for 0; for the free-cast path CR 107.3b is explicit that 0 is the only legal choice. `StackResolver` resolves that at cast time and feeds it to both `SpellOnStackComponent` and `SpellCastEvent`, so the board wipe filters correctly and the game log's cast line carries `(X=0)` — previously absent for the same reason. Spells with no `{X}` still report null, so their log lines are unchanged.

Guarded against the two paths that share the X slot: mana payment already computes `action.xValue ?: 0`, so X=0 matches what was actually paid, and a `PayXLife` additional cost can never co-exist with an `{X}` mana cost (`Costs.kt`).

## Leatherhead, Swamp Stalker spent her hexproof counter every time

"You may remove a counter from her" hardcoded `Counters.HEXPROOF`, so once anything had put +1/+1 counters on her (Ouroboroid's begin-combat trigger) there was no prompt at all and her hexproof went every time. The card wants any kind, controller's choice.

It also closes an adjacent hole in the same path: both counter-removal executors no-op on a permanent with no counters and report success, which armed the "when you do" payoff for free — Leatherhead would have kept destroying an artifact every combat after her last counter was gone. `ReflexiveTriggerEffectExecutor.isActionFeasible` now gates on the counters actually being there, alongside the existing sacrifice/discard/evidence gates.

## Review follow-ups (second commit)

**The removal needed a floor, not just a ceiling.** Modelled as `RemoveCountersUpTo(1, …)`, accepting the "may" and then answering 0 at every prompt removed nothing and still reported success — so CR 603.12's reflexive "when you do" fired for a "you do" that never happened. That is the same free-payoff hole the feasibility gate closes, one step further in. `RemoveAnyNumberOfCountersEffect` gains `minTotal` to match its `maxTotal`, exposed as `Effects.RemoveCounterOfAnyKind`. Each prompt's minimum is the share of the floor the kinds still to come can't cover, so the controller keeps a free choice of *which* kind while the floor is reachable and is held to it on the last kind that can pay; the resumer coerces into `[min, max]`, so enforcement is server-side. A prompt whose bounds coincide is applied rather than asked, so a permanent with one kind of counter raises no prompt at all.

**The feasibility gate failed closed on targets it couldn't resolve.** It read counters through the stateless resolver and scored an unresolvable target as zero. Feasibility runs before the action, so a pipeline slot the same composite fills a step earlier isn't stored yet — treating that as "no counters" silently deleted the whole ability. It now uses the state-aware resolver and fails open on "don't know", matching the walker's own `else -> true` policy.

**"That player controls" was filtered as "an opponent controls."** Leatherhead and Dawning Purist both print the damaged player, but filtered on any opponent — the two coincide only in a two-player game. Both now use `controlledByTriggeringPlayer()`, the primitive Rustmouth Ogre already uses for the identical template.

Also re-blesses the ONS and TMT card snapshots. TMT was stale as of the first commit, which changed Leatherhead's definition without updating the golden — `CardDefinitionSnapshotTest` was failing on the branch until now.

## Tests

`DayOfBlackSunScenarioTest` (4) covers the undeclared-X wipe, an explicit X=2, the cast event's reported X, and a no-`{X}` control case. `LeatherheadSwampStalkerScenarioTest` (7) covers kind selection, the zero-answer floor, the single-kind no-prompt path, and the no-counters-left case. `ReflexiveTriggerFeasibilityTest` (+5) covers both new `isActionFeasible` branches, present and absent, plus the fail-open pipeline-target case.

Verified with `:rules-engine:test` over the seven affected classes (37 passed) and `:mtg-sets:test --rerun-tasks` (`CardDefinitionSnapshotTest` 308/308). A full `test` gate ran 10,637/10,638; the one failure was `ConniveTargetingTest` timing out after 120s under parallel-module contention on a memory-limited box — it passes in isolation, its sibling `ConniveTest` passed in the same run, and it touches none of the changed code paths.

The fail-open branch was confirmed load-bearing by temporarily reverting it: exactly one test failed, that one.

**Not covered:** the multiplayer half of the "that player controls" fix. `GameTestDriver.initGame` hard-codes two `PlayerConfig`s, and two players is exactly where `controlledByTriggeringPlayer()` and `opponentControls()` agree. The swap is proven correct-by-precedent and proven not to regress two-player play, but the divergence it fixes has no test; extending the driver to three players is a larger change than this warranted.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1132,6 +1132,15 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
   with `maxTotal` set, not a separate effect: one `ChooseNumber` prompt per kind, each capped at
   `min(kind's count, remaining budget)`; prompting stops once the budget is spent. Used by Heartless Act's
   "Remove up to three counters from target creature."
+- `RemoveCounterOfAnyKind(target, count = 1)` — "**remove a counter** from it": the player chooses which
+  *kind* comes off, but not *whether*. The floored form of the same effect (`minTotal = maxTotal = count`).
+  Each prompt's minimum is the share of the floor the kinds still to come can't cover, so the choice stays
+  free while the floor is reachable and is enforced on the last kind that can pay it; a prompt whose min and
+  max coincide is applied without asking, so a permanent carrying a single kind raises no prompt at all.
+  **Reach for this, not `RemoveCountersUpTo(1, …)`, whenever a payoff hangs off the removal** — `IfYouDo` /
+  `ReflexiveTriggerEffect` ("remove a counter … when you do, …", Leatherhead, Swamp Stalker). A bare ceiling
+  lets the player answer 0 to every prompt and still report success, firing the payoff for free against
+  CR 603.12. Use `RemoveCountersUpTo` only where the card really does say "up to".
 - **Player-scoped counters (CR 122.1, 107.14).** Every counter type above lives on a permanent/object. Poison,
   energy, and rad counters instead live directly on a **player entity**, reusing the same `CountersComponent` —
   no separate component or data model. `AddCountersExecutor` already resolves player-shaped targets (`that
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/dsl/Effects.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/dsl/Effects.kt
@@ -1762,6 +1762,24 @@ object Effects {
         com.wingedsheep.sdk.scripting.effects.RemoveAnyNumberOfCountersEffect(target, maxTotal = maxCount)
 
     /**
+     * "Remove a counter from [target]" — the player picks *which kind*, but not *whether*. The
+     * floored form of [RemoveCountersUpTo]: exactly [count] counters come off, chosen across
+     * whatever kinds the permanent carries.
+     *
+     * Reach for this, not `RemoveCountersUpTo(1, …)`, whenever a payoff hangs off the removal
+     * ("remove a counter … when you do, …", Leatherhead, Swamp Stalker; "… if you do, draw a card",
+     * Mister Hyde, Monster Within). A bare ceiling lets the player answer 0 to every prompt and
+     * still succeed, which fires the payoff for free. Use [RemoveCountersUpTo] only where the card
+     * really does say "up to".
+     */
+    fun RemoveCounterOfAnyKind(
+        target: EffectTarget = EffectTarget.ContextTarget(0),
+        count: Int = 1
+    ): Effect = com.wingedsheep.sdk.scripting.effects.RemoveAnyNumberOfCountersEffect(
+        target, maxTotal = count, minTotal = count
+    )
+
+    /**
      * "[player] gets N [counterType] counters" (CR 122.1 — counters placed on a player rather
      * than a permanent). Sugar for [AddCounters] targeting the player directly; no new plumbing —
      * `AddCountersExecutor` already resolves player-shaped targets the same way it does for
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/effects/CounterEffects.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/effects/CounterEffects.kt
@@ -137,32 +137,51 @@ data class RemoveCountersEffect(
 
 /**
  * Remove counters from a target permanent. The controller chooses how many counters of each kind
- * to remove (0 up to the current count of that kind). With [maxTotal] set, no more than that many
- * counters may be removed in total across all kinds ("remove up to N counters"); left null, there
- * is no cap ("remove any number of counters").
+ * to remove. [maxTotal] caps the total across all kinds ("remove up to N counters"); left null,
+ * there is no cap ("remove any number of counters"). [minTotal] is the matching *floor* — the
+ * number the controller must remove in total once the effect resolves at all.
  *
  * "Remove any number of counters from target creature you control." ([maxTotal] = null.)
  * "Remove up to three counters from target creature." (Heartless Act, [maxTotal] = 3.)
+ * "Remove a counter from it." ([minTotal] = [maxTotal] = 1 — the kind is the player's choice, the
+ * count is not.)
+ *
+ * The floor matters because a bare ceiling silently makes every such removal optional: "remove a
+ * counter" modelled as `maxTotal = 1` alone lets the controller answer 0 to every prompt and still
+ * report success, which hands a free payoff to any "if you do" / "when you do" clause hanging off
+ * it (CR 603.12 — the reflexive ability triggers only when the action is actually taken).
  *
  * At resolution time, the executor enumerates each counter kind currently on the target and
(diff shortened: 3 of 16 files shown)


bolav:
## Review

Reviewed `10d9dd3835` (the two playtest fixes) against `origin/main`, which the branch already sat directly on — the merge was a no-op, so no conflicts to flag. Findings below, each with what was done about it in the follow-up commit.

### What was good

- **No new SDK types.** Both halves compose what exists — `RemoveCountersUpTo` for the card, two branches on the existing `isActionFeasible` `when`, one binding in `StackResolver`. That's the right instinct and the reason this review is short on SDK objections.
- **The X binding is at the correct layer.** `StackResolver.castSpell` is the single funnel every cast path goes through (AI, synthesized free casts, alternative costs), so binding there covers all of them; changing `ManaValueAtMostX` instead would have broken legal-action enumeration. Verified: the resolution-only siblings really do fail closed (`PredicateEvaluator.kt:499,518`) while `ManaValueAtMostX` fails open (`:407`), exactly as the commit message says.
- **The two paths that share the X slot were checked and are safe** — payment already computes `action.xValue ?: 0`, so X=0 matches what was paid, and `Costs.kt:569` forbids a `PayXLife` card from also carrying `{X}`. Guarding on `castCost.hasX` so non-`{X}` spells still report null is the right conservatism, and the Doom Blade control test locks it in.
- Both test files are one-card-one-file in `rules-engine/scenarios/`. Oracle text verified against Scryfall for both cards; CR 601.2b and 603.12 verified against the local Comprehensive Rules.

### Blocking

**1. "Yes, then remove zero" still fired the destroy.** `LeatherheadSwampStalker.kt:58`

The card comment claimed taking off zero was "already covered here by the `optional = true` decline". Declining skips everything, but *accepting* and then answering 0 at every prompt did not:

- `RemoveAnyNumberOfCountersExecutor` built every `ChooseNumberDecision` with `minValue = 0`.
- `MiscContinuationResumer.kt:684` guarded the removal with `if (chosen > 0)`; at 0 no `CountersRemovedEvent` was emitted and it fell straight through.
- `CoreAutoResumerModule.kt:254` then resumed `ReflexiveTriggerTargetContinuation` and called `buildReflexiveTriggeredEvent` **unconditionally** — the paused path had no success check at all.

So Leatherhead destroyed an artifact or enchantment every combat, forever, never spending a counter — a strictly better version of the bug the commit's third hunk closes. CR 603.12's own example is explicit that the reflexive ability triggers only when the action is actually taken.

Mister Hyde was not the precedent the comment assumed: it survives this because it wraps the same primitive in `Effects.IfYouDo(..., successCriterion = SuccessCriterion.CountersRemoved)`. `ReflexiveTriggerEffect` has no equivalent gate and can't borrow `IfYouDo` — the point of the reflexive shape is a separate stack object with its own target.

**Fixed** by giving `RemoveAnyNumberOfCountersEffect` a `minTotal` floor to match its `maxTotal` ceiling, exposed as `Effects.RemoveCounterOfAnyKind`. Each prompt's minimum is the share of the floor the later kinds can't cover, so the choice of *which* kind stays free while the floor is reachable and binds on the last kind that can pay; the resumer coerces into `[min, max]` so it's enforced server-side. Same ceiling-without-floor trap as `dynamicChooseCount` / `dynamicMinChooseCount` on modal spells.

**2. The new engine gate had no test.** `ReflexiveTriggerEffectExecutor.kt:220-228`

The commit claimed `isActionFeasible` now suppresses the may-clause when the counters are gone, but neither test file put a counterless Leatherhead into combat, and the `RemoveCountersEffect` branch (Slumbering Walker's shape) was untouched by any test. `ReflexiveTriggerFeasibilityTest.kt` already existed for exactly these branches and wasn't opened.

**Fixed** — five tests added there covering both branches present and absent, plus a permanent short of the full count.

### Important

**3. The feasibility gate failed closed on target shapes it couldn't resolve.** `ReflexiveTriggerEffectExecutor.kt:246`

`countersOn` called the **stateless** `resolveTarget` overload, whose `else -> null` swallows `EnchantedCreature`, `EquippedCreature`, `ChosenCreature`, `TargetController` — and returns null for a `PipelineTarget` stored by an earlier step of the *same* composite, since feasibility runs before execution. In every one of those cases it returned 0 and the entire may-ability silently vanished.

Nothing broke today (Leatherhead and Slumbering Walker both use `EffectTarget.Self`), but it was primed for the next Equipment or `SelectTarget`-then-remove card, and it contradicted the method's own stated policy of not gating on shapes it doesn't recognize.

**Fixed** — `countersOn` returns `Int?` and uses the state-aware overload; callers fail open on `null`. Confirmed load-bearing by temporarily reverting it: exactly one test failed, the new fail-open one.

**4. `opponentControls()` doesn't mean "that player controls".** `LeatherheadSwampStalker.kt:62`

Scryfall confirms "destroy target artifact or enchantment **that player controls**" — the player just dealt combat damage. The filter was `opponentControls()`, i.e. *any* opponent, so in multiplayer Leatherhead could destroy a permanent belonging to someone she never hit. The line was pre-existing, but the commit added a comment explicitly asserting the correct semantics, which made it worth closing rather than enshrining.

**Fixed** on Leatherhead and on Dawning Purist, which had the same approximation. Both now use `controlledByTriggeringPlayer()` — the primitive Rustmouth Ogre already uses for the identical template, and which works inside a reflexive trigger (Dreadmaw's Ire).

Caveat worth recording: `GameTestDriver.initGame` hard-codes two `PlayerConfig`s, and two players is exactly where the two filters agree — so the swap is proven not to regress two-player play, but the divergence it fixes has no test.

### Minor

- `StackResolver.kt:261` — **CR 107.3b** is the on-point rule for the free-cast case ("the only legal choice for X is 0"); 601.2b covers announcement generally. **Fixed**: both are now cited.
- `LeatherheadSwampStalker.kt:49` — bestow doesn't place counters. **Fixed**: dropped from the comment.
- `DayOfBlackSunScenarioTest.kt:118` — the Doom Blade case tests the `hasX` guard rather than Day of Black Sun; an engine-level `XValueBindingTest` would be the more honest home. Left as-is, not worth churn.
- The Leatherhead tests inject the hexproof counter via `replaceState` rather than letting `EntersWithCounters` place it, so they'd still pass if that replacement broke. Left as-is; fine for a focused test.

### Also found while verifying

The branch was shipping a **stale snapshot golden**. `TMT.json` still described Leatherhead's old `RemoveCounters/hexproof` action with no `descriptionOverride`, and the commit touched nothing under `mtg-sets/src/test/resources/` — so `CardDefinitionSnapshotTest` was failing on the branch as delivered. Re-blessed in the follow-up commit; only ONS and TMT moved, and only those two cards within them.

PR #1795, opened 2026-08-12
Gate every controller-grant marker behind its condition

bolav:
# Gate every controller-grant marker behind its condition

Player-level grants ("you have shroud", "you can't lose the game") sit outside the Rule 613 layer
system: `StaticAbilityHandler` stamps a marker component **once**, as the permanent enters, and
readers later scan the battlefield for it. Because the stamp happens once, an "as long as …" gate
can't be resolved at stamp time — it has to travel on the marker and be re-evaluated on every read.

Only hexproof did that. The eight siblings matched the bare type, so wrapping one in a
`ConditionalStaticAbility` stamped **no marker at all** and the ability was *silently inert* — not
stuck on, not stuck off, simply absent, with no compile error or warning. That is the bug Captain
America, Super-Soldier hit for hexproof, and it was waiting for the next card to hit for any of the
other eight.

## What changed

- **`ControllerGrantMarker`** — one interface carrying the condition, with the stamp/read rules that
  make the silent no-op unrepresentable.
- **`ControllerGrants`** — the single place a gate is evaluated. `ControllerHexproof` delegates to
  it; `ControllerShroud` is new and collapses the duplicated battlefield scans in `TargetValidator`,
  `TargetEnumerationUtils`, `TargetFinder` and `ClientStateTransformer`.
- **`StaticAbilityHandler`** routes all nine grants through `controllerGrant<A>()`. Protection gates
  per *scope*, so `GrantsControllerProtectionComponent` now holds `List<ProtectionGrant>` rather than
  `List<ProtectionScope>` — one card can carry several `GrantProtectionToController` abilities and
  gate them independently.

## Fixes found along the way

**Controller grants now follow the projected controller.** Control change is a Layer 2 floating
effect that never writes back to `ControllerComponent` (`EffectApplicator`'s `ChangeController`
branch), so every one of these scans was matching on a stale base controller: steal a True Believer,
Sigarda or Platinum Angel and the grant kept protecting the player it was taken from, while its gate
was evaluated as the thief. `ControllerGrants.granterController` resolves it through the projection
now — the same rule `ReplacementEffectProcessor` already follows for `Player.You` replacements.

**Empty-library and effect-driven losses are team-aware.** `DrawCardPrimitive` and `LoseGameExecutor`
each had their own can't-lose scan and neither was team-aware, so in Two-Headed Giant a teammate's
Platinum Angel stopped the 704.5a life SBA but not an empty-library draw or a `LoseGameEffect`.
CR 810.8a makes it team-wide; both now call the shared `playerCantLoseGame`.

**Player shroud is visible in the client.** `ClientStateTransformer` only ever checked
`PlayerShroudComponent`, so a True Believer's controller was untargetable but shown as plain — the
badge that matters most, since shroud stops the player's *own* spells (CR 702.18) and the client
would otherwise offer a self-target the server then rejects. Shroud and hexproof both read through
the facades now, which also removes the transformer's own battlefield scan.

## Tests

- **`ConditionalControllerGrantsTest`** covers all nine grants bare and gated, asserts the gate flips
  at read time, and fails if a new marker is added without a case.
- **`ControllerGrantFollowsProjectedControllerTest`** steals a True Believer with Control Magic and
  asserts shroud both leaves the owner *and* reaches the thief — a half-fix would leave it granting
  to nobody.
- **`PlayerGrantVisibilityTest`** covers the badge appearing, staying controller-only, leaving with
  the permanent, tracking a gate that flips without the granter re-entering, and not duplicating.

## Verification

`just test` — BUILD SUCCESSFUL across all modules, 0 failures. `game-server` is included because
`ClientStateTransformer` changed. The suites this touches all pass: `ConditionalControllerGrantsTest`,
`CaptainAmericaSuperSoldierScenarioTest`, `TrueBelieverTest`, `SigardaHostOfHeronsScenarioTest`,
`TwoHeadedGiantTeamLossTest`, `TeamVsTeamTest` and `ReplacementFollowsProjectedControllerTest`.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -5133,6 +5133,17 @@ staticAbility {
   enchanted permanent is a land, it has '{T}: Add two mana of any one color'" is a
   `ConditionalStaticAbility(GrantActivatedAbility({T}: AddAnyColorMana(2)), EnchantedPermanentMatches(Land))` — the
   granted mana ability switches on/off continuously with the host's type.
+  The same holds for the **player-level controller grants** — `GrantShroudToController`,
+  `GrantHexproofToController`, `GrantProtectionToController`, `OpponentsCantMakeYouSacrifice`,
+  `GrantCantLoseGame`, `GrantOpponentsCantWinGame`, `GrantCantLoseGameFromLife`, `StationUsingToughness`,
+  `CantBeTargetedByOpponentAbilities`. Those are stamped as marker components once, as the permanent
+  enters, rather than projected each pass, so the gate travels on the marker and every reader
+  re-evaluates it against current state via `ControllerGrants` (Captain America, Super-Soldier:
+  `ConditionalStaticAbility(GrantHexproofToController, Conditions.SourceHasCounter(Named(Counters.SHIELD)))`).
+  Adding a new marker of this kind means stamping it through `StaticAbilityHandler.controllerGrant<A>()`
+  and reading it through `ControllerGrants` — a bare `any { it is A }` at the stamp makes the gated form
+  *silently inert*, and a bare `has<M>()` at the read leaves it stuck on. `ConditionalControllerGrantsTest`
+  enforces both.
 - `CompositeStaticAbility(abilities)` — bundles several component static abilities into **one** printed static ability
   whose single continuous effect spans multiple Rule 613 layers (CR 613.6). Use it when one printed ability grants a
   *combination* of type/subtype/color/keyword/P/T changes to the same objects — e.g. Bello, Bard of the Brambles: "each
@@ -5671,7 +5682,9 @@ staticAbility {
   [scope]" while this permanent is on the battlefield (Absolute Virtue: "You have protection from each
   of your opponents."). The continuous, static counterpart of the one-shot `GrantPlayerProtection`
   (The One Ring) — for a player only the **D**amage and **T**argeting legs of DEBT apply. Projected to
-  `GrantsControllerProtectionComponent(scopes)`; `PlayerProtectionRules.isProtectedFromSource` unions
+  `GrantsControllerProtectionComponent(grants)` — a list of `ProtectionGrant(scope, condition)`, gated
+  **per scope**, so one permanent can carry several of these abilities and wrap only some in a
+  `ConditionalStaticAbility`. `PlayerProtectionRules.isProtectedFromSource` unions
   these battlefield-sourced scopes with any one-shot `PlayerProtectionComponent` on the player, so the
   protection appears/disappears with the permanent (no cleanup). General over any `ProtectionScope`
   (single color, `Everything`, `EachOpponent`, …), mirroring the controller-grant statics
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/DynamicAmountEvaluator.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/DynamicAmountEvaluator.kt
@@ -5,6 +5,7 @@ import com.wingedsheep.engine.handlers.effects.LkiPolicy
 import com.wingedsheep.engine.handlers.effects.TargetResolutionUtils
 import com.wingedsheep.engine.handlers.effects.lkiPolicyFor
 import com.wingedsheep.engine.handlers.effects.lkiSnapshotFor
+import com.wingedsheep.engine.mechanics.ControllerGrants
 import com.wingedsheep.engine.mechanics.layers.ProjectedState
 import com.wingedsheep.engine.state.GameState
 import com.wingedsheep.engine.state.ZoneKey
@@ -1128,10 +1129,12 @@ class DynamicAmountEvaluator(
     ): Boolean {
         val projected = state.projectedState
         val controller = projected.getController(entityId) ?: fallbackControllerId ?: return false
+        // Scanned here rather than through ControllerGrants.grantedTo so the controller match
+        // stays on the *projected* controller, as it always has — a control-change effect
+        // re-points the grant. The gate still goes through ControllerGrants.
         return state.getBattlefield().any { permanentId ->
-            val perm = state.getEntity(permanentId) ?: return@any false
-            perm.has<GrantsStationUsingToughnessComponent>() &&
-                projected.getController(permanentId) == controller
+            projected.getController(permanentId) == controller &&
+                ControllerGrants.isActiveOn<GrantsStationUsingToughnessComponent>(state, permanentId)
         }
     }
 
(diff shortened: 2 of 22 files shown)


bolav:
## Review — `gate-controller-grants`

Reviewed at `04176bf` (the original commit); findings marked **fixed** were addressed in the
follow-up commit on this branch, so the diff you see already contains them. Findings marked **open**
are left for the author's judgement.

### Verdict

The diagnosis is right and the shape is right: this is engine plumbing over one new marker interface,
not a new SDK type, and it turns a whole class of silent no-ops into something you have to opt out of.
Two issues blocked at review time — a base-vs-projected controller read and a wrong CR number — both
now fixed.

### What's good

- **The framing.** "Not stuck on, not stuck off, simply absent" is the correct characterisation, and
  the fix targets the representation (`ControllerGrantMarker` + `controllerGrant<A>()`) rather than
  patching the eight call sites one at a time.
- **No new SDK type.** `ControllerGrants` is an engine object over an existing interface;
  `ProtectionGrant` is a two-field record. Nothing here forces a new `Effect`/`StaticAbility` on the
  next card.
- **Per-scope protection gating is the right call**, and `controllerGrants<A>()` as a deliberate
  plural counterpart, with the "protection from red, and from blue as long as you control an Island"
  example, justifies why this one marker is shaped differently.
- **`isActive` short-circuits on `condition == null`** before touching `projectedState`, so
  unconditional grants pay nothing on the targeting-enumeration hot path.
- **The shroud de-dup is real** — verified no leftover copies of the scan remain in `game-server` or
  `ai`.
- **Test clues name the failure mode**, not the assertion ("…stamped no marker — the ability is
  silently inert"). That's the message a future author actually needs.

### Blocking — both fixed

**1. `ControllerGrants.anyGranting` read the base controller while `isActive` read the projected one.**
`ControllerGrants.kt`. Control change is a Layer 2 floating effect (`GainControlExecutor`,
`EffectApplicator`'s `ChangeController` branch) and never writes back to `ControllerComponent` — the
repo documents exactly this at `ReplacementEffectProcessor.kt:410-412`. So Mind Control on an
opponent's Sigarda left the *opponent* sacrifice-immune and not you; same for Shalai, True Believer,
Platinum Angel, Marina Vendrell's Grimoire, Herald of Eternal Dawn. Within a single read the two
halves disagreed: `anyGranting` awarded the grant to the old controller while `isActive` evaluated
its "you control an Island" gate as the new one.

Pre-existing at most call sites, but `anyGranting` was new code and the single line every grant
funnels through, and the diff had already noticed the divergence and worked around it in
`DynamicAmountEvaluator` rather than fixing it.

**Fixed** — `ControllerGrants.granterController()` resolves through the projection with a base
fallback (needed: `TeamVsTeamTest`/`TwoHeadedGiantTeamLossTest` hand-build containers with no
projection entry). `PlayerProtectionRules` follows, and `DynamicAmountEvaluator`'s bespoke scan
collapses back to a single `grantedTo(...)`. Covered by
`ControllerGrantFollowsProjectedControllerTest`, which asserts both directions.

**2. CR 704.5c is poison, not empty library.** `DrawCardPrimitive.kt`. The diff rewrote that KDoc
line and kept "Handles empty-library loss (Rule 704.5c)"; the inline comment repeated it. 704.5b is
*"If a player attempted to draw a card from a library with no cards in it since the last time
state-based actions were checked, that player loses the game"*; 704.5c is ten-or-more poison
counters.

**Fixed** — both citations now read 704.5b.

### Important — open

**3. The team-reach fix has no test.** `DrawCardPrimitive` and `LoseGameExecutor` both now cite
CR 810.8a and both changed behaviour, but `TwoHeadedGiantTeamLossTest:136` only covers the 704.5a
life SBA and `:95` covers decking *without* a grant. Per the project rule that a cited rule needs a
test that exercises it, add two cases: teammate controls the can't-lose grant and (a) the other
teammate draws from an empty library, (b) a `LoseGameEffect` resolves targeting them. Both fail on
`origin/main` and pass here — exactly the pair worth locking in.

**4. The roster-completeness test enforces nothing.** `ConditionalControllerGrantsTest.kt:300-317`.
`knownMarkers` is hand-maintained in the same file as `grantCases`, so adding a ninth marker to
*neither* list passes silently. Yet three KDocs assert otherwise (`StaticAbilityHandler.kt:361`,
`BattlefieldComponents.kt:779`, and the doc reference). The test comment concedes reflection would
need classpath scanning — but it doesn't, because all eight implementors sit in one package in one
module. Make the interface sealed:

```kotlin
sealed interface ControllerGrantMarker { val condition: Condition? }
```

then the roster check becomes authoritative and the KDoc claims become true:

```kotlin
ControllerGrantMarker::class.sealedSubclasses.toSet() shouldBe grantCases.map { it.marker.kotlin }.toSet()
```

Without this, either fix the test or soften the three claims — as written they're a name lie about
the guarantee.

### Minor — open

- **`isActiveOn`'s KDoc says "for the self-scoped markers"** (`ControllerGrants.kt`) and names only
  `CantBeTargetedByOpponentAbilitiesComponent`, but `DynamicAmountEvaluator` calls it for
  `GrantsStationUsingToughnessComponent`, which is controller-scoped. Reword to "when the caller has
  already decided which permanent to ask about."
- **`addContinuousEffectComponentFromAbilities`** (the token path) stamps no markers at all, so a
  token with "you have shroud" gets nothing — a second, quieter hole in the "stamp it through
  `controllerGrant<A>()`" invariant that `ControllerGrantMarker`'s KDoc states unconditionally.
- **`playerCantLoseGameFromLife`'s KDoc** asserts "it is not a team-wide 'can't lose' grant."
  CR 810.8a is unqualified — *"If an effect says that a player can't lose the game, that player's
  team can't lose the game"* — and Marina Vendrell's Grimoire is a can't-lose effect with a
  restricted cause. Pre-existing and arguable, but the comment states it as settled.
- **The same three-paragraph explanation appears six times** (`ControllerGrantMarker`,
  `ControllerGrants`, `ControllerShroud`, `ControllerHexproof`, `controllerGrant<A>()`, the test
  header) plus the doc reference. The *why* here is genuinely non-obvious so it earns a thorough
  home — but one canonical copy with the rest pointing at it would drift less.
- The doc reference says a new marker means stamping through `controllerGrant<A>()` — worth
  mentioning `controllerGrants<A>()` for the multi-instance case, since `GrantProtectionToController`
  is in the list right above it.

### Also fixed in the follow-up

**Player shroud was invisible in the client.** `ClientStateTransformer` checked only
`PlayerShroudComponent`, so a True Believer's controller was untargetable but shown as plain, while
hexproof got a badge. Shroud and hexproof now both read through the facades, which also removed the
transformer's own battlefield scan — the fifth copy, and the last base-controller read of a grant —
and collapsed the two duplicate hexproof blocks into one. `ControllerHexproof.isGrantingNow` had no
callers left and is gone. Covered by `PlayerGrantVisibilityTest`.

PR #1796, opened 2026-08-12
Add an ability-source predicate for stack targets: CardPredicate.AbilitySourceMatches (+2 MSH cards)

bolav:
# Add an ability-source predicate for stack targets (Echo, Perceptive Prodigy · Scientist Supreme of A.I.M.)

Marvel Super Heroes has two cards that copy an ability on the stack, restricted by what *kind of
object* produced it:

- **Echo, Perceptive Prodigy** [51] — "Copy target activated or triggered ability you control **from
  a creature source**."
- **Scientist Supreme of A.I.M.** [225] — "…**from an artifact source**."

Everything else about them already shipped: `Targets.ActivatedOrTriggeredAbilityYouControl` and
`Effects.CopyTargetSpellOrAbility` back Gogo, Master of Mimicry and Peter Parker's Camera today.

## The gap

An ability on the stack is its own object and carries none of its source's characteristics
(CR 113.3b/c); it has no `CardComponent` at all. `PredicateEvaluator.matchesCardPredicate` bails at
that missing component, so `GameObjectFilter.Artifact` applied to an ability entity is
unconditionally false. There was no way to express "the ability's source is a creature".

## The design

`CardPredicate.AbilitySourceMatches(subfilter: GameObjectFilter)` — resolved in the evaluator's
stack branch, immediately alongside its sibling `CardPredicate.TargetsMatching`, *before* the
`CardComponent` null-check. It reads `ActivatedAbilityOnStackComponent.sourceId` /
`TriggeredAbilityOnStackComponent.sourceId` — the source per **CR 113.7** — and re-runs the ordinary
filter machinery against that entity, inheriting the outer `PredicateContext`.

**Why a `CardPredicate` rather than a one-off target type.** The clause is a *narrowing of an
existing target*, not a new kind of target: the requirement is still "activated or triggered ability
you control", with one more condition on it. Modelling it as a predicate means it composes with
everything the SDK already has — `And`/`Or`/`Not`, controller predicates, and any subfilter at all
(`Creature`, `Artifact`, `Creature.youControl()`, a subtype, an `anyOf` union) — so the next card
that says "from an enchantment source" or "from a Villain source" is a filter argument, not another
`TargetRequirement`. `TargetsMatching` set exactly this precedent for "spell that targets a
creature"; this is the same shape pointed at the source instead of the targets.
`CantBeTargetedBySourceTypeAbilities` (Artifact Ward) is the engine-side precedent for keying a
restriction off an ability's source, and the source-characteristic resolution here matches its
`SourceTypeTargeting.sourceCardTypes` exactly.

Surface added: `GameObjectFilter.abilitySourceMatches(...)`, `TargetFilter.abilitySourceMatches(...)`,
and the ready-made `Targets.ActivatedOrTriggeredAbilityYouControlFrom(sourceFilter)`.

## Last known information

Once an ability is on the stack it exists independently of its source, and by the time it can be
targeted the source is routinely gone — a dies trigger's source is already in the graveyard. The
match therefore reads the source's **projected** characteristics while it is on the battlefield and
falls back to its **printed** ones once it has left, which is what CR 113.7a ("if the source is no
longer in the zone it's expected to be in at that time, its last known information is used") and
CR 608.2b ("if the source of an ability has left the zone it was in, its last known information is
used during this process") call for. That is equivalent resolution to what `SourceTypeTargeting` does
for Artifact Ward (a parallel implementation, not shared code). A scenario test kills a creature with
Lightning Bolt and proves Echo can still copy the resulting dies trigger.

### Deleted tokens — the Clue/Food case

A **token** source is a harder case: CR 704.5d ("If a token is in a zone other than the battlefield,
it ceases to exist") makes `TokensInWrongZonesCheck` delete the entity outright, so once a Clue has
sacrificed itself to pay for its own draw ability there is no `getEntity(sourceId)` row left to read.
That is not a corner case for this card — MSH itself prints four Clue/Food makers (Agent 13, Sharon
Carter; Panther Pounce; Ant-Man's Army; Heroic Feast), and cracking a Clue is the line Scientist
Supreme of A.I.M. is printed for.

The timing is worth stating precisely, because it makes the bug look intermittent from the table:
state-based actions are applied on the engine's post-resolution pass, so the token is still a
readable graveyard entity the instant its ability hits the stack, and only ceases to exist once
something else resolves — an opponent's response, a second ability of yours. Copy it immediately and
it worked; let one thing resolve first and the ability silently stopped being offered.

It is handled, using the store the engine already has rather than a new one:
`ActivatedAbilityOnStackComponent.lastKnownSourceSnapshot` is an `EntitySnapshot` frozen at
activation whenever the cost sacrifices or exiles the source (`costExilesOrSacrificesSelf`) — the
Clue / Food / Blood / Treasure shape verbatim. It now also carries the source's **projected type
line**, keywords and token-ness, captured through a `projectedTypeLine` helper shared with
`ZoneTransitionService` so the two capture sites can't drift. When `state.getEntity(sourceId)` is
null, the `AbilitySourceMatches` branch evaluates the subfilter against that snapshot via
`PredicateEvaluator.matchesSnapshot`. Reading the *projected* type line also gets an
animated-artifact / crewed-Vehicle source right on that path.

`matchesSnapshot` is deliberately **partial**: a snapshot freezes card types, sub/supertypes,
keywords, P/T, controller and token-ness, and a predicate it cannot answer reports *unanswerable*
(tri-state) rather than guessing — which keeps `Not`/`And`/`Or` honest and degrades to the
pre-existing "no match" answer.

**Remaining limits**, documented in the predicate KDoc, the DSL reference and `LOOP_UNIT.md`:

1. A **nontoken** source that left the battlefield some other way (it died, it was exiled) is read
   off the card in its new zone, so a type a continuous effect had granted it is no longer visible —
   its printed types answer.
2. A deleted **token** source is recoverable only for an *activated* ability whose cost sacrificed or
   exiled it; a triggered ability whose token source has ceased to exist has no equivalent snapshot
   on its stack object.
3. A snapshot answers only the characteristics it freezes, so a subfilter asking for a mana value or
   a color does not match a deleted-token source.

## Player-facing verification — and a bug it turned up

Scenario tests drive `GameAction`s directly, so they prove nothing about whether a card is
*playable*. Checking that here found a real defect that predates this unit:

`TargetEnumerationUtils.findValidSpellTargets` — the enumeration behind the legal actions the server
sends to clients and the AI — filtered **every** stack-zone target down to `state.isSpellOnStack(...)`
unconditionally. The authoritative `TargetFinder.findSpellTargets` has a seam for this
(`filterPermitsAbilitiesOnStack`); the enumerator did not. Consequences:

- No ability-targeting card ever had an ability **offered** as a legal target — not Echo, and not the
  already-shipped **Gogo, Master of Mimicry** or **Peter Parker's Camera**. The engine accepted the
  action if it was submitted; the UI had nothing to click.
- `ActivatedAbilityEnumerator` gates `holdPriority` on "top of stack ∈ `validTargets`", so those
  cards also never stopped auto-pass — the ability you meant to copy resolved before you got priority.

Fixed by extracting the question into one shared `StackObjectTargeting.permitsAbilities(filter)`
consulted by **both** readers, with `AbilitySourceMatches` counted as an ability-naming predicate.
"Target spell" filters stay spell-only (CR 112.1 vs 113.3b/c) — the existing
`TargetSpellExcludesAbilitiesTest` regression still passes and is re-asserted directly against the
new seam.

Both new cards also set `holdPriority = true` so auto-pass stops while a copyable ability of yours is
on top.

**Client:** no change was needed, verified by reading the two paths. `StackZone.handleStackItemClick`
already routes *both* the action-pipeline (`targetingState.validTargets`) and decision-path
(`decisionSelectionState.validOptions`) clicks for stack objects, and force-expands any collapsed
pile containing a targetable item; `TargetingOverlay` only diverts to the pile picker for
`Graveyard`/`Exile`, so a `Stack` requirement falls through to the click-the-board banner;
`ClientStateTransformer.transformAbilityOnStack` gives every ability on the stack a `ClientCard`, and
`useStackCards` includes it. Not exercised in a browser — see "Not done" below.

## Verified CR citations

Every rule number below was grepped out of the local Comprehensive Rules
(`MagicCompRules_20260619.txt`) and read before being written down; nothing is cited from memory.

- **113.3b / 113.3c** — activated and triggered abilities exist on the stack as their own objects.
- **113.7** — "The source of an activated ability on the stack is the object whose ability was
  activated… of a triggered ability… the object whose ability triggered."
- **113.7a** — "Once activated or triggered, an ability exists on the stack independently of its
  source… if the source is no longer in the zone it's expected to be in at that time, its last known
  information is used."
- **608.2b** — "If the source of an ability has left the zone it was in, its last known information
  is used during this process."
- **707.10 / 707.10b / 707.10c** — copying an ability; the copy has the same source; "you may choose
  new targets for the copy".
- **112.1** — a spell is a card on the stack (the spell-vs-ability distinction the enumeration seam
  turns on).
- **117.5** — "Each time a player would get priority, the game first performs all applicable
  state-based actions…" (why the deleted-token window opens only after something resolves).
- **704.5d** — "If a token is in a zone other than the battlefield, it ceases to exist" (the
  deleted-token case above; **111.7** states the same rule from the token's side). An earlier draft of
  this PR cited 704.5s for it — that number is the Saga final-chapter sacrifice and was wrong. Note
  the repo has four *pre-existing* wrong 704.5s citations for the token rule
  (`TokensInWrongZonesCheck.kt`, `DeathAndLeaveTriggerDetector.kt`, `TriggerDetector.kt`); they are
  left alone as out of scope for this unit, but worth a sweep so the next agent doesn't copy them.

## Files

| Area | Change |
|---|---|
| `mtg-sdk` | `CardPredicate.AbilitySourceMatches`; `GameObjectFilter.abilitySourceMatches`; `TargetFilter.abilitySourceMatches`; `Targets.ActivatedOrTriggeredAbilityYouControlFrom` |
| `rules-engine` | `PredicateEvaluator` stack branch + snapshot fallback (`matchesSnapshot`) + 4 exhaustive-`when` arms; new `StackObjectTargeting`; `TargetFinder` uses it; `TargetEnumerationUtils.findValidSpellTargets` fixed; `EntitySnapshot.projectedTypeLine` shared with `ZoneTransitionService`; `ActivateAbilityHandler` freezes the source's type line/keywords/token-ness |
| `mtg-sets` | `msh/cards/EchoPerceptiveProdigy.kt`, `msh/cards/ScientistSupremeOfAim.kt` |
| tests | `EchoPerceptiveProdigyScenarioTest`, `ScientistSupremeOfAimScenarioTest`, `AbilitySourceTargetingTest` (mechanic-level: the seam + the enumeration regression) |
| docs | `docs/card-sdk-language-reference.md` — predicate, chained builder, target facade, enumeration note |
| backlog | `cards.md` both ticked, header 239 → 241; `mechanics.md` section marked SHIPPED, 33 → 31 blocked |

## Verification

The branch was gated three times, on three different trees, because two rounds of review findings
landed as new commits after the first gate. Reading the table top-down gives the history; **the
bottom block is the one that describes the code being merged.**

**Tree 1 — `5df00a8467`, the original implementation (superseded).**

| Command | Result |
|---|---|
| `just test` | BUILD SUCCESSFUL in 15m 46s — 12200 tests PASSED, 0 FAILED (`build/pr/loop-msh-u08-gate.log`) |
| `just test-rules` (earlier iteration) | BUILD SUCCESSFUL in 15m 8s, 0 failed (`build/pr/loop-msh-u08-gate-rules.log`) |

This run does **not** cover the merged code: the round-3 correction (`daec416da0`) went on to rewrite
`PredicateEvaluator` (+161 lines, `matchesSnapshot`), `ActivateAbilityHandler`, `EntitySnapshot` and
`ZoneTransitionService`, and added two scenario tests.

**Tree 2 — `daec416da0`, after the round-3 corrections.**

| Command | Result |
|---|---|
| `just test-class ScientistSupremeOfAimScenarioTest` | **6/6 PASSED** (`build/pr/loop-msh-u08-gate-correction-targeted.log`) — includes both new last-known-information cases; the token test asserts `state.getEntity(clue) == null` before checking the offer, so it fails on the pre-fix code |
| `just test` | **0 test failures** across `mtg-sdk`, `mtg-sets`, `ai`, `game-server`, `gym`, `gym-server`, `gym-trainer`, `mtg-search`, `mtgish-tooling` — but the run ended with *"Gradle build daemon disappeared unexpectedly"* while `:rules-engine:test` was still in flight, at ~146 MiB available RAM (`build/pr/loop-msh-u08-gate-correction.log`). That is the documented shared-box OOM (`rules-engine` starves first), **not** a test failure. No `clean`, no retry loop — the two uncovered modules were re-run separately below |
| `:rules-engine:test` (reduced footprint) | **BUILD SUCCESSFUL in 12m 27s, 0 failures** (`build/pr/loop-msh-u08-gate-correction-rules.log`) — the module carrying every behavioural change |
| `:mtg-sets:test` (reduced footprint) | **BUILD SUCCESSFUL in 1m 12s, 0 failures** (`build/pr/loop-msh-u08-gate-correction-sets.log`) — the card-snapshot golden, so `MSH.json` is confirmed unmoved |

Together those three runs cover every module. Three comment-only edits (two KDoc blocks, one doc
paragraph) landed while the `just test` run was executing; they change no bytecode, and the two
reduced-footprint runs were started after them.

**Tree 3 — `c6d333a27e`, after the round-4 review — the code being merged.**

| Command | Result |
|---|---|
| `:rules-engine:test` (reduced footprint) | **BUILD SUCCESSFUL in 4m 34s — 10765 tests PASSED, 0 failures** (`build/pr/loop-msh-u08-gate-round4-rules.log`). All 15 of this unit's tests green, including both last-known-information cases and the enumeration-parity assertions |

Round 4 changed one line of behaviour-neutral plumbing in `ActivateAbilityHandler`
(`captureEntitySnapshots(ids, projectedState)` + a hand-rolled `wasToken`/`name` copy → the existing
`captureEntitySnapshots(ids, state)` overload, which fills exactly those two fields the same way),
plus three comment-only fixes. `manual-scenarios/…/loop-msh-u08-new-cards-castable.json` was also
added; nothing compiles or loads it under test.

`:rules-engine` is the only module round 4 touched, and the change is inside a function body — no
signature moved — so the downstream modules (`ai`, `game-server`, `gym*`) can't be affected at
compile or behaviour level and were not re-run; tree 2's runs still cover them.

**Everything below is unaffected by which tree it ran on.**

| Command | Result |
|---|---|
| `just rebless-cards` | exit 0. `git diff --stat` on the snapshots: **only** `mtg-sets/src/test/resources/snapshots/cards/MSH.json`, +150 / −0 — the two new cards and nothing else moved |
| `just check-card-printing "Echo, Perceptive Prodigy"` | `ok: canonical is in the card's earliest real printing and all scaffolded reprints have rows.` (msh 51 canonical; pw26 promo not scaffolded) |
| `just check-card-printing "Scientist Supreme of A.I.M."` | `ok: canonical is in the card's earliest real printing and all scaffolded reprints have rows.` (msh 225 canonical) |
| `just check-backlog` | `all 14 card-count headers in sync` |

New tests, all green:

- `EchoPerceptiveProdigyScenarioTest` — vigilance; copy-and-retarget a creature-source ability; an
  artifact-source ability is *neither enumerated nor accepted*; the last-known-information case (kill
  the creature with Lightning Bolt, copy its dies trigger from the graveyard).
- `ScientistSupremeOfAimScenarioTest` — copy-and-retarget an artifact-source ability and pay 2 life; a
  creature-source ability rejected; once-per-turn; your-turn-only; and two last-known-information
  cases added in round 3 — a **nontoken** artifact that sacrificed itself to pay for its own ability
  (source in the graveyard), and a **Clue token** whose entity CR 704.5d has deleted outright.
- `AbilitySourceTargetingTest` — the `StackObjectTargeting.permitsAbilities` seam (including that
  "target spell" filters stay spell-only) and the enumeration regression, asserted on a plain
  `Targets.ActivatedOrTriggeredAbilityYouControl` card so it stands independent of this unit's cards.
- Pre-existing `TargetSpellExcludesAbilitiesTest` still passes both ways.

Every "is it offered?" assertion goes through `LegalActionEnumerator.enumerate` — the same
computation the server sends the client — not just `submitSuccess`.

No web-client change was made, so `npm test` was not run.

## Not done

- No manual playthrough in the web client, and no UX pass from either seat. The client analysis above
  is by code reading only.
- No e2e/Playwright run.
- No AI/gym self-play pass over the two cards.
- The token-source last-known-information gap **was** fixed in round 3 (the `lastKnownSourceSnapshot`
  fallback + a Clue-token scenario test). What is still documented-not-fixed is narrower: a deleted
  token source is recoverable only for an *activated* ability whose cost consumed it, not for a
  triggered ability; a nontoken source that left the battlefield some other way reads its printed
  types; and the snapshot arm answers only card types, sub/supertypes, keywords, token-ness and
  controller.

This PR came out of an autonomous loop: independent review is still pending — a reviewer agent will
comment its findings and a corrector agent may push follow-up commits, so this is not final at open
time.

### DIFF
--- a/LOOP_UNIT.md
+++ b/LOOP_UNIT.md
@@ -27,9 +27,26 @@
 
 ## Gate
 
-`just test` — BUILD SUCCESSFUL in 15m 46s, 12200 PASSED / 0 FAILED (`build/pr/loop-msh-u08-gate.log`).
-Plus `just rebless-cards` (only MSH.json moved, +150/−0), `just check-card-printing` ok for both
-cards, `just check-backlog` in sync.
+Three trees, because two review rounds landed as new commits after the first gate. **Only the last
+block describes the code as it stands.**
+
+1. **`5df00a8467`** (original, superseded) — `just test` BUILD SUCCESSFUL in 15m 46s, 12200 PASSED /
+   0 FAILED (`build/pr/loop-msh-u08-gate.log`). Does not cover the merged code: `daec416da0` went on
+   to rewrite `PredicateEvaluator`, `ActivateAbilityHandler`, `EntitySnapshot` and
+   `ZoneTransitionService` and add two scenario tests.
+2. **`daec416da0`** (after round-3 corrections) — `just test-class ScientistSupremeOfAimScenarioTest`
+   6/6 PASSED (`-gate-correction-targeted.log`); `just test` 0 failures across every module *except*
+   `:rules-engine`, where the run ended in the documented shared-box daemon OOM at ~146 MiB available
+   (`-gate-correction.log`, environmental, not a test failure); re-run at reduced footprint,
+   `:rules-engine:test` BUILD SUCCESSFUL 12m 27s / 0 failures (`-gate-correction-rules.log`) and
+   `:mtg-sets:test` BUILD SUCCESSFUL 1m 12s / 0 failures (`-gate-correction-sets.log`). The three
+   runs together cover every module.
+3. **`HEAD`** (after round-4 review) — one behaviour-neutral plumbing change in
+   `ActivateAbilityHandler` (switch to the existing state-aware `captureEntitySnapshots` overload)
+   plus comment-only fixes. **Re-gate pending** — see `build/pr/loop-msh-u08-round4-correction.md`.
+
+Unaffected by the tree: `just rebless-cards` (only MSH.json moved, +150/−0),
+`just check-card-printing` ok for both cards, `just check-backlog` in sync.
 
 ## Things I'm unsure about — please look
 
--- a/LOOP_UNIT.md
+++ b/LOOP_UNIT.md
@@ -41,9 +41,13 @@ block describes the code as it stands.**
    `:rules-engine:test` BUILD SUCCESSFUL 12m 27s / 0 failures (`-gate-correction-rules.log`) and
    `:mtg-sets:test` BUILD SUCCESSFUL 1m 12s / 0 failures (`-gate-correction-sets.log`). The three
    runs together cover every module.
-3. **`HEAD`** (after round-4 review) — one behaviour-neutral plumbing change in
-   `ActivateAbilityHandler` (switch to the existing state-aware `captureEntitySnapshots` overload)
-   plus comment-only fixes. **Re-gate pending** — see `build/pr/loop-msh-u08-round4-correction.md`.
+3. **`c6d333a27e`** (after round-4 review, the code as it stands) — one behaviour-neutral plumbing
+   change in `ActivateAbilityHandler` (switch to the existing state-aware `captureEntitySnapshots`
+   overload) plus comment-only fixes. `:rules-engine:test` at reduced footprint **BUILD SUCCESSFUL
+   in 4m 34s, 10765 PASSED / 0 failures** (`build/pr/loop-msh-u08-gate-round4-rules.log`), all 15 of
+   this unit's tests green. Only `:rules-engine` was touched and no signature moved, so downstream
+   modules were not re-run — tree 2 covers them. Accounting in
+   `build/pr/loop-msh-u08-round4-correction.md`.
 
 Unaffected by the tree: `just rebless-cards` (only MSH.json moved, +150/−0),
 `just check-card-printing` ok for both cards, `just check-backlog` in sync.
--- a/LOOP_UNIT.md
+++ b/LOOP_UNIT.md
@@ -1,67 +1,51 @@
-# loop-msh-u07 — Ward with a non-listed cost
-
-Feature unit: two new `WardCost` variants plus the two MSH cards they unblock.
-
-## SDK
-
-- `WardCost.PlayerCounters(counterType, amount)` — counters placed on the **paying player**
-  (CR 122.1). Facade `KeywordAbility.wardPlayerCounters(Counters.POISON, 5)`. The only ward cost with
-  no affordability gate.
-- `WardCost.Choice(options)` — OR disjunction, sibling of `Composite`'s AND. Facades
-  `KeywordAbility.wardChoice(...)` and the named `wardDiscardOrPay("{2}")`. Modelled on
-  `AdditionalCost.Choice` / `PayCost.Choice`; the prompt follows `CostPaymentService.choicePrompt`
-  (payable options only + trailing decline, reduced list stored on the continuation).
-- `WardCost.clause` — new property: the self-contained verb phrase ("discard a card", "pay {2}") as
-  opposed to `description`'s object phrase. Only `Choice` reads it, so no existing card's rendered
-  text changes.
-
-## Engine
-
-- `WardCounterEffectExecutor`: two new branches; `canPayWardCost` is now the single source of truth
-  for "unpayable → counter without a prompt" and the four pre-existing inline can-pay checks were
-  folded into it.
-- Two continuations (`CounterUnlessPlayerCountersContinuation`, `WardCostChoiceContinuation`) in
-  `ManaContinuations.kt`, registered in `Serialization.kt`, resumed in
-  `ManaPaymentContinuationResumer`. The choice resumer routes the chosen option through the existing
-  `chargeNextWardPartOrNull`, so the spell-left-the-stack guard and composite chaining stay
-  single-sourced.
-- Counters are placed via the ordinary `AddCountersEffect` executor with the payer as controller —
-  replacement effects, `CountersAddedEvent` and the ten-poison SBA all follow for free.
+# loop-msh-u08 — Ability-source predicate on stack targets
 
 ## Cards
 
-- **The Serpent Society** [226] — deathtouch; `wardPlayerCounters(Counters.POISON, 5)`; OTHER-bound
-  dies trigger filtered by `withKeyword(DEATHTOUCH)` (matched against LKI) → `Effects.Sacrifice`
-  nontoken over `Player.EachOpponent`.
-- **Titania, Rugged Rumbler** [235] — `Costs.additional.DiscardOrPay("{2}")` +
-  `KeywordAbility.wardDiscardOrPay("{2}")`. Same printed shape on both rails, deliberately
-  matching facade names; the types stay separate because the additional cost's mana leg folds into
-  the spell's mana cost at cast time and the ward's does not.
+- **Echo, Perceptive Prodigy** (MSH 51) — vigilance + `{1}, {T}: Copy target activated or triggered
+  ability you control from a creature source.` Composes `Targets.ActivatedOrTriggeredAbilityYouControlFrom(Creature)`
+  (existing ability-on-stack target + the new `CardPredicate.AbilitySourceMatches`) with the existing
+  `Effects.CopyTargetSpellOrAbility`, plus `holdPriority`.
+- **Scientist Supreme of A.I.M.** (MSH 225) — `Pay 2 life: Copy target activated or triggered ability
+  you control from an artifact source. Activate only during your turn and only once each turn.`
+  Same pieces with `GameObjectFilter.Artifact`, `Costs.PayLife(2)` and
+  `ActivationRestriction.OnlyDuringYourTurn` + `OncePerTurn`.
 
-## Also
+## SDK / engine
 
-- `docs/card-sdk-language-reference.md` — ward section extended for both variants and the two-rail
-  note.
-- `backlog/sets/marvel-super-heroes/cards.md` — both ticked, count 239 → 241.
-- `backlog/sets/marvel-super-heroes/mechanics.md` — section rewritten as SHIPPED; blocked count
-  33 → 31.
-- `mtgish-tooling` `CardStructure.kt` — `wardKeywordLine` learned `_Cost: "Or"` via a new
-  `wardCostExpr` leg renderer; existing single-cost branches unchanged, no corpus ward uses `Or`, so
-  no golden moves. The IR has no player-counter cost tag, so `PlayerCounters` is not taught.
+- `CardPredicate.AbilitySourceMatches(subfilter)` — new stack-branch predicate, sibling of
+  `TargetsMatching`. Redirects the match onto the ability's `sourceId` (CR 113.7) and evaluates the
+  subfilter there. Builders: `GameObjectFilter.abilitySourceMatches`, `TargetFilter.abilitySourceMatches`,
+  `Targets.ActivatedOrTriggeredAbilityYouControlFrom`.
+- `StackObjectTargeting.permitsAbilities` — the "may this stack target requirement offer abilities?"
+  seam, extracted from `TargetFinder` so the enumerator can share it.
+- **Bug fixed, not part of the brief:** `TargetEnumerationUtils.findValidSpellTargets` filtered every
+  stack target down to spells unconditionally, so *no* ability-targeting card ever had an ability
+  **offered** as a legal target — Gogo, Master of Mimicry and Peter Parker's Camera included. Since
+  `ActivatedAbilityEnumerator` gates `holdPriority` on "top of stack ∈ validTargets", those cards
+  also never stopped auto-pass. Both readers now go through `StackObjectTargeting`.
 
-## Things worth a second opinion
+## Gate
 
-`WardPlayerCountersTest`, `WardCostChoiceTest` (engine-level, one per mechanic),
-`TheSerpentSocietyScenarioTest`, `TitaniaRuggedRumblerScenarioTest` (one per card).
+`just test` — BUILD SUCCESSFUL in 15m 46s, 12200 PASSED / 0 FAILED (`build/pr/loop-msh-u08-gate.log`).
+Plus `just rebless-cards` (only MSH.json moved, +150/−0), `just check-card-printing` ok for both
+cards, `just check-backlog` in sync.
 
-## Unsure / worth a reviewer's eye
+## Things I'm unsure about — please look
 
-- `WardCost.clause` duplicates information with `description`. The alternative — making `description`
-  the full clause everywhere and dropping the per-renderer verb prefixes — is cleaner but rewrites
-  the rendered text of every existing ward card and moves a lot of snapshots; I chose not to.
-- `canPayWardCost` for `Composite` is a snapshot check (all parts payable *now*). Paying an earlier
-  part could in principle make a later one unpayable; the per-part handler still counters at that
-  point, so behaviour is right, but a `Choice` over a `Composite` could offer a leg that later fails.
-  No printed card has that shape.
-- The ward `Choice` picker labels are generated from `clause` ("Discard a card" / "Pay {2}" /
-  "Counter spell"). Not seen in a running client.
+1. **Token sources are unmatchable once gone.** CR 704.5s deletes a token's entity, so a sacrificed
+   Clue's "Sacrifice this artifact: draw a card" is *not* "from an artifact source" for Scientist
+   Supreme. Non-token sources are fine (the entity survives in the graveyard with its printed types).
+   Closing it needs a general last-known store for deleted tokens, which the engine doesn't have —
+   `EntitySnapshot`/`LastKnownPermanentComponent` both die with the entity. Documented in the
+   predicate KDoc and the DSL reference rather than silently approximated.
+2. Related, smaller: a source whose *type* came from a continuous effect (animated land, crewed
+   Vehicle) and has since left reads its printed types.
+3. `ChosenTarget.Spell` gets no CR 608.2b filter re-check in `StackResolver.validateTargets` (only
+   "still on the stack"). Pre-existing, not touched here; it means the source restriction is enforced
+   at targeting time only. Arguably correct in practice (the ability can't stop being from a creature
+   source), but worth a second opinion.
+4. No client change was needed: `StackZone` already routes both the action-pipeline
+   (`targetingState`) and decision (`decisionSelectionState`) click paths for stack objects, and
+   `TargetingOverlay` only diverts to the pile picker for `Graveyard`/`Exile`. Verified by reading;
+   not exercised in a browser.
--- a/LOOP_UNIT.md
+++ b/LOOP_UNIT.md
@@ -33,18 +33,29 @@ cards, `just check-backlog` in sync.
 
 ## Things I'm unsure about — please look
 
-1. **Token sources are unmatchable once gone.** CR 704.5s deletes a token's entity, so a sacrificed
-   Clue's "Sacrifice this artifact: draw a card" is *not* "from an artifact source" for Scientist
-   Supreme. Non-token sources are fine (the entity survives in the graveyard with its printed types).
-   Closing it needs a general last-known store for deleted tokens, which the engine doesn't have —
-   `EntitySnapshot`/`LastKnownPermanentComponent` both die with the entity. Documented in the
-   predicate KDoc and the DSL reference rather than silently approximated.
-2. Related, smaller: a source whose *type* came from a continuous effect (animated land, crewed
-   Vehicle) and has since left reads its printed types.
+1. ~~**Token sources are unmatchable once gone.**~~ **Closed in review round 3.** The claim that the
+   engine had no last-known store for a deleted token was wrong:
+   `ActivatedAbilityOnStackComponent.lastKnownSourceSnapshot` already freezes an `EntitySnapshot` of
+   the source whenever the activation cost sacrifices/exiles it — exactly the Clue / Food / Blood /
+   Treasure shape. It just wasn't carrying the type line. It now does (via the shared
+   `projectedTypeLine` helper), and the `AbilitySourceMatches` branch falls back to it through
+   `PredicateEvaluator.matchesSnapshot` when `state.getEntity(sourceId)` is null. A cracked Clue's
+   draw ability is a legal, *offered* target for Scientist Supreme; scenario test added. Timing
+   nuance found while testing: SBAs run on the engine's post-resolution pass, so the token is still
+   readable the instant its ability hits the stack and only vanishes once something else resolves —
+   the test reproduces that sequence and asserts the entity is gone before checking the offer. The
+   rule is
+   **CR 704.5d** ("If a token is in a zone other than the battlefield, it ceases to exist"), not
+   704.5s (that one is the Saga final-chapter sacrifice).
+2. Related, smaller, still open: a **nontoken** source whose *type* came from a continuous effect
+   (animated land, crewed Vehicle) and has since left the battlefield some other way reads its
+   printed types — only the self-sacrifice/self-exile cost path freezes the projected type line.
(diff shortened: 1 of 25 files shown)


bolav:
# Review round 4 — `loop-msh-u08` (Echo, Perceptive Prodigy · Scientist Supreme of A.I.M.)

**Verdict: no blocking code findings. The SDK shape is right, the round-3 corrections hold up under
re-verification, and the behavior is correct for everything the tests cover. One Important finding is
a reporting problem, not a code one: the PR body advertised a gate that was run against the
pre-correction tree.**

Reviewed **in place** on `loop-msh-u08` (clean tree). `git rev-list --count HEAD..origin/main` = 0 and
`origin/main`'s tip is `7851afaf42` from the same day, so `main` is already contained — no merge, no
worktree, no merge commit produced. `build/pr/loop-msh-u08-review.md` (round 3) and
`-correction.md` were read first so this doesn't re-litigate settled ground.

All findings below were fixed in `c6d333a27e`; the accounting is in
`build/pr/loop-msh-u08-round4-correction.md`.

---

## What's good — keep this

- **The predicate shape survives a second look.** `CardPredicate.AbilitySourceMatches(subfilter:
  GameObjectFilter)` is parameterized by an arbitrary filter, not a source-*type* enum, so "creature
  source" and "artifact source" are one primitive with two arguments and the next card ("from an
  enchantment source", "from a source you control") costs nothing. No new `Effect`, no new
  `TargetRequirement`, no executor.
- **`matchesSnapshot` mirrors `matches` field-for-field, and I checked that claim rather than taking
  it.** `GameObjectFilter` has exactly four fields (`cardPredicates`, `statePredicates`,
  `controllerPredicate`, `anyOf`) and `matchesSnapshot` (`PredicateEvaluator.kt:319`) handles all four
  with the same conjunction/union semantics as `matches` (`:88`). Nothing is silently ignored, so the
  "partial by construction" framing is accurate rather than aspirational — a filter field it can't
  answer makes the match fail, it doesn't fall through to true.
- **The tri-state `Boolean?` is the right call and the `Not` reasoning is sound.**
  `matchesSnapshotPredicate` returning `null` for unanswerable predicates, with `And`/`Or`
  propagating unknown correctly, is what stops `Not(unsupported)` from becoming a free match.
- **`projectedTypeLine` extraction is a genuine dedupe.** `ZoneTransitionService.buildProjectedTypeLine`'s
  old body and the new `EntitySnapshot.kt:792` helper are identical line for line, so the delegation
  is behavior-preserving and the two capture sites can no longer drift.
- **`StackObjectTargeting` is the right seam** and `AbilitySourceTargetingTest` locks both halves:
  `SpellOnStack`/`CreatureSpellOnStack` stay spell-only, ability-naming filters don't, and enumeration
  parity is asserted on a plain `Targets.ActivatedOrTriggeredAbilityYouControl` card — i.e. the
  pre-existing Gogo / Peter Parker's Camera bug is covered independently of the new predicate.
- **The token test proves the snapshot arm, not a lucky live read.**
  `ScientistSupremeOfAimScenarioTest` asserts `driver.state.getEntity(clue) shouldBe null` *before*
  checking the offer, and reproduces the real sequence (a second ability resolves so the SBA pass
  sweeps the token). That test fails on the pre-fix code, which is the property that matters.
- **Card fidelity is exact.** Both `cardDef`s verified against Scryfall live: name, mana cost, type
  line, P/T, rarity, collector number, artist, flavor text and full oracle text (including the
  reminder) match character-for-character. `just check-card-printing` exits 0 for both — canonical in
  `msh`, the card's earliest real printing.
- **Every CR number in the diff verified** against `/workspace/MagicCompRules_20260619.txt`: 112.1,
  113.3b, 113.3c, 113.7, 113.7a, 117.5, 608.2b, 608.2h, 704.5d, 707.10b, 707.10c — all correct,
  including the round-3 `704.5s → 704.5d` fix.
- The playtest scenario's six card names all resolve to real definitions, and Lifespark Spellbomb
  genuinely sacrifices itself, so it exercises the LKI path as the commit message claims.

---

## Important

### I1 — The PR body's gate table describes a tree that no longer exists

`build/pr/loop-msh-u08-body.md:170-171` and `LOOP_UNIT.md:30`.

Both reported only `just test` — *BUILD SUCCESSFUL in 15m 46s, 12200 PASSED* — from round 1, against
`5df00a8467`. The round-3 correction then rewrote `PredicateEvaluator` (+161 lines),
`ActivateAbilityHandler`, `EntitySnapshot` and `ZoneTransitionService` and added two scenario tests.
`build/pr/loop-msh-u08-correction.md:136-162` records what actually happened on the corrected tree,
and it is not a clean single run: `just test` reached 0 failures across `mtg-sdk`, `mtg-sets`, `ai`,
`game-server`, `gym*`, `mtg-search` and `mtgish-tooling` but died with *"Gradle build daemon
disappeared unexpectedly"* mid-`:rules-engine:test` at ~146 MiB available, and coverage was completed
by a separate `:rules-engine:test` (12m 27s, 0 failures) and `:mtg-sets:test` (1m 12s, 0 failures)
with the reduced-footprint flags.

Coverage is in fact complete — this is purely a reporting gap, and it's the exact failure mode
`CLAUDE.local.md` warns about ("re-run `just test` on the new base rather than repeating the old
green result in the PR body"). Fix: replace the gate table with the three round-3 runs, state the
daemon-OOM as environmental, and note that the one commit since (`9c909cbe35`) adds only
`manual-scenarios/sets/msh/loop-msh-u08-new-cards-castable.json`, which nothing compiles or loads
under test.

---

## Minor

### m2 — `CR 112.7a` doesn't exist; it's the pre-2021 number for `113.7a`

`ActivateAbilityHandler.kt:1236`, on a line this diff rewrote. Rule 112 ends at **112.4** in
`MagicCompRules_20260619.txt` (and in the 20260417 copy); the text the comment means — *"Once
activated or triggered, an ability exists on the stack independently of its source"* — is **113.7a**,
which the diff already cites correctly everywhere else. There are 63 occurrences repo-wide, so this
is a stale-numbering convention rather than something this unit introduced; sweeping them is a
follow-up, but the one line the diff re-touched is fair to fix now.

### m3 — The snapshot `.copy(...)` re-implements an overload that already exists

`ActivateAbilityHandler.kt:1249-1268`. `captureEntitySnapshots(ids, state)` (`EntitySnapshot.kt:230`)
already fills `wasToken` (via `TokenComponent`) and `name` (via `CardComponent`) — exactly two of the
five fields hand-rolled here, using the same lookups. Switching the call to
`captureEntitySnapshots(listOf(action.sourceId), currentState)` resolves through the same import,
leaves only `typeLine`, `keywords` and `cardDefinitionId` in the `.copy(...)`, and collapses three
separate `currentState.getEntity(action.sourceId)` calls into one.

### m4 — `matchesSnapshot` is public, and its tri-state discipline has two holes for defaulted fields

`PredicateEvaluator.kt:319`, `:358-363`. `EntitySnapshot.wasToken` and `.keywords` default to
`false`/`emptySet()` with no "not captured" state, so `IsNontoken` and `HasKeyword` answer a
confident `false`/`true` for a snapshot that never recorded them — the one case where the careful
"unanswerable → `null`" contract quietly doesn't hold. Harmless today because the sole caller
populates both (and m3's overload would make that structural), but `matchesSnapshot` is `public` and
the next caller won't know. Either state the invariant on the KDoc ("caller must pass a snapshot
captured with the state-aware overload") or make it `internal`.

### m5 — One `704.5s` survives in a file this diff edits

`EntitySnapshot.kt:99` — *"so dies/leaves triggers resolve for tokens after 704.5s cleanup"*.
Pre-existing and outside the changed hunks, but it's in a file the diff touches, and round 3 already
established the correct number is `704.5d`. Cheap to sweep with the other four the previous review
listed.

---

## Recommendation

Merge-ready on the code. Before opening the PR: rewrite the gate table (**I1**) so it describes the
corrected tree and the split run honestly, and fix `112.7a → 113.7a` (**m2**) while you're in that
comment. **m3–m5** are the corrector's discretion — m3 is the one worth taking, since it also turns
m4 from a documentation promise into a structural guarantee.

---

### Note on this round's verification

The gate was **not** re-run during the review pass: `free -h` showed 1.8 GiB available with five JVMs
already resident (~74% of RAM), i.e. a sibling container's build in flight, and `CLAUDE.local.md` is
explicit that adding a competing JVM there is how the box gets OOM-killed. Everything above is
source-level verification plus the runs recorded in `build/pr/`. The round-4 fix commit does touch
`rules-engine` source, so it carries its own `:rules-engine:test` — see the correction file.

bolav:
# Round-4 correction — `loop-msh-u08` (Echo, Perceptive Prodigy · Scientist Supreme of A.I.M.)

Accounting against the round-4 `review-changes` pass (1 Important, 4 Minor; no blocking code
findings). **All 5 fixed.** Changes are a new commit on top of `9c909cbe35`; no history rewritten.

---

## Important

### I1 — the PR body advertised a gate run against a tree that no longer exists — **FIXED**

`build/pr/loop-msh-u08-body.md` and `LOOP_UNIT.md` both reported only the round-1 `just test`
(15m 46s, 12200 PASSED) against `5df00a8467` — before the round-3 correction rewrote
`PredicateEvaluator` (+161 lines), `ActivateAbilityHandler`, `EntitySnapshot` and
`ZoneTransitionService` and added two scenario tests.

Both files now carry a three-tree gate history instead of one line, with the superseded run marked
as such:

1. `5df00a8467` — the original `just test`, explicitly labelled as *not* covering the merged code;
2. `daec416da0` — the round-3 runs, including the fact that `just test` reached 0 failures on every
   module except `:rules-engine`, where it ended in the documented shared-box daemon OOM at ~146 MiB
   available, and the two reduced-footprint re-runs (`:rules-engine:test` 12m 27s / 0 failures,
   `:mtg-sets:test` 1m 12s / 0 failures) that closed the gap;
3. `HEAD` — this round.

Two stale claims in the same file were corrected while I was in there: the "Not done" bullet still
said *"the token-source last-known-information gap is documented, not fixed"* (round 3 fixed it —
the bullet now names the three narrower limits that genuinely remain), and the
`ScientistSupremeOfAimScenarioTest` bullet didn't mention the two last-known-information cases
round 3 added.

## Minor

### m2 — `CR 112.7a` is not a rule — **FIXED**

`ActivateAbilityHandler.kt:1236` → `CR 113.7a`. Verified against
`/workspace/MagicCompRules_20260619.txt`: rule 112 ends at **112.4** ("Spells"), and the text the
comment means — *"Once activated or triggered, an ability exists on the stack independently of its
source"* — is **113.7a**. `112.7a` is that rule's pre-2021 number.

Only the line this diff already owned was changed. The other 62 repo-wide occurrences (including
`ActivateAbilityHandler.kt:1268`, three lines below) are pre-existing and left alone per
AGENTS.md's "focus on your own work" — **worth a follow-up sweep**, together with the four stale
`704.5s` citations the round-3 review listed (`TokensInWrongZonesCheck.kt`,
`DeathAndLeaveTriggerDetector.kt`, `TriggerDetector.kt`).

### m3 — the snapshot `.copy(...)` re-implemented an existing overload — **FIXED**

`ActivateAbilityHandler.kt:1254`. `captureEntitySnapshots(ids, state)`
(`EntitySnapshot.kt:230`) already fills `wasToken` (via `TokenComponent`) and `name` (via
`CardComponent`) — exactly two of the five fields the round-3 code hand-rolled, from the same
lookups. Switched the call from the `(ids, projectedState)` overload to it (same import, no new
one), which drops those two fields from the `.copy(...)` and collapses three separate
`currentState.getEntity(action.sourceId)` calls into one. `typeLine`, `keywords` and
`cardDefinitionId` still layer on top.

Behaviour-neutral by construction: the `(ids, state)` overload delegates to
`captureEntitySnapshots(ids, state.projectedState)` — the exact call that was there — and then
copies in `wasToken`/`name` with the same expressions that were being written inline.

### m4 — `matchesSnapshot`'s tri-state discipline has two holes for defaulted fields — **FIXED (documented)**

`PredicateEvaluator.kt:149`. `EntitySnapshot.wasToken` and `.keywords` default to `false` /
empty with no "was never captured" state, so a snapshot taken by the projection-only
`captureEntitySnapshots` overload answers `IsToken` false, `IsNontoken` true and every `HasKeyword`
false *with full confidence* rather than reporting unknown — the one place the "unanswerable →
`null`" contract can't hold. The KDoc now states the caller invariant explicitly (pass a snapshot
from the state-aware overload, then layer `typeLine`/`keywords` on it), names why the null arm
can't cover those two fields, and records that there is exactly one caller today which does it
correctly.

Kept `public` rather than narrowed to `internal`: `PredicateEvaluator` is a public engine surface
and `matchesSnapshot` is the natural companion to the public `matches`. m3 makes the invariant
structural at the only call site, so the doc is a guard for the *next* caller.

### m5 — a stale `704.5s` in a file this diff edits — **FIXED**

`EntitySnapshot.kt:99` → `704.5d`. Same verification as m2: `704.5d` is *"If a token is in a zone
other than the battlefield, it ceases to exist"*; `704.5s` is the Saga final-chapter sacrifice.

---

## Gate

`:rules-engine:test` at reduced footprint (`-Pkotlin.compiler.execution.strategy=in-process`,
`-Dorg.gradle.jvmargs="-Xmx3g -XX:MaxMetaspaceSize=768m"`) —
**BUILD SUCCESSFUL in 4m 34s, 10765 tests PASSED, 0 failures**
(`build/pr/loop-msh-u08-gate-round4-rules.log`). The five `grep -i failed` hits in that log are test
*names* containing the word ("nothing to return is not a failure", "DrawFailedEvent is emitted…",
`FailedFordingScenarioTest`, "no infinite failure"); all five report PASSED.

All 15 of this unit's tests green:

- `EchoPerceptiveProdigyScenarioTest` 4/4 — including *"a dead creature's dies trigger is still a
  creature source"*;
- `ScientistSupremeOfAimScenarioTest` 6/6 — including both last-known-information cases, the
  nontoken self-sacrifice and *"a Clue token's draw ability is still an artifact source after
  CR 704.5d deletes the token"*;
- `AbilitySourceTargetingTest` 5/5 — the `permitsAbilities` seam and the enumeration-parity
  assertion.

**Scope of the run.** `:rules-engine` is the only module round 4 touched, and the m3 change is
inside a function body with no signature moved, so `ai` / `game-server` / `gym*` cannot be affected
at compile or behaviour level; they were not re-run and tree 2's runs still cover them. `mtg-sets`
likewise — nothing in this round changes a printed characteristic, so `MSH.json` is untouched.

**On the environment.** The round-4 gate was initially reported as blocked behind "a sibling
container's build", inferred from `free -h` showing 1.8 GiB available. That was wrong: the five
resident JVMs were **idle** daemons (load average 0.09, every one at ~0% CPU) holding 8.8 of 9.6 GiB
— a Gradle daemon lives ~3 h past its last build and pins its whole `-Xmx`. `./gradlew --stop`
reclaimed 10 GiB instantly and the gate then ran in 4m 34s rather than the ~12 min the round-3 runs
took under real memory pressure. `CLAUDE.local.md` has been updated with the diagnosis rule (check
CPU/load, not `free -h`; stop idle daemons rather than waiting on a threshold that never clears) and
with the fact that `scripts/gradle-locked`'s lock is a no-op on this Linux box.

PR #1797, opened 2026-08-12
Add improvise (CR 702.126) on a shared tap-for-generic payment rail

bolav:
# Improvise (CR 702.126) on a shared tap-for-generic payment rail

Adds the **improvise** keyword and the two MSH cards it unblocks. The interesting part is not the
keyword — it is that improvise and waterbend turned out to be the same mechanism, so they were
converged onto one rail instead of growing a fourth parallel payment field.

## Cards

- **Ironheart, Clever Champion** [msh 60] — `{4}{U}` Legendary Artifact Creature — Human Hero 3/4.
  `keywords(Keyword.IMPROVISE, Keyword.FLYING)` plus
  `GrantKeywordToOwnSpells(Keyword.IMPROVISE, GameObjectFilter.Noncreature)`. The grant needed no new
  primitive — `GrantedKeywordResolver` already handles cost keywords.
- **Arc Reactor** [msh 243] — `{5}` Artifact. `keywords(Keyword.IMPROVISE)`, an `EntersTapped()`
  replacement effect, and `Effects.AddColorlessMana(3)` behind `Costs.Tap`.

Oracle text for both was checked against Scryfall; `just check-card-printing` is clean for both
(MSH is the earliest real printing).

## The design call

The backlog triage proposed a **fourth** parallel payment field (`improvisedArtifacts`) alongside
delve / convoke / harmonize / waterbend. That is not what shipped. Improvise and waterbend are the
same mechanism with different eligibility, so:

- `AlternativePaymentChoice.waterbendPermanents` became **`tapForGenericPermanents`** — one carrier
  for "tap permanents you control, each paying `{1}` generic".
- The eligibility rule became a value: **`TapForGeneric.IMPROVISE`** (artifacts) /
  **`TapForGeneric.WATERBEND`** (artifacts or creatures), consumed by one
  `AlternativePaymentHandler.applyTapForGeneric` and one
  `CostEnumerationUtils.findTapForGenericPermanents` / `canAffordWithTapForGeneric`.
- `LegalAction`, the DTO and the client renamed to match, plus a `tapForGenericLabel` so the single
  HUD (`TapForGenericSelector`, `tapForGeneric` pipeline phase) names the mechanic being paid.

**Adding a further keyword of this shape is now one enum entry** — not a new payment field, handler
branch, or client UI.

`AlternativePaymentHandler.supportsAlternativePayment` (instance + companion) was deleted rather
than extended: both listed only `DELVE || CONVOKE`, so both were already wrong for
harmonize/waterbend, and neither had a caller in any module.

## Rules

Verified against the Comprehensive Rules, not from memory:

- **CR 702.126a** — "For each generic mana in this spell's total cost, you may tap an untapped
  artifact you control rather than pay that mana." Generic only; a tap can never pay a colored pip.
- **CR 702.126b** — neither an additional nor an alternative cost, and applies only *after* the
  total cost is determined. So improvise never changes mana value and has no cap beyond that
  generic.
- **CR 702.126c** — multiple instances on the same spell are redundant.

## Improvise alongside another alternative payment

The grant is by card *type* ("Noncreature spells you cast have improvise"), and delve spells are
noncreature almost to a card, so improvise routinely lands on spells that print delve or convoke —
Treasure Cruise, Dig Through Time, Murderous Cut, Temporal Cleansing and 17 more are implemented
today. `CastSpellHandler` applies delve/convoke first and improvise second, and enumeration now
agrees: `canAffordWithDelve` / `canAffordWithConvoke` take an optional `tapForGenericPermanents`
pool, ORed in as an extra disjunct.

The disjunct matters rather than folding the pool into the existing calls: an artifact that taps
for more than `{1}` (Arc Reactor) is worth more as a mana source than as an improvise tap, so the
no-taps configuration has to stay reachable on its own. Convoke also drops artifact creatures
already in its own pool, matching `applyTapForGeneric` skipping whatever convoke tapped.

This was caught in review — before the fix, the affordability check was an exclusive `else if`
chain, so the convoke/delve arm won and improvise was never counted. Because an unaffordable cast
is *dropped* from the legal actions rather than greyed out, the cast simply disappeared: a legal
play the engine would have accepted could not be reached from the client.

## Two known scope limits

Both are documented in `docs/card-sdk-language-reference.md` and in-code, with what closing them
costs:

1. **`{X}` generic is not credited.** Per CR 601.2b/601.2f, X is locked in before the total cost is
   determined, so CR 702.126a's "generic mana in this spell's total cost" *includes* the X-derived
   generic (the Whir of Invention ruling). The engine credits taps only against the printed
   generic, and `maxAffordableX` deliberately ignores improvise to stay on the under-offering side.
   Four printed cards need it — Whir of Invention, Universal Surveillance, Saheeli's Directive,
   Battle at the Bridge — and none is implemented, so this lands with the first of them.
2. **`CastFromZoneEnumerator` doesn't offer improvise** on graveyard/exile casts. Closing it is the
   same metadata post-pass `CastSpellEnumerator.applyImproviseMetadata` already runs.

## AI

`Strategist.withAutomaticTapForGeneric` fills the payment only when the taps are *needed*
(`LegalAction.tapForGenericRequired`), and only with artifacts. Filling an optional improvise
payment can lose the cast: a tapped artifact stops being a mana source but credits only `{1}`, so
auto-tapping Arc Reactor trades three mana for one and can make the AI's own action unpayable.

## Tests

- `ImproviseScenarioTest` — the mechanic: generic-only vs the colored pip, the tap cap, CR 702.126c
  redundancy via a *granted* second instance, eligibility, and improvise combining with delve.
- `IronheartCleverChampionScenarioTest` — printed improvise, the grant reaching noncreature spells,
  the grant *not* reaching creature spells, and scoping to spells you cast.
- `ArcReactorScenarioTest` — full improvise of `{5}`, enters-tapped, and the mana ability.
- `ImprovisePaymentAiTest` — both directions of the auto-payer decision.
- `pipelinePhases.test.ts` — the tap phase for improvise and waterbend under auto-tap.
- Playtest scenario at `manual-scenarios/sets/msh/loop-msh-u09-new-cards-castable.json`.

## Gate

`just test -x :mtgish-tooling:test` — **BUILD SUCCESSFUL, zero failures.** `tsc --noEmit` clean;
`vitest pipelinePhases.test.ts` green.

`:mtgish-tooling:test > EmitterGoldenTest > HOB` is excluded because it **fails on clean
`origin/main`** — `hob.emitted.golden.txt` predates `Keyword.STORIED` landing and was never
re-blessed. Verified by checking out `origin/main` in a scratch worktree and reproducing it there.
Unrelated to this branch and deliberately not fixed here.

MSH is now at **249 / 276** implemented; the blocked-mechanic count drops to 23.


bolav:
# Review — `loop-msh-u09` (improvise)

Reviewed against `loop-msh-u08` as base. All findings below were fixed on the branch before it was
opened; the resolution is recorded under each.

## Verdict

Behavior is correct and the SDK shape is the highlight of this unit: the brief proposed a fourth
parallel payment field and the author converged waterbend and improvise onto one rail instead.
That is the right call and it is executed consistently across all seven layers. One reachable
enumeration gap, confirmed with a repro, plus three minor cleanups.

## What's good

- **The convergence.** `AlternativePaymentChoice.waterbendPermanents` → `tapForGenericPermanents`
  plus a `TapForGeneric` eligibility enum turns "add another tap-for-generic keyword" into one enum
  entry, instead of a field + handler branch + DTO + HUD. This is exactly the "composition before
  new types" bar, and the enum carrying its own `label` is what lets one HUD serve both mechanics
  without the client knowing either name.
- **Rule citations hold up.** Checked CR 702.126a/b/c against the Comprehensive Rules — the KDoc
  quotes are verbatim, including the "isn't an additional or alternative cost" clause that
  justifies the no-cap design. CR 601.2b/601.2f in the `{X}` TODO are also correctly attributed.
- **Oracle text matches Scryfall exactly** for both cards (cost, type line, P/T, all three lines),
  and `just check-card-printing` is clean for both.
- **Every cited rule has a paired test.** CR 702.126a gets both the generic-only case and the
  colored-pip refusal; the cap gets "taps beyond the generic stay untapped"; CR 702.126c gets a
  *granted* second instance on a printed-improvise spell.
- **`tapForGenericRequired` is real design, not defensive padding.** The Arc Reactor self-defeat —
  an auto-payer taps a `{T}: Add {C}{C}{C}` rock to save `{1}` and makes its own cast unpayable —
  is a genuine trap, and `ImprovisePaymentAiTest` pins both directions.
- **The two known gaps (`{X}`, and `CastFromZoneEnumerator`) are documented** in the SDK reference
  and in-code, with the reason they are deferred and what closing them costs. That is the right way
  to leave a gap.

## Findings

### Important — improvise silently dropped when the spell also has convoke or delve

The affordability check in `CastSpellEnumerator` was an exclusive `else if` chain, so `hasConvoke`
or `hasDelve` won and the improvise disjunct was never evaluated. Because an unaffordable cast is
`continue`d past rather than emitted greyed-out, the action disappeared entirely — not shown as
unaffordable, simply absent.

Reachable in this repo today, and it is Ironheart that makes it so: "Noncreature spells you cast
have improvise" reaches all 20 implemented noncreature convoke/delve spells (Treasure Cruise, Dig
Through Time, Murderous Cut, Empty the Pits, Temporal Cleansing, Collective Inferno, …).

Confirmed with a throwaway scenario test — a `{6}{U}` delve sorcery, 2 graveyard cards, 4
artifacts, one Island, and a beacon granting improvise:

```
> Ironheart-granted improvise combines with delve on the same spell FAILED
    delve 2 + improvise 4 + one Island covers {6}{U}
    <null> should not equal <null>
> the handler itself accepts the combined delve + improvise payment PASSED
```

The second case is the diagnosis: `CastSpellHandler` applies delve and then improvise in sequence
and accepts the payment happily. Only enumeration refused to offer it, so the player could not make
a legal play.

**Fixed.** `canAffordWithDelve` / `canAffordWithConvoke` take an optional `tapForGenericPermanents`
pool, ORed in at the call site as an extra disjunct rather than folded into the existing calls —
an artifact that taps for more than `{1}` is worth more as a mana source than as an improvise tap,
so the no-taps configuration has to stay reachable on its own. Convoke drops artifact creatures
already in its own pool, matching `applyTapForGeneric` skipping what convoke tapped. Two regression
tests added to `ImproviseScenarioTest`: the delve+improvise case, and a guard that a delve spell
payable on mana alone stays payable.

### Minor — the waterbend HUD changed behavior with no test covering it

`pipelinePhases.ts` dropped `waterbend` from `hasAlternativePaymentPhase`, so waterbend casts no
longer force the `manaSource` confirm step under auto-tap. The justification comment argued only
the improvise case, and the new test only exercised `tapForGenericLabel: 'improvise'`.

**Fixed.** Added a waterbend-labelled case asserting the intended phase list, so the change to the
older mechanic is pinned rather than incidental.

### Minor — tombstone comment for `supportsAlternativePayment`

Deleting the stale unused predicate is right; the six-line comment explaining what was removed and
why is exactly the `// removed` marker AGENTS.md §"No backwards-compat hacks" rules out. `git blame`
and the commit message carry that reasoning.

**Fixed.** Comment removed.

### Minor — `tapForGenericRequired` is deliberately engine-only but didn't say so

It is on `LegalAction`, read by `Strategist`, and correctly absent from `LegalActionInfo` since no
client needs it. The KDoc said "Exists for automatic payers", which reads like an oversight to
anyone auditing the enricher for unmapped fields.

**Fixed.** The KDoc now states it is engine-side by design and names its one consumer.

## Verification after the fixes

`just test -x :mtgish-tooling:test` — BUILD SUCCESSFUL, zero `FAILED` lines. `tsc --noEmit` clean,
`vitest pipelinePhases.test.ts` 7/7.

`:mtgish-tooling:test > EmitterGoldenTest > HOB` is excluded because it fails on clean
`origin/main` (`hob.emitted.golden.txt` predates `Keyword.STORIED`), verified by reproducing it in
a scratch worktree at `origin/main`. Unrelated to this branch and left alone.

PR #1798, opened 2026-08-12
u10: Mjölnir, Hammer of Thor + "Equip [quality]" (CR 702.6c)

bolav:
Depends on https://github.com/wingedsheep/argentum-engine/pull/1797

# u10 — Mjölnir, Hammer of Thor + "Equip [quality]" (CR 702.6c)

**Base is `loop-msh-u09`, not `main`** — u09 lands first. The diff below is the u10 delta only
(`git diff loop-msh-u09...HEAD`): 32 files, +1577 / −296.

Implements **Mjölnir, Hammer of Thor** (MSH #146), the set's only "Equip [quality]" card, and in
doing so converges every other card in the catalog that was hand-rolling that shape.

```
{3}{R} · Legendary Artifact — Equipment · Mythic
When Mjölnir enters, it deals 4 damage to up to one target creature.
Double all damage equipped creature would deal.
Equip worthy {1} (A creature is worthy if it's a legendary non-Villain that's red and/or white.)
{2}{R}, Discard this card: It deals 2 damage to each creature.
```

## SDK surface — three additions, all small

**1. `equipAbility(cost, genericCostReduction, quality, targetFilter)`** (`mtg-sdk/.../dsl/CardBuilder.kt`)

CR 702.6c: an equip ability may further restrict its targets, and "may legally target only a
creature that's controlled by the player activating the ability and that has the chosen quality."

No new target machinery. The restriction is an ordinary `TargetFilter` on an ordinary
`TargetCreature`, the same rail `auraTarget` and every other filtered target already ride —
`TargetEnumerationUtils.findValidTargets` and `StackResolver.validateTargets` pick it up with zero
new code. `quality` is wording; `targetFilter` is the rules half and must stay controller-scoped.

**2. `ActivatedAbility.equipQuality` + printed-line rendering in `describeWithCost`**

An equip ability is *printed* as `Equip {3}` (CR 702.6a) or `Equip Human {1}` (CR 702.6c), never as
the attach effect's generated `{1}: Attach this equipment to Human creature you control`. It now
renders that way — and because the line is **derived** rather than a per-card `descriptionOverride`,
the legal-action enumerator re-renders it against the *effective* cost, so Éowyn's discount and Forge
Anew's free first equip show the `{0}`/`{2}` the player actually pays. A static override structurally
cannot do that. Non-mana equip costs take the printed em dash: `Equip—Pay 3 life`.

> ⚠️ **This changes the menu text for every Equipment in the catalog**, not only the cards this PR
> touches. That is deliberate — it's how the ability is printed — but it is the widest user-visible
> change here and deserves a look. Nothing in the repo asserted the old generated string except one
> `ForgeAnewScenarioTest` line, updated in this PR to the stronger `description shouldBe "Equip {0}"`.

**3. `SourceFilter.EquippedCreature`** (`mtg-sdk/.../scripting/events/EventFilters.kt`)

The source-side mirror of `RecipientFilter.EquippedCreature`, which already existed.
`DamageUtils.damageSourceMatches` handles it in a shared branch with `EnchantedCreature`, exactly as
`damageRecipientMatches` already did for the `RecipientFilter` pair. That matcher is shared by
prevention, doubling and damage modification, and by both the combat and noncombat paths — so
"Double all damage equipped creature would deal" needed no further plumbing.

"Worthy" itself is **not** an SDK concept. Scryfall returns exactly one printed card using the term,
so it is spelled out on Mjölnir from existing predicates
(`.legendary().notSubtype(Subtype.VILLAIN).withAnyColor(RED, WHITE).youControl()`) and the card's own
reminder text carries the definition.

## Convergence — seven cards, and a real bug fix

Seven printed "Equip [quality]" cards were hand-rolling the shape as bare `activatedAbility { }`
blocks. **Five of them never set `isEquipAbility`**: Blackblade Reforged (DOM), Bilbo's Ring and
Dúnedain Blade (LTR), Ghostfire Blade (KTK), Pirate Hat (LCI). Everything keyed off that flag —
`CastPermissionUtils.applyEquipCostReduction`, `applyFreeFirstEquipDiscount`,
`ActivatedAbilityEnumerator`'s instant-speed permission — silently skipped their restricted halves,
so Forge Anew's free first equip, Éowyn's discount and Leonin Shikari's instant-speed equip did not
apply to them. That is a pre-existing bug this PR fixes, and it is why five extra set snapshots move.

Two more (**Thinking Cap**, MKM; **Wizard's Staff**, HOB) *did* set the flag, so the engine saw them,
but froze their cost in a `descriptionOverride` — the other half of the same problem. Both converged.
The Irencrag's redundant `descriptionOverride = "Equip {3}"` is dropped for the same reason.

**Ghostfire Blade is not a printed quality variant.** It prints one `Equip {3}` plus "costs {2} less
to activate if it targets a colorless creature". Its `{1}` ability is our *model* of that reduction —
behaviourally equivalent, and older than the facade's `genericCostReduction` rail, which is how it
would be written today. It rides the same rail and is labelled accordingly, but the comment and the
test both say plainly that this is our modelling, not its Oracle text.

Non-mana equip costs (Dissection Tools, Demonmail Hauberk, Dark Knight's Greatsword, Bloodthorn
Flail) stay hand-rolled — the facade parses a mana cost only, and those already set the flag.

## Client — the damage-doubling badge

`DoubleDamage` badges were recipient-side only, evaluating just the `RecipientFilter`. Mjölnir leaves
`recipient` at `Any`, so an unattached Mjölnir told **both** players "damage dealt to you is doubled"
— false twice over: an unequipped Equipment doubles nothing, and even equipped it doubles one
creature's *outgoing* damage. Attachment-scoped sources now badge the equipped creature instead, via
`DamageUtils.damageDoublersAffectingSource` → `ClientStateTransformer.buildCardActiveEffects`, and
only while attached. That helper reads the maintained `AttachmentsComponent` reverse index rather
than scanning the battlefield, because it runs for every card on every state push.

## Tests

- **`MjolnirHammerOfThorScenarioTest`** (rules-engine) — one file, one card, 12 cases. The
  load-bearing one asserts the *legal-action enumerator* offers only worthy creatures
  (`LegalActionInfo.validTargets` is verbatim what the client highlights), across seven bodies that
  each vary one axis of "worthy". Plus: combat and noncombat doubling with unequipped controls
  pinning the undoubled number; CR 702.6c's targeting-only restriction (a host that stops being
  worthy stays equipped); the ETB damage pinned at exactly 4 on a 2/5; the from-hand discard sweep;
  and three badge cases (unattached / attached / follows the Equipment when it moves).
- **`EquipQualityVariantTest`** (mtg-sets) — three catalog-wide *properties*, not a card list, so a
  new Equipment never needs it edited: every flagged equip is sorcery-speed and targets a single
  creature you control; every mana-cost equip renders as its printed line with no frozen override;
  and anything whose target filter is narrower than plain creature-you-control declares a matching
  `equipQuality`. That last one is what found Thinking Cap and Wizard's Staff.

Every CR number cited in code, comments or this body was verified against
`MagicCompRules_20260619.txt`: 301.5b, 616.1 (with 614.5), 702.6a, 702.6c, 702.6d, 704.5n.

## Gate

- `:rules-engine:test` — **10795 passed, 0 failed**, `BUILD SUCCESSFUL`, exit 0.
- `:mtg-sdk:test` + `:mtg-sets:test` — passed, `EquipQualityVariantTest` 3/3.
- `just rebless-cards` — eight snapshots move (DOM, HOB, KTK, LCI, LTR, MKM, MSH, WOE) and the churn
  is exactly `equipQuality` on the eight restricted equips plus the two removed overrides. No other
  card moved.
- `just check-card-printing "Mjölnir, Hammer of Thor"` — ok, canonical in the earliest real printing.
- `npx tsc --noEmit` (web-client) — clean.

Not done: no web-client playthrough, no e2e run, no UX pass from either seat. A manual playtest
scenario is included at `manual-scenarios/sets/msh/loop-msh-u10-new-cards-castable.json` — one board
where the active player can cast everything new on this branch in a single turn, with Thor as the
only worthy equip target and the other creatures each failing exactly one clause.

## Reviewer, look here

- **The catalog-wide menu-text change** flagged above. It is the call I'd most want a second opinion
  on.
- **`equipCost` metadata on a restricted equip.** `equipAbility` sets it unconditionally, so a card
  with both a restricted and a plain equip keeps the last one's cost — unchanged for every converged
  card, since the plain equip is declared last in each. Mjölnir has only the restricted equip, so its
  `equipCost` is `{1}`. It's metadata only (read by `CardValidator` and `CardLinter`, never by the
  engine, AI or client), but it does mean `equipCost` is not always the unrestricted cost.
- **Three unconverged `SourceFilter` matchers** in `DamageUtils` (`DamageBonusComponent`, the
  max-damage cap, `ReplaceDamageWithMill`) still hand-roll `Any`/`Matching`-only subsets. Pre-existing
  drift from the shared matcher; left alone as out of scope.
- **The Aura half of the attachment-scoped source pair is unexercised** — no catalog card pairs
  `DoubleDamage` with `SourceFilter.EnchantedCreature`. It shares a branch with the Equipment half, so
  it is generality carried by the shared matcher rather than behaviour under test. Documented as such.


bolav:
## Review — u10 · Mjölnir, Hammer of Thor + "Equip [quality]"

Reviewed the u10 delta against `loop-msh-u09` (`git diff loop-msh-u09...HEAD`, 32 files). Two rounds:
round 1 against `fc57b9ce9b`, round 2 against the tip after the badge-fix and playtest commits. **All
findings from both rounds are fixed on the branch** — round 1 in `9d97a9c2bb`, round 2 in
`65e1e4f3d6`. Posting the record so the reasoning is on the PR rather than only in the commits.

**Verdict: approve.** Behaviour is right and the SDK shape is right. `equipAbility` grew no second
targeting rail — the restriction is an ordinary `TargetFilter` on an ordinary `TargetRequirement`,
which is the opposite of the duplicated-rail pattern earlier units in this run kept producing.

### What's good

- **Projection is clean.** The "worthy" check runs through `PredicateEvaluator.matches(state,
  projected, …)`, resolving `IsLegendary` off `projectedValues.types`, `HasColor` off
  `projectedValues.colors` and `NotSubtype` off `projectedValues.subtypes`, with the controller
  predicate going through `matchesControllerPredicate`. A type-, colour- or control-changing effect is
  read correctly; no base-state read anywhere in the new path.
- **The convergence is a bug fix, not scope creep.** Five of the seven converged cards never set
  `isEquipAbility`, and `applyEquipCostReduction` / `applyFreeFirstEquipDiscount` /
  `ActivatedAbilityEnumerator` all early-return on that flag — so Forge Anew, Éowyn and Leonin Shikari
  genuinely skipped their restricted halves. Confirmed green against `ForgeAnewScenarioTest`,
  `EowynLadyOfRohanScenarioTest`, `PirateHatScenarioTest` and both Bilbo's Ring tests.
- **Every CR citation verified** against `MagicCompRules_20260619.txt` — 301.5b, 702.6a, 702.6c,
  702.6d, 704.5n, and 616.1f + 614.5 backing the "each replacement applies once" claim. Oracle text is
  byte-identical to Scryfall.
- **The player-facing claim is actually tested.** `MjolnirHammerOfThorScenarioTest` asserts on
  `LegalActionInfo.validTargets` from the enumerator, which is verbatim what the client highlights —
  not just on validation rejecting a bad target.

### Findings — round 1 (fixed in `9d97a9c2bb`)

1. **Ghostfire Blade was described as a printed CR 702.6c card in three places.** It isn't: it prints
   one `Equip {3}` plus a target-conditional cost reduction. Behaviourally equivalent under Éowyn /
   Forge Anew / cost increases, so a citation defect rather than a behaviour bug. → Reframed as our
   *model* of the reduction in the card comment, the test and `mechanics.md`.
2. **Nothing tested noncombat damage from the equipped creature**, though the card comment, the SDK
   doc and the PR body all claimed it worked. → Added, including damage to a *permanent* so the
   `RecipientFilter.Any` default is exercised.
3. **The ETB test asserted only that a 2/2 died**, which 2, 4 or 8 damage all satisfy, and its comment
   claimed a case it couldn't reach (an Equipment doesn't enter attached — CR 301.5b). → Now pins
   marked damage at exactly 4 on a 2/5 that survives it.
4. **A disjunctive assertion** (`error != null || nothing attached`) that would pass either way. →
   Split.
5. **`quality` and `targetFilter` are unchecked against each other.** → KDoc now says so explicitly
   and names the catalog test as the only guard.

### Findings — round 2 (fixed in `65e1e4f3d6`)

**Important — `damageDoublersAffectingSource` scanned the whole battlefield per card.** It runs from
`ClientStateTransformer.buildCardActiveEffects` for every card on every state push, so the view path
was quadratic in battlefield size — and it did the scan even for cards in hand. The engine maintains
`AttachmentsComponent.attachedIds` as the reverse link for both Auras and Equipment, and
`TriggerAbilityResolver` documents it as exactly the O(1) lookup for this question. → Rewritten to
read the index, with an early return for the un-attached common case. The forward
`AttachedToComponent` check is kept so the badge agrees with what `damageSourceMatches` reads at
damage time.

**Two more cards were still hand-rolling "Equip [quality]"** — Thinking Cap (`Equip Detective {1}`,
MKM) and Wizard's Staff (`Equip Wizard {1}`, HOB). Unlike the original five they *did* set
`isEquipAbility`, so the engine saw them, but both froze their cost in a `descriptionOverride`, which
is the other half of the same problem. → Converged. The Irencrag's redundant
`descriptionOverride = "Equip {3}"` dropped for the same reason.

**The menu-text regression on the converged cards, resolved properly.** Round 1 accepted losing the
printed wording ("Equip Human {1}" → the generated attach text) as a consistency tradeoff. It didn't
need to be a tradeoff: `describeWithCost` now renders the printed keyword line from the new
`equipQuality` field against the *effective* cost, which restores the wording **and** makes it
discount-aware — something a static `descriptionOverride` structurally cannot be. Note this changes
the menu text for every Equipment in the catalog; that's the one call in this PR I'd want a second
opinion on.

**`EquipQualityVariantTest` hardcoded six card names**, so every future quality-equip card would have
to edit it. → Rewritten as three catalog-wide properties. It keys "is this restricted?" off the
*target filter*, not the prompt label — a hand-rolled equip picks its own label
(`"target Wizard you control"`), which is precisely why a label-based check missed the two cards
above.

**Doc-accuracy.** `isAttachmentScopedSource`'s KDoc claimed "every other `SourceFilter` either matches
anything or is controller-relative". `SourceFilter.Self` is neither — true of today's catalog (every
other `DoubleDamage` card uses `SourceFilter.Matching(…youControl())`), but stated as a type-level
guarantee it isn't. → Restated as the catalog fact, naming `Self` as the next shape that would need
the same treatment.

**Minor — playtest scenario didn't match its own commit message.** `a516b0ec2d` names "Skyward
Spider" (never in the file), and the committed board's **Frodo Baggins** (`{G}{W}` Legendary Halfling
Scout) is legendary, non-Villain and white — i.e. *worthy*. So there were two legal equip targets and
the "legendary but wrong colour" axis wasn't on the board at all. → Swapped for Hercules, Prince of
Power (mono-green legendary Hero). Thor is now the only worthy target and each other creature fails
exactly one clause. The `a516b0ec2d` message is still wrong; read the file, not that message.

**Minor — `getTooltipBorderColor` had no `double-damage` case**, so a red-orange badge opened a purple
tooltip while every other custom badge style has a matching pair. → Added.

**Minor — the Aura half of the pair is unexercised.** No catalog card pairs `DoubleDamage` with
`SourceFilter.EnchantedCreature`, and it shares a branch with the Equipment half. → Documented as
shared-matcher generality rather than behaviour under test, rather than adding a test that would
exercise identical code.

### Verification run for this review

- `:rules-engine:test` — **10795 passed, 0 failed**, `BUILD SUCCESSFUL`, exit 0.
- `:mtg-sdk:test` + `:mtg-sets:test` — passed; `EquipQualityVariantTest` 3/3.
- Targeted regression pass over the classes most exposed to the badge change
  (`TwinflameTyrant`, `TheRollercrusherRide`, `LightningArmyOfOne`, `ProjectedControllerInReplacements`,
  `ForgeAnew`, `EowynLadyOfRohan`, `PirateHat`, both Bilbo's Ring tests) — 27 passed, 0 failed.
- Snapshots reblessed; churn is exactly `equipQuality` on the eight restricted equips plus the two
  removed overrides. No other card moved.
- One pre-existing assertion needed updating: `ForgeAnewScenarioTest` asserted the old generated
  prefix `description.startsWith("{0}:")`. Its intent (a free first equip lists as `{0}`) is
  unaffected and its load-bearing `manaCostString shouldBe "{0}"` still passes; the description
  assertion is now `description shouldBe "Equip {0}"`, which is stronger — it pins that the derived
  line actually absorbs the discount.

PR #1799, opened 2026-08-12
Let the AI choose a value for {X}

bolav:
Lets the built-in AI choose a value for `{X}`.

X is announced as part of casting (CR 601.2b), so the legal-action enumerator — which runs *before*
the cast — cannot pick it. It hands over `maxAffordableX`, `minX` and a deliberately permissive
`validTargets`, and leaves the choice to whoever is casting. A human makes it in the client's
X-selection phase. The AI had no equivalent.

For an `ActivateAbility` that was survivable: submitted bare it reaches the engine's own choose-X
pause, which `DecisionResponder` answers by simulating each value. A `CastSpell` has no such pause —
`xValue` left null binds to 0 as the spell goes on the stack — so **every X spell in a deck was dead
weight**. An unexpanded Mind Spring drew nothing; an unexpanded Genesis Wave revealed nothing.

`XCostSelection` is the AI's X-selection phase. It proposes the X values worth considering and
narrows each one's target list to what that X actually reaches, handing the Strategist ordinary
candidates that nothing downstream has to know were ever open.

## Which X values

Two shapes, because "what is a good X" has two different answers:

- **X gates target legality** ("mana value X or less", "mana value X", "power X" — Blue Sun's
  Twilight, Repeal, Spell Blast, Ent-Draught Basin): the candidates are the values that make some
  currently-legal target legal. Sweeping the affordable range would mostly produce X values nothing
  matches. For the "or less" form the candidate is the target's own mana value — a larger X hits the
  same permanent for more mana, so it is dominated.
- **X is free of the targets** (Fireball, Genesis Wave, Day of Black Sun, "up to X target
  creatures"): more X is more effect, so the top affordable values are the interesting ones. X=0 is
  never a candidate — it is the enumerator's own default.

Each candidate then carries a target list narrowed to that X, mirroring `applyXFilters` /
`resolveMaxByX` in the client's `pipelinePhases.ts`. An X that empties a mandatory target slot is
dropped rather than offered; an action with no legal X at all is dropped entirely, because
submitting the bare one would cast it at X=0 and fizzle.

Simulation, not this file, decides *which* candidate is cast. The only judgement here is which
handful are worth a simulation.

## Enumerator fix

The target-gated half depends on four `xConstrains*` flags, and only
`ActivatedAbilityEnumerator` populated all four. `CastSpellEnumerator` and `CastFromZoneEnumerator`
set `xConstrainsTargetManaValue` and `xConstrainsTargetCount` but never the `…ManaValueExactly` /
`…Power` siblings — so Repeal and Spell Blast, both single-requirement `CastSpell`s filtering on
`ManaValueEqualsX`, looked like a free X. The AI swept to the largest affordable value and the cast
was illegal.

This adds the two missing flags at all 14 construction sites. It fixes the **web client** too: its
X-selection phase reads exactly the same flags, so it was failing to narrow for the same cards.

## Shapes that keep their own rule

- A **targeted activated ability** is never pre-bound — the engine's choose-X pause simulates each
  value, which beats anything decided here, and it has to pick targets in the same breath anyway.
  One predicate (`bindsXWithoutTheEnginesHelp`) governs both the expansion path and the
  single-legal-action shortcut so the two can't drift.
- The **Momir avatar** keeps `momirXCandidates`: its X is a format-strategy question — skip the weak
  early activations, then aim at the 8-drop band — not a "more is better" sweep.

## Details worth a look

- **Face-down permanents have no mana cost** (CR 708.2a), which is what `PredicateEvaluator` applies
  for `ManaValueAtMostX` / `ManaValueEqualsX`. Mana values come out of the projection, so a morph
  contributes 0 rather than its printed cost — otherwise the AI derives X from a card the engine
  cannot see and picks a target it then rejects.
- **The candidate cap applies after narrowing.** `candidateXValues` returns the uncapped ordered
  proposal; `expandToX` narrows lazily and then takes `MAX_X_CANDIDATES`, so an uncastable value
  costs a filter rather than a simulation slot.
- **`PlayoutPolicy` samples X rather than maximizing it.** A rollout policy must not be
  deterministic — always taking the largest affordable X would make all R playouts of that line
  identical, collapsing R samples into one. Narrowing reads the board and plays nothing forward, so
  this stays inside the file's no-simulation rule.

## Tests

- `XCostSelectionTest` (35) — the candidate set and the narrowing, kept off the simulation path on
  purpose so the answers don't depend on the evaluator. Covers both X shapes, the face-down rule,
  the optional-vs-mandatory target distinction in both the flat and multi-requirement views, and
  `bindBestX` / `sampleX`.
- `XCostSpellAiTest` (4) — Goldvein Hydra end to end through `chooseAction` for the free-X case;
  Repeal through real legal-action enumeration into `expandToX` for the target-gated case. The
  Repeal half deliberately stops short of `chooseAction`: Repeal is an instant, and whether the AI
  *wants* to cast it in a given window is the hold policy's business, which would make it a test of
  the evaluator wearing a wiring test's name.
- `RepealScenarioTest` (3) — the enumerated cast advertises `xConstrainsTargetManaValueExactly`.

## Verification

`just test` — BUILD SUCCESSFUL, all ten module test tasks green.

## Known gap

Nothing covers the AI casting an X-cost *instant* at a window it would genuinely choose. That needs
an opponent's-end-step scenario and a read on the hold policy's intent; left out rather than guessed
at.


bolav:
## Review — `ai-choose-x-for-spells`

Reviewed `263a7c9ef2` (the original `Let the AI choose a value for X`) against `origin/main`, which
the branch already contained. Findings below; all ten are fixed in `d080707b3f`, so this comment is
the record of what was wrong rather than a request for changes.

### Verdict

Behavior right, shape right. X selection is lifted out of `Strategist` into a small, well-documented
`XCostSelection` that proposes X values and narrows targets, leaving *scoring* to simulation. No new
SDK types — no `Effect`, `Trigger`, `Condition`, `Modification` — so the SDK-elegance bar isn't
engaged and `card-sdk-language-reference.md` correctly needs no update. The blocking problem was
that the two spells the new docs named as the headline examples of the target-gated branch were not
actually reached by it.

### What's good

- **The split is the right one.** `candidateXValues` proposes, simulation decides, `narrowToX`
  mirrors the client's `applyXFilters` / `resolveMaxByX` — including the "X-driven cap *replaces*
  the placeholder rather than clamping it" rule and the `minTargets = min(static, X)` clamp, which
  match `pipelinePhases.ts:1004-1023` exactly.
- **The premise checks out.** `CastSpell.xValue ?: 0` at `CastSpellHandler.kt:877/1199/2169`, versus
  `ActivateAbilityChooseManaXContinuation` on the activate path — the asymmetry the KDoc leans on is
  real.
- **CR 601.2b is correct** ("If the spell has a variable cost that will be paid as it's being cast …
  the player announces the value of that variable").
- **Descending sweep order** is right given the Strategist's anytime contract, and honoring `minX`
  fixes "X can't be 0" abilities that the old `maxOf(1, maxX - MAX_X_CANDIDATES + 1)` ignored.
- **`bindBestX` in `PlayoutPolicy` is a genuine fix, not plumbing:** `FastDecisionResponder.kt:129`
  answers a `ChooseNumberDecision` with `minValue`, so every X ability reaching that pause inside a
  rollout was played for the minimum.

### Blocking / Important

**1. The target-gated branch was unreachable for Repeal and Spell Blast.** `XCostSelection` gates on
`xConstrainsManaValueExactly` / `xConstrainsPower`, but `CastSpellEnumerator` only ever set
`xConstrainsTargetManaValue` and `xConstrainsTargetCount` on the flat fields — the other two came
only from `ActivatedAbilityEnumerator:960-961`. Both `Repeal` (`gpt/cards/Repeal.kt:29`) and
`SpellBlast` (`lea/cards/SpellBlast.kt:26`) are single-requirement `CastSpell`s filtering on
`manaValueEqualsX()`, so they fell into the free-X sweep: largest affordable X, un-narrowed target
list, rejected by `PredicateEvaluator.kt:556-568`. Not a regression (they were dead at X=0 before),
but the docs claimed coverage that didn't exist. *Fixed:* both flags populated at all 14
`CastSpellEnumerator` / `CastFromZoneEnumerator` sites — which also fixes the web client, whose
X-selection phase reads the same flags — plus an assertion in `RepealScenarioTest` and a wiring test
in `XCostSpellAiTest`.

**2. `manaValueOf` ignored the face-down rule the engine enforces.** It read `CardComponent.manaValue`
flat, while `PredicateEvaluator.kt:566/576` applies `if (isFaceDown) 0 else card.manaValue`
(CR 708.2a — a face-down permanent has no mana cost). Against a morph board the AI derived X from a
printed cost the engine cannot see; for the `EqualsX` form the pick was rejected and the legal X=0
never offered. *Fixed,* with three tests.

**3. The `requiresTargets` floor made an optional single target fatal.** `flatRequirement` raised
`minTargets` to 1 whenever `action.requiresTargets`, but `CastSpellEnumerator.kt:1165` hardcodes
`requiresTargets = true` while setting `minTargets = firstReq.effectiveMinCount` — 0 for "up to one
target …". So an X that emptied an optional slot returned null, contradicting `narrowRequirement`'s
own KDoc ("a requirement that may be left empty is never fatal"), which the multi-requirement path
honored. The test that pinned it only exercised `targetRequirements`, never the flat path.
*Fixed:* the floor is gone (`LegalAction.minTargets` already defaults to `targetCount`, itself 1, so
a genuinely mandatory action is floored without help), plus the flat-shape twin of that test.

**4. The single-action shortcut disagreed with the expansion path.** `Strategist.kt:168` called
`bindBestX` unconditionally, while `expandXCostAbilities` deliberately bails on a targeted activated
ability so the engine's choose-X pause — which `DecisionResponder.respondNumber` answers *by
simulating* — keeps handling it. *Fixed:* one `bindsXWithoutTheEnginesHelp` predicate now governs
both.

### Minor

**5. `bindBestX` always took the head inside a rollout,** so every X spell in every playout was cast
for the maximum affordable X. `PlayoutPolicy`'s own doc says a playout "must be stochastic … a
deterministic policy makes all R playouts of one candidate identical", and the neighbouring choice
is a softmax, not a max. *Fixed:* new `sampleX` draws uniformly and threads the generator.

(I checked whether the same always-max bias leaks into the main path via tie-breaking — `maxByOrNull`
keeps the first max, so descending order breaks ties toward the biggest X — but
`RawBoardFeatures.kt:44` scores `untappedLandDifference`, so overpaying is already penalized. Not a
finding.)

**6. `.take(MAX_X_CANDIDATES)` ran before `narrowToX` filtered,** so five unsatisfiable proposals
could drop a spell that a sixth would have cast. *Fixed:* `candidateXValues` returns the uncapped
proposal and the new `expandToX` narrows lazily, then caps.

**7. `targetGatedXValues` unions proposals across requirements where it arguably should intersect.**
*Not fixed in code, deliberately:* with (6) in place `narrowToX` already drops any X that empties a
mandatory requirement and the cap applies after that, so an unsatisfiable proposal costs a filter
rather than a slot. An intersection filter would duplicate the rule. Documented instead, with a test
(`a proposal that cannot be cast is dropped rather than filling a candidate slot`).

**8. `XCostSpellAiTest.aiRegistry` duplicated `ScenarioTestBase.cardRegistry`** exactly, and that
field is `protected`. *Fixed.*

**9. `docs/ai/architecture.md` described only the expand-and-simulate path,** though `bindBestX` has
two non-simulating callers. *Fixed,* along with the enumerator-flag dependency and the face-down
rule.

**10. `narrowToX` ran both branches when `targetRequirements` was non-null** — the enumerator
populates `validTargets` *and* `targetRequirements` for multi-requirement actions and
`TargetSelection.targetInfosFor` prefers the latter, so the flat work was discarded. Harmless on its
own, but it is the mechanism that made (3) fire on multi-requirement actions too. *Fixed:* the flat
view is derived from the already-narrowed requirement 0.

### On the tests

The original suite was well-targeted but blind to (1) and (2) by construction: `XCostSelectionTest`
hand-built `LegalAction`s with flags no enumerated `CastSpell` ever carried, and its board had no
face-down permanents. A Repeal case driven through *real enumeration* is what would have caught (1),
and that is what was added.

One test I wrote and then rewrote, worth recording: driving Repeal through `chooseAction` fails —
the AI passes. That is correct behavior, not a bug. Repeal is an instant and the hold policy saves
it for the opponent's turn, so asserting "the AI casts Repeal here" tests the evaluator under a
wiring test's name. It now runs the real enumerated action through `expandToX`.

### Gate

`just test` — BUILD SUCCESSFUL, all ten module test tasks. `XCostSelectionTest` 17 → 35,
`XCostSpellAiTest` 2 → 4, `RepealScenarioTest` 2 → 3, `AIPlayerTest` 14/14, `PuzzleSuiteTest` and
`BlockSurvivalPropertyTest` unmoved.

Worth noting how close this came to a false green: on the first full run `:ai:test` failed early and
aborted the build **before `:rules-engine:test` ever ran**, leaving the enumerator change — the
largest and riskiest edit — unverified at the point the run would otherwise have been called done.
Per-module gates were re-run to close that.

### Known gap

Nothing covers the AI casting an X-cost *instant* at a window it would genuinely choose. That needs
an opponent's-end-step scenario and a read on the hold policy's intent; left out rather than guessed
at.

PR #1808, opened 2026-08-13
Add Azog, Moria's Ruin

wingedsheep:
## Summary

- add Azog, Moria's Ruin to The Hobbit with its optional ETB target, frozen last-known power, destruction, affected-controller amass, and controller-sensitive draw
- make unresolved relational single-player `ForEach` references a no-op instead of widening to every active player
- document that relational iteration contract and add a manual client scenario

## Capability design and reuse

Azog composes existing `StoreNumber`, `Destroy`, `ForEachPlayer(Player.ControllerOf(...))`, `Amass`, and conditional/draw primitives. The only engine change corrects the general unresolved-reference semantics used by any optional-target “its controller” rider; no card-specific engine code or new SDK/mtgish IR type was added. Existing targeting and `SelectCardsDecision` UI paths cover the player flow.

## Verification

Passed:
- `just test-class ForEachRelationalPlayerTest`
- `just test-class AzogMoriasRuinScenarioTest` (projected/LKI power, opponent/self controller branches, no target, affected-controller multi-Army choice)
- `just rebless-cards` (only Azog added to `HOB.json`)
- `just check-card-printing "Azog, Moria's Ruin"`
- `just check-backlog`
- image URI HTTP 200
- `git diff --check`

Full gate:
- `just test` ran; 10,893 of 10,894 rules-engine tests passed and the other module suites reported passing before the gate stopped.
- One unrelated existing failure remains: `FlourishingBloomKinScenarioTest` (`Expected value to not be null, but was null`). Neither Flourishing Bloom-Kin nor its test is touched by this diff.

Checks not performed:
- no E2E run because this adds no new visual or decision type
- no mtgish emitter work because no new SDK primitive or IR capability was introduced

Independent review is pending.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -2191,6 +2191,10 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
   - `ForEachPlayerEffect(players, effects)` → `IterationSpace.Players(players)` — per matching
     player; `controllerId` rebound so `Player.You` is the current player, `opponentId` recomputed,
     fresh `storedCollections` (Winds of Change, Bend or Break, One Ring to Rule Them All).
+    Relational single-player references such as `Player.ControllerOf("target creature")` iterate
+    exactly that player, using last-known controller information if the permanent has left. If the
+    reference is undefined (for example, an optional target was not chosen), the loop is a no-op;
+    it never widens to every active player.
   - `ForEachInCollectionEffect(collection, effect)` → `IterationSpace.Collection(name)` — per
     entity of a named pipeline collection; `pipeline.iterationTarget` bound so `EffectTarget.Self`
     is the current entity; outer collections preserved (Fight or Flight).
--- a/manual-scenarios/cards/a/azog-morias-ruin.json
+++ b/manual-scenarios/cards/a/azog-morias-ruin.json
@@ -1,0 +1,35 @@
+{
+  "player1Name": "Azog",
+  "player2Name": "Moria Defender",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Azog, Moria's Ruin"],
+    "battlefield": [
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"}
+    ],
+    "graveyard": [],
+    "library": ["Swamp", "Grizzly Bears", "Swamp"]
+  },
+  "player2": {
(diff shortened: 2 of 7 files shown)


wingedsheep:
Verdict: 0 blocking, 1 important, 0 minor. The card definition and the general unresolved-player-loop correction are composition-first and appear rules-correct, but the dedicated card coverage misses the projected-controller/LKI seam that is central to Azog's implementation.

What's good

- Azog composes existing `StoreNumber`, `Destroy`, `ForEachPlayer`, `Amass`, condition, and draw vocabulary; there is no card-specific SDK or executor shape.
- The optional-target no-op is fixed at the general `ForEach` player-reference boundary and has a focused regression test.
- Oracle text, metadata, canonical-printing placement, snapshot scope, affected-player Army choice, projected power, and the no-target ruling all check out. Both focused test classes and the printing check pass on the reviewed head; the reported full-suite failure is outside this diff.

Important

- `rules-engine/src/test/kotlin/com/wingedsheep/engine/scenarios/AzogMoriasRuinScenarioTest.kt:58` — The two controller branches only use normally owned/controlled creatures, so neither would fail if Azog read the base `ControllerComponent` instead of Layer-2 projected control or lost that projected controller after destruction. This is the exact high-risk composition here: `Conditions.TargetMatchesFilter(...youControl())` is evaluated live, then `Player.ControllerOf(...)` must recover the same controller from LKI after `Destroy`. Add a scenario where player 1 steals a player-2-owned creature with a Threaten-style effect, casts Azog targeting it, and assert player 1 both amasses and draws while player 2 does neither. The generic projection and `TargetControllerLastKnownInfoScenarioTest` coverage is useful, but it does not pin Azog's combined conditional-plus-player-rebind pipeline.

Minor

- None.

Recommendation

Add the stolen-target Azog scenario above, rerun `just test-class AzogMoriasRuinScenarioTest`, and keep the implementation shape unchanged if it passes.

wingedsheep:
Review finding accounting: fixed 1 of 1 findings; declined 0.

Added a dedicated Azog scenario in which player 1 steals a player-2-owned Grizzly Bears with Threaten, verifies projected control differs from the base ControllerComponent, then targets it with Azog. The test proves player 1 amasses and draws through the destruction/LKI boundary while player 2 does neither.

Gate: just test-class AzogMoriasRuinScenarioTest — passed (5 tests, BUILD SUCCESSFUL).

PR #1809, opened 2026-08-13
Add five Murders at Karlov Manor cards

wingedsheep:
Five MKM cards built on the set's investigation mechanics — suspect, collect evidence, and investigate. All five compose existing `Effects.*` / `Patterns.*`; no new SDK vocabulary, so `docs/card-sdk-language-reference.md` is unchanged.

| Card | What it does | Composed from |
|---|---|---|
| **Frantic Scapegoat** {R} 1/1 | Haste; suspects itself on entry, then hands the suspicion to a creature that enters later | `Triggers.OneOrMorePermanentsEnter(excludeSource)` + `SelectFromCollectionEffect` over the captured batch + `Effects.Suspect` / `NoLongerSuspected` |
| **Presumed Dead** {1}{B} | +2/+0 and a dies-return-and-suspect trigger, until end of turn | `Effects.ModifyStats` + `GrantTriggeredAbilityEffect` (same shape as Undying Malice) |
| **Torch the Witness** {X}{R} | Twice X damage; investigates if excess was dealt | `DynamicAmount.Multiply(XValue, 2)` + `Conditions.IfTargetTookExcessDamage` |
| **Evidence Examiner** {G}{U} 2/2 | Optional collect evidence 4 at combat; investigates on any collection | `Triggers.BeginCombat` + `MayEffect(Effects.CollectEvidence)` + `Triggers.WheneverYouCollectEvidence` |
| **Analyze the Pollen** {G} | Optional collect evidence 8 upgrades a basic-land tutor to creature-or-land | `collectEvidence(8)` + `ConditionalEffect(WasEvidenceCollected, …, elseEffect = …)` over `Patterns.Library.searchLibrary` |

Nothing dropped from the batch.

### Details worth a reviewer's eye

- **Frantic Scapegoat** — "if this creature is suspected" is an intervening-if (`triggerCondition`), so once the suspicion is handed off the ability stops going on the stack entirely; and the un-suspect is gated on the *selection* being non-empty, not on the ability resolving, which is what makes a declined "may" leave the Goat suspected. Selection uses `useTargetingUI` since the candidates are permanents already in play.
- **Presumed Dead** — the granted suspect is `Duration.Permanent` even though the grant is `EndOfTurn`. Sharing the grant's duration would silently un-suspect the returned creature at cleanup, against the printed rulings.
- **Analyze the Pollen** — the nonbasic land in the test is Commercial District, a `Land — Mountain Forest`: it has basic land *types* but is not *basic*, so the un-upgraded mode must not offer it.

### Verification

Run from the branch worktree, via `just`:

- `just rebless-cards` — MKM golden only, **526 insertions / 0 deletions**, exactly these five cards. No unrelated card moved.
- `just test-class "com.wingedsheep.engine.scenarios.*ScenarioTest"` — the whole scenarios suite green; the only failures in the run were two of my own Evidence Examiner tests (a test-harness issue: `passUntilPhase` leaves begin-of-step triggers queued, so the prompt needed a `resolveStack` first), fixed and re-run green.
- `just check-card-printing` — all five report canonical placement in MKM, their earliest real printing.

15 scenario tests across five files, one file per card.


wingedsheep:
### Review findings addressed

Pushed `729c955` (on top of a merge of current `main`):

- **`FranticScapegoat.kt`** — cited `CR 603.3b`, which does not exist. Corrected to **`CR 603.6a`** ("Each time an event puts one or more permanents onto the battlefield, all permanents on the battlefield … are checked for any enters-the-battlefield triggers that match the event"), verified against the 2026-08-07 Comprehensive Rules. The other four citations (`701.59b`, `701.60a`, `603.4`, `120.10`) all check out.
  *Note:* the same bad `603.3b` citation exists in `TriggerDetector.kt` (~line 3005) — that's where I copied it from. Out of scope for this PR, but worth a follow-up.
- **Two test gaps closed.** Both implementations were already correct; the tests just didn't prove it:
  - `FranticScapegoatScenarioTest` — an opponent's creature entering must not trigger the batch ability. Every prior test deployed only on one side, so a filter missing the "you control" scope would have passed all of them.
  - `PresumedDeadScenarioTest` — the creature must return under its **owner's** control. The other tests use the caster's own creature, where owner and controller coincide.

All 17 scenario tests across the five cards pass locally.

### Heads-up: a pre-existing flaky test may trip CI here

`FlourishingBloomKinScenarioTest > finding only one Forest puts it onto the battlefield, never into your hand` failed in one of my local full-suite runs. **It is unrelated to this PR** — that file is not in this diff, and the cause is in the test itself:

`GameInitializer.kt:161` seeds the shuffle from entropy when no seed is supplied (`config.seed ?: System.nanoTime()`), and `initMirrorMatch` supplies none. That test's deck is `Plains 39 / Forest 1`, so whenever the single Forest is drawn into the opening hand the library holds no Forest, the search raises no decision, and `find.shouldNotBeNull()` fails — roughly a 1-in-5 chance per run.

It passes on clean `main` too, which is exactly what a ~20% flake looks like. I have deliberately **not** touched it, since it isn't my change. The fix would be `withRngSeed`/an explicit `GameConfig.seed`, or a deck where the Forest can't be drawn away.

PR #1810, opened 2026-08-13
[agent-loop: gpt-5.6-sol] Add 2 The Hobbit cards

wingedsheep:
Agentic-loop batch driven by gpt-5.6-sol.

- Bard, King of Dale — composes the existing per-card draw replacement and token-creation multiplier.
- My Precious // Allure of Power — composes Adventure, an additional sacrifice cost, static Equipment grants, and a composite equip cost.

Kíli the Resourceful was dropped during review because its whole-equip-cost replacement needs a reusable engine fix and dedicated feature PR.

Verification:
- `just build` — passed after review fix
- `just check-card-printing` for both included cards — passed
- HOB snapshot audited: only these two card trees were added

Not checked: no manual playthrough, no UX pass, and no e2e.


wingedsheep:
Review verdict: changes requested. Bard and My Precious compose existing SDK primitives cleanly, their Oracle metadata and canonical placement match Scryfall, and the full build is green.

Blocking — `KiliTheResourceful.kt:44`: `FreeFirstEquipEachTurn` does not faithfully cover Kíli’s wording for equip costs containing nonmana atoms. `CastPermissionUtils.applyFreeFirstEquipDiscount` replaces mana atoms with `{0}` but retains atoms such as `PayLife`; therefore Kíli plus My Precious incorrectly charges 2 life for the first equip, even though `{0}` is paid rather than the equip cost. This needs a reusable engine fix and scenario coverage, so under the batch policy Kíli must leave this existing-primitives PR and receive its own feature PR.

Recommendation: remove Kíli and its HOB snapshot entry from this batch; ship Bard and My Precious after re-gating. Implement Kíli separately with the generalized whole-equip-cost replacement and tests.

wingedsheep:
Review follow-up: the blocking Kíli finding is resolved by removing Kíli and its snapshot entry from this existing-primitives batch. Re-review of the remaining Bard and My Precious diff found no blocking, important, or minor issues. `just build` passes after the fix, and the HOB golden adds only those two card trees. Recommendation: merge once required GitHub checks are green.

PR #1835, opened 2026-08-15
Add four Marvel Super Heroes cards with the engine vocabulary they need

bolav:
Four Marvel Super Heroes cards, each with the engine vocabulary it needed. The units were built
separately and are squashed to one commit apiece, so the PR reads as four self-contained changes:

| Commit | Card | New vocabulary |
|---|---|---|
| `Add Loki Laufeyson…` | Loki Laufeyson | last-known information for source-relative dynamic caps |
| `Add Bullseye, Death Dealer…` | Bullseye, Death Dealer | none — composes existing primitives |
| `Add dynamic spell cost reductions…` | The Scarlet Witch | `CostReductionSource.Dynamic(amount: DynamicAmount)` |
| `Add a filter-level counter-history predicate…` | Kid Loki | `StatePredicate.ReceivedCounterThisTurn` |

## Loki Laufeyson — LKI for source-relative dynamic caps

`{1}, {T}` arms a pending copy rider whose mana-value cap is Loki's own power, and a power-up
ability grows him. The cap is resolved as the matching spell is cast (when the delayed trigger's
condition is checked), so pumping him after arming the rider widens what it copies.

The rider outlives its source (CR 113.7a), so Loki may be gone when the spell is cast, and CR 608.2h
then makes the cap a last-known read: armed at 2/1, powered up to 4/3, killed in response, the cap
is 4. The engine already owned that rule — `DynamicAmountEvaluator` has a `LIVE_THEN_LKI` branch —
but `PredicateEvaluator.evaluateDynamicCap` rebuilds a minimal `EffectContext` and carried no
snapshot, so the branch found nothing and fell through to base characteristics. `PendingSpellCopy`
now carries `lastKnownSourceSnapshot`, stamped by `ZoneTransitionService` at the leaves-battlefield
choke point out of the `EntitySnapshot` it already assembles.

Stamped at **departure** rather than at rider creation, because the source's power can change in
between and CR 608.2h wants the last known state. First departure wins, so a later blink reusing the
same `EntityId` cannot overwrite it.

## Bullseye, Death Dealer — no new vocabulary

The ETB is a real CR 603.12 reflexive trigger, so the damage target is chosen after the
sacrifice-or-discard is paid: a `ReflexiveTriggerEffect` over `ChooseActionEffect`, one
`EffectChoice` per printed option with a `FeasibilityCheck` hiding options the controller can't
perform. With neither payable the "may" prompt is suppressed outright, so the reflexive can never
fire without a cost being paid.

The activated ability's cost is a choice between two independently payable non-mana costs, which the
engine has no vocabulary for on activated abilities (`AdditionalCost.Choice` is cast-time only). It
takes Bloodthorn Flail's decomposition: two abilities identical but for which half of the cost they
pay. That introduces a CR 601.2c-before-601.2h announcement-order difference (CR 602.2b routes
activation through 601.2b–i); it cannot change a legal line.

Captain America, Living Legend and Cloak and Dagger, Entwined were dropped from that batch — both
need new SDK vocabulary and belong in their own PRs.

## The Scarlet Witch — `CostReductionSource.Dynamic`

No `CostReductionSource` could read the permanent the modifier is printed on, and the
greatest-property matcher ignores name and state predicates, so both obvious workarounds reduce by 0.

`Dynamic(amount: DynamicAmount)` is the general variable-amount source, evaluated at cast time with
the carrying permanent as the amount's `EntityReference.Source` — deliberately the same vocabulary
`ReduceActivatedAbilityCost` takes on the activated-ability rail. "Twice this creature's power",
"your life total" and counter counts need no new members.

"You" for the amount is the **source's** controller, not the caster — under
`SpellCostTarget.AnyCaster` those differ — read from projected state so control-change effects are
respected. Floored at 0 per CR 107.1b, per source, so a shrunk source reduces nothing and can never
eat another source's discount.

Coexists with main's `CostReductionSource.AttachedPermanentProperty` (Glamdring, Foe-hammer), which
landed independently and needed the same "pass the modifier's own source permanent into the
evaluator" plumbing. This reuses that existing `abilitySourceId` parameter rather than adding a
second one: `AttachedPermanentProperty` reads the permanent the source is attached to, `Dynamic`
reads the source itself.

## Kid Loki — `StatePredicate.ReceivedCounterThisTurn`

"Each creature you control that you've put one or more +1/+1 counters on this turn has hexproof"
needs the counter-history question at *filter* scope; only the self-scoped
`Conditions.SourceReceivedCounterThisTurn` existed.

The new predicate adds no tracking — `ReceivedCountersThisTurnComponent` already recorded both the
counter kind and the placed-by-controller flag. The engine read lives once in
`handlers/predicates/ReceivedCounterThisTurnPredicate.kt` and is wired into all four exhaustive
`StatePredicate` dispatch sites, so target filtering, group-static projection, trigger gating and
untap filtering all give the same answer.

**Net SDK surface goes down**: `SourceReceivedCounterThisTurn` is deleted and its facade
reimplemented as `SourceMatches` over the new predicate, so the source- and filter-scoped readings
share one evaluator. The facade signature is unchanged, but Beast, Erudite Aerialist and Fractal
Tender re-serialize — hence the MSH/SOS snapshot moves.

CR 122.6a counter-placement facts are threaded through every call site, not just the four enters-with
ones. Explore, move-counters-each-kind, the three counter continuations, devour's pre-entry placement
and wither all supply the counter kind and the placer, so Kid Loki covers a creature you made
explore. Tokens created with initial counters stamp the same marker (CR 122.6) without emitting a
`CountersAddedEvent`.

### Follow-up hardening in the same commit

Review found that `DamageUtils.markCounterPlacedOnCreature` accepted a *nullable* counter kind, and
six call sites passed none — `DoubleCountersExecutor`, `MoveAllLastKnownCountersExecutor`, two blight
paths in `CastSpellHandler` and two in `CostHandler`. Those stamped the marker with empty type sets,
which no type-scoped counter-history filter could ever match. The kind is now required, and the two
executors mark once per kind inside their loop rather than once outside it.

**No shipped card reaches that gap today** — every doubler in the catalog pairs `AddCounters` with
`DoubleCounters` so the kind is already recorded, and The Millennium Calendar doubles on a
non-creature. So this is hardening against a future card, not a live bug fix, and no scenario test
was added for behaviour that cannot currently be triggered.

## Rules

Every CR number cited in the diff was verified against `MagicCompRules_20260619.txt`: 122.6, 122.6a,
122.5, 608.2h, 113.7a, 107.1b, 603.12, 601.2b, 601.2f, 602.2b, 202.3, 202.3e, 702.6a, 702.6c,
702.193, 707.10c, 707.10d, 707.10e, 700.4, 121.2.

## Gate

`scripts/gradle-locked :rules-engine:test --no-build-cache --rerun-tasks` over the four units plus
the counter / blight / explore suites — **25 classes, 119 tests, 0 failures, 0 errors**.

Per-unit: Loki Laufeyson 8, Bullseye 5, The Scarlet Witch 8, `DynamicCostReductionTest` 12, Kid Loki
7, Beast Erudite Aerialist 5.

Loki's LKI behaviour is mutation-checked: removing the single line that threads the snapshot into the
`EffectContext` turns exactly one test red — the pumped-then-killed case, the only one where
last-known and printed power disagree. The other two LKI cases are controls guarding the opposite
direction and stay green under mutation by construction.

`--no-build-cache --rerun-tasks` is deliberate: these commits were rebased, and Gradle otherwise
serves a stale `FROM-CACHE` compile and passes falsely.

## What was NOT done

- **No manual playthrough in the web client**, from either seat.
- **No UX pass** and **no e2e**.
- The full multi-module `test` suite was not re-run after the final rewrite; the gate above is scoped
  to the modules these commits reach (`rules-engine`, plus the card snapshots via `mtg-sets`).
- `BeastEruditeAerialistScenarioTest` covers a *pre-existing* card that this PR re-serializes; it was
  added because the source-scoped half of the consolidation had no test anywhere, not because that
  card changed behaviour.

This PR came out of an autonomous loop. An independent review has been posted as a comment; its
findings were applied before this PR was opened.


bolav:
## Independent review — 4-unit stack vs `main`

**SCOPE: full** (engine + SDK changed) · **FINDINGS: 1 blocking, 1 important, 1 minor** — all three
addressed before this PR was opened; see the accounting at the bottom.

### Verdict

The SDK shape is right — all three additions generalize rather than special-case, and one of them
deletes a type on the way in. Behaviour is correct and the tests are unusually good. One Blocking
issue: commit contamination in the Scarlet Witch commit.

### What's good

**All 21 CR citations verify** against `MagicCompRules_20260619.txt` — 122.6/122.6a/122.5, 608.2h,
113.7a, 107.1b, 603.12, 601.2b/f, 602.2b, 202.3/202.3e, 702.6a/c, 702.193, 707.10c/d/e, 700.4,
121.2. Not one had drifted, which is rare enough to call out.

**`StatePredicate.ReceivedCounterThisTurn` is a model case.** Parameterized on both axes real cards
vary (counter kind, placed-by-controller), reuses the existing `ReceivedCountersThisTurnComponent`
with no new tracking, and *deletes* `SourceReceivedCounterThisTurn`, re-expressing the facade as
`SourceMatches` over the new predicate. Net SDK surface goes down while capability goes up.

**One read, four dispatch sites.** `receivedCounterThisTurn(container, predicate)` is the single
implementation behind `PredicateEvaluator`, `AffectsFilterResolver` (group-static projection),
`TriggerMatcher` and untap filtering — the "one condition, both contexts" rule, and why the layer
integration is trustworthy.

**`CostReductionSource.Dynamic` chose the right generality.** Taking a full `DynamicAmount` rather
than an `EntityNumericProperty` means "twice this creature's power" and "your life total" need no new
member, and it matches the vocabulary `ReduceActivatedAbilityCost` already uses on the other cost
rail.

**Honest gap-documentation.** `TriggerMatcher` states plainly that the zone-change path doesn't read
LKI, that it fails closed, and where to add it when a card needs it.
`BeastEruditeAerialistScenarioTest` was added because the *pre-existing* source-scoped half of the
consolidation had no test anywhere.

### Blocking

**1. The Scarlet Witch commit contained two unrelated files.** `git log --diff-filter=A` showed it
adding `backlog/sets/marvel-super-heroes/TODO.md` (+38) and
`manual-scenarios/cards/l/loki-laufeyson.json` (+41). Neither is that unit's work: `TODO.md` is
loop-run planning that was untracked on `main` and documents a card (Baron Helmut Zemo) not in this
stack at all, and the JSON is a Loki artifact belonging to the first commit. Cause: a `git add -A`
during conflict resolution swept both untracked files in.

### Important

**2. `LOOP_UNIT.md` churned 183 lines across the stack.** It is tracked on `main`, so not new
contamination, but every unit rewrote it with per-unit loop notes and each commit carried a near-total
rewrite of the previous unit's text. In a PR against `main` that is pure noise and conflicts with
every other loop branch.

### Minor

**3. Dead justification in `ReceivedCounterThisTurnPredicate.kt`.** The comment claimed "placement
paths that can't name a counter kind stamp the marker with empty type sets", justifying a
marker-presence shortcut for the any-kind reading. It also masked an asymmetry: `(null, false)`
answered from marker presence while `(null, true)` consulted a type set, so the two any-kind readings
could disagree.

### Recommendation

Fix #1; #2 and #3 at the author's discretion. The engine work, the tests and the rules research are
in good shape.

---

## Accounting — all three applied before this PR was opened

**#1 fixed.** The stack was rebuilt from `main` commit-by-commit. `TODO.md` is in no commit and back
to untracked on disk; `loki-laufeyson.json` is now added by the Loki commit and the Scarlet Witch
commit does not touch it.

**#2 fixed.** `LOOP_UNIT.md` is untouched by all four commits, and the file at the branch tip is
byte-identical to main's blob.

**#3 fixed — and it was undergraded.** Making the counter kind mandatory made the compiler surface a
second recording path: `DamageUtils.markCounterPlacedOnCreature` took a nullable kind and **six** call
sites passed none (`DoubleCountersExecutor`, `MoveAllLastKnownCountersExecutor`, two blight paths in
`CastSpellHandler`, two in `CostHandler`), each stamping the marker with empty type sets that no
type-scoped filter could match. The kind is now required and the two executors mark once per kind
inside their loop.

Correcting the review's own severity call: this was graded Minor and "latent", and the plumbing gap
was real — but **no shipped card reaches it end-to-end**, because every doubler in the catalog pairs
`AddCounters` with `DoubleCounters` (so the kind is already recorded) and The Millennium Calendar
doubles on a non-creature. It is hardening against a future card, not a live bug fix, and no scenario
test was added for behaviour that cannot currently be triggered.

The rewrite touched only those items: the diff against the pre-fix stack is the three fixes plus the
six call sites, with `mtg-sdk`, `mtg-sets` and `docs` completely untouched.

**Re-gated after the rewrite**: 25 classes, 119 tests, 0 failures, 0 errors.

PR #1836, opened 2026-08-15
Collapse `TriggeredAbility.optional` into the consent gate it already built

wingedsheep:
`TriggeredAbility.optional` was a second spelling of `Gate.MayDecide`, and the argument for that was
not made here — it was already in the engine. `TriggerProcessor.putOnStack` read the flag and
*built* `GatedEffect(Gate.MayDecide, then = effect, otherwise = elseEffect)` from it before the
ability ever reached the stack, skipping the wrap when the effect already `ownsConsentGate()`. One
spelling lowered into the other on every game. The flag is gone; the gate is the model.

`optional = true` survives as a **DSL shorthand** on `TriggeredAbilityBuilder`, lowered in `build()`
to `MayEffect(effect, otherwise = elseEffect)`. That keeps 149 card definitions unedited while the
serialized model has exactly one shape — which is what the differential fold existed to paper over.
Setting it on an effect that already owns a consent gate is now a build-time `require` rather than a
double prompt. Outside the DSL there is no shorthand: `TriggeredAbility.create` takes no `optional`,
and the five call sites that passed it (three Legions Slivers, Grothama, soulbond) write
`MayEffect(...)`.

## Settling `TriggerProcessor.kt:708` first, as asked

```kotlin
val effectiveMinTargets = if (ability.optional) 0 else req.effectiveMinCount
```

**Verdict: a third meaning, and a wrong one. Deleted rather than rehoused.**

It is load-bearing — it was the *only* consent mechanism for the 54 targeted `optional` abilities,
which is exactly what the neighbouring comment at :686 said ("carries its consent solely through the
target-selection decision's `minTargets = 0`"). But the claim it makes is not "the controller may
decline the effect", it is "every target slot is *up to* N", and that is wrong twice:

- **CR 603.3d** chooses targets for a "you may" trigger like any other, and the choice to decline
  comes on resolution. Reclamation Sage's "you may destroy target artifact or enchantment" targets
  mandatorily.
- An ability whose target is mandatory and has no legal target is **removed from the stack**. With
  `minTargets = 0` it went on the stack targetless and resolved doing nothing — indistinguishable for
  most of these cards, not indistinguishable in general.

It also collided with the *real* "up to N", which is `optional` on the **requirement**
(`TargetCreature(optional = true)`). `EtbSuspectUpToOneTargetCreatureTest`'s card was written with
the ability flag for "up to one target creature" precisely because the two were spelled the same
way; it is fixed here, and the two statements now have two spellings.

So there is nowhere for it to live, because it should not have existed. A slot's minimum is now the
requirement's, full stop.

The second read (`:694`) fell out with it. The single-legal-player-target auto-select was disabled
for optional abilities because skipping the target decision skipped the consent riding on it (the
Ebon Dragon fail-open). Consent is now asked *before* targeting is reached, so the guard is
unnecessary and the optimization is back.

## What changes at the table

The 54 targeted "you may" triggers move onto the may-then-target path that the 214 hand-written
`MayEffect` triggers already used. That is a strict improvement in four ways, none of which is new
code:

- an explicit yes/no instead of "select zero targets to decline" — a decline you could stumble into
- a proper fizzle when no legal target exists (CR 603.3d)
- eligibility for **run batching** (`BatchYesNoDecision`) — `batchKeyOf` requires `asMayDecide()`, so
  the flag spelling could never batch
- eligibility for **remembered auto-answers** (`autoAnswerFor`), same reason

Two behaviours are preserved deliberately:

- **Entrails Feaster** ("you may exile a creature card from a graveyard … If you don't, tap this") is
  the one targeted optional with an `elseEffect`. Its else moves inside the gate, and `asMayDecide()`
  already refuses gates carrying an `otherwise`, so it routes through `processTargetedTrigger` and
  the gate asks at resolution — CR-correct. The no-legal-target branch now reads the else from the
  gate as well as from `elseEffect`, so it still taps with every graveyard empty.
- **Feasibility stays derived, never stored.** `withImpliedMayFeasibility` stamps it onto a no-target
  `Gate.MayDecide` from the effect, so Yawgmoth Demon and Devouring Sugarmaw still skip an
  unanswerable "you may sacrifice" without any card recording a fact its text does not state. A gate
  that declares its own feasibility (Provisions Merchant) keeps it.

## Assay

`Triggers.abilityFor`/`scriptFor` and `Granted`'s pair lose the lowering entirely — a trigger's
effect clause is now the same value a spell's clause is — and **`Folds.liftTriggerConsent` is
deleted**.

The `gated == MayEffect(gated.then)` fail-closed check goes with it, and is not missed: nothing is
discarded any more, so nothing can be dropped.

## Gates

| Gate | Result |
|---|---|
| `just build` | green (full, including all tests) |
| `just test-scenarios` | green — the real check, since the flag was read on every triggered ability |
| `just assay-gate` | MISMATCH 0, AMBIGUOUS 0, redundant readings 0 |
| `just assay-differential` | 2,431 compared / 2,425 confirmed / **6 divergent** |
| `just rebless-cards` | 109 abilities across 106 cards reshape `"optional": true` → a `Gate.MayDecide` wrapper; 109 removed, 109 added, no other key moved |

**On the divergence count.** The task quoted 2 as today's number. My branch point (`a4a62e1a45`) is
not that tree: I ran the differential on a clean detached worktree at that exact commit and it
reports **6** — 2,431 compared, 2,425 confirmed, and the *same six cards* (Mm'menon, Zombie Master,
Lavaborn Muse, Donatello, Tattered Ratter, Kalastria Highborn). After this change it is byte-for-byte
the same. So the number is unmoved and the fold really was redundant. (Two of the six, Mm'menon and
Donatello, are the `TriggerBinding.OTHER` cards fixed in the shared checkout's uncommitted work,
which is not in this branch; the other four are unrelated to consent — a grant filter's
`IsCreature`/`IsPermanent`, and the bare-tribal-noun family.)

## Test changes

Eleven scenario tests asserted the old flow and now assert the new one. Four are worth reading rather
than skimming, because they are where the semantics changed:

- **`EbonDragonScenarioTest`** — was the regression test for the auto-select fail-open. Its guard was
  "the target decision offers `minTargets = 0`"; it is now "the may-question is raised at all", which
  is what the auto-select can no longer skip.
- **`SerpentAssassinTest`** — "can be declined by selecting no targets" becomes "can be declined at
  the may-question", and the sibling test now asserts `minTargets == 1`: a targeted "you may" is not
  an "up to one".
- **`EntrailsFeasterTest`** — targets first, may at resolution. All four tests pass, including
  "no creature cards in any graveyard - gets tapped automatically", which is what proves the
  else-from-the-gate path.
- **`EtbSuspectUpToOneTargetCreatureTest`** — the test card is re-authored with an optional
  *requirement*, which is what "up to one target creature" always meant.

Docs: `card-sdk-language-reference.md`'s `optional` section rewritten (shorthand, not a field; the
three consequences); `oracle-assay/README.md` records the finding as closed and marks the Legions
band's "~10 cards" bullet resolved.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -3945,20 +3945,37 @@ table because the prompt looks identical. Rather than mis-place the gate silentl
 throws, and `EffectOncePerTurnLoweringTest` sweeps every card in the pool for the shape so the
 failure lands at build time rather than mid-game.
 
-**`optional` = "you may [effect]"; `elseEffect` adds "If you don't, [elseEffect]."** For a
-**targeted** trigger, `optional` lets the player choose 0 targets to decline, and `elseEffect` runs
-on decline or when no legal targets exist (Entrails Feaster: "you may exile a creature card from a
-graveyard … if you don't, tap this"). A **no-target** optional trigger lowers to a
-`GatedEffect(Gate.MayDecide, then = effect, otherwise = elseEffect)` resolved by the unified gated
-executor — a resolution-time yes/no whose "no" runs `elseEffect` (or nothing when there is none,
-e.g. Song of Stupefaction's "you may mill two cards"); the wrap is skipped when `effect` already
-carries its own consent gate (a `May*`-gated `GatedEffect`), so an authored `Effects.May` never
-double-prompts. The may-action's feasibility is
-derived from `effect` (a `SacrificeEffect` needs the controller to control a matching permanent), so
-an impossible "may" skips the prompt and runs `elseEffect` directly — the no-target analogue of "no
-legal targets → else" (Yawgmoth Demon: "you may sacrifice an artifact. If you don't, tap this
-creature and it deals 2 damage to you" — with no artifact the tax just applies). Always-feasible
-bodies (gain life, draw, add a counter) always prompt.
+**`optional` = "you may [effect]" — a DSL shorthand, not a field on the model.** `build()` lowers it
+to `MayEffect(effect, otherwise = elseEffect)`, i.e. a `GatedEffect(Gate.MayDecide, …)` around the
+effect, and clears `elseEffect`. `TriggeredAbility` has **no** `optional` flag: it had one, the
+engine read it and constructed exactly this gate before the ability reached the stack, and two
+spellings of one fact meant a card could be written either way and a differential fold had to bridge
+them. Writing `MayEffect(...)` by hand is equivalent and is the only option outside the DSL
+(`TriggeredAbility.create` takes no `optional`). Setting `optional = true` on an effect that already
+owns a consent gate is rejected at build time rather than prompting twice.
+
+What that gets you, uniformly, for targeted and untargeted triggers alike:
+
+- **The yes/no is its own decision.** For a **no-target** trigger the unified gated executor asks it
+  at resolution and runs the gate's `otherwise` on "no" (or nothing when there is none, e.g. Song of
+  Stupefaction's "you may mill two cards"). For a **targeted** trigger `TriggerProcessor` asks it as
+  the ability goes on the stack and only then selects targets, so declining never costs you a target
+  choice first — and the trigger becomes eligible for batching and for a remembered auto-answer, both
+  of which key on the gate.
+- **Targets follow CR 603.3d.** A slot's minimum is the *requirement's*: "target creature" stays
+  mandatory and the ability is removed from the stack when nothing is legal. "Up to one target
+  creature" is an optional **requirement** (`TargetCreature(optional = true)`), which is a different
+  statement and now has a different spelling.
+- **`elseEffect` keeps its own meaning.** On a mandatory ability it is still the branch for target
+  selection producing nothing. A "you may … If you don't, …" ability carries that clause inside the
+  gate; the no-legal-target path reads it from there too, so Entrails Feaster ("you may exile a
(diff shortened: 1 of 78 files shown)


wingedsheep:
Merged `origin/main` (through `1eae31bf9b`) into the branch. Two conflicts, both in Assay grammar and
both the same shape:

- `Triggers.abilityFor` and `Granted.abilityFor` — main refactored the `optional` derivation into a
  `val may = (effect as? GatedEffect)?.takeIf { … }` local. That whole derivation is what this branch
  deletes, so the resolution is to take this branch's side. Main's *other* change to `Triggers.kt`,
  the `Steps.triggeredStep` slot on `filteredTriggerRule` (the third-anaphor / Tattered Ratter fix
  from `0a111d637d`), is in a different hunk, auto-merged, and kept.

Everything else auto-merged, including the 48 card snapshots. Re-ran `just rebless-cards` after the
merge and it produced **no drift** — the merged goldens are already what the merged code emits.

`just test-scenarios` green after the merge (0 failures).

**One pre-existing failure on `main`, disclosed rather than touched:**
`SubtypesTest > the bare noun parses and prints back as the adjective form`. I verified it fails on a
clean checkout of `origin/main` at both `1eae31bf9b` (the commit merged here) and `768239a2dc` (the
tip at the time of writing), so it is not caused by this branch. It belongs to the in-flight
bare-tribal-noun migration and is someone else's to finish — not reverted, not patched.

PR #1842, opened 2026-08-16
Add four MSH cards and their engine vocabulary: active damage history, first-time-tapped, targeted proliferate, equip-ability mana

bolav:
Four Marvel Super Heroes cards and the engine vocabulary each one needed, plus the review fixes for the bundle.

MSH goes 254 → 258 / 276.

## Cards

- **Red Guardian, Super-Soldier** (#34) — `{2}{W}` 2/2, flash, "destroy target creature an opponent controls that dealt damage this turn".
- **Captain America, Living Legend** (#210) — `{1}{W}{U}` 3/4, vigilance, "whenever a creature you control becomes tapped during your turn, if it's the first time that creature has become tapped this turn, untap it".
- **Powerful Broker** (#179) — `{2}{G}` 3/3, `{T}`: targeted single-object proliferate, sorcery speed.
- **Ronin, Shadow Stalker** (#112) — `{2}{B}` 3/3, a once-per-turn life-for-mana ability restricted to Equipment spells and equip abilities, plus `{T}`, sacrifice an attached Equipment: −4/−4.

All four Oracle texts match Scryfall exactly, and all four canonicals sit in MSH (their earliest real printing).

## Engine vocabulary

**Active-voice damage history.** `StatePredicate.HasDealtDamage(thisTurnOnly)` replaces the old `data object`. One per-permanent marker (`HasDealtDamageComponent`, now carrying the turn it last dealt damage) answers both windows: presence is the lifetime reading, the turn stamp against `state.turnNumber` is the per-turn one. Nothing needs end-of-turn cleanup — a stale stamp simply stops matching — and no damage path can record one window without recording the other.

The old `dealtDamageThisTurn()` filter builder was the **passive** predicate ("was dealt damage"). It is renamed `wasDealtDamageThisTurn()` and the active one is `hasDealtDamageThisTurn()`. The short name is retired rather than reused, so an un-rebased caller becomes a compile error at the exact line instead of silently flipping to the opposite set of permanents.

**Per-permanent first-time-tapped window.** `TappedEvent.firstThisTurn`, computed in the `tap()` atom from `HasBecomeTappedComponent`, matched by `EventPattern.TapEvent.firstTimeEachTurn`. Per-*permanent*, not per-*ability* — which is exactly what `oncePerTurn` cannot express, and the two are tested head-to-head.

Regeneration is folded into the same atom: CR 701.19a defines it as "…remove all damage marked on it and **its controller taps it**", which is a real tap transition. It was open-coded and event-free, so it silently skipped every "becomes tapped" trigger (Deeproot Pilgrimage as much as this card). All three call sites now propagate the emitted event, and `TapEventEnforcementTest`'s allowlist no longer grandfathers any real tap.

**Targeted proliferate.** `Effects.Proliferate(target?)`. With no argument it is proliferate proper (CR 701.34a) — recipients chosen at resolution, nothing targeted. With a target it is the single-object form: announced with the ability (CR 601.2c), respondable, subject to hexproof/shroud/protection, and countered if the target is illegal on resolution (CR 608.2b). Both shapes share one placement routine, which also picked up the `canReceiveCounters` prohibition the untargeted path had been ignoring.

**`TargetPermanentOrPlayer`.** The general "… or player" requirement, with a `permanentFilter` so the same type covers "target permanent or player", "target artifact or player", and so on. All five dispatch sites got a branch (validator, finder, enumeration utils, `ChangeTargetExecutor`, `ReselectTargetRandomlyExecutor`), each mirroring its `TargetCreatureOrPlayer` sibling rather than inventing new legality logic. Its KDoc points at `backlog/target-union-with-arms.md`, which scopes collapsing this family into one arms-based requirement — this type is the interim shape and the intended migration target.

**`ManaRestriction.EquipAbilityActivationOnly`.** Reads `SpellPaymentContext.isEquipAbilityActivation`, a fact about the *ability* rather than its source. `SubtypeSpellsOrAbilitiesOnly("Equipment")` cannot express this: it admits every activated ability of an Equipment source, so it also pays Iron Man Armor's `{2}`: animate. A card type cannot separate them either — the equip ability and the others share one source. `buildAbilityPaymentContext` gained a required (nullable, undefaulted) `ability` parameter so every activation site had to be revisited; all eight were.

**Source-relative cost filters.** `…attachedToSource()` in an activated ability's sacrifice cost is `StatePredicate.IsAttachedToSource`, which is unconditionally false without a source in the `PredicateContext`. Enumeration and payment now both supply the ability's own source, so "Sacrifice an Equipment attached to Ronin" works — and so does **Faunsbane Troll**'s "Sacrifice an Aura attached to this creature", which had been unpayable and untested since it shipped.

## Review fixes (last commit)

- **Captain America's intervening-`if` is now checked twice.** The clause is a printed intervening-"if" (CR 603.4), so `InterveningIfClassificationTest` — added upstream in `28d8fdbe93`, after these units were written — was red on it. Carrying a frozen copy of the event flag onto the stack object would have turned the gate green while leaving the second check decorative, since it can only repeat the trigger-time answer. `HasBecomeTappedComponent` therefore gained a per-turn **count**: a bare turn stamp cannot tell "tapped once" from "tapped twice", and that distinction *is* the second check. The card now carries the clause once per check — the event rider for the trigger-time one, `Conditions.TriggeringPermanentBecameTappedOnlyOnceThisTurn` for the resolution one. Untap the creature and tap it again in response and the ability is removed from the stack.
- **Freya Crescent converged onto `EquipAbilityActivationOnly`.** It prints the same clause as Ronin and was still on the over-broad spelling; its own KDoc named the gap and said the fix needed a flag that did not exist yet. This branch built the flag. It had no test at all, so it has one now.
- **`OneOrMoreBecomeTapped(firstTimeEachTurn)` dropped.** It shipped a guessed semantics for zero cards. Since the flag lives on the public `TapEvent`, removing only the DSL parameter would have left the pairing silently ignorable — `TapEvent` now rejects `batch && firstTimeEachTurn` outright, so the first real card decides the reading instead of inheriting one.
- **Card scenario tests moved to their era modules** per AGENTS.md (Faunsbane Troll → 2023, the four MSH cards → 2026). Mechanic tests stay in `rules-engine`.

## Testing

New scenario/mechanic tests: `RedGuardianSuperSoldier`, `CaptainAmericaLivingLegend`, `PowerfulBroker`, `RoninShadowStalker`, `FreyaCrescent`, `FaunsbaneTroll`, plus `HasDealtDamagePredicate`, `FirstTimeTappedThisTurn`, `TargetedProliferate`, `EquipAbilityManaRestriction`, `RegenerationTapEvent`.

| Suite | Result |
|---|---|
| `:mtg-sdk` `:rules-engine` `:ai` `:mtgish-tooling` `:mtg-sets` | ✅ `BUILD SUCCESSFUL` |
| `:mtg-sets:2023:tests` (forced `--rerun`) | ✅ 1266 tests, 0 failed |
| `:mtg-sets:2025:tests` (forced `--rerun`) | ✅ 1584 tests, 0 failed |
| `:mtg-sets:2026:tests` (forced `--rerun`) | ✅ 1179 tests, 0 failed |

The three era suites were forced to re-execute rather than accepting Gradle's `UP-TO-DATE`. Card snapshots were re-blessed and move exactly two cards — Freya's restriction and Captain America's `interveningIf`.

`ConniveTargetingTest` failed once mid-review with a 120 s `TimeoutCancellationException`, only when six module suites ran alongside the ten-module scenario fan-out. It passes in a split run; its sole contact with this diff is an `AbilityCost.Tap`, and the change there is an O(1) component write.

Not covered: `game-server` and `web-client`. Nothing in the diff reaches either and no new `GameEvent` type was added (`TappedEvent` gained a defaulted field, which `ClientEvent.kt` already branches on), so no gate was run for them.

## Compatibility notes

- `HasDealtDamageComponent` and `HasBecomeTappedComponent` no longer have a zero-argument serial form. No server path is affected — replays re-simulate from a compact record rather than storing `GameState` — but a hand-saved scenario JSON from an older build will not decode.
- `TapEvent`'s new `require` rejects `batch && firstTimeEachTurn` at card-load time rather than at compile time. It is unreachable from the DSL today.
- `ProliferateEffect` became a `data class` with an all-default field, so its `@SerialName("Proliferate")` wire shape is unchanged.


bolav:
## Review — SDK elegance, correctness, tests

Reviewed at `58bf143043` (the four-unit bundle) with `origin/main` merged in; every finding below has since been addressed in `1b37964d27`, and the resolution is recorded under each one. Kept as the record of what was checked.

### Verdict

Behaviour right, SDK shapes right. One blocking gate failure, one over-broad card left behind by a new primitive, three minor items.

### What's good

- **`StatePredicate.HasDealtDamage(thisTurnOnly)`** is the model refactor here: one marker answers both windows, so no damage path can record one without the other, and the per-turn window expires by arithmetic rather than by a cleanup entry someone forgets. Retiring the bare `dealtDamageThisTurn()` name so a stale caller fails to **compile** rather than silently flip voice is exactly right — that rename would otherwise have been a landmine.
- **Regeneration routed through `tap()`** (CR 701.19a verified). All three production callers propagate the returned event; `MoveCollectionExecutor:651` was the only one that needed fixing. `TapEventEnforcementTest` was *tightened*, not loosened, and its KDoc now names its own two scan holes.
- **`ProliferateExecutor.addOneOfEachKind`** unifies the targeted and untargeted placement, and picks up `canReceiveCounters` on the untargeted path as a side effect — a real pre-existing bug, with a test for each branch.
- All five `TargetRequirement` dispatch sites got a `TargetPermanentOrPlayer` branch, each mirroring its `TargetCreatureOrPlayer` sibling rather than inventing new legality logic.
- Every CR citation checks out against `MagicCompRules_20260808`: 701.34a, 701.26a, 701.19a, 702.6a/c/e, 601.2c, 608.2b, 614.17b, 400.7, 603.4, 603.3d, 506.4b, 500.3, 605.1a/3b, 120.1, 702.194a. The docs even fix a stale `701.27` → `701.34` for proliferate. All four Oracle texts match Scryfall exactly; all four canonicals are in MSH.
- The `mtgish` emitter handles the substring hazard correctly (`blob.replace("WasDealtDamageThisTurn", "")` before testing the active tag) and labels the guessed IR tag as unverified rather than claiming it.

### Blocking

**`InterveningIfClassificationTest` fails on Captain America.** The gate (upstream `28d8fdbe93`, which *is* in this branch's history) reports:

> `[MSH] Captain America, Living Legend: prints an intervening "if" but models it as a triggerRestriction`

The card's KDoc anticipated the divergence and argued it honestly, but the gate is deliberately allowlist-free, so "documented" is not a resolution.

The trap in fixing it: moving `Conditions.IsYourTurn` into `interveningIf` would pass the gate, but "during your turn" is not the printed CR 603.4 clause — that is precisely the silent misfiling the gate exists to catch, in the other direction. Equally, carrying `TappedEvent.firstThisTurn` onto the stack object and re-reading it would satisfy the field the gate inspects while leaving the second check decorative, since a frozen flag can only repeat the trigger-time answer.

**Resolved** in `1b37964d27` by making the second check real. `HasBecomeTappedComponent` gained a per-turn `timesThisTurn` count — a bare turn stamp cannot distinguish "tapped once" from "tapped twice", which is the entire content of the second check. The clause is now carried once per check: the existing event rider for the trigger-time one, and `Conditions.TriggeringPermanentBecameTappedOnlyOnceThisTurn` (composition over the existing `EntityMatches(TriggeringEntity, filter)` primitive — no new `Condition` type) for the resolution one. `triggerRestriction = IsYourTurn` stays where it belongs. New test: hold the trigger on the stack, untap and re-tap underneath it, assert the creature stays tapped — the only case that can tell a real second check from a frozen flag.

### Important

**Freya Crescent is now the odd one out** — `mtg-sets/2025/.../fin/cards/FreyaCrescent.kt:55`. It prints *the same clause* as Ronin's second half and uses `SubtypeSpellsOrAbilitiesOnly("Equipment")`, whose KDoc says verbatim:

> *"A dedicated equip-only restriction would require threading an `isEquipAction` flag through the activated-ability legal-action enumerator, which is out of scope for this card."*

This PR builds exactly that flag. Leaving Freya behind means the corpus carries two encodings of one printed clause, one of them known-wrong — it pays Iron Man Armor's `{2}`: animate, the very example the new atom's KDoc cites as its reason to exist.

**Resolved:** migrated to `AnyOf(SubtypeSpellsOnly(setOf("Equipment")), EquipAbilityActivationOnly)`, the "out of scope" paragraph deleted. It had no test at all, so it now has one, built around an inline Equipment whose equip and non-equip abilities both cost `{1}` — so neither cost nor card type can be what separates them, only the restriction.

### Minor

**`OneOrMoreBecomeTapped(firstTimeEachTurn = …)`** — `Triggers.kt`. The KDoc states outright that no printed card combines the batch wording with a first-time clause and that the chosen "narrow the batch" semantics is an analogy. That is an SDK parameter added for zero cards with guessed semantics, and two tests pinned the guess.

**Resolved:** parameter dropped. Since the flag lives on the public `TapEvent`, removing only the DSL sugar would have left the pairing silently ignorable, so `TapEvent` now rejects `batch && firstTimeEachTurn` in `init` (the `SpellPaymentContext` idiom already in this diff). The unreachable detector narrowing and both tests are gone; one test pins the refusal.

**`TargetPermanentOrPlayer` doesn't cite the `TargetUnion` plan** — `TargetRequirement.kt`. `backlog/target-union-with-arms.md` scopes collapsing exactly this family (`AnyTarget`, `TargetCreatureOrPlayer`, `TargetPlayerOrPlaneswalker`, `TargetOpponentOrPlaneswalker`) into one arms-based requirement. This diff adds a fifth member — a better one, since it has the criteria slot the proposal argues for — but the next "X or player" author would read the KDoc, see no mention of the plan, and add a sixth.

**Resolved:** KDoc points at the backlog doc; the doc gained a table row and a paragraph naming this type the interim shape and first migration candidate.

**Card scenario tests live in `rules-engine`, not their era module.** `FaunsbaneTrollScenarioTest` covers a 2023/WOE card with no new engine vocabulary; the four MSH card tests belong beside `AgentMariaHillScenarioTest` in `mtg-sets/2026/tests/`. Precedent inside this same loop is mixed, so this was author's discretion — but each file added there is compile weight on the module that already starves the daemon first.

**Resolved:** all five moved. Mechanic tests (`FirstTimeTappedThisTurn`, `TargetedProliferate`, `EquipAbilityManaRestriction`, `RegenerationTapEvent`, `HasDealtDamagePredicate`) stay in `rules-engine`, which is correct for them.

**Serialization note (no change made).** `HasDealtDamageComponent` lost its zero-argument serial form for a code-discipline reason, while `TappedEvent.firstThisTurn` was explicitly *given* a default so older events still decode. Nothing in the server path breaks — replays re-simulate rather than store `GameState` — but it is worth a line in the PR body rather than a code change. Now recorded there.

### Also fixed in passing

`FirstTimeTappedThisTurnScenarioTest`'s class KDoc still described regeneration as an un-fixed exception that skips the tap event — stale since this branch routed it through `tap()`.

### Verification

`:mtg-sdk` `:rules-engine` `:ai` `:mtgish-tooling` `:mtg-sets` all green; the 2023 / 2025 / 2026 era suites forced to re-execute rather than accepting `UP-TO-DATE` (1266 / 1584 / 1179 tests, zero failures). Snapshot re-bless moves exactly two cards.

`ConniveTargetingTest` failed once mid-review with a 120 s `TimeoutCancellationException` — only when six module suites ran alongside the ten-module scenario fan-out. It passes in a split run, and its sole contact with this diff is an `AbilityCost.Tap` whose change is an O(1) component write. Recorded rather than written off.

Not covered by any of this: `game-server` and `web-client`. Nothing in the diff reaches either, and no new `GameEvent` type was added, so no gate was run for them.

PR #1851, opened 2026-08-16
Add four Marvel Super Heroes cards, each with the engine primitive it needs

bolav:
# Four Marvel Super Heroes cards, each with the engine primitive it needs

Four MSH cards — Captain Mar-Vell, Space-Born [12], Invisible Woman, Sue Storm [17], Loki, God of
Mischief [65], Cloak and Dagger, Entwined [211] — and the four independent pieces of engine
vocabulary they exposed. MSH goes from 258/276 to 262/276.

Five commits, 47 files. Each card composes existing SDK primitives; **no new `Effect`, `Trigger`,
`Condition` or `StaticAbility` type is introduced** — every new capability is a field on an existing
pattern or a parameter on an existing recipe.

Oracle text, mana cost, type line, P/T, rarity, collector number and artist verified against
Scryfall for all four. `scripts/check-card-printing.py` exits 0 for each (canonical in MSH, the
earliest real printing; Loki's `pmei` promo correctly ignored). Every CR number cited in code,
comments or commit messages was checked against `MagicCompRules_20260808.txt`.

### Captain Mar-Vell, Space-Born [12] — conditional flash grants

> Flying, vigilance
> Cosmic Awareness — As long as an opponent has cast a spell this turn, you may cast spells as
> though they had flash.

`GrantFlashToSpellType(Any, controllerOnly = true)` behind
`SpellsCastThisTurn(EachOpponent) >= 1` — pure composition, but the grant was inert. Flash
permission is read at cast-legality time and never reaches the layer system, and both read sites
matched the raw ability type, so a grant wrapped in the `ConditionalStaticAbility` that
`staticAbility { condition = … }` produces granted flash *never* rather than conditionally.

`FlashTypeGrants` now owns the whole "may this be cast as though it had flash?" decision — the
card's own `conditionalFlash`, the turn-scoped player grant, and battlefield `GrantFlashToSpellType`
statics including conditional ones, whose gate is re-asked on every read against the granting
permanent. `CastPermissionUtils.hasGrantedFlash` and `CastZoneResolver.hasGrantedFlash` were
byte-identical copies; both are now thin delegates, so enumeration and the authoritative cast-time
re-check cannot drift.

Two pre-existing bugs in that scan are fixed while consolidating it, since the rewrite would
otherwise carry them forward:

- it iterates `state.controlledBattlefield` rather than the ownership-keyed zone map, so a **stolen**
  granter grants to its new controller (CR 109.5) and a **phased-out** one grants nothing
  (CR 702.26b) — and, being the projected view, this is also a base→projected correction;
- it reads `effectiveStaticAbilities(classLevel)` rather than `script.staticAbilities`, so a Class
  card's level-gated grant is no longer inert.

Both are covered by tests. The condition context uses the granter's controller as `you`, the same
player the `controllerOnly` check uses, so the two gates always agree about who that is.

### Invisible Woman, Sue Storm [17] — batched counter-placement triggers

> Whenever you put one or more +1/+1 counters on **one or more** other Heroes you control, you may
> create a 0/4 colorless Wall creature token with defender.

`EventPattern.CountersPlacedEvent` matched one `CountersAddedEvent` at a time and the counter
executors emit one event per recipient, so an effect countering three Heroes fired the ability three
times — CR 603.2c says an ability triggers only once each time its trigger event occurs.

Adds `batch: Boolean = false`, following the existing `TapEvent`/`UntapEvent`/`DealsDamageEvent`
batch shape exactly: the per-event `TriggerMatcher` branch skips batch patterns, a new
`TriggerDetector.detectCountersPlacedBatchTriggers` pass fires them once, and it reads the existing
`TriggerCategory.COUNTERS_ADDED` bucket rather than adding a category. Counter type, `placedBy`
(CR 122.6a), the recipient filter, `firstTimeEachTurn` and SELF/OTHER binding all *narrow* the batch
instead of discarding it; the first matching recipient is bound as the triggering entity and the
trigger's counter count is the batch total.

The two multiplicities share their narrowing rather than each spelling it out:
`TriggerMatcher.matchesCountersPlacedAxes` is the one implementation both call, so a batch trigger
and its per-permanent twin can never disagree about which placements count. That extraction also
drops a wart — the old inline branch shadowed the injected `predicateEvaluator` with a fresh
`PredicateEvaluator()` per call.

### Loki, God of Mischief [65] — player targets and an abilities-only axis

> Whenever a player or permanent becomes the target of an ability you control, draw a card. This
> ability triggers only once each turn.

`StackResolver.emitBecomesTarget` returned early for `ChosenTarget.Player`, so CR 601.2c ("The
chosen objects and/or players each become a target of that spell") was half implemented. It now
emits for player targets, stamped with a new `BecomesTargetEvent.targetIsPlayer`, and folds players
into the same per-turn targeting tracker permanents use. All four target-declaration sites share the
one helper, whose `sourceIsSpell` parameter is now required so every call site must state whether a
spell or an ability did the targeting. The `when` over `ChosenTarget` is exhaustive now rather than
falling through an `else`.

`EventPattern.BecomesTargetEvent` gains `includePlayerTargets` (widens *what got targeted*) and
`abilitiesOnly` (narrows *what did the targeting*, mirror of `spellsOnly`). Player targets are
opt-in and the guard sits ahead of the target-filter check, so no existing becomes-target wording
starts firing on a targeted player; ward is unaffected, being SELF-bound off a battlefield permanent.

Three illegal combinations now fail loudly at load rather than half-working:
`spellsOnly + abilitiesOnly`, `byYou + byOpponent`, and `includePlayerTargets` with a non-`Any`
filter (a player carries no card data for a filter to read, so that pair used to fire on permanents
only while `description` still promised the player half).

**Known bug, audited and pinned rather than fixed.** The three retarget/reselect sites
(`resumeChangeSpellTarget`, `ContestedRetargetLogic.advance`, `ReselectTargetRandomlyExecutor`) emit
nothing for *any* target kind, so ward and every other becomes-target trigger miss every redirect
today. That is wrong under the rules, not an open question: CR 115.9c counts the targets chosen when
a spell or ability was put on the stack "(as modified by effects that changed those targets)", and
CR 603.2e fires "becomes" triggers at the moment the named event happens. It is pre-existing and
orthogonal to the player axis, so `BecomesTargetPlayerAndAbilityAxesTest` characterizes it and says
in as many words that it locks in current-and-**wrong** behaviour and should be **inverted, not
deleted**, when a follow-up fixes it.

### Cloak and Dagger, Entwined [211] — returning exiles to the zone they came from

> When Cloak and Dagger enter, choose target opponent and up to one target creature they control.
> They reveal their hand. You may exile a nonland card from their hand or the chosen creature until
> Cloak and Dagger leave the battlefield.

One exile clause reaching two different zones, and one leaves-the-battlefield trigger that has to
put each back where it belongs — CR 610.3's "this second one-shot effect returns the object to its
previous zone".

Adds `CardDestination.ToZoneExiledFrom`, a **per-card** `MoveCollection` destination backed by a new
`ExiledFromZoneComponent` recording where each object was exiled from. The executor groups the
collection by recorded zone and runs each group through the ordinary `ToZone` path, so owner
routing, Aura targeting, reveals and events behave identically. A card returning to the battlefield
comes back as a new object (CR 400.7) under its owner's control (CR 610.3c); a card exiled **from
exile** (CR 406.7 — "it doesn't change zones") is left where it is, with no move and no event; a
card recorded as exiled from the **stack**, or with no record at all, takes the `fallback`.

The stamp is written by `ZoneTransitionService`'s `Zone.EXILE` entry branch — every effect-driven
exile — and explicitly at the direct-`addToZone` sites that bypass it: exile as a cost
(`CostPaymentService` and `CastSpellHandler`'s three additional-cost branches),
`ExileOpponentsGraveyardsExecutor`, and the graveyard sweep in `SbaZoneMovementHelper`. It is
best-effort by design, so `ToZoneExiledFrom`'s `fallback` is load-bearing rather than decoration,
and that is documented on both. The same service clears it on the way out, as do the two exits that
reuse the entity id (`StackResolver`'s cast-from-exile — which covers *every* cast from exile, not
just suspend — and `ReturnOneFromLinkedExileExecutor`), so it can't be read stale.

`ExilePatterns.returnLinkedExile` now takes a destination and `returnLinkedExileToHand` is a thin
facade over it rather than a third parallel fixed-destination recipe; both build byte-identical
effect trees to before, which two pinning tests check.

`SuccessCriterion.Auto.canInfer` returns false for the new destination (there is no single zone to
snapshot), so `CardValidator` rejects an `IfYouDo` over it at card load rather than failing open.

## Review fixes (`641f4bc82b`)

The first four commits were reviewed; this commit applies the findings.

**Blocking — the MSH golden snapshot was stale, and `:mtg-sets:test` was red.** The committed block
for Invisible Woman carried `"optional": true` beside a bare `CreateToken`, but `CardBuilder.build()`
lowers a triggered ability's `optional` into `MayEffect` — the model has no such field — so the live
tree is `Gated{Gate.MayDecide}`. It had been blessed against an older model and never re-blessed
after the rebase. Re-blessed with the official `-DupdateSnapshots=true`; only `MSH.json` moved.

**Cloak and Dagger re-implemented an existing recipe.** Its hand branch was a hand-rolled copy of
`HandPatterns.revealHandAndExileChosen` — the reveal/gather/select/move tree Cruelclaw's Heist and
Soul Search already share — differing only in needing `linkToSource` (so a hand exile joins the
source's linked-exile pile; `ExileUntilLeaves` only accepts battlefield permanents and graveyard
cards, so it can't reach a hand) and in hoisting the reveal out of the "you may" (the reveal is
unconditional; bundling it in would make it conditional too). Both are parameters now
(`linkToSource`, `revealHand`) and the card composes the recipe: 26 lines and 9 imports gone. The
only serialized difference is the pipeline key name, which is all the snapshot diff shows — the
5 scenario tests pass unchanged, as do Cruelclaw's Heist and Soul Search.

**`describeObjectsForEvent` had no test and a KDoc example wrong twice over.** Predicates render in
insertion order and `withSubtype` appends, so `Creature.youControl().withSubtype(HERO)` renders
"creature Heroes you control", not the documented "Hero creatures you control" — and since the head
noun is therefore the *subtype*, the `-o` rule pluralized "Hero" to "Heros". Masked today only
because the card overrides its description, but the string is computed at load for every batch
pattern and is what the next batch card without an override would show. Split out
`pluralizeHeadNoun` with a curated irregular table built by walking the 204-entry `Subtype` catalog
(`-f → -ves`, the genuine irregulars, WotC's invariant plurals), excluding entries the existing
"ends in -s" rule already gets right and inventing nothing for types whose plural couldn't be
confirmed. New `DescribeObjectsForEventTest` (12 cases) makes the documented examples executable and
pins the end-to-end batch description.

The **word order** is deliberately unchanged: it is `describeObjectForEvent`'s corpus-wide
convention (singular renders "a creature Hero you control"), and reordering would move generated
text well beyond batch triggers. Documented at the function.

**`Triggers.countersPlacedOn(firstTimeEachTurn)` now defaults to `!batch`** rather than to `true`
defended by prose — a batch template essentially never carries a printed "for the first time this
turn" rider, and inheriting one silently narrows it. Required moving `batch` ahead of it in the
signature; all 17 call sites use named arguments. The combination stays expressible, and the engine
test that deliberately exercises `batch + firstTimeEachTurn` still passes.

`docs/card-sdk-language-reference.md` updated for every SDK addition and both signature changes.

## Gate

Run on this tree at `641f4bc82b`, with `origin/main` already an ancestor.

**Focused on what the diff reaches** — `:mtg-sdk:test` and `:mtg-sets:test` in full, the nine touched
engine and scenario classes, and the two existing `revealHandAndExileChosen` callers:

| Scope | Result |
|---|---|
| `:mtg-sdk:test` + `:mtg-sets:test` (full) + 9 touched classes + Cruelclaw's Heist / Soul Search | **892 passed, 0 failed** |

That includes `CardDefinitionSnapshotTest` (320 cases, MSH golden now matching), `FacadeBoundaryTest`,
`PatternsNamespaceTest`, `CardLintTest` and `SuccessCriterionValidationTest`.

**Full `test` across every module: 3 failures, all environmental.**

| Test | Failure |
|---|---|
| `AbuJafarTest > Abu Ja'far blocking and dying …` | `TimeoutCancellationException` @ 120000 ms |
| `AccursedCentaurTest > Accursed Centaur ETB requires sacrificing …` | `TimeoutCancellationException` @ 120000 ms |
| `AIPlayerTest > AI can evaluate board state` | `TimeoutCancellationException` @ 120000 ms |

None is an assertion failure — all three are the 120 s Kotest coroutine timeout, hit while four
concurrent test JVMs drove the load average to ~16 on an 11 GiB box. **All three pass standalone on
this same tree**, re-run individually after the box went quiet:

| Re-run | Result |
|---|---|
| `AbuJafarTest` | 3 passed |
| `AccursedCentaurTest` | 3 passed |
| `AIPlayerTest` | 17 passed |

None of the three is reachable from this diff (an Antiquities dies-trigger, an ETB sacrifice prompt,
and the AI board evaluator), and all three have flaked the same way in earlier full runs on this
box — `build/pr/u11-recheck-{AbuJafarTest,AccursedCentaurTest,AIPlayerTest}.log` record the same
timeout-then-pass-standalone pattern.

**Not covered:** `web-client` (no TypeScript change in this diff), and no e2e or manual playthrough.

## Known gaps

- **Cloak and Dagger's second target slot is the two-player approximation.** The printed text is "up
  to one target creature **they** control"; the filter is `TargetFilter.CreatureOpponentControls`,
  which coincides exactly at two players but is laxer in a multiplayer pod. `ControllerPredicate`
  already has the faithful `ControlledByTargetPlayer` and `TargetValidator` already resolves it
  against sibling choices — what's missing is the enumeration half: `TargetFinder.findLegalTargets`
  takes one `TargetRequirement` with no view of the others' choices, so the predicate resolves to
  `false` for every candidate. Same approximation Demonic Junker, Kitesail Larcenist and Unstable
  Glyphbridge already document.
- **Neither `ChooseAction` branch carries a `FeasibilityCheck`,** and neither can today — the two
  available variants read the *choosing* player's own zones, while these conditions are about the
  targeted opponent's hand and about whether the optional target was chosen. So a branch with
  nothing to exile is still offered and resolves as a no-op; both dead branches are covered in the
  scenario test.
- **`moveToZonesExiledFrom` abandons un-run groups if one pauses.** Only two destinations can pause
  (a library group under `CardOrder.ControllerChooses`, and a battlefield group containing an Aura),
  and they are ordered last; the shipped facade uses the default `CardOrder.Preserve`, so it is
  unreachable through the public API. TODO'd at the call site for a follow-up that resumes the
  remaining groups from the continuation.
- **The batch pass can't filter no-op counter placements.** A permanent that can't have counters put
  on it stays out of the batch only when the placing executor checks
  `ProjectedState.canReceiveCounters` first. `AddCountersExecutor` and its single-target siblings do;
  `AddCountersToCollectionExecutor` and `DistributeCountersAmongTargetsExecutor` do not. That gap is
  per-executor and pre-dates this pass (those executors also add the counters to the component), so
  it is recorded rather than papered over.
- **Four units on one branch.** This branch carries u30, u31, u26 and u24 plus the review-fix commit.
  They are stacked for merge-conflict convenience, not real dependency — u31 and u24 collide only
  textually in `EventPattern.kt`/`Triggers.kt`. Splitting them into four PRs is reasonable if you'd
  prefer to review them separately.
- **No manual playthrough in the web client, no UX pass, no e2e.** `BecomesTargetEvent` maps to no
  `ClientEvent` and no card adds a new decision shape, so no client change was needed. Playtest
  scenarios for all four cards are in `manual-scenarios/sets/msh/`.


bolav:
## Review — SDK elegance, correctness, rules accuracy, tests

Reviewed the branch with `origin/main` merged in (it was already an ancestor, so no merge commit).
Findings below were raised against `a76a405a0e`; **all four actionable ones are fixed in
`641f4bc82b`**, the last commit on this PR. Verdicts and reasoning are kept so the record survives.

### Verdict

Behavior is right and the SDK shape is right: four cards, four Oracle-exact implementations, and
**no new `Effect`/`Trigger`/`Condition`/`StaticAbility` type** — every new capability is a field on
an existing pattern. One blocking issue (a stale golden snapshot) and three smaller ones, all now
addressed.

### Verified independently

- All four cards match Scryfall Oracle text, mana cost, type line, P/T, rarity, collector number and
  artist exactly.
- `scripts/check-card-printing.py` exits 0 for all four — canonical in MSH, the earliest real
  printing (Loki's `pmei` promo correctly skipped).
- Every CR number cited in code, comments or commit messages checks out against
  `MagicCompRules_20260808.txt`: 610.3, 610.3c, 400.7, 406.7, 603.2c, 603.2e, 601.2c, 602.2b,
  603.3d, 115.9c, 115.10a, 122.6/122.6a, 109.5, 111.7, 702.26b, 702.8a, 704.5d, 303.4f.
- Claims in comments that I could falsify, and which hold: `ChosenTarget` really has exactly four
  members, so the `when` is genuinely exhaustive; `CleanupPhaseManager` really iterates *all*
  entities, so player targeting markers do clear; `TriggeredAbilityFiredThisTurnComponent` really is
  stamped in `processSingleTrigger` at put-on-stack, matching the corrected doc comment;
  `removeFromCurrentZone` covers *every* cast from exile, so the origin stamp can't ride onto a
  permanent; the batch detector sits in the standard row, so `capOncePerTurnTriggers` and
  `filterByTriggerCondition` apply to it; and the Loki test asserts **Loki's controller** drew, not
  the targeted player — the assertion that rules out a `triggeringPlayerId` mis-binding.

### What's good

- **The SDK bar is met.** `abilitiesOnly`/`includePlayerTargets` on `BecomesTargetEvent`, `batch` on
  `CountersPlacedEvent`, a `destination` parameter on `returnLinkedExile` — four cards, zero new
  types. `returnLinkedExileToHand` becoming a thin facade over the parameterized recipe, with two
  pinning tests proving byte-identical trees, is the right way to do that refactor.
- **`TriggerMatcher.matchesCountersPlacedAxes`** extracted so both multiplicities share one
  narrowing — a batch and its per-permanent twin *cannot* disagree. It also fixes a latent wart: the
  old inline branch shadowed the injected `predicateEvaluator` with a fresh `PredicateEvaluator()`
  per call. Reads `state.projectedState` throughout.
- **The `FlashTypeGrants` consolidation** collapses two byte-identical copies into one and picks up
  two genuine pre-existing fixes on the way (`controlledBattlefield` → CR 109.5 stolen granter and
  CR 702.26b phased-out; `effectiveStaticAbilities(classLevel)` → Class-gated grants), each with its
  own test. `controlledBattlefield` reads projected state, so this is also a base→projected
  correction.
- **Illegal combinations fail loudly** rather than half-working — `spellsOnly + abilitiesOnly`,
  `byYou + byOpponent`, `includePlayerTargets` with a non-`Any` filter. And
  `SuccessCriterion.Auto.canInfer` returns `false` for `ToZoneExiledFrom`, so `CardValidator` rejects
  it at load rather than failing open. Fail-closed by construction.
- **The redirect characterization test is honest.** It states that it locks in current-and-wrong
  behaviour, cites 115.9c/603.2e for why, and instructs a future unit to invert rather than delete
  it. That is the right way to pin a known bug.

### Blocking — fixed in `641f4bc82b`

**1. `mtg-sets/src/test/resources/snapshots/cards/MSH.json` was stale; `:mtg-sets:test` was red.**
The golden block for Invisible Woman had `"optional": true` beside a bare `CreateToken`, but
`CardBuilder.kt:1477` is explicit that "the model has no `optional` flag; `build` lowers this one"
into `MayEffect`, which serializes as `Gated{Gate.MayDecide}`. The committed golden could not be what
the builder produces — it was blessed against an older model and never re-blessed after the rebase,
exactly the risk `LOOP_UNIT.md` flags for itself ("Gate re-run owed"). Diff was 30 lines, one card.

*Fixed:* re-blessed via the official `-DupdateSnapshots=true`. Only `MSH.json` moved.

### Important — both fixed in `641f4bc82b`

**2. `CloakAndDaggerEntwined.kt` re-implemented `HandPatterns.revealHandAndExileChosen` inline.**
That recipe (`HandPatterns.kt:716`) already builds exactly this tree — `RevealHandEffect(target)` +
`GatherCardsEffect(FromZone(HAND, player, Nonland))` + `SelectFromCollectionEffect(ChooseExactly(1),
Chooser.Controller, prompt)` + `MoveCollectionEffect(ToZone(EXILE, player))`, with
`player = effectTargetToPlayer(…)` resolving to the same `Player.ContextPlayer(0)` the card
hardcoded. It differed in exactly two ways, one parameter each: `linkToSource = true` on the final
move, and the reveal needing to sit *outside* the `MayEffect` because the printed reveal is
unconditional. As written, the next "reveal their hand, exile one until ~ leaves" card copies twenty
lines again.

*Fixed:* added `revealHand = true` and `linkToSource = false` to the pattern; the card now composes
it, shedding 26 lines and 9 imports. The only serialized difference is the pipeline key name — which
is all the snapshot diff shows, and the strongest evidence the substitution is behavior-preserving.
Cruelclaw's Heist and Soul Search, the two existing callers, pass unchanged.

**3. `describeObjectsForEvent` (`EventPattern.kt`) was untested and its documented example wrong.**
The KDoc claimed `Creature.youControl().withSubtype(HERO) -> "Hero creatures you control"`. It
didn't. `GameObjectFilter.Creature` is `cardPredicates = [IsCreature]` (`ObjectFilter.kt:121`) and
`withSubtype` *appends* (`:293`), so `baseParts` is `"creature Hero"` (descriptions `"creature"` then
`"Hero"`, `CardPredicate.kt:32`/`:223`). The head-noun rule then took `"Hero"` and, because `-o`
fell through to `else`, produced **`"creature Heros you control"`** — wrong word order *and* a
misspelled plural, on the one subtype the only shipped batch card uses. Latent only because the card
overrides its description, but the string is computed at card load for every batch pattern and is
what the next batch card without an override inherits. Zero call sites outside this one, zero tests.

*Fixed:* split out `pluralizeHeadNoun` with a curated irregular table built by walking the 204-entry
`Subtype` catalog — `-f → -ves`, the genuine irregulars, WotC's invariant plurals — excluding
entries the existing "ends in -s" rule already gets right, and inventing nothing for types whose
plural couldn't be confirmed. New `DescribeObjectsForEventTest` (12 cases) makes the documented
examples executable and pins the end-to-end batch description. **Word order left unchanged on
purpose**: it is `describeObjectForEvent`'s corpus-wide convention (singular renders "a creature
Hero you control"), and reordering would move generated text well beyond batch triggers.

### Minor

**4. `Triggers.countersPlacedOn`'s `firstTimeEachTurn` trap — fixed in `641f4bc82b`.** The facade
defaulted it to `true` while `batch` implies you almost never want it, defended only by prose in two
places. Now defaults to `!batch`, which removes the trap without banning the combination (the engine
test deliberately exercises it). Required moving `batch` ahead of it in the signature; all 17 call
sites use named arguments.

**5. Four units on one branch — not fixed, your call.** The branch carries u30, u31, u26 and u24 plus
the review-fix commit, but `LOOP_UNIT.md` documents only u24 and its "Base:" section still narrates
the pre-rebase chain `u26 → u31 → u30 → u28 → origin/main`, which no longer exists — `origin/main` is
the direct base now. Three of the four add independent engine vocabulary and are stacked for
merge-conflict convenience rather than real dependency (u31 and u24 collide only textually in
`EventPattern.kt`/`Triggers.kt`), so per AGENTS.md each would warrant its own PR.

**6. `MoveCollectionExecutor.moveToZonesExiledFrom` abandons un-run groups on a pause — no action
needed.** Correctly reasoned and TODO'd. Confirmed there are exactly two pause paths in
`moveToZone`: `CardOrder.ControllerChooses` + `Zone.LIBRARY` (`:318`) and an Aura needing an enchant
target (`:557`, CR 303.4f). Both are ordered last, and the shipped facade's `MoveCollectionEffect`
uses the default `CardOrder.Preserve`, which rules out the library pause entirely — so it is
unreachable through the public API, and the failure mode is inert (cards sit in exile) rather than
corrupting. The real fix is separate engine work: return a continuation capturing the remaining
`orderedZones` instead of passing `moveToZone`'s own continuation straight up. Leave the TODO.

### Gate

Run on this tree at `641f4bc82b`, with `origin/main` already an ancestor.

**Focused on what the diff reaches** — `:mtg-sdk:test` and `:mtg-sets:test` in full, the nine touched
engine and scenario classes, and the two existing `revealHandAndExileChosen` callers:

| Scope | Result |
|---|---|
| `:mtg-sdk:test` + `:mtg-sets:test` (full) + 9 touched classes + Cruelclaw's Heist / Soul Search | **892 passed, 0 failed** |

That includes `CardDefinitionSnapshotTest` (320 cases, MSH golden now matching), `FacadeBoundaryTest`,
`PatternsNamespaceTest`, `CardLintTest` and `SuccessCriterionValidationTest`.

**Full `test` across every module: 3 failures, all environmental.**

| Test | Failure |
|---|---|
| `AbuJafarTest > Abu Ja'far blocking and dying …` | `TimeoutCancellationException` @ 120000 ms |
| `AccursedCentaurTest > Accursed Centaur ETB requires sacrificing …` | `TimeoutCancellationException` @ 120000 ms |
| `AIPlayerTest > AI can evaluate board state` | `TimeoutCancellationException` @ 120000 ms |

None is an assertion failure — all three are the 120 s Kotest coroutine timeout, hit while four
concurrent test JVMs drove the load average to ~16 on an 11 GiB box. **All three pass standalone on
this same tree**, re-run individually after the box went quiet:

| Re-run | Result |
|---|---|
| `AbuJafarTest` | 3 passed |
| `AccursedCentaurTest` | 3 passed |
| `AIPlayerTest` | 17 passed |

None of the three is reachable from this diff (an Antiquities dies-trigger, an ETB sacrifice prompt,
and the AI board evaluator), and all three have flaked the same way in earlier full runs on this
box — `build/pr/u11-recheck-{AbuJafarTest,AccursedCentaurTest,AIPlayerTest}.log` record the same
timeout-then-pass-standalone pattern.

**Not covered:** `web-client` (no TypeScript change in this diff), and no e2e or manual playthrough.

PR #1856, opened 2026-08-16
Four MSH cards: a repeatable optional payment, per-object coloured pips, a re-evaluated base-P/T set, and one CR 302.6 gate

bolav:
Four Marvel Super Heroes cards, each paired with the minimum engine vocabulary it actually needed.
In every case that turned out to be an extension of an existing type rather than a new one — no new
`Effect`, no new `Modification`, no new `StaticAbility`.

MSH goes 262 → 266 of 276.

## The cards

| Card | Oracle line that needed engine work |
|---|---|
| **Ms. Marvel, Kamala Khan** `{2}{U}` | "Until end of turn, Ms. Marvel gains 'Ms. Marvel's base power is equal to the number of cards in your hand.'" |
| **Namor the Sub-Mariner** `{1}{U}{U}` | "Whenever you cast a noncreature spell with one or more blue mana symbols in its mana cost, create that many 1/1 blue Merfolk creature tokens." |
| **Shang-Chi, Master of Kung Fu** `{1}{G}` | "You may activate abilities of creatures you control as though those creatures had haste." |
| **Hawkeye, Master Marksman** `{1}{R}` | "…you may pay {1} up to three times. When you do, choose up to that many —" |

Everything else on all four cards was existing vocabulary.

## The four primitives

**1. `SetBaseStatsEffect(reevaluateContinuously = true)`** — a flag on the existing layer-7b atom
that carries the `DynamicAmount` into the projection instead of snapshotting it at resolution, so a
quoted "base power is equal to …" clause keeps tracking. It lowers onto the *existing*
`SetPowerToughnessDynamic` modification, whose `power`/`toughness` became independently nullable so
"base **power** is equal to …" leaves printed toughness alone.

That clause is **not** a CDA — CR 604.3a fails on criterion 2 (printed on the card it affects) and
criterion 4 (not an ability an object grants to itself) — so it belongs in layer 7b per CR 613.4b,
with the timestamp of the grant. Four limits of the flag are documented on the effect; the
projection-scope one is enforced at card load (below).

**2. `ManaCost.coloredSymbolCount(colors)`** — extracted from the devotion evaluator (deleting
`DynamicAmountEvaluator.manaSymbolColors`) and hung off a new `ManaSymbol.colors`, then shared by
both new per-object reads so they can never disagree:

- `CardPredicate.ColoredManaSymbolsAtLeast(colors, min)` — the trigger's "one or more blue mana
  symbols in its mana cost" gate, via `GameObjectFilter.coloredManaSymbolsAtLeast`. Also fits
  Omnath, Locus of All at `min = 3` over all five colours.
- `EntityNumericProperty.ColoredManaSymbolCount(colors)` — the "that many" count, via
  `DynamicAmounts.coloredManaSymbolsOf(entity, …)`.

Hybrid and Phyrexian pips count for their colour(s) (CR 107.4e/f); generic, `{C}` and `{X}` count
for none whatever X was announced (CR 107.4b). This reads the **printed** cost (CR 202.1/202.1a) —
additional costs, increases and reductions only build the spell's *total* cost (CR 601.2f), so none
of them move the count. Face-down objects have no mana cost (CR 708.2a) and never match. Wired into
all five object-evaluation sites: `PredicateEvaluator` (including the cast-record path, which
returns false — a record stores the resolved mana value, not the printed cost), `TriggerMatcher`,
`AffectsFilterResolver`, `CastZoneResolver`, `CostCalculator`.

**3. `SummoningSicknessRules.blocksTapOrUntapCost`** — CR 302.6's ability half, in one place. The
open-coded `has<SummoningSicknessComponent>() && !hasKeyword(HASTE)` was duplicated at fourteen
sites across the mana solver, both ability enumerators, the cost helpers and the activation handler.
CR 302.6 gates a creature's `{T}`/`{Q}` abilities *and* its ability to attack on one condition;
haste (CR 702.10b/c) lifts both, and Shang-Chi's
`AbilityFlag.MAY_ACTIVATE_ABILITIES_AS_THOUGH_HASTY` lifts **only the ability half** — so it is
deliberately not `Keyword.HASTE`, and combat's `AttackRestrictionRules` keeps its own plain-haste
check and never consults the shared rule. `SummoningSicknessGateEnforcementTest` fails the build if
a new open-coded gate (`has<…>` or `get<…>`) appears outside an allowlist of eight files, each
annotated with why it is not a tap gate. The handler's `{Q}` half was also brought in line with the
enumerator, bare and inside a `Composite`.

**4. `PayManaCostRepeatedlyEffect`** (`Effects.PayRepeatedly(cost, upTo)`) — "pay {1} up to three
times", read back with `DynamicAmounts.timesPaid()`. Two properties worth calling out:

- **The offered ceiling is affordability- and colour-aware**, tested as `cost * n` through the
  *same* auto-tap predicate the payment itself uses (new `canAutoPayManaCost`), deliberately **not**
  `ManaSolver.canPay`. `canPay` counts mana the auto-tapper refuses to spend on the player's behalf
  (sacrificing a Treasure, Springleaf Drum's sub-cost), so gating with it would offer a count the
  payment then errors on. Because both go through one predicate, the prompt can never offer a
  payment that fails mid-resolution.
- **The floor is one repetition, not zero.** Declining belongs to the wrapper —
  `ReflexiveTriggerEffect(optional = true)` for "you **may** pay … **when you do**" (CR 603.12), or
  `Gate.MayPay` for "if you do" — so the two questions stay one "do you want to?" and one "how
  many?". Per CR 603.12a the reflexive half triggers **once**, not once per repetition, which is why
  the repetitions are charged as a single `cost * n` payment. A cap of exactly 1 pays with no prompt.

The count is published to the resolution pipeline rather than as an X because an `EffectResult` has
no X channel — a sub-effect hands back collections, stored numbers and chosen values, and that is
the whole list.

## Tests

One test file per card, plus mechanic-level suites for the primitives:

- `HawkeyeMasterMarksmanScenarioTest` (8), `MsMarvelKamalaKhanScenarioTest` (6),
  `NamorTheSubMarinerScenarioTest` (16), `ShangChiMasterOfKungFuScenarioTest` (7)
- `PayManaCostRepeatedlyTest` (12) — the affordability cap, its colour-awareness, the Treasure
  exclusion, both decline paths, the count surviving the CR 603.12 round-trip, the no-prompt shortcut
- `SetBaseStatsContinuousScenarioTest` (10) — snapshot vs re-evaluated, counters and pump on top,
  duration expiry, the granter leaving play, two competing 7b sets in both timestamp orders
- `ActivateAbilitiesAsThoughHastyTest` (10) — both `{T}` and `{Q}`, the attack half staying shut,
  printed haste as the control, the permission leaving play mid-turn
- `ColoredManaSymbolCountTest` (11), `ManaCostColoredSymbolCountTest` (22),
  `CostEnumerationUtilsTest` (+3), `SummoningSicknessGateEnforcementTest` (1)

Every CR number cited in code, KDoc or docs was checked against the Comprehensive Rules. One
pre-existing citation was wrong and is fixed in passing: face-down objects having no mana cost is
CR 708.2a, not 711.4.

## Review round

This branch was reviewed and the findings applied in `Address review: …`:

- The projection-scope guard was a resolution-time `require()` in `SetBaseStatsExecutor` while its
  KDoc called it a load error. `contextScopedReferenceIn` and its deny list moved to
  `mtg-sdk/.../scripting/values/ProjectionScopedAmounts.kt`, and `SetBaseStatsEffect.init` now runs
  it, so an amount the projector cannot re-evaluate fails as the `cardDef` is built rather than
  mid-game. Gated on the flag, so the JSON scan costs nothing in the default snapshot mode.
- The hygiene test matched only `has<…>`, so the same gate written `get<…>() != null` bypassed it in
  any file. Both spellings now match.
- `ActivatedAbilityEnumerator` gated a bare `AbilityCost.Untap` but not one inside a `Composite`,
  so it could offer a `{Q}` activation the handler's re-check then rejected. Closed.
- Added a test pinning "up to that many" as a ceiling rather than a quota (pay twice, fire one
  arrow, decline the second) — the KDoc asserted this and nothing tested it.
- Documented a fourth limit of `reevaluateContinuously`: the clause is a layer-7b floating effect,
  not an ability the creature has, so `LoseAllAbilities` cannot strip it.

`docs/card-sdk-language-reference.md` is updated for every SDK addition and for the corrections
above.

## Verification

**12,702 tests, 0 failures, 0 errors** across `mtg-sdk`, `rules-engine`, `mtg-sets`, `game-server`
and all nine per-era scenario modules — run against the final tree. Per-module numbers are in the
review comment on this PR, along with what the gate does **not** cover: the `web-client` enum
addition is unverified by any TypeScript tool in this environment, and `just check-card-printing`
could not run because the `justfile` on `main` does not parse under `just` 1.36.0 (all four cards
are first printings with a single Scryfall printing each).

### DIFF
--- a/backlog/sets/marvel-super-heroes/cards.md
+++ b/backlog/sets/marvel-super-heroes/cards.md
@@ -3,7 +3,7 @@
 **Set Size:** 276 booster cards (collector numbers 1-276; basic lands 277-296 and showcase
 variants 297+ are excluded)
 **Release Date:** June 26, 2026
-**Implemented:** 262 / 276
+**Implemented:** 263 / 276
 - [x] Agent 13, Sharon Carter
 - [x] Agent Maria Hill
 - [x] Agent of Atlas
@@ -70,7 +70,7 @@ variants 297+ are excluded)
 - [ ] Leader, Super-Genius
 - [x] Loki, God of Mischief
 - [x] Mister Fantastic, Reed Richards
-- [ ] Ms. Marvel, Kamala Khan
+- [x] Ms. Marvel, Kamala Khan
 - [x] Multiversal Incursion
 - [ ] Namor the Sub-Mariner
 - [x] Pym Particles
--- a/backlog/sets/marvel-super-heroes/cards.md
+++ b/backlog/sets/marvel-super-heroes/cards.md
@@ -3,7 +3,7 @@
 **Set Size:** 276 booster cards (collector numbers 1-276; basic lands 277-296 and showcase
 variants 297+ are excluded)
 **Release Date:** June 26, 2026
-**Implemented:** 263 / 276
+**Implemented:** 264 / 276
 - [x] Agent 13, Sharon Carter
 - [x] Agent Maria Hill
 - [x] Agent of Atlas
@@ -72,7 +72,7 @@ variants 297+ are excluded)
 - [x] Mister Fantastic, Reed Richards
 - [x] Ms. Marvel, Kamala Khan
 - [x] Multiversal Incursion
-- [ ] Namor the Sub-Mariner
+- [x] Namor the Sub-Mariner
 - [x] Pym Particles
 - [x] Rewrite History
 - [x] Secret Invasion
--- a/backlog/sets/marvel-super-heroes/cards.md
+++ b/backlog/sets/marvel-super-heroes/cards.md
@@ -3,7 +3,7 @@
 **Set Size:** 276 booster cards (collector numbers 1-276; basic lands 277-296 and showcase
 variants 297+ are excluded)
 **Release Date:** June 26, 2026
-**Implemented:** 264 / 276
+**Implemented:** 265 / 276
 - [x] Agent 13, Sharon Carter
 - [x] Agent Maria Hill
 - [x] Agent of Atlas
@@ -190,7 +190,7 @@ variants 297+ are excluded)
 - [x] Rick Jones, Destined Sidekick
 - [x] Savage Land Dinosaur
 - [x] Serpent Specialist
-- [ ] Shang-Chi, Master of Kung Fu
+- [x] Shang-Chi, Master of Kung Fu
 - [x] She-Hulk, Jade Defender
 - [x] Super Strength
 - [x] The Thing, Ben Grimm
--- a/backlog/sets/marvel-super-heroes/cards.md
+++ b/backlog/sets/marvel-super-heroes/cards.md
@@ -3,7 +3,7 @@
 **Set Size:** 276 booster cards (collector numbers 1-276; basic lands 277-296 and showcase
 variants 297+ are excluded)
 **Release Date:** June 26, 2026
-**Implemented:** 265 / 276
+**Implemented:** 266 / 276
 - [x] Agent 13, Sharon Carter
 - [x] Agent Maria Hill
 - [x] Agent of Atlas
@@ -133,7 +133,7 @@ variants 297+ are excluded)
 - [x] Death to Our Enemies
 - [ ] Evil's Thrall
 - [x] Fin Fang Foom
-- [ ] Hawkeye, Master Marksman
+- [x] Hawkeye, Master Marksman
 - [x] Hawkeye, Young Avenger
 - [x] Hawkeye's Bow
 - [x] Hex Magic
--- a/backlog/sets/marvel-super-heroes/mechanics.md
+++ b/backlog/sets/marvel-super-heroes/mechanics.md
@@ -610,21 +610,32 @@ gate lands inside the "may" rather than around the whole composite.
 > end of turn, Ms. Marvel gains "**Ms. Marvel's base power is equal to the number of cards in your
 > hand.**"
 
-The granted clause is a **characteristic-defining ability**: base power must keep tracking hand size
-for the rest of the turn. `Effects.SetBasePower` is a one-shot resolution-time *set* —
-`SetBaseStatsEffect` documents its `power` as "evaluated at resolution time" and is deliberately
-distinct from the projector's `SetPowerToughnessDynamic`, which is "re-evaluated per affected entity
-at projection time". So her power froze at whatever the hand was when the trigger resolved.
-Reproduced: hand 8 → power 11 (8 base + Giant Growth's 3); after a draw, hand 9 → power still 11.
+Base power must keep tracking hand size for the rest of the turn. `Effects.SetBasePower` used to be a
+one-shot resolution-time *set* — `SetBaseStatsEffect` documented its `power` as "evaluated at
+resolution time" — so her power froze at whatever the hand was when the trigger resolved. Reproduced:
+hand 8 → power 11 (8 base + Giant Growth's 3); after a draw, hand 9 → power still 11.
 
-**Implemented then removed from this branch.** The right shape is
-`Effects.GrantStaticAbility(<power-only dynamic CDA>, EffectTarget.Self, Duration.EndOfTurn)` —
-`GrantStaticAbility` already exists, but there is no power-only dynamic CDA to hand it:
-`SetBasePowerToughnessDynamicStatic` (`mtg-sdk/.../scripting/StatsStaticAbilities.kt`) sets **both**
-stats from one `DynamicAmount`, which would clobber her printed toughness of 4. Needed: a
-`SetBasePowerDynamicStatic` mirroring the existing toughness-only `SetBaseToughnessForCreatureGroup`,
-plus its `StaticAbilityHandler` branch onto the projector's existing `SetPowerToughnessDynamic` path.
-Narrow and reusable — any "power is equal to X" grant wants it.
+**Fixed (u35).** `SetBaseStatsEffect` now takes `reevaluateContinuously`; when true the executor
+carries the `DynamicAmount` into the floating effect as a nullable-halves
+`SetPowerToughnessDynamic` modification, which the projector re-reads on every pass. The card is
+`Effects.SetBasePower(EffectTarget.Self, DynamicAmounts.cardsInYourHand(), Duration.EndOfTurn,
+reevaluateContinuously = true)`.
+
+Two corrections to the earlier triage, for anyone reading it as precedent:
+ - The granted clause is **not** a characteristic-defining ability. CR 604.3a requires the ability to
+   be printed on the card it affects (criterion 2) and to not be one the object grants to itself
+   (criterion 4); this one fails both. It is a plain layer 7b effect (CR 613.4b).
+ - `Effects.GrantStaticAbility(<power-only dynamic CDA>, …)` would **not** have worked *as
+   `GrantStaticAbility` is wired*. Grants made that way land in `GameState.grantedStaticAbilities`,
+   which is read at points of use (combat, cast permission, ability legality) and is deliberately
+   never fed into the layer projector — see the KDoc on `GrantedStaticAbility` — so a base-P/T static
+   handed out that way would have compiled and done nothing. A projected route for runtime-granted
+   statics does exist (`StaticAbilityHandler.lowerToContinuousEffectData`, used by
+   `BecomeArtifactExecutor` so The Irencrag's granted `ModifyStats(3,3)` reaches the layer system),
+   and routing a `GrantStaticAbility` variant through it would have worked and would generalise to
+   any quoted static ability. The flag on the existing layer-7b atom is the smaller change for this
+   card; the lowering route is the one to reach for if a card ever needs to hand out a whole quoted
+   static ability.
 
 Her other two lines are fine today: `NoMaximumHandSize` and
 `Triggers.youCastSpellTargeting(Creature.youControl())`.
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1065,17 +1065,53 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
   turn). Pass `Duration.WhileSourceTapped("…")` for the Antiquities "tap-locked" buffs (Ashnod's Battle
   Gear `+2/-2`, Tawnos's Weaponry `+1/+1`): the bonus persists for as long as the source artifact remains
   tapped and the one-way latch drops it when it untaps.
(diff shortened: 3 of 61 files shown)


bolav:
## Review — SDK elegance, correctness, rules accuracy

Reviewed with `origin/main` merged in (it was already an ancestor, so no merge commit was needed).

### Verdict

Behavior is right and the SDK shape is right. Four cards, each paired with the minimum engine
vocabulary it actually needed — and in every case that turned out to be an extension of an existing
type rather than a new one. All four cards match Scryfall Oracle text verbatim; every CR number
cited across code, KDoc and docs checks out. Six findings were raised and all six are applied in
`Address review: load-time projection-scope check, and three gap fixes`.

### What's good

**The CR 302.6 centralization is the strongest change here.** Fourteen open-coded
`has<SummoningSicknessComponent>() && !hasKeyword(HASTE)` sites collapse into
`SummoningSicknessRules.blocksTapOrUntapCost`, and a hygiene test fails the build if a new bypass
appears. The KDoc is honest about the split that motivates it — the flag lifts CR 302.6's ability
half but not its attack half, and `AttackRestrictionRules` deliberately keeps reading plain haste.
`ManaSolver` even keeps two separate variables (`canAttack` off plain haste, `tapBlockedBySickness`
off the shared rule) so the auto-tap attacker heuristic cannot be polluted by the grant. The
hygiene test's "Known limit, stated plainly" section names its own per-file granularity rather than
overselling.

**No new `Effect` or `Modification` types where an existing one stretched.**
`reevaluateContinuously` is a flag on the existing `SetBaseStatsEffect` lowering onto the existing
`SetPowerToughnessDynamic`. `ManaCost.coloredSymbolCount` is *extracted from* the devotion
evaluator — deleting `DynamicAmountEvaluator.manaSymbolColors` — and then shared by the predicate
and the amount, so Namor's "does it match" and "how many" sides structurally cannot disagree.

**The `contextScopedReferenceIn` deny list is correct**, which is not obvious and I checked it
against every `@SerialName` in `DynamicAmount`, `EntityReference` and `Player`. The non-obvious
calls are all right: `DistinctColorsManaSpent` / `ManaSpentFromSubtype` / `CraftedMaterials*` read
`context.sourceId` plus entity components, so they are projector-safe and correctly *omitted*;
`Player.ControllerOf` takes a `targetDescription: String` and resolves through
`context.targets.firstOrNull()`, so it is genuinely context-scoped and correctly *denied* — which
also makes the KDoc's "until an affected-entity-controller player reference exists" an accurate
statement rather than an excuse. Choosing JSON traversal over a hand-written `when`, so no nesting
site can be missed, is the right call, and the test pins the classification at depth.

**`canAutoPayManaCost` vs `ManaSolver.canPay`** is a real bug class caught rather than stumbled
into, and the test pins it with actual Treasures.

**Wiring is complete.** All five `CardPredicate` dispatch sites got the new branch. New continuation
registered in `engineSerializersModule`; `CardLinter` knows `storeCountAs` is a NUMBER-space write;
the `AbilityFlag` is mirrored into `web-client` *with* a display name and a comment explaining why
it is excluded from `displayableKeywords`. No new `GameEvent`, so no `ClientEvent.kt` gap.

**`mechanics.md` corrects its own earlier triage** rather than quietly overwriting it, including the
finding that `GrantStaticAbility` lands in `grantedStaticAbilities` — read at points of use, never
projected — so the "obvious" rewrite would have compiled and done nothing. Independently confirmed;
it is why no such rewrite was proposed.

### Findings (all applied)

**1. Important — the projection-scope check claimed to be load-time and was not.**
`StatsEffects.kt` said a context-scoped amount under `reevaluateContinuously = true` "is a load
error", but the check was a `require()` inside `SetBaseStatsExecutor.execute`, throwing
`IllegalArgumentException` mid-game; nothing caught it at load. *Applied:*
`contextScopedReferenceIn` and its deny list moved to
`mtg-sdk/.../scripting/values/ProjectionScopedAmounts.kt` and the check into
`SetBaseStatsEffect.init`, guarded on the flag so the JSON scan costs nothing in snapshot mode. The
test's `shouldThrow` moved from `executor.execute(...)` to the `Effects.SetBasePower(...)` call, plus
a new assertion that a *nested* offender inside `Add`/`Multiply` is caught. (The move also surfaced a
second test importing the function from the old engine package.)

**2. Minor — the hygiene regex only matched `has<…>`.** `get<SummoningSicknessComponent>() != null`
is the same gate written differently and would have passed in *any* file, not just the eight
allowlisted ones — a hole the per-file-granularity note did not cover. *Applied:* pattern is now
`\.(?:has|get)<…>`.

**3. Minor — `Untap` inside a `Composite` was enumerated ungated.** The handler gated `{Q}` in both
shapes; `ActivatedAbilityEnumerator`'s composite sub-`when` handled only `Tap`, so the enumerator
could offer an activation the authoritative re-check then rejected. Documented as latent (no shipped
card uses `Costs.Untap`), but the fix was three lines. *Applied:* the branch is added and the
handler comment no longer describes a divergence.

**4. Minor — a KDoc claim with no test.** Hawkeye's "paying twice and then taking one mode (or none)
is legal" is the whole reason `dynamicChooseCount` is a ceiling with no `dynamicMinChooseCount`, and
the tests only covered pay-N → N-modes. *Applied:* a test pays twice, fires one arrow, asserts the
decline option is offered, that the unchosen arrows did not fire, and that both repetitions were
still paid.

**5. Minor — `LOOP_UNIT.md` shipped stale**, describing a base and a stack the branch no longer had.
*Applied:* dropped from the branch's history entirely; the four original commits were rewritten and
verified patch-identical (`git patch-id --stable`) once that file is excluded.

**6. Minor — one unlisted limit of `reevaluateContinuously`.** Because the clause is a layer-7b
floating effect rather than an ability the creature has, `LoseAllAbilities` cannot strip it. Paper
settles "gains '…'" against Humility by layer-6 timestamp, so the two agree whenever the grant is
the later effect — the common case — and diverge only when ability-removal lands afterwards.
*Applied:* documented as a fourth limit in the KDoc and the language reference, pointing at
`StaticAbilityHandler.lowerToContinuousEffectData` as the route for a whole quoted static.

### Rules accuracy

Every CR number in the diff was verified against the Comprehensive Rules, including the ones easiest
to misremember: **302.6** (summoning sickness, both halves), **702.10b/c** (haste lifting each),
**107.4b/c/e/f** (generic/`{X}`, colorless, hybrid-is-all-its-colors, Phyrexian), **700.5**
(devotion, including the two-colour clause that makes a dual pip count once), **613.4a/4b** (7a CDA
vs 7b set), **604.3a** criteria 2 and 4, **611.2c/2d**, **208.3a**, **601.2f/2i**, **700.2d**,
**207.2c**, **202.1/202.1a**, **708.2a**. Two worth singling out:

- **CR 603.12a** is quoted correctly for the load-bearing case — "if a resolving spell or ability
  includes a choice to pay a cost multiple times … paying that cost one or more times causes the
  reflexive triggered ability to trigger **only once**." That is exactly why the repetitions are
  charged as a single `cost * n` payment.
- A **pre-existing** citation is fixed in passing: face-down objects having no mana cost is
  **CR 708.2a**, not 711.4.

All four cards match Scryfall Oracle text verbatim (name, cost, type line, P/T, rules text).

### Gate

`scripts/gradle-locked` — **12,702 tests, 0 failures, 0 errors**, all of it against the final tree
(post-fix, post-rebase):

| Module | Classes | Tests | Failures |
|---|---:|---:|---:|
| `mtg-sdk` | 38 | 462 | 0 |
| `rules-engine` | 489 | 2,835 | 0 |
| `mtg-sets` (catalog + snapshots) | 16 | 359 | 0 |
| `game-server` | 67 | 491 | 0 |
| `mtg-sets:1993-1999:tests` | 138 | 408 | 0 |
| `mtg-sets:2000-2002:tests` | 154 | 593 | 0 |
| `mtg-sets:2003-2007:tests` | 165 | 533 | 0 |
| `mtg-sets:2008-2016:tests` | 168 | 456 | 0 |
| `mtg-sets:2017-2022:tests` | 271 | 696 | 0 |
| `mtg-sets:2023:tests` | 400 | 1,266 | 0 |
| `mtg-sets:2024:tests` | 655 | 1,771 | 0 |
| `mtg-sets:2025:tests` | 589 | 1,592 | 0 |
| `mtg-sets:2026:tests` | 432 | 1,240 | 0 |

The four new card suites and every new engine suite are in there: Hawkeye 8, Ms. Marvel 6, Namor 16,
Shang-Chi 7; `PayManaCostRepeatedlyTest` 12, `SetBaseStatsContinuousScenarioTest` 10,
`ActivateAbilitiesAsThoughHastyTest` 10, `ColoredManaSymbolCountTest` 11,
`ManaCostColoredSymbolCountTest` 22, `SummoningSicknessGateEnforcementTest` 1.

Getting there took several attempts, all for environmental reasons worth recording so the next run
doesn't repeat them. This box has ~11 GiB shared across containers while `gradle.properties` asks for
roughly 10 GiB, and `scripts/gradle-locked` does not actually serialize anything here (it needs
`shlock`, which is macOS-only). One full-suite attempt lost its daemon to the OOM killer, leaving
four orphaned JVMs holding 6 GiB. Shrinking the daemon with `-Xmx3g` made things *worse*, because
test executors are forked JVMs whose heap comes from the test task, not the daemon — combined with
`--continue` that put five executors in flight at once and five died with exit 13. What worked was
`--max-workers=2` for the bulk and `--max-workers=1` for the four oldest era modules. Throughout,
`:ai`'s only failures were `TimeoutCancellationException` at the 120 s guard — the contention
signature, never an assertion, and unreachable from this diff. **`:ai` was not re-run after the
fixes**; nothing in the diff touches it.

**Not covered by any gate, stated plainly:**

**Not covered by any gate, stated plainly:**

- The `web-client/src/types/enums.ts` addition is **unverified by any tool**. No TypeScript gate was
  run; the Kotlin green says nothing about it. It is a two-line enum + display-name addition, but it
  is unchecked, not passing.
- `just check-card-printing` could **not** run: the `justfile` on `main` has
  `prune-worktrees DAYS="7" *ARGS:`, which `just` 1.36.0 rejects (variadic after a defaulted
  parameter), so every recipe fails to parse. All four cards are first printings in a 2026 set with a
  single Scryfall printing each, so the risk is low — but the check is unrun, not passed.

PR #1876, opened 2026-08-17
[agent-loop: claude-opus-5[1m]] Connive as a keyword action, heal, a power-up lockout and an extra-activation permission — Leader, Wolverine, Kang, Wonder Man

bolav:
Four Marvel Super Heroes cards, each of which needed one piece of engine or SDK vocabulary that
didn't exist yet, plus a review follow-up commit. MSH 266 → 270 / 276.

| Card | New vocabulary |
|---|---|
| Leader, Super-Genius (#64) | connive becomes a *named, replaceable, observable* keyword action |
| Wolverine, Fierce Fighter (#240) | `HealOtherDamage` — heal, CR 701.69a |
| Kang the Conqueror (#62) | a turn-scoped power-up lockout |
| Wonder Man, Hollywood Hero (#160) | a once-only limit that is *raised* rather than waived |

**This branch supersedes `loop-msh-u18`, `loop-msh-u19` and `loop-msh-u20`.** Those were the original
per-unit branches; the first three commits here are patch-identical to their tips (verified with
`git patch-id --stable`), rebased onto `origin/main` and consolidated because u21's permission
generalizes u20's keyword and its CR 101.2 test needs u20's lockout. Nothing is stacked — this
branch is based directly on `main`.

## Design: generalize, don't duplicate

Each of the three new primitives replaced or extended an existing type rather than sitting beside a
near-identical one.

- **`ModifyExplore` → `ModifyKeywordAction`.** Leader is structurally Twists and Turns (LCI) with
  connive in place of explore — a keyword-action prefix replacement plus a recurring source of that
  action. The two printed cards differ only in *which* action and *what* the prefix is, both already
  carried by `appliesTo` / `prefixEffect`, so the type is one across keyword actions instead of one
  per action. Twists and Turns is migrated in the same change; the prefix lookup both executors
  shared moved to `KeywordActionReplacements`.
- **`IgnoreExhaustActivationLimit` → `ExtraOnceOnlyActivations`.** Exhaust (CR 702.177) and power-up
  (CR 702.193) desugar to the *same* `ActivationRestriction.Once`, and the engine already had a
  permission that lifted it. Rather than add a second, near-identical static on the same three
  activation-legality paths, the existing one gained two axes: `kind` (which keyword — load-bearing,
  or Wonder Man would silently re-arm every exhaust ability on the board) and `extraActivations`
  (`null` waives, N raises by N and sums). Elvish Refueler is the `null` case, Wonder Man the `1`
  case. `ExhaustActivationWaiver` → `OnceOnlyActivationAllowance` follows on the engine side.
- **The Kang lockout is a rider on `TakeExtraTurnEffect`, not a new effect.** "During that turn"
  names the turn *that* effect creates, so when `PreventExtraTurns` (Ugin's Nexus) stops the extra
  turn the rider must not apply either. A sibling effect in the `Composite` could re-check that, but
  only by duplicating a precondition the extra-turn executor legitimately owns.

`HealOtherDamage` is the one genuinely new type: nothing in the SDK removed marked damage, and it is
the only member of the damage family that leaves the *amount* alone. Its engine half,
`DamageUtils.healMarkedDamage`, became the single home the Pyramids shield, regeneration
(CR 701.19a) and the cleanup step (CR 514.2) all route through, replacing three open-coded
`.without<DamageComponent>()` strips.

## Rules corners worth a reviewer's eye

- **CR 510.2** — all combat damage in a step is simultaneous, so the heal fires at most once per
  recipient per step: a double-blocked Wolverine keeps *both* blockers' damage and dies. The
  first-strike step is a separate event, so its damage is healed away — the card's real trick.
- **CR 701.50f** — a permanent connives even when the draw and the discard were both impossible.
  The event is the pipeline's tail, so it lands after the discard decision resolves.
- **CR 614.5** — `replacementsApplied` stops a replacement applying to its own re-issue; two
  separate sources still each contribute their prefix.
- **CR 101.2** — Kang's prohibition beats Wonder Man's permission. The lockout is a flat gate ahead
  of any `ActivationRestriction`, so a raised ceiling has nothing to raise.
- **CR 400.7 / 708.2a** — a body that leaves and returns gets a fresh allowance; a face-down
  permission source grants nothing its card printed, while an ability *granted* to it still applies.

Two limits are documented rather than fixed, both pre-existing in the skip-based extra-turn model
and both called out in `TakeExtraTurnExecutor`'s KDoc: in a pod of three or more `TurnManager`
consumes at most one pending skip per boundary, so `turnNumber + 1` can name an opponent's ordinary
turn; and with no extra-turn queue, a second extra-turn effect resolving after Kang's puts the
lockout on the wrong turn (CR 500.7).

## Review follow-up (last commit)

A review of the first four commits produced three changes, none of them behavioural:

- The CR 708.2a face-down guard and `AbilityActivatedEverComponent`'s legacy-state fallback both
  shipped untested; each now has a test, and each was mutation-checked (breaking the guard reddens
  exactly the new cases and nothing else).
- `checkActivationRestriction` carried both a nullable, defaulted `abilityId` and a required
  `ability`. All eight call sites passed `ability.id` for both, so nothing was wrong — but that
  pairing is the drift the required parameter was added to prevent. The ability is now the single
  source of both.

## Verification

`:rules-engine:test` 2886 · `:ai:test` 504 · `:mtg-sdk:test` 462 · corpus `:mtg-sets:test` 363 ·
scenario suites 1993-1999 408, 2000-2002 593, 2003-2007 533, 2023 1266, 2025 1592, 2026 1257 —
all zero failures. Snapshots reblessed and independently verified purely additive (no line
disappears from `MSH.json` or `SPM.json`); `check-card-printing` clean — MSH is the only printing of
all four cards, so each canonical is in its earliest real set. All four cards match Scryfall Oracle
exactly, including Wonder Man's printed "only . . . once?" joke reminder, which is deliberately not
normalised.

Two 120-second `TestHangGuard` timeouts appeared during parallel runs on a loaded box
(`AbattoirGhoulScenarioTest`, `AgelessSentinelsTest`, plus `AIPlayerTest` and `AbuJafarTest`
earlier); every one passes in under a minute when run alone. They are contention, not this change.

**Not covered by the above:** `game-server`, `gym*`, the `oracle-assay` gates, and `web-client` —
this diff doesn't reach them, but no gate here is evidence about them either.

`docs/card-sdk-language-reference.md` is updated in the same commits as the SDK changes.

<sub>If a local run shows `MtgSetCatalogTest` reporting `Big Score -> [BLC, SNC]` together with a
BLC snapshot mismatch, that is unrelated to this branch: it is an orphaned
`mtg-sets/2024/build/classes/.../blc/cards/BigScoreKt.class` left behind by incremental compilation
when `c8ffc1ce0f` moved that card's canonical to SNC. Deleting the stale class file fixes both, with
no source change.</sub>


bolav:
## Review — u18–u21 (Leader, Wolverine, Kang, Wonder Man)

Reviewed against `docs/sdk-design-principles.md`, with the branch based directly on `origin/main`
(`ac9eebca92`) — no merge commit needed.

### Verdict

The behavior is right and the SDK shape is right. Three times over, this branch made the call the
elegance bar asks for — generalize the existing type rather than add a sibling next to it — and
followed through on the engine side and on the already-shipped cards that used the old shape. No
blocking issues. The two findings that mattered (a rule citation without a test, and a documented
invariant without a test) are fixed in the follow-up commit; the rest are noted below as author's
discretion.

### What's good

- **The three generalizations are carried through, not just started.** `ModifyExplore` →
  `ModifyKeywordAction`, `IgnoreExhaustActivationLimit` → `ExtraOnceOnlyActivations`, and the
  extra-turn lockout as a rider on `TakeExtraTurnEffect` rather than a new effect. The engine
  mirrors each: `collectExploreReplacementPrefixes` → `KeywordActionReplacements`,
  `ExhaustActivationWaiver` → `OnceOnlyActivationAllowance`. Twists and Turns and Elvish Refueler
  are migrated in the same change and both stay green.
- **`HealOtherDamage` clears the new-type bar honestly.** Nothing in the SDK removed marked damage,
  and `PreventDamage` / `CapDamage` / `ModifyDamageAmount` all touch the amount where this one
  deliberately doesn't. The better half is the follow-through: `DamageUtils.healMarkedDamage`
  becomes the single home regeneration, the Pyramids shield and cleanup all route through — a net
  *reduction* in open-coded `.without<DamageComponent>()`, with the battlefield-exit strip correctly
  left out as a CR 400.7 object reset rather than a heal.
- **Two deliberate "make omission a compile error" choices**, both correct and both argued in the
  KDoc: `OnceOnlyAbilityKind` carrying no default (a defaulted two-valued discriminator would let
  `ExtraOnceOnlyActivations(extraActivations = 1)` silently mean the wrong keyword), and
  `checkActivationRestriction`'s `ability` being required.
- **Test quality is unusually high.** Every heal accumulation assertion is paired with a
  no-replacement control creature; CR 510.2's double block and the separate first-strike step are
  both exercised; CR 701.50f is tested with *both* library and hand emptied; CR 101.2 gets a pair
  asserting the exact rejection message; and `EffectWalkerTest` pins an AI-scorer regression (the
  wrapper hiding `DrawCardsEffect` from all 16 conniving cards) that would otherwise have been
  entirely silent.
- **The verification claims hold up under independent checking.** I re-derived the "purely additive"
  snapshot claim by diffing trimmed line multisets for `MSH.json` and `SPM.json` — not one line
  disappears. All 20 CR citations I sampled verify against `MagicCompRules_20260808.txt`. All four
  cards match Scryfall exactly (mana cost, type line, P/T, collector number, rarity, artist, oracle
  text), and MSH is the only printing of each, so canonical placement is right.

### Fixed in the follow-up commit

1. **The CR 708.2a face-down guard had no test.** `OnceOnlyActivationAllowance` suppresses printed
   statics on a face-down permanent while keeping runtime-granted ones — a correct fix, but also a
   real behaviour change against the deleted `ExhaustActivationWaiver` (a face-down Elvish Refueler
   used to grant its waiver and now doesn't), folded into a rename. Both halves now have a test, and
   they fail differently, so an implementation that skipped the whole permanent is caught too.
2. **`AbilityActivatedEverComponent`'s `abilityIds` fallback had no test** despite its KDoc calling
   it load-bearing rather than a dead branch. It now decodes the real pre-`activationCounts` wire
   form, so a rename of the field is caught as well as a dropped fallback.
3. **`checkActivationRestriction` carried `abilityId` and `ability` redundantly**, with the `Once`
   branch reading `ability.id` while `OncePerTurn` / `MaxPerTurn` read `abilityId`, and
   `|| abilityId == null` guards that would silently disable two restrictions if it were ever
   omitted. Folded into `ability`.

Both new tests were mutation-checked: stubbing the face-down guard to `if (false)` reddens both
face-down cases, and dropping the `abilityIds` fallback reddens the legacy-state case and only it.

### Remaining — author's discretion

4. **Persistence back-compat is argued in one place and dropped in two others.** The same branch
   renames `@SerialName("ModifyExplore")` → `"ModifyKeywordAction"` and
   `"IgnoreExhaustActivationLimit"` → `"ExtraOnceOnlyActivations"`, and both types are serialized
   into live `GameState` (`ReplacementEffectSourceComponent`, registered in `Serialization.kt`;
   `GameState.grantedStaticAbilities`). By the exact argument the `activationCounts` KDoc makes
   about games in flight across a deploy, those renames fail deserialization outright — a harder
   failure than the one the fallback prevents. There is repo precedent for bare `@SerialName`
   renames, so this is probably accepted practice; the ask is only that one branch not argue both
   ways at once.
5. **The connive trigger's filter has no mirror test.** `ConniveReplacementTest` proves the
   *replacement* filter resolves "you" to the source's controller; nothing asserts that an
   opponent's creature conniving does **not** fire your `WheneverCreatureYouControlConnives` watcher.
   Same file, same "Connive Watcher" card, one more case.
6. **The permission is read off the activating player's battlefield, not the source's controller's.**
   Fine for the ordinary path — you only activate what you control — but on
   `ActivatedAbilityEnumerator`'s `AnyPlayerMay` path an opponent activating your power-up would
   draw the permission from *their* board, which "each power-up ability of permanents you control"
   doesn't say. Unreachable today (no printed `AnyPlayerMay` power-up); worth one sentence in
   `OnceOnlyActivationAllowance` naming that path, since the existing "no who axis" note only covers
   the ordinary case.
7. **`ConniveEffectExecutor` cites CR 701.50b slightly beyond what it does.** It skips both the
   replacement lookup and the connive event when the subject won't resolve, under a comment citing
   the LKI rule. I checked all 17 `Effects.Connive` call sites — every one is `Self`,
   `TriggeringEntity`, or a bound cast-time target, so I couldn't construct a reachable case. Fine
   as-is; the citation just claims a little more than the code delivers.

### Gate

`:rules-engine:test` 2886 · `:ai:test` 504 · `:mtg-sdk:test` 462 · corpus `:mtg-sets:test` 363 ·
scenario suites 1993-1999 408, 2000-2002 593, 2003-2007 533, 2023 1266, 2025 1592, 2026 1257 —
all zero failures.

Two process notes from running it:

- **A plain `test` invocation can serve `:rules-engine:test` and the era scenario suites as
  UP-TO-DATE**, reporting a green that never executed. Everything above was re-run with `--rerun`,
  and every new test class was confirmed to have a fresh timestamp.
- **Four `TestHangGuard` 120-second timeouts** appeared across parallel runs on a loaded box
  (`AIPlayerTest`, `AbuJafarTest`, `AbattoirGhoulScenarioTest`, `AgelessSentinelsTest`). All four
  pass in under a minute when run alone — contention, not this change.

Not covered by any of it: `game-server`, `gym*`, the `oracle-assay` gates, and `web-client`.

<sub>🤖 Review produced with [Claude Code](https://claude.com/claude-code)</sub>

PR #1879, opened 2026-08-17
[agent-loop: claude-opus-5[1m]] A double-faced-card predicate, CantBeAttackedBy, a capped free-cast loop and end-of-your-next-turn — Nick Fury, Storm, Doom Reigns Supreme, Evil's Thrall

bolav:
Four Marvel Super Heroes cards, each needing one piece of engine or SDK vocabulary that didn't
exist yet, plus a review-fix commit. MSH 270 → 274 / 276.

| Card | New vocabulary |
|---|---|
| Nick Fury, Agent of S.H.I.E.L.D. (#25) | `CardPredicate.IsDoubleFaced` — and Rule 712 face tracking on non-cast battlefield entries |
| Storm, Windrider (#230) | `CantBeAttackedBy` generalized from `CantBeAttackedWithout`; trigger-time cast-target capture |
| Doom Reigns Supreme (#96) | `maxCasts` — the capped "cast up to N of them" free-cast loop |
| Evil's Thrall (#128) | `Duration.EndOfYourNextTurn` |

Nothing is stacked — this branch is based directly on `main`.

## Design: generalize or parameterize, don't add a type

Each of the four additions either replaced an existing type with a more general one, added a
parameter to an existing one, or read state no primitive could read.

**`CantBeAttackedWithout` → `CantBeAttackedBy` (a deletion, not an addition).** The old static
carried a `requiredKeyword` field plus an optional narrowing filter, so "creatures **with** flying
can't attack you" needed a third shape. The replacement drops the keyword field entirely and lets
the filter carry the whole clause, positive or negative:

```kotlin
CantBeAttackedBy(GameObjectFilter.Creature.withKeyword(Keyword.FLYING))                    // Storm
CantBeAttackedBy(GameObjectFilter.Creature.withoutKeyword(Keyword.FLYING))                 // Form of the Dragon
CantBeAttackedBy(GameObjectFilter.Creature.sharingChosenColorWithSource()
    .withoutKeyword(Keyword.FLYING))                                                       // Teferi's Moat
```

Three cards, one static, three filters. The rule (`CantBeAttackedByDefenderRule`) also picked up two
corrections while it was open: it now fires only when the chosen defender **is** the restricting
permanent's controller — attacking a planeswalker that player controls stays legal, per Form of the
Dragon's 2014-02-01 ruling (CR 506.3 / 508.1b) — and it skips face-down permanents, which have no
abilities (CR 708.2). Both directions are pinned by tests.

**`maxCasts` is a parameter, not an effect.** "You may cast up to two spells from among the exiled
cards" is `CastAnyNumberFromCollectionWithoutPayingCostEffect` with a bound, reached through
`Effects.CastUpToNFromCollectionWithoutPayingCost(from, 2)`. The budget rides on the existing
continuation, so a cast that pauses for targets / X / modes resumes with the count already spent.
The cap is spent on a cast that *initiates*, not on the pick: a chosen card whose required target
has no legal choice can't be cast at all (CR 601.2c), so it stays in exile and the count is
untouched — though it is dropped from the pool, so the loop can't re-offer it forever.

**`CardPredicate.IsDoubleFaced` is modelled on `HasAdventure`,** a printed whole-card layout
characteristic (CR 712.1) stamped onto `CardComponent.isDoubleFaced` at entity creation. It reads
the same in every zone and **stays true on the back face**, because it is carried across a flip
rather than re-derived from the face that's up — a back face's own definition has no back face of
its own. Every copy path is decided explicitly rather than inherited by accident:

- **Tokens never match.** A token is not a card (CR 111.1) and layout is not a copiable value
  (CR 707.2), so all four token-copy executors clear the flag. Such a token *is* a double-faced
  token and can still transform (CR 707.8a) — that is `DoubleFacedComponent` face-tracking state,
  a different question.
- **A permanent that becomes a copy keeps its own answer**, which is exactly CR 712.9's pair of
  examples: a Clone copying Wildblood Pack can't transform, a Kruin Outlaw copying Elite Vanguard
  can.

**`Duration.EndOfYourNextTurn`** is strictly longer than the existing `UntilYourNextTurn`, which
ends at the *beginning* of your next turn. The window is a turn-number floor
(`ActiveFloatingEffect.expiresAfterTurn = turnNumber + 1`) plus a controller guard, which is what
keeps it right across extra turns, skipped turns and multiplayer seats — the same mechanism
`MayPlayPermission.expiresAfterTurn` already uses for the play-permission wording. It is wired for
the floating-effect path only; the KDoc and the language reference say so, and say what it would
take to widen it.

## An engine bug found on the way

`DoubleFacedComponent` was only ever stamped by the **cast** pipeline. A double-faced card put onto
the battlefield by an *effect* — reanimated, fetched out of a library, returned from exile — arrived
with no face tracking at all, so a printed `Transform` on it was silently a no-op. The shared
`stampDoubleFacedFrontFace` helper now runs on every non-cast entry via
`ZoneTransitionService.applyBattlefieldEntry`, and from `PlayLandHandler` for the land special
action that deliberately bypasses that funnel (Balamb Garden, SeeD Academy). Face-down entries are
excluded (CR 708.2), and `DayNightService`'s hand-rolled duplicate of the same stamp was folded into
the helper. `DoubleFacedEntryAndPredicateTest` covers reanimation, the land play, and the
shipped Balamb Garden case.

Separately, `CantBeBlockedByRule` no longer skips the host when the host is itself the attacker: the
printed read only ever picked up `Scope.Self` clauses, so a battlefield-scoped group clause on a
creature's own card ("creatures you control can't be blocked by …") never applied to that creature.
`excludeSelf` is honoured, so an "other creatures you control" wording still leaves the host out.
Wall Crawl is the only other battlefield-scoped user and is an Enchantment, so Storm is the only
card whose behaviour changes.

## Trigger-time target capture

`Triggers.youCastSpellTargeting(filter)` now records the targets that satisfied its own gate into
`IterationSpace.TRIGGER_CAPTURED_COLLECTION`, the same engine-seeded slot a batched ETB payoff
reads. So "those creatures gain flying until end of turn" is a plain `ForEachInCollectionEffect`
with no effect that reads the stack, and the clause deciding *whether* it triggers and the clause
deciding *what it acts on* are one computation that can't drift.

The capture is a snapshot taken when the ability triggers, which is the load-bearing part: the
trigger sits on top of the spell that caused it and exists independently of it (CR 113.7a), so an
opponent with priority can answer that spell first. `SpellCastTargetCaptureTest` pins both ways it
can be answered — countering it (a live read would find nothing, because countering strips
`TargetsComponent`) and retargeting it (a live read would find the wrong creature).

## Verification

- `just test` equivalent (`scripts/gradle-locked test`, since `just` can't currently parse the
  `justfile` on this box) — green apart from three coroutine-timeout flakes caused by the run
  oversubscribing the machine (`AbuJafarTest`, `AccursedCentaurTest`, `AIPlayerTest`); each re-run
  isolated and green.
- That run aborts scheduling once a module fails, so the newer era test modules were run separately:
  `:mtg-sets:2026:tests:test :mtg-sets:2003-2007:tests:test :mtg-sets:1993-1999:tests:test` →
  BUILD SUCCESSFUL, all four new card scenario tests plus `FormOfTheDragonTest` green.
- After the review-fix commit: `:rules-engine:test :mtg-sets:test :mtg-sdk:test` → BUILD SUCCESSFUL,
  0 failures, including both MSH snapshot tests.
- All CR numbers cited in code, comments and docs verified against `MagicCompRules_20260808.txt`.
- All four cards' oracle text, cost, type line, P/T, rarity, collector number, artist and flavour
  text match Scryfall exactly; all four are MSH-only printings, so each canonical lands in its
  earliest real printing.

Not covered by any gate: the four `manual-scenarios/*.json` dev fixtures. No TypeScript changed.

## Known gap, not addressed here

`CantBeAttackedBy`'s description renders through a small pluralizer that pluralizes the *first* type
noun it finds, so a filter whose description leads with a compound type phrase would render
"Artifacts creature can't attack you". No shipped filter hits it — every current caller leads with
"creature" — and it only affects the user-visible attack-rejection message, so it is left for a
follow-up rather than widening this branch.


bolav:
## Review — `loop-msh-u39`

Reviewed against `origin/main` (already an ancestor of this branch's tip, so no merge was needed and
no merge commit was produced). Primary lens: SDK elegance.

**Verdict.** Behaviour is right and the SDK shape is close to exemplary — the diff *removes* a
specialized type, adds one parameter where a new effect type was tempting, and the three genuinely
new primitives all clear the novelty bar. One blocking problem was found and has since been fixed in
`b7f106ba` (the last commit on this branch); the remaining notes are recorded below for the record.

---

### What's good

- **`CantBeAttackedWithout` → `CantBeAttackedBy` is the model refactor for this codebase.** It
  deletes the specialized `requiredKeyword` field and lets the filter carry the whole clause, so
  Form of the Dragon, Teferi's Moat and Storm become three *filters* over one static instead of
  three shapes. It also picked up the `applyTextReplacement` override the old type was missing.
- **`maxCasts` is a nullable parameter on the existing effect**, not a `CastUpToNEffect` plus
  executor, with a facade that rejects a non-positive value.
- **`isDoubleFaced` follows the `hasAdventure` precedent**, and every copy path is reasoned
  individually: the four token-copy executors clear it (CR 111.1 / 707.2 / 707.8a), the clone path
  keeps `originalCardComponent.isDoubleFaced` — precisely CR 712.9's Clone/Kruin Outlaw pair — and
  `buildCardComponentForDfcFace` carries it across a flip because a back face's own definition has
  no back face. All six `CardComponent` construction sites were updated, checked against the
  existing `hasAdventure` sites.
- **`stampDoubleFacedFrontFace` fixes a real pre-existing bug:** a double-faced card put onto the
  battlefield by an effect had no `DoubleFacedComponent`, so "transform it" was silently a no-op.
  The extraction also folded `DayNightService`'s duplicate stamp into it.
- **Dropping the `hostId == ctx.attackerId` skip in `hostScopedRestrictions` is correct** — the
  printed read filters on `Scope.Self`, so a battlefield-scoped clause on the attacker's own card
  never reached it. Blast radius verified: Wall Crawl is the only other battlefield-scoped user and
  is an Enchantment, so Storm is the only card whose behaviour changes.
- **`CantBeAttackedByDefenderRule` now fires only for player defenders**, matching Form of the
  Dragon's 2014-02-01 ruling, and skips face-down permanents (CR 708.2). Both directions tested,
  including a Teferi's Moat planeswalker case.
- **Trigger-time target capture reuses `TRIGGER_CAPTURED_COLLECTION`** rather than inventing a
  channel, and doesn't clobber an event-supplied capture (`capturedTargets ?: ctx.capturedEntityIds`).
- **Every CR number in the diff checks out** against `MagicCompRules_20260808.txt` — 712.1/.8a/.9/
  .12/.13/.14, 707.2/.8a, 111.1, 508.1b/c, 506.3, 514.2, 601.2c, 708.2, 305.1, 113.7a, 701.27a/c/g,
  302.6, 603.12. No mismatches.
- **All four cards match Scryfall exactly** (oracle text, cost, type line, P/T, rarity, collector
  number, artist, flavour), and all four are MSH-only printings, so canonical placement is right.
- **Test quality is high.** The Time Warp extra-turn case, "a Villain that only ties doesn't
  extend", mana-value-0 versus `MAX`-over-empty, the uncastable pick that doesn't spend a cast, the
  paused-cast budget round trip, and ETB-fires-on-the-front-face-then-flips are real rule corners,
  not filler.

---

### Blocking — fixed in `b7f106ba`

**The committed MSH golden could not be regenerated by the serializer.** Doom Reigns Supreme's block
in `mtg-sets/src/test/resources/snapshots/cards/MSH.json` carried `"triggerCondition"` where
generated output has `"triggerRestriction"`, and that was the *only* difference between the golden
and `build/snapshots-actual`. `triggerCondition` is a computed `val` on `TriggeredAbility`
(`TriggeredAbility.kt:170`, derived from `interveningIf` + `triggerRestriction`), not a serialized
constructor property, and no other set's golden contains the key — so that hunk had been hand-edited
rather than blessed. `:mtg-sets:test` was red on it. Re-blessed; the regenerated file differs by
exactly that one key, and no other set moved.

---

### Important — one fixed, one open

**Fixed in `b7f106ba`: the split-shield `timestamp`.** `CombatContinuationResumer`'s rewrite to
`originalShield?.copy(...)` now restates `timestamp = timestamp` explicitly. *Correction to the
original review note:* this was never a behaviour change — `timestamp` is derived as
`originalShield?.timestamp ?: state.timestamp` a few lines above, so the copy already carried the
same value. Restating it pins the invariant if that derivation ever moves.

**Open: `pluralAttackerSubject` pluralizes the wrong word for a compound type phrase.**
`mtg-sdk/.../scripting/CombatStaticAbilities.kt` takes `indexOfFirst`, so
`GameObjectFilter.ArtifactCreature.description` ("artifact creature") renders as
"**Artifacts creature** can't attack you" in the attack-rejection message. No shipped card hits it —
every current caller leads with "creature" — but the next `CantBeAttackedBy` with a compound type
will. It is also the SDK's third hand-rolled pluralizer, alongside `pluralizeHeadNoun`
(`EventPattern.kt:2834`) and `DynamicAmount.pluralize`; neither of those handles trailing
qualifiers, so a third helper is defensible, but it should delegate the pluralization and target the
end of the leading type run:

```kotlin
private fun pluralAttackerSubject(filter: GameObjectFilter): String {
    val words = filter.description.split(" ")
    val first = words.indexOfFirst { it.lowercase() in PLURALIZABLE_TYPE_NOUNS }
    if (first < 0) return filter.description.replaceFirstChar { it.uppercase() }
    // "artifact creature without flying" -> pluralize the *last* word of the leading type run.
    var head = first
    while (head + 1 < words.size && words[head + 1].lowercase() in PLURALIZABLE_TYPE_NOUNS) head++
    return words.mapIndexed { i, w -> if (i == head) pluralizeHeadNoun(w) else w }
        .joinToString(" ").replaceFirstChar { it.uppercase() }
}
```

Reusing `pluralizeHeadNoun` also lets `PLURALIZABLE_TYPE_NOUNS` drop its "all regular `+s` plurals"
caveat. Deliberately left for a follow-up rather than widening this branch.

---

### Minor — all fixed in `b7f106ba`

- **`stampDoubleFacedFrontFace`'s KDoc claimed a token no-op the function doesn't implement.** It
  guards only on an existing `DoubleFacedComponent` and on the definition having a back face. The
  KDoc now documents the real guards and why there deliberately *is* no token guard: face tracking
  is a question about the object (a token copy of a double-faced permanent is a double-faced token
  and can transform, CR 707.8a / 712.9), while the card-level question is
  `CardComponent.isDoubleFaced`, which the token-copy executors clear.
- **The "retarget" half of the capture claim was untested.** Both the `capturedCastTargets` KDoc and
  the language reference cite "counters **or retargets**"; only countering was exercised. Added
  `retargeting the triggering spell in response does not move the capture`, which redirects the
  spell out from under its own trigger and asserts both halves — the spell's `TargetsComponent`
  moves to the new creature, the captured target does not.
- **`maxCasts` with `payManaCost = true` was undocumented.** The facade only offers the free form,
  but the raw constructor allows both; there the "did the cast initiate" precondition asks about
  legal targets, not affordability, so a pick abandoned for want of mana would still spend one of
  the N. Documented in the effect KDoc and the language reference.

---

### Unrelated observation, not addressed

While writing the retarget test I found that after a **permanent** retarget lands, the redirected
spell resolves without affecting either creature (the original stayed at base power plus its
counter, the new target was untouched, the spell went to the graveyard). The redirect mechanism
itself works — the spell's `TargetsComponent` is correctly rewritten, which is what the new test
asserts — so this is downstream of it. The existing coverage for `Effects.ChangeTarget`
(`ReturnTheFavorScenarioTest`) only exercises **player** retargets. None of this branch's code is
involved, so it is pre-existing; worth its own investigation.

---

### Gate

- `scripts/gradle-locked test` (the `just test` equivalent — `just` can't currently parse the
  `justfile` on this box): green apart from three coroutine-timeout flakes from the run
  oversubscribing the machine (`AbuJafarTest`, `AccursedCentaurTest`, `AIPlayerTest`), each re-run
  isolated and green, plus the MSH golden above.
- That run aborts scheduling once a module fails, so the newer era test modules were run separately:
  `:mtg-sets:2026:tests:test :mtg-sets:2003-2007:tests:test :mtg-sets:1993-1999:tests:test` →
  BUILD SUCCESSFUL. All four new card scenario tests green, plus `FormOfTheDragonTest`,
  `TeferisMoatScenarioTest`, and the four new engine tests (`CantBeAttackedByRestrictionTest`,
  `CantBeBlockedByGroupScopeTest`, `DoubleFacedEntryAndPredicateTest`, `SpellCastTargetCaptureTest`).
- After the fixes: `:rules-engine:test :mtg-sets:test :mtg-sdk:test` → BUILD SUCCESSFUL, 0 failures,
  including both MSH snapshot tests.

Not covered by any gate: the four `manual-scenarios/*.json` dev fixtures. No TypeScript changed, so
there is no web-client gap to report.

PR #1886, opened 2026-08-17
Finish Marvel Super Heroes: the last cards, then prove the set is done

bolav:
# Finish Marvel Super Heroes: the last cards, then prove the set is actually done

MSH's backlog read `276 / 276`. That is a claim that every name has *a* `CardDefinition` — not that
the definitions carry the right power, the right oracle text, or that the cards work when played.
This branch closes the set out and then closes that gap.

Seven commits: three finish the remaining cards (World War Hulk, Baron Helmut Zemo, and a dynamic
amount rendering fix), four are the verification pass and the archive.

```
Set:              MSH — Marvel Super Heroes
Completeness:     276/276 draft cards (collector numbers 1–276), 0 missing,
                  6 reprint rows, 30 basic-land printings
Field fidelity:   270 registered CardDefinitions × 13 fields vs a fresh Scryfall dump — 0 discrepancies
                  9 fixed on the way, all display-class:
                    4 × reminder text inlined that Oracle does not carry (connive, ward ×2, crew)
                    1 × reminder text missing that Oracle does carry (indestructible)
                    1 × missing flavor text
                    2 × keyword order
                    (the 9th is Reptil, which also needed a script-side description fix)
DSL fidelity:     270 cards reviewed in 26 batches (8–14 cards each) — 8 divergences, all fixed
                  assay-differential --set MSH: 0 divergent of 34 read whole
                  (226 declined, 5 multi-face, 4 script-slot, 1 no-fold — i.e. 87% not covered)
Behaviour:        84 of 270 cards exercised by ≥1 scenario test; 186 with none
                  2 new tests added (Ant-Man's Army, Avengers: Under Siege), both mutation-checked
                  self-play: 4 games across a W/G/R and a U/B deck, ~90 turns
Tokens:           27 token-creation sites, all resolving MSH (tmsh) art in the `normal` form
                  token-art-gaps: no MSH rows; TokenArtCoverageTest green
Waived:           none
```

## What the gates do not cover

"0 discrepancies" means: against a dump taken 2026-08-17, on those thirteen fields, for the cards in
`set.cards`. It says nothing about rulings, legalities, or whether a card plays correctly. The DSL
pass is **26 reviewers' judgement**, not a proof. `assay-differential` read only 34 of 270 cards
whole — a green differential here is a floor, not the pass. A clean `token-art-gaps` means no token
falls back to generic art; it does not mean the art is the one printed on the card you meant.

**Combat was never exercised by self-play** — see harness bug B below. Attack, block and
combat-damage triggers across the whole set are unverified by that stage.

## The eight DSL divergences fixed

Gameplay:

- **Frozen in Ice** — "can't become untapped" was `DOESNT_UNTAP`, the narrower "doesn't untap during
  your untap step". Now `CANT_BECOME_UNTAPPED`, which also blocks explicit untap effects and untap
  costs. Copied from Stop Cold, whose printed wording really *is* the narrower one; Spider-Woman,
  Secret Agent in this same set already gets it right.
- **Mister Hyde, Monster Within** — mode 2 used `RemoveCountersUpTo(1)`, a ceiling, so a player who
  chose the *mandatory* "Remove a counter" mode could remove zero. Now the floored
  `RemoveCounterOfAnyKind`, which names this card in its own KDoc.
- **Thunderbolts Conspiracy** — "return it … **with** a finality counter on it" did a Move and *then*
  a separate `AddCounters`, so the creature entered without the counter. Now rides the zone change
  via `addCounterType` (CR 614.1c), matching Hellcat and Grim Reaper in this same set.
- **Tribal filters narrowed to creatures** — a bare "Squirrels" / "Villain" / "Heroes" means any
  permanent with that type. The Unbeatable Squirrel Girl (X count), Villainous Hideout (connive
  target — CR 701.50a instructs a *permanent* to connive) and The Thing, Ben Grimm (damage source).

Display:

- **Loki, Laufeyson / Squirrel Girl / Villainous Hideout** — `descriptionOverride` strings omitted
  their printed cost prefix, so the action menu rendered the rider with no cost.
- **Reptil, Dinomorpher** — its ability description still said "vigilance and reach" after the oracle
  text was corrected. Ability descriptions are *not* among the 13 fields the Scryfall harness
  compares, which is exactly why this survived a green field check.

A sweep of all 33 tribal-filter sites found four more creature-scoped ones — all `+N/+N` anthems,
where the narrowing is unobservable because a noncreature permanent has no power/toughness. Left
alone deliberately.

## Findings routed elsewhere, not fixed here

`backlog/archived/sets/marvel-super-heroes/marvel-super-heroes-engine-gaps.md` records eleven. The
two worth reading:

**1. `DamageType` is silently unenforced on `ModifyDamageAmount`.**
`rules-engine/.../handlers/effects/DamageUtils.kt` (~1863) checks restrictions, source and recipient
but never `damageEvent.damageType`. The `DoubleDamage` loop 60 lines above it *does*, with the same
six-line guard, and combat damage reaches both through the same
`applyStaticDamageAmplification(isCombatDamage = true)`. Four cards in the corpus use
`ModifyDamageAmount`; **Hawkeye, Young Avenger is the only one that pairs it with a non-`Any`
DamageType**, so this latent engine bug is exposed by exactly one card and that card is in this set.
Effect: Hawkeye's "+X to *noncombat* damage" also amplifies your creatures' combat damage.

**2. The Kingpin of Crime snapshots a set that CR 611.2c requires to stay dynamic.**
"Until end of turn, creatures you control with toughness greater than their power assign combat
damage equal to their toughness" modifies no characteristic and no controller, so per CR 611.2c it
modifies the rules of the game and must reach creatures that become toughness-heavy later. Ours
snapshots at resolution. A faithful fix needs a floating filtered continuous effect. The card's KDoc
claimed the snapshot "matches the printed wording's timing" — corrected to state it as a limitation.

## Two harness bugs that limit what self-play can prove

Both were found by driving MSH through the gym HTTP API, and neither is an MSH defect. Both self-play
reports initially blamed cards; both were wrong, and the two scenario tests in the last commit are
what settled it.

**A. `gym-server` never registers the predefined tokens.**
`gym-server/.../config/GymBeansConfig.kt` builds its `CardRegistry` from set cards plus basic lands
and never calls `register(PredefinedTokens.allTokens)` — `game-server`'s `GameBeansConfig.kt:37`
does. `CreatePredefinedTokenExecutor` then bails to an `EffectResult.error` rather than throwing, so
**every predefined-token card in the corpus — Treasure, Food, Clue, Map, Incubator — silently mints
nothing when driven through the gym HTTP API.** That is why inline `CreateToken` cards (Squirrel,
The Tiger God, the Villain tokens) were observed working in the very same games. One-line fix.

**B. The gym action contract cannot express any choice beyond an action id.**
`LegalAction` carries `validAttackers`, `mandatoryAttackers`, `validAttackTargets`,
`additionalCostInfo` and `tapForPower`. The gym's `LegalActionView` serialises only
actionId/kind/description/affordable/manaCost, and `StepBody` is `data class StepBody(val actionId:
Int)`. Nothing under `gym/` references `validAttackers` at all. So declaring attackers, equipping,
and paying teamwork's "tap creatures with total power N" are all **silent no-ops that return HTTP 200
with an unchanged state digest** — 47 and 65 of them across the two runs.
`docs/gym-self-play-testing.md` §4.3 tells you to "drive combat to exercise attack/block/damage
triggers"; that instruction currently cannot be followed.

## Backlog

Archived to `backlog/archived/sets/marvel-super-heroes/`, with an engine-gaps file alongside.
`MarvelSuperHeroesSet` declares no `incomplete` override and `MtgSet` defaults it to `false`, so
nothing was needed there. Two entries in `mechanics.md` had since shipped and are marked so:
`ModifyCounterPlacement.placedByYou` now exists, and boast ships as the `isBoast` authoring flag
(CR 702.142). `scripts/check-card-counts.py --check` is green.

The Scryfall dump and `MshCardFieldVerificationTest` were throwaway and are not committed.


bolav:
## Review — `finish-msh` (7 commits, 93 files) vs `main`

Reviewed in place on the rebased branch. **Findings 1, 2 and 6 below have since been fixed on the
branch**; 3, 4 and 5 are open and left to the author's judgement.

### Verdict

Behavior is right and the SDK shape is right. Two new cards, three genuinely new pieces of engine
vocabulary (boast marker, sum-gated graveyard exile, one-shot free-cast rider), and 26 card
corrections that check out field-for-field against Scryfall.

### What's good

- **`GraveyardTotalExileResolver` is the right generalization.** Collect evidence was a special case
  of "exile any number of graveyard cards whose summed measure reaches N" and is now implemented as
  one, with `CollectEvidenceResolver` reduced to what is actually *collect evidence* (the name, the
  event, the pool/measure choice). Reachability, legality, auto-selection and the exile itself can't
  drift apart because there is one of each.
- **`CardMeasure` was correctly kept separate from `VariableCostMeasure`.** The KDoc's reason — one
  reads projected battlefield state, the other reads intrinsic card characteristics where
  `TOTAL_POWER` is meaningless — is the right call, not a hedge.
- **Fail-closed is applied uniformly**: the enumerator drops the action, `canAfford`/`canPayCost` say
  no, and `PayOrSuffer` plus the spell additional-cost paths report unpayable rather than offering a
  cost nothing can pay. A client-submitted selection that doesn't pay is *rejected*, never silently
  swapped for the engine's pick — the right choice for a client-supplied field.
- **A real pre-existing bug fixed in passing:** the collect-evidence mana-value floor was never on
  the `AdditionalCostInfo` DTO nor forwarded by `LegalActionEnricher`, so the client read a field the
  server never sent and its Confirm gate was inert.
- **`evaluateForDisplay` returning `null` instead of `0`** is the correct fix for "+0/+0", and
  passing chosen targets into the ability-on-stack context (`ClientStateTransformer`) fixes the other
  half, where the value was in fact known.
- **The card corrections are real.** All 28 changed/new MSH cards match Scryfall exactly on
  `oracle_text`, `mana_cost` and `type_line`; `check-card-printing` is clean for both new cards.
  Every CR number cited in new code verifies against `MagicCompRules_20260808`: 702.142a, 701.59a,
  118.9/118.9a, 107.3b, 107.4e/f, 202.3, 601.2f/h, 602.2b, 707.12, 707.10a, 614.1c, 611.2c.
- **`marvel-super-heroes-engine-gaps.md`** is a genuinely useful artifact — documented deferrals with
  file/line, observable symptom and the fix shape, rather than silence.

### Issues

#### Blocking

**1. The branch didn't compile after the rebase.** ~~`CostPaymentService.kt`~~ — `selectionCandidates()`
landed on `main` after this branch was written, and its `when (val atom = c.atom)` had no
`ExileFromGraveyardForTotal` branch. Neither side conflicts textually, so the rebase applied cleanly
and only the compiler caught it. **Fixed**: the atom joins the other activated-ability-only atoms
returning `null` — consistent with `canAfford` already reporting it unaffordable as a `PayCost`,
since its real selection lives in `CostHandler`/`GraveyardTotalExileResolver`.

#### Important

**2. `exileMinTotalManaValue` was strictly redundant with `exileMinTotalWeight` + `exileCardWeights`**
— two DTO fields, two `costType` branches client-side and two ways to compute a running total, for
one gate. **Fixed** by collapsing to the weight table: `CollectEvidenceResolver.costInfo` now ships
the mana values it already had as weights, the client's `CollectEvidence` and `ExileForTotal` cases
are one branch, and `TargetingOverlay` has a single floor and a single weight lookup. The unit label
that would otherwise have been lost now comes from `CardMeasure.unitLabel`, so the measure names its
own unit ("mana value", "black mana symbols") and the client never has to know which cost it is
rendering. Two vitest cases pin both costs through the shared branch.

**3. `GraveyardTotalExileResolver.candidates` evaluates the filter with no source in context**
(`PredicateContext(controllerId = playerId)`). A source-relative filter ("cards that share a creature
type with this permanent") would silently match nothing. Zemo's colour filter doesn't care, but the
class KDoc documents the projected-vs-base decision without mentioning this gap — and the free-cast
rider two files over goes out of its way to thread `sourceId` for exactly this reason. Either thread
a `sourceId: EntityId?` through `candidates`/`canPay`/`costInfo`, or record the limitation the way
`GrantNextSpellAffinityEffect`'s doc records its own. **Open.**

#### Minor

**4. Client coverage.** Now partly addressed — the sum-gated picker has tests as of the fix for #2 —
but the Confirm-gate arithmetic in `TargetingOverlay` itself is still only covered indirectly.

**5. Scope.** AGENTS.md: *"A card needing new engine vocabulary still gets a PR of its own."* This
branch bundles two such cards plus 26 corrections plus the backlog archive. Defensible for a
set-finishing branch and each unit is a separate commit — flagged so splitting stays a deliberate
choice rather than an accident. **Open.**

**6. Wrong CR citation.** `CR 112.7a` appeared throughout the LKI plumbing; section 112 is *Spells*,
and the rule that an ability exists on the stack independently of its source (so last-known
information is used) is **113.7a**. **Fixed** across `rules-engine`, `mtg-sdk` and
`docs/card-sdk-language-reference.md`. The same wrong number still appears in ~15 card files under
`mtg-sets/` and in two archived backlog gap docs; those were left alone deliberately to avoid
touching other work in flight, and want their own sweep.

### Gates

| Gate | Result |
|---|---|
| `npm run typecheck` | green |
| `npm test` (vitest) | green — 40 files, 570 tests |
| `:mtg-sdk:test` `:rules-engine:test` `:mtg-sets:2026:tests:test` `:mtg-sets:test`, plus `:game-server` / `:ai` compile | green — 4m42s, 27 tasks executed, 0 failures |

All four Kotlin suites re-ran against the fixes (result XML timestamps confirm they weren't served
from cache). Not covered: `:gym*`, e2e Playwright, and the AI's own suites.

Scryfall verification was run over every changed card (`oracle_text` / `mana_cost` / `type_line`) and
every CR number in new code was checked against the local Comprehensive Rules text.

PR #1888, opened 2026-08-17
Fix two engine bugs and two gym-harness bugs found verifying MSH

bolav:
Depends on: https://github.com/wingedsheep/argentum-engine/pull/1886

Four bugs surfaced while verifying Marvel Super Heroes, each fixed with a test that fails without
the fix. Two are engine/SDK fidelity, two are the gym harness lying to self-play. Based on
`finish-msh`; the diff is the four commits below and nothing else.

## 1. `DamageType` was silently unenforced on `ModifyDamageAmount`

`DamageUtils.applyStaticDamageAmplification`'s `ModifyDamageAmount` loop checked `restrictions`,
`damageSourceMatches` and `damageRecipientMatches` — but never the pattern's `damageType`. The
`DoubleDamage` loop 60 lines above it does, and combat damage reaches both through the same
`applyStaticDamageAmplification(isCombatDamage = true)`.

Four cards in the corpus use `ModifyDamageAmount`; **Hawkeye, Young Avenger** (MSH #131) is the only
one pairing it with a non-`Any` `DamageType`, so his "+X to *noncombat* damage" also pumped your
creatures' combat damage. The fix is the same six-line `damageTypeMatches` guard.

`HawkeyeYoungAvengerScenarioTest` covers both sides — a burn spell to an opponent gets the bonus, two
attacking creatures do not. Mutation-checked: disabling the guard fails exactly the combat case.

## 2. The Kingpin of Crime snapshotted a set CR 611.2c requires to stay dynamic

> Whenever you attack, you may pay 2 life. If you do, until end of turn, creatures you control with
> toughness greater than their power assign combat damage equal to their toughness rather than their
> power.

That effect changes no characteristic and no controller, so under **CR 611.2c** it modifies the rules
of the game and its affected set keeps re-evaluating. Ours froze the group at resolution
(`ForEachInGroup` + `GrantKeyword(ASSIGNS_COMBAT_DAMAGE_AS_TOUGHNESS)`).

The faithful shape needed **no new SDK vocabulary**. The sentence is Bedrock Tortoise's printed static
ability for a turn, so the Kingpin now grants *himself*
`AssignDamageEqualToToughness(AllCreaturesYouControl, onlyWhenToughnessGreaterThanPower = true)` until
end of turn, and `CombatDamageUtils` reads granted static abilities alongside printed ones — the
point-of-use read every granted `StaticAbility` kind needs, in the shape `BlockPhaseManager` already
uses for `CantBeBlockedByMoreThan`. Because that read happens against the **final** projected power
and toughness, the set is re-decided per creature per damage step.

A floating group-flag effect (the shape the gap write-up proposed) was built first and does **not**
work: `dynamicGroupFilter` is resolved in layer 6, before layer 7 applies, so it cannot see a
toughness pumped after the trigger resolved — and that is the only observable case for this card,
since a creature entering later can't join an already-declared combat. It was reverted.

Residual, documented on the card: the grant is anchored to the Kingpin, so it stops applying if he
leaves the battlefield mid-turn, where the printed effect would outlive him.

`TheKingpinOfCrimeScenarioTest` covers pay/decline, the damage-as-toughness attack, and the
pumped-after-resolution creature. Mutation-checked: without the granted-ability read, both positive
cases fail.

## 3. gym-server never registered the predefined tokens

`GymBeansConfig` built its `CardRegistry` from set cards and basic lands only, never calling
`register(PredefinedTokens.allTokens)` the way `game-server`'s `GameBeansConfig` does.
`CreatePredefinedTokenExecutor` resolves a token *by name* and returns an `EffectResult.error` when
the name is missing — no exception, no log. So every Treasure / Food / Clue / Map / Incubator card in
the corpus silently minted nothing over the gym API, and self-play read that as a broken card rather
than a broken harness.

`GymBeansConfigTest` asserts the registry resolves every predefined token. Mutation-checked: without
the registration it fails on `Treasure`.

## 4. The gym action contract couldn't express any choice beyond an action id

`DeclareAttackers` is enumerated **once**, carrying an empty attacker map; the candidates ride on the
legal action (`validAttackers`, `validAttackTargets`) because the choice belongs to the player.
`StepBody` was a lone `actionId`, so stepping that action declared no attackers — a legal move, an
accepted `200`, and a permanently attack-free game. Blockers, a spell's targets and X had the same
shape. Combat was therefore unreachable over the gym API, which caps what self-play can claim and
makes `docs/gym-self-play-testing.md` §4.3 unfollowable.

- `ActionParams` (`attackers`, `blockers`, `targets`, `xValue`) + `ActionParameterizer`, which folds
  them into the enumerated `GameAction`. Target ids are resolved to the `ChosenTarget` variant their
  current zone implies.
- `GymEnv.step` takes params; the deckbuild env rejects any (its actions are fully described by their
  id).
- `LegalActionView` exposes `validAttackers` / `mandatoryAttackers` / `validAttackTargets` /
  `validBlockers`, without which params can't be built.
- `GameEnvironment.lastRejection` records an engine rejection and `GameGymEnv` turns it into a `400`:
  an illegal declaration leaves state untouched, which is otherwise indistinguishable from "attacked
  with nobody". Params an action can't use are a `400` too — never dropped.
- Schema hash bumped to `v1.3-action-params`.

`GameGymEnvCombatTest` drives a real game to combat damage through the contract, and covers both
rejection paths plus the "no params means attack with nobody" baseline. The self-play guide and the
gym-server README document the shape; the guide's action `kind` names were also stale (`PASS` is
really `PassPriority`).

## Notes for review

- Both self-play reports that started this were **wrong** about the cards; chasing them is what
  found bugs 3 and 4 in the harness instead. One reported rationale was also simply false (that MSH
  was the only user of a `CreateTreasure` overload — four other sets use it), which is why the card
  claims were settled with mutation-checked scenario tests rather than acted on.
- `backlog/archived/sets/marvel-super-heroes/marvel-super-heroes-engine-gaps.md` now marks gaps 1 and
  2 fixed, with the reasoning above; the other nine stand as recorded.
- `docs/card-sdk-language-reference.md` documents `AssignDamageEqualToToughness` as a granted-ability
  kind wired into a read site, and when a rules modification must *not* use
  `Patterns.Group.grantKeywordToAll`.

## Verification

| Gate | Result |
|---|---|
| `:rules-engine:test` | green |
| `:mtg-sets:2026:tests:test` (Hawkeye, Kingpin, whole 2026 module) | green |
| `BedrockTortoiseScenarioTest` (printed `AssignDamageEqualToToughness`) | green |
| `:gym:test` (incl. `GameGymEnvCombatTest`) | green |
| `:gym-server:test` (incl. `GymBeansConfigTest`) | green |

Each fix was mutation-checked individually: the guard, the token registration and the granted-ability
read were each disabled in turn, and only the intended tests went red.

**Not covered:** nothing here touches TypeScript or the e2e suite, so no claim is made about them;
the gym trainer / self-play loop was not exercised end-to-end against a live server, only through the
env contract in `:gym`'s tests.


bolav:
## Review — `verify-bugs`

Three real bugs, three correct fixes, at the right layers, with no new SDK vocabulary. The branch was
**red on `main`** when this review started, though: the MSH snapshot golden was never re-blessed after
the Kingpin rewrite. That and every other finding below is now fixed in a follow-up commit on the
branch; this comment records what was found and why, so the reasoning survives the diff.

### Gate

Run fresh (not served from cache) after the fixes:

| Module | Tests | Failures |
|---|---|---|
| `:mtg-sdk` | 466 | 0 |
| `:rules-engine` | 2948 | 0 |
| `:mtg-sets` (corpus snapshot + JSON round-trip) | 363 | 0 |
| `:mtg-sets:2026:tests` | 1307 | 0 |
| `:gym` | 54 | 0 |
| `:gym-server` | 11 | 0 |
| **Total** | **5149** | **0** |

Not covered, and not reached by this diff: `:ai`, `:game-server`, `web-client`.

---

## What's good

- **The `DamageType` fix is a literal symmetry restoration.** The new guard matches the `DoubleDamage`
  loop 50 lines above it, verbatim. Blast radius checked: `EventPattern.DamageEvent.damageType`
  defaults to `DamageType.Any`, and of the four `ModifyDamageAmount` cards in the corpus (Valley
  Flamecaller, Far Fortune, Fated Firepower, Hawkeye) only Hawkeye sets a non-`Any` type — so exactly
  one card's behaviour changes, as the commit claims.
- **The Kingpin rewrite is the right shape.** Reusing `AssignDamageEqualToToughness` +
  `GrantStaticAbility` instead of inventing a group-dynamic grant is the composition-first bar, and
  the point-of-use read follows the `BlockPhaseManager` / `CantBeBlockedByMoreThan` precedent rather
  than establishing a new pattern. `matchesBattlefield` reads `projected.getController` and routes the
  filter through `predicateEvaluator.matches(state, projected, …)` — projection discipline is correct.
- **CR 611.2c is right**, verified verbatim against the Comprehensive Rules, including the quoted "can
  affect objects that weren't affected when that continuous effect began". Both cards' oracle text
  matches Scryfall exactly.
- **The rule-corner test earns its citation.** "A creature that becomes toughness-heavy after the
  trigger resolved is covered" is precisely the case that separates the new shape from the
  `ForEachInGroup` snapshot — not a decorative test.
- **`GymBeansConfigTest` guards the drift class, not the symptom.** Asserting every
  `PredefinedTokens.allTokens` entry resolves will catch the next token added too. `GameBeansConfig`
  does register them, as the KDoc claims, and those two are the only registry builders in the repo.
- **The `ActionParams` premise is real**, not speculative: `CombatEnumerator.kt:79` emits
  `DeclareAttackers(playerId, emptyMap())`. `SchemaHash` correctly bumped. `EntityId`-keyed maps over
  kotlinx JSON already have precedent (`DistributionResponse.distribution`), so the wire format is fine.

---

## Blocking

### 1. The MSH golden was stale — `CardDefinitionSnapshotTest` was failing

`mtg-sets/src/test/resources/snapshots/cards/MSH.json` still held the old `ForEachInGroup` +
`GrantKeyword ASSIGNS_COMBAT_DAMAGE_AS_TOUGHNESS` tree. Commit `dba23ee` changed the Kingpin's
compiled tree and didn't re-bless.

Re-blessed and reviewed: one card's subtree, no other set touched. The new tree is faithful — the
absent `onlyWhenToughnessGreaterThanPower` and `duration` keys are correct, both are defaults
(`= true`, `Duration.EndOfTurn`), so the exporter's default-omission isn't dropping meaning.

The process point matters more than the file. That commit's body advertises *"Gate: `:rules-engine:test`,
`:mtg-sets:2026:tests:test`, `BedrockTortoiseScenarioTest` — all green."* That gate cannot see this
failure. **Any change to a `cardDef` must also run `:mtg-sets:test`**, because the corpus snapshot
lives there and nowhere else.

---

## Important

### 2. CR 508.1a is the wrong rule — it is 508.1d

`TrainingObservation.kt` and `docs/gym-self-play-testing.md` both cited 508.1a for "must attack if
able". 508.1a is the *choice* rule (untapped, not a battle, haste-or-continuously-controlled). Attack
*requirements* are 508.1d: *"effects that say a creature attacks if able… if the number of
requirements being obeyed is fewer than the maximum possible… the declaration of attackers is
illegal."* The prose was right; only the number was wrong. Fixed in both places.

### 3. A rejected action in `step-batch` returned 500, not the documented 400

`EnvWorkerPool.invokeAll` did `futures.map { it.get() }`, so the new `IllegalArgumentException`
arrived wrapped in `ExecutionException`, which `GymExceptionHandler` doesn't map → 500. It was
inconsistent even with itself: the `tasks.size == 1` fast path calls directly and *did* give 400.

A latent hazard for stale IDs before; this PR makes engine rejections a routine path and documents
`params` on `step-batch`, so it became load-bearing. Now unwrapped (`throw e.cause ?: e`), with
`EnvWorkerPoolTest` covering both the 400 and 409 mappings. Mutation-checked: restoring the plain
`it.get()` turns both new assertions red.

### 4. Three of the four `ActionParams` fields were untested

`GameGymEnvCombatTest` covered attackers only. The `DeclareBlockers` branch, the
`CastSpell` / `ActivateAbility` `targets` + `xValue` branches, and all of `resolveTarget` — whose
four-way zone dispatch is the only logic in the file that reads live state — had no coverage.

Added: an end-to-end test that casts Portal's **Blaze** (`{X}{R}`, "deals X damage to any target")
through `step` with `targets` + `xValue` and asserts the named player takes exactly X; a direct
assertion of `resolveTarget`'s player / card-in-zone / not-in-any-zone dispatch; and two tests on the
`DeclareBlockers` branch. Mutation-checked: making the `CastSpell` branch ignore params turns the
Blaze test red.

### 5. `GameGymEnvCombatTest` ran an unseeded game

`GameConfig.seed` exists and its KDoc says outright: *"Supply an explicit value for reproducible runs
(replays, MCTS, the cross-engine parity harness, **tests**)."* With a 20-card deck and a 400-step
budget, an unlucky shuffle buries every Raging Goblin and the life assertion fails. Now seeded.

Seeding surfaced a second problem worth recording: the shared driver loop cannot identify a spell that
is completable from an ID alone. `minTargets` / `maxTargets` are enumerator placeholders — **Raging
Goblin reports `minTargets=1, maxTargets=1` despite targeting nothing** — so filtering on them skips
castable creatures entirely. The driver now tries candidates in order and falls through on rejection,
which is also what a real client would do.

### 6. The blockers half of the observation was incomplete

`ObservationBuilder` projected `validAttackers` / `mandatoryAttackers` / `validAttackTargets` /
`validBlockers`, but `LegalAction` also carries `blockerMaxBlockCounts` and
`mandatoryBlockerAssignments` and neither was exposed. A client could see *which* creatures may block
but still not build a declaration satisfying mandatory blocks or per-blocker limits — it just got a
400 it had no way to predict, the same class of gap this PR fixes for attackers. Both now projected
and documented (CR 509.1a for the default one-attacker limit, CR 509.1c for required blocks).

---

## Minor

7. **`docs/card-sdk-language-reference.md` had two contradictory `GrantStaticAbility` bullets.** The
   duplication pre-dated this change, but the diff edited *both* copies and left them out of sync —
   one listed four wired read sites, the other five. The stale copy is deleted. The inserted prose
   also spliced mid-sentence in both (*"…read sites today; **Granting** `AssignDamageEqualToToughness`…"*
   and *"…either. **granting** another `StaticAbility` kind…"*); the paragraph now sits after the
   sentence it was cutting in half.

8. **`assignsDamageAsToughness` allocated per permanent per damage assignment.** It built a filtered +
   mapped list inside the battlefield loop, for every creature assigning combat damage. Now indexed
   once with `groupBy { it.entityId }` above the loop, which is empty in the common case.

9. **`ActionParameterizer.apply` hand-wrote eleven `reject(...)` calls.** Replaced with a single
   `allowOnly(action, …)` whitelist, so a new `ActionParams` field is inapplicable everywhere by
   default rather than silently dropped by every branch that forgot to name it.

10. **`AssignDamageEqualToToughness.description` hardcoded equipment flavour** — *"As long as equipped
    creature's toughness…"* regardless of `filter`. Harmless on the Kingpin, which has a
    `descriptionOverride`, but wrong for Doran, Bedrock Tortoise and Tapestry Warden too, and wrong for
    anything that surfaces a granted static ability's own text. Now branches on `filter.scope`:
    singular wording for the `Self` / `AttachedTo` shapes, and the filter's own description for a
    battlefield group, matching what `CantAttack` / `CantBlock` already do.

11. **The decision branch didn't check `lastRejection`.** `GameGymEnv.executeResolved` closed the
    silent-no-op hole for `ResolvedAction.Legal` but left it open for `ResolvedAction.Decision` — a
    rejected `SubmitDecision` still read as a successful step. The rationale in the new KDoc applies
    identically there. Both branches now go through one `failOnRejection`.

12. **`repeat(400) { if (terminated) return@repeat }`** continued rather than broke, spinning through
    the remaining iterations after the game ended. The extracted driver returns instead.

---

## Residuals, deliberately left

- The Kingpin grant is anchored to the Kingpin, so it stops applying if he leaves the battlefield
  mid-turn where the printed effect would outlive him. Documented on the card; `grantedStaticAbilities`
  is keyed by entity id and the read loops the battlefield, so there is no cheap fix.
- `GroupFilter.buildDescription` renders a battlefield group as `all you control creatures` — the
  controller predicate leads. That is repo-wide (every ability using `filter.description` inherits it),
  so it belongs in its own change rather than being patched inside one ability.

PR #1896, opened 2026-08-17
Run OnEnterRunEffect when an effect puts a permanent onto the battlefield

bolav:
# Run `OnEnterRunEffect` when an effect puts a permanent onto the battlefield

## The bug

An earthbent Multiversal Passage that was exiled and returned by its granted "when this dies or is
exiled, return it to the battlefield tapped" trigger came back inert: its "as this land enters,
choose a basic land type" clause never ran, so the land had no subtype, no intrinsic mana ability,
and nothing its controller could do with it.

That clause is an `OnEnterRunEffect` replacement, and the engine only consulted it in
`PlayLandHandler` — i.e. only when the land was *played*. Every other route onto the battlefield
(a return from exile or the graveyard, reanimation, blink) skipped it.

## The fix

Wire it into `MoveToZoneEffectExecutor` through a new helper on the existing as-enters seam,
`PermanentEntryReplacements.runOnEnterRunEffect`. No new SDK type: this reuses the
`OnEnterRunEffect` replacement and the executor pipeline that already exist.

- The effect runs with a fresh `EffectContext` rooted at the entering permanent, so
  `EffectTarget.Self` resolves to it rather than to whatever moved it.
- It carries the caller's `resolutionDepth`, so the registry's runaway-recursion backstop still
  counts a self-perpetuating entry loop.
- A pause forwards the entry events already collected (including the entry `ZoneChangeEvent`), so
  the entry's ETB triggers are deferred by the normal resume path instead of being lost.
- Skipped for face-down entries (CR 708.2) and for a battlefield-to-battlefield redirect, which is
  not a new entry.
- `ZonesExecutors` gains late-bound registry recursion for this, mirroring `PermanentExecutors`.

## Review round

A review of the first commit produced eight findings; all are addressed in the follow-up commit.

**Correctness**

- `MoveToZoneEffectExecutor` now propagates `triggersAlreadyProcessed` from the replacement's
  result. Dropping it would double-fire cast triggers if a future `OnEnterRunEffect` nested a cast
  (`CastFromCollectionWithoutPayingCost` routes through `CastSpellHandler`, which stacks its own).
  `error` is still deliberately not propagated, with the reasoning recorded inline: the permanent
  did enter, only the as-enters clause failed, and surfacing an error would make the enclosing
  composite treat the whole move as the failed step (CR 609.3).
- The `effectExecutor` constructor parameter is now required rather than nullable. The one caller
  that cannot reach the battlefield — the bounce-to-hand reuse in
  `ReturnSpellOrPermanentToOwnersHandExecutor` — passes a throwing stub, so if that destination
  ever changes it fails loudly instead of silently skipping a rules-required replacement.

**A doc claim that was wrong**

The original scope note claimed "fetches that move one card" were covered. They are not: a library
search that puts a *single* card onto the battlefield (`SearchDestination.BATTLEFIELD`, e.g.
Wayfarer's Bauble) routes through `MoveCollectionExecutor`, which the same note correctly lists as
unwired. Corrected in `docs/card-sdk-language-reference.md` and in both KDoc copies.

**Accuracy of the shared-seam docs**

`PermanentEntryReplacements`' object KDoc read as though all three as-enters replacements applied
on all three listed paths. They don't — each caller uses a different subset. It now says which,
and names the next gap explicitly: `EntersWithChoice` on the move path, where a reanimated
Shapeshifter or Sorcerous Spyglass still enters with its choice never made. Same shape of bug, one
replacement over.

**Deduplication**

The `OnEnterRunEffect` lookup was duplicated in `PlayLandHandler` and `PermanentEntryReplacements`.
`MultiversalPassage`'s KDoc explicitly depends on the "first one wins" rule, so the two copies
could have drifted on it. Single-sourced as `PermanentEntryReplacements.onEnterRunEffectFor`.

**Gaps now recorded where the code is, not only in the docs**

- `MoveCollectionExecutor` carries a `KNOWN GAP` comment next to its `EntersWithReplacements`
  call, explaining both what is missing and why closing it needs a continuation (the replacement
  can pause, and this is a loop over cards).
- An Aura carrying this replacement would still skip it, because `attachAuraOnEnter` returns
  first. No card does today; noted in the code and the scope list.
- CR 614.12a says an as-enters choice is made *before* the permanent enters, whereas the engine
  places it first so `EffectTarget.Self` resolves. The observable consequence — entry triggers are
  matched against pre-choice characteristics, so "whenever a Mountain enters" won't see the type a
  Multiversal Passage just chose — is now written down. Both entry paths behave identically here,
  so it is consistent rather than path-dependent.

## Still out of scope

Recorded in the SDK reference's "Scope today" list rather than fixed here:

- `MoveCollectionExecutor` — the batch move, which also covers single-card library searches onto
  the battlefield. Needs a continuation to pause mid-loop.
- `StackResolver.enterPermanentOnBattlefield` — the spell-resolution path. Nothing needs it until
  the first non-land permanent uses this replacement; the only two cards using it today
  (Multiversal Passage, Game Trail) are lands.
- Auras, on any path.

## Tests

`MultiversalPassageScenarioTest` gains two cases:

1. **Returned by an effect asks for the basic land type again.** Uses Sandman, Shifting Scoundrel's
   graveyard ability, which is `CompositeEffect(MoveToZoneEffect, MoveToZoneEffect)` — structurally
   identical to the earthbend return trigger that produced the original bug
   (`Effects.kt`, `earthbendReturnTapped`). Asserts the land comes back as the chosen Mountain and
   taps for `{R}`, rather than as a typeless land that can't do anything.
2. **The entry's landfall trigger survives the as-enters pause.** This one exists because the
   riskiest line in the change is the claim that the paused result carries the entry
   `ZoneChangeEvent` up to the resume path. Without it, a refactor returning the pause with no
   events stayed green. Verified by mutation: dropping the entry events from the paused return
   fails this test and only this test.

## Verification

| Gate | Result |
|---|---|
| `:rules-engine:test` | 2923 tests, 0 failures, 0 errors |
| `:mtg-sets:2025:tests:test` | 1594 tests, 0 failures, 0 errors |

Both suites re-run after the review fixes, with fresh test-result XML confirmed (the first attempt
at the targeted run came back all-`UP-TO-DATE`, i.e. a cached false green, and was forced).

**Not covered by these gates:** `mtg-sets` scenario suites outside the 2025 era, `game-server`, and
the web client — none of which this diff touches.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -10178,8 +10178,12 @@ The priority groups are (CR 616.1a–f):
   executes via the normal effect-executor pipeline at entry time (so `EffectTarget.Self` resolves to
   the entering permanent) and may pause for player input. Compose with atomic pausable effects like
   `Effects.MayRevealCardFromHand` to build SOI shadow lands or other "as ~ enters" choices.
-  **Scope today:** only wired into the land-play path (`PlayLandHandler`). When the first non-land
-  permanent needs this, also wire it into `StackResolver.enterPermanentOnBattlefield`.
+  **Scope today:** the land-play path (`PlayLandHandler`) and every single-card "put onto the
+  battlefield" effect (`MoveToZoneEffectExecutor` → `PermanentEntryReplacements.runOnEnterRunEffect`),
+  which covers reanimation, blink, an earthbend/"return it to the battlefield" trigger, and fetches
+  that move one card. Not yet wired into the spell-resolution path
+  (`StackResolver.enterPermanentOnBattlefield` — needed when the first non-land permanent uses this)
+  or into the multi-card `MoveCollectionExecutor` batch move.
 - `EntersWithCounters(counterType?, count, selfOnly?, condition?, otherOnly?, appliesTo?)` /
   `EntersWithDynamicCounters(counterType?, count, otherOnly?, appliesTo?)` — "[permanent] enters with
   N counters." `EntersWithCounters` takes a fixed `count: Int` (Master Biomancer, Metallic Mimic);
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -10178,12 +10178,23 @@ The priority groups are (CR 616.1a–f):
   executes via the normal effect-executor pipeline at entry time (so `EffectTarget.Self` resolves to
   the entering permanent) and may pause for player input. Compose with atomic pausable effects like
   `Effects.MayRevealCardFromHand` to build SOI shadow lands or other "as ~ enters" choices.
-  **Scope today:** the land-play path (`PlayLandHandler`) and every single-card "put onto the
-  battlefield" effect (`MoveToZoneEffectExecutor` → `PermanentEntryReplacements.runOnEnterRunEffect`),
-  which covers reanimation, blink, an earthbend/"return it to the battlefield" trigger, and fetches
-  that move one card. Not yet wired into the spell-resolution path
-  (`StackResolver.enterPermanentOnBattlefield` — needed when the first non-land permanent uses this)
-  or into the multi-card `MoveCollectionExecutor` batch move.
+  **Scope today:** the land-play path (`PlayLandHandler`) and the single-card `MoveToZoneEffect`
+  path (`MoveToZoneEffectExecutor` → `PermanentEntryReplacements.runOnEnterRunEffect`), which covers
+  reanimation, blink, and an earthbend/"return it to the battlefield" trigger. Not yet wired into:
+  - `MoveCollectionExecutor` — the batch move. This is **not** only multi-card: a library search that
+    puts a *single* card onto the battlefield (`SearchDestination.BATTLEFIELD`, e.g. Wayfarer's
+    Bauble) routes here too, so tutoring up one of these lands still skips the clause. Closing it
+    needs a continuation, because the replacement can pause mid-loop.
+  - `StackResolver.enterPermanentOnBattlefield` — the spell-resolution path; needed when the first
+    non-land permanent uses this replacement.
+  - Auras, on any path — `MoveToZoneEffectExecutor` hands an entering Aura to its choose-a-host
+    continuation before reaching the replacement.
+
+  **Entry order caveat.** CR 614.12a says an as-enters choice is made *before* the permanent enters;
+  the engine places the permanent first and then runs the effect, so `EffectTarget.Self` resolves.
+  The observable difference is that the entry's triggers are matched against the pre-choice
+  characteristics — a "whenever a Mountain enters" trigger won't see the type a Multiversal Passage
+  just chose. Both entry paths behave the same way here, so it is consistent, not path-dependent.
 - `EntersWithCounters(counterType?, count, selfOnly?, condition?, otherOnly?, appliesTo?)` /
   `EntersWithDynamicCounters(counterType?, count, otherOnly?, appliesTo?)` — "[permanent] enters with
   N counters." `EntersWithCounters` takes a fixed `count: Int` (Master Biomancer, Metallic Mimic);
--- a/mtg-sets/2025/tests/src/test/kotlin/com/wingedsheep/engine/scenarios/MultiversalPassageScenarioTest.kt
+++ b/mtg-sets/2025/tests/src/test/kotlin/com/wingedsheep/engine/scenarios/MultiversalPassageScenarioTest.kt
@@ -6,10 +6,14 @@ import com.wingedsheep.engine.core.OptionChosenResponse
 import com.wingedsheep.engine.core.YesNoDecision
 import com.wingedsheep.engine.state.components.battlefield.TappedComponent
 import com.wingedsheep.engine.state.components.player.ManaPoolComponent
+import com.wingedsheep.engine.state.components.stack.ChosenTarget
 import com.wingedsheep.engine.support.GameTestDriver
 import com.wingedsheep.engine.support.TestCards
 import com.wingedsheep.mtg.sets.definitions.spm.cards.MultiversalPassage
+import com.wingedsheep.mtg.sets.definitions.spm.cards.SandmanShiftingScoundrel
+import com.wingedsheep.sdk.core.Color
 import com.wingedsheep.sdk.core.Step
+import com.wingedsheep.sdk.core.Zone
 import com.wingedsheep.sdk.model.Deck
 import com.wingedsheep.sdk.model.EntityId
 import com.wingedsheep.sdk.scripting.AbilityId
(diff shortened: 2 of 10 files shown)


bolav:
# Review — `multiverse-passage-bug`

Reviewed with `origin/main` merged in (it was already the direct parent, so no merge commit was
needed). Findings below were raised against commit `f1782a2` and are **all addressed** in the
follow-up commit `b28cc32` on this branch; each is annotated with what changed.

## Verdict

The behavior is right and the SDK shape is right. This adds **no new SDK type** — it reuses the
existing `OnEnterRunEffect` replacement and the existing `PermanentEntryReplacements` seam, and the
late-bound recursion copies the established `PermanentExecutors` / `LibraryExecutors` pattern. The
findings were doc/comment accuracy and one missing test, not design.

## What's good

- **The right seam, no new vocabulary.** One lookup plus one recursive call into the executor
  pipeline that already exists. This is what the SDK-elegance bar asks for, and it is rare to see a
  bug of this shape fixed without a bespoke type.
- **The pause plumbing is actually correct, and I traced it.** `MoveToZoneEffectExecutor` returns
  `transitionResult.events + extraEvents`; `CompositeEffectExecutor` forwards
  `allEvents + result.events`; `PassPriorityHandler.resolveTopOfStack` detects triggers on a paused
  resolution's events and inserts a `PendingTriggersContinuation` beneath the in-flight frames. The
  comment's claim that the entry `ZoneChangeEvent` survives to fire ETB triggers after the choice
  holds.
- **`resolutionDepth` is carried into the fresh context** so the `MAX_RESOLUTION_DEPTH` backstop
  still counts a self-perpetuating entry loop. Easy to drop; wasn't.
- **Guards are the right ones and in the right terms** — `actualDestZone` rather than
  `effect.destination`, face-down skipped, battlefield→battlefield excluded.
- **The test proxy is faithful to the original repro.** Earthbend's granted return is
  `CompositeEffect(MoveToZoneEffect(from=GRAVEYARD), MoveToZoneEffect(from=EXILE))`
  (`Effects.kt`, `earthbendReturnTapped`) — structurally identical to Sandman's two-`Move`
  composite, so the Sandman test exercises the same path the bug came from.
- **Honest commit message**: states the mutation check and names both remaining gaps rather than
  implying full coverage.

## Issues

### Important

**1. `PermanentEntryReplacements` object KDoc overstated what each caller does.** It read
"Applies … `EntersWithChoice`, `EntersAsCopy`, and the generic `OnEnterRunEffect` … Callers share
this seam: PlayLandHandler, TokenFromDefinition, and MoveToZoneEffectExecutor." Each caller uses a
*different* subset: `MoveToZoneEffectExecutor` calls only `runOnEnterRunEffect`, and
`pauseForEntersWithChoice` is called from `PlayLandHandler`, `TokenFromDefinition`,
`CreateTokenCopyOfSourceExecutor` and the resumers — never from the move path. So a reanimated
`EntersWithChoice` permanent (Shapeshifter, Sorcerous Spyglass) still enters with the choice never
made: the *exact* bug this PR fixes, one replacement over.

> **Fixed.** The KDoc now lists the subset per caller and names `EntersWithChoice` on the move path
> as the next gap to close.

**2. "fetch" was claimed as covered and isn't.** A library search that puts a *single* card onto the
battlefield (`SearchDestination.BATTLEFIELD`, e.g. Wayfarer's Bauble) routes through
`MoveCollectionExecutor`, not `MoveToZoneEffectExecutor` — the very executor the doc correctly lists
as unwired two lines later. So "every single-card 'put onto the battlefield' effect … and fetches
that move one card" contradicted itself.

> **Fixed** in `docs/card-sdk-language-reference.md` and in both KDoc copies. The scope list now
> spells out that the `MoveCollectionExecutor` gap is not only about multi-card moves.

**3. No test for the trigger-deferral claim.** The comment made a specific, load-bearing assertion
about events surviving the pause, and nothing exercised it — replacing the return with
`EffectResult.paused(resultState, decision)` (no events) left the whole suite green.

> **Fixed.** New case: a landfall witness on the battlefield, Multiversal Passage reanimated, life
> asserted unchanged at the pause and incremented after the choice resolves. Mutation-verified —
> dropping the entry events from the paused return fails this test and only this test.

### Minor

**4. `MoveCollectionExecutor` had no in-code note** that `OnEnterRunEffect` is deliberately not run
alongside its `EntersWithReplacements.applyOnEntry` call. The gap was recorded in the doc and the
commit message, but someone reading the executor wouldn't see either.

> **Fixed.** A `KNOWN GAP` comment now sits beside that call, including why closing it needs a
> continuation rather than a two-line call.

**5. The `OnEnterRunEffect` lookup was duplicated** in `PlayLandHandler` and
`PermanentEntryReplacements`. `MultiversalPassage`'s KDoc explicitly depends on the "first one wins"
rule, so two independent copies could drift on it.

> **Fixed.** Single-sourced as `PermanentEntryReplacements.onEnterRunEffectFor`; `PlayLandHandler`
> calls it and its now-unused import is gone.

**6. The nullable `effectExecutor` silently skipped a rules-required replacement.** The only null
caller was `ReturnSpellOrPermanentToOwnersHandExecutor`, which really does only move to hand — safe
now, but the failure mode was silence, and the KDoc was the only thing preventing a future
battlefield destination there from resurrecting this bug.

> **Fixed.** The parameter is required; the bounce executor passes a throwing stub, and the unit
> test passes one too.

**7. `onEnterResult.triggersAlreadyProcessed` was dropped.** Swallowing `error` is fine and matches
"do as much as possible". Dropping the flag is not: a future `OnEnterRunEffect` nesting a cast
(`CastFromCollectionWithoutPayingCostEffect`) would double-fire cast triggers — the exact hazard
`EffectResult` documents on that field.

> **Fixed.** The flag rides along; the deliberate non-propagation of `error` is now justified inline
> with CR 609.3 (verified against the local Comprehensive Rules — an effect attempting something
> impossible does only as much as possible).

**8. Auras skip the new block**, because `attachAuraOnEnter` returns first. No such card exists
today.

> **Fixed** as documentation: noted in the code and in the scope list.

**9. Known limitation worth one doc line.** Entry is physical-then-choose, whereas CR 614.12a says
the choice is made *before* the permanent enters. Consequence: triggers are detected against the
pre-choice state, so a "whenever a Mountain enters" trigger wouldn't see the chosen type. This
matches the play-land path exactly, so it's consistent rather than a regression — but it reads as a
bug when someone hits it.

> **Fixed** as documentation: an "Entry order caveat" paragraph in the SDK reference.

## Verification

| Gate | Result |
|---|---|
| `:rules-engine:test` | 2923 tests, 0 failures, 0 errors |
| `:mtg-sets:2025:tests:test` | 1594 tests, 0 failures, 0 errors |

Both re-run after the fixes, with fresh test-result XML confirmed. The first targeted run came back
all-`UP-TO-DATE` — a cached false green — and was forced with `--rerun-tasks`.

**Not covered:** `mtg-sets` scenario suites outside the 2025 era, `game-server`, and the web client.
None are touched by this diff.

## Recommendation

Merge. Everything raised is addressed on the branch.

PR #1904, opened 2026-08-18
Make granted "add one mana of any color" abilities real mana abilities

bolav:
## What

Granted "{T}: Add one mana of any color" abilities were not mana abilities, and the engine path they take when they have to ask for a colour was missing most of what happens after a permanent is tapped for mana.

Reported symptom: **Badgermole Cub added no {G} when a creature tapped for mana via Enduring Vitality.**

## The card half

Six grants built their `ActivatedAbility` with the raw constructor, whose `isManaAbility` defaults to `false`. The engine keys the entire tap-for-mana pipeline off that flag, so each one was an ordinary activated ability that used the stack: it couldn't be activated while paying a cost, the auto-tap solver never saw it, and "whenever you tap a permanent for mana" statics never fired.

Fixed by setting `isManaAbility = true` and `TimingRule.ManaAbility` on all six, matching Cryptolith Rite / Citanul Hierophants:

Abundant Growth · Nature's Embrace · New Horizons · Huatli, Poet of Unity (chapter II) · Enduring Vitality · Great Divide Guide

## The engine half

An any-color mana ability activated without a `manaColorChoice` pauses for the colour decision, so `ActivateAbilityHandler` returns **before** the whole post-effect pass — and nothing supplies a `manaColorChoice` from `legalactions/` or `ai/`, so this is the path the AI and the gym always take. `ColorChoiceContinuationResumer` re-applied exactly one payoff (the mirror-colour bonus), silently dropping everything else, including the Cub's fixed `+{G}`.

Rather than porting payoffs one at a time, the whole post-effect sequence moved into a new **`ManaAbilityResolutionPipeline`** (`applyLandManaDampening` + `finishTapBonuses`), and both callers now use it. `AdditionalManaOnSourceTapOnColorChoice` is deleted — one implementation, not two. `ActivateAbilityHandler` sheds ~330 lines.

Newly working on the pausing path:

| Behaviour | Concrete case |
|---|---|
| Damping Sphere replacement | New Horizons grants a land "add two mana of any color" |
| `AdditionalManaOnTap` auras | Fertile Ground / Elvish Guidance on an enchanted land |
| `AdditionalManaOnSourceTap` riders | Overabundance's 1 damage (its filter is any `Land`) |
| `AdditionalSourceTriggers` doubling | the bonus fired once instead of N+1 |
| `LandTappedForManaEvent` | emitted, at parity with the synchronous path |

The synchronous path is unchanged by construction: the code moved verbatim, same order, same event list.

**Known limit, unchanged by this PR:** the generic `landTappedForMana` *trigger* is still not wired to off-stack mana resolution. That was already true on both paths — see Groundchuck and Dirtbag, which uses `AdditionalManaOnSourceTap` for exactly that reason. This PR brings the *event* to parity, not the trigger.

## Stopping the third batch

`docs/card-sdk-language-reference.md` already warned about this exact footgun, and named Cryptolith Rite, Joiner Adept and Citanul Hierophants as prior victims. It didn't prevent the six above. So the invariant is now enforced instead of documented:

**`CardLinter.UnflaggedManaAbility`**, gated corpus-wide by the existing `CardLintTest`. By CR 605.1a an activated ability that could add mana, doesn't target, isn't a loyalty ability and moves no card to or from a library **is** a mana ability — the classification is a consequence, not a choice. The check is structural over the serialized tree, so a granted ability is covered wherever it nests, and it stops at nested abilities and embedded card definitions (Avatar Roku's `{8}` creates a token whose firebending adds mana; the `{8}` ability adds none).

It immediately found **two more**, both plain `activatedAbility { }` blocks the raw-constructor audit could not see:

- **Priest of Yawgmoth** (ATQ) — "{T}, Sacrifice an artifact: Add {B} equal to its mana value"
- **Tarrian's Soulcleaver** (LCI back face) — "{T}: Add {B}" on a land

The allowlist for the new rule is empty.

### One hole that only surfaced by fixing the Priest

Correctly flagging Priest of Yawgmoth broke its scenario test: as a real mana ability it resolves inline, and the inline `EffectContext` never carried `sacrificedPermanents` — so `EntityReference.Sacrificed(0)` read nothing and the ability produced **0** mana instead of the artifact's mana value. Threaded the snapshots through. Without this, making the card correct would have quietly broken it.

## Tests

- `EnduringVitalityScenarioTest` — the grant resolves off the stack, and is seen by a "tap a creature for mana" static. The payoff test activates **without** a `manaColorChoice`, so it genuinely pauses and covers the resume path.
- `BadgermoleCubTest` — the bonus survives the colour-choice pause, asserting `green == 1` so a double-application fails too.
- `ManaAbilityColorChoicePipelineTest` (new, engine-level, named for the mechanic) — the pausing path against Damping Sphere, a mirror bonus **with its rider**, an attached aura, and the land-tapped event. Mutation-checked: disabling the dampening call turns it red.
- `HuatliPoetOfUnityScenarioTest` updated — its grant now pauses for the R/G/W choice instead of using the stack.

Also in this PR: `CR 605.3a` → `605.3b` for "doesn't use the stack" (605.3a is the timing permission — verified against the Comprehensive Rules); five of the six repeated "the raw constructor defaults to false" KDoc paragraphs removed in favour of the lint; and the seven golden snapshots the flags changed are re-blessed.

## Verification

Full `test` suite: **BUILD SUCCESSFUL in 10m 44s**, zero failures, every module including all nine era suites.

Two notes on how that green was reached, for the record. An earlier full run died with *"Gradle build daemon disappeared"* after leaving six orphaned test workers at load average 25 — the environment's known OOM, not a code failure; two tests (`AbuJafarTest`, `AIPlayerTest`) had failed in it as 120-second timeouts and both pass in the clean serialized run. And `:mtg-sets:test` reported `UP-TO-DATE` in the final run, so its two gates were verified in dedicated runs beforehand: the snapshot re-bless, and `CardLintTest`. No TypeScript is touched, so no client gate applies.


bolav:
## Review of `babb8f5438` (`/review-changes`)

Posted for the record. Every finding below is addressed by the follow-up commit on this branch, **`be1ea2e12f` — "Share one mana-ability resolution pipeline across both entry points"**; resolutions are noted inline.

### Verdict

The diagnosis is right and the card-side fix is complete: after the change, a scan of the whole corpus finds **zero** remaining `ActivatedAbility(...)` constructions with a mana effect and no `isManaAbility`. The engine half is also a real hole correctly identified. But as submitted the change shipped **red** — six golden snapshots stale — and the engine fix widened a *partial copy* of the mana-ability pipeline rather than making that pipeline shared, so five other post-resolution steps still didn't run on the path these six cards now take.

Gates run at review time (branch 0 behind `origin/main`, clean tree, reviewed in place):

| Gate | Result |
|---|---|
| `scripts/test-class EnduringVitalityScenarioTest --rerun-tasks --no-build-cache` | ✅ 5m15s |
| `scripts/test-class BadgermoleCubTest --rerun-tasks --no-build-cache` | ✅ 4m37s |
| `scripts/test-class HuatliPoetOfUnityScenarioTest --rerun-tasks --no-build-cache` | ✅ 4m49s |
| `:mtg-sets:test --tests "*CardDefinitionSnapshotTest"` | ❌ **AVR, XLN, VOW, LCI, DSK, TLA fail** |

### What's good

- **Root cause is the real one, and the fix is corpus-complete** for the raw-constructor shape.
- **The `Mirror` → `OnColorChoice` rename fixes a name that had started to lie** once the object handled both `color = null` and fixed-colour forms.
- **`BadgermoleCubTest`'s new test exercises the path that was broken**, deliberately omitting `manaColorChoice` so the effect really pauses — and asserting `green shouldBe 1`, so a double-application would fail it too. Verified the premise: nothing under `rules-engine/legalactions/` or `ai/` ever sets `manaColorChoice`, so the AI/gym path genuinely always pauses.
- The commit message is unusually good — symptom, mechanism, blast radius, and why nothing double-applies.

### Blocking

**1. Six golden card snapshots are stale — `:mtg-sets:test` is red.** `CardDefinitionSnapshotTest` serializes `timing` and `isManaAbility` whenever non-default (compare `SOI.json:1062-1063` for Cryptolith Rite). None of the six changed cards' snapshots carry them, and no `.json` is in the diff. Confirmed by running it: all six sets fail `card trees match golden`.

> **Resolved.** Re-blessed with `-DupdateSnapshots=true`; seven files (the six plus `ATQ`, from a card the new lint surfaced).

### Important

**2. The colour-choice resume path is a partial duplicate of the mana-ability pipeline; the change extends the duplicate instead of sharing the pipeline.** `ActivateAbilityHandler` returns early on `effectResult.isPaused`, so *everything* after it is skipped — and the resumer re-adds exactly one of those steps. Still missing on the path these six cards now take:

- **Damping Sphere** — New Horizons grants a land "{T}: Add two mana of any color"; with Damping Sphere out that should become one `{C}`, and on the paused path it doesn't.
- **`resolveAdditionalManaOnTap`** — Fertile Ground or Elvish Guidance on the same land as Abundant Growth: the aura's bonus never fires.
- **`LandTappedForManaEvent`** — Mana Flare-style watchers never see the tap.
- **`onSourceTap.rider`** — Overabundance's `sourceFilter = GameObjectFilter.Land` (any land, any controller) matches a land tapped via any of these four land grants; the new file applies its mirror mana but drops the `DealDamageEffect(1, …)`.
- **`countAdditionalSourceTriggerDoublers`** — the inline pass does `repeat(1 + extraFirings)`; the new object applies the bonus exactly once, so a doubler stops doubling as soon as the producer is an any-colour source.

The KDoc's "the two paths are mutually exclusive … so nothing is applied twice" is accurate, but papers over the second path being a *subset*, not a mirror. Note there was by then a third copy of this rule in `ManaSolver.augmentWithSourceTapBonusMana`. Suggested fix: extract the block into a shared `finish(...)` and call it from both sites.

> **Resolved, via the extraction rather than the minimum.** New `ManaAbilityResolutionPipeline` (`applyLandManaDampening` + `finishTapBonuses`); both callers use it, `AdditionalManaOnSourceTapOnColorChoice` deleted, `ActivateAbilityHandler` ~330 lines lighter. New `ManaAbilityColorChoicePipelineTest` pins Damping Sphere, a mirror bonus with its rider, an attached aura, and the land-tapped event; mutation-checked.
>
> One correction to this finding: the generic `landTappedForMana` **trigger** was never wired to off-stack mana resolution on *either* path (Groundchuck and Dirtbag documents that and works around it), so listing it as a paused-path gap overstated it. Only the *event* was missing, and only the event is now at parity.

**3. Wrong CR number.** `EnduringVitalityScenarioTest` clues "A mana ability never uses the stack (CR 605.3a)". 605.3a is the *timing permission*; "doesn't go on the stack … resolves immediately" is **605.3b** (verified against `MagicCompRules_20260808.txt:2696-2698`).

> **Resolved.**

**4. The footgun is documented in eight places and prevented in none.** `docs/card-sdk-language-reference.md:1217` already states this rule and already names Cryptolith Rite, Joiner Adept and Citanul Hierophants as having shipped without the flags. This is the second batch, fixed the same way: six hand-written flag pairs. Suggested a `CardLinter` rule (and/or a `Patterns` facade) so the invariant can't be forgotten again.

> **Resolved.** `CardLinter.UnflaggedManaAbility`, gated by the existing `CardLintTest`, narrowed to CR 605.1a's carve-outs and stopped from descending into nested abilities and embedded token definitions. It immediately found two more real instances the raw-constructor scan could not see — **Priest of Yawgmoth** (ATQ) and **Tarrian's Soulcleaver** (LCI back face) — both plain `activatedAbility { }` blocks. Fixing the Priest then exposed a further engine hole: the inline mana-ability `EffectContext` never carried `sacrificedPermanents`, so its "add {B} equal to the sacrificed artifact's mana value" read nothing and produced 0. Also fixed.

### Minor

**5. Six near-identical KDoc paragraphs** all restating "the raw constructor defaults to `isManaAbility = false`". Per AGENTS.md, comments carry non-obvious *why*, and this why now lives in the language reference.

> **Resolved** — down to one, pointing at the lint.

**6. The Enduring Vitality payoff test never pauses.** It passes `manaColorChoice = Color.BLUE`, so the effect resolves synchronously and the *inline* pass supplies the bonus — the `(getPendingDecision() as? ChooseColorDecision)?.let { … }` block is unreachable.

> **Resolved** — the `manaColorChoice` is gone, so the test covers the resume path from the DSK side too.

**7. Dead pause guard.** `applyForResolvedTap` returns `ExecutionResult.success(...)` on every path, so `if (mirrored.isPaused) return mirrored` can't fire.

> **Resolved** — removed with the file.

PR #1905, opened 2026-08-18
Make Taster of Wares exile the chosen card, and end its cast window with the creature

bolav:
Taster of Wares never exiled anything, and its cast-from-exile permission ignored the "for as long as you control this creature" clause. This fixes both, and adds the `MayPlayExpiry` variant the clause needs.

**Oracle text** (verified against Scryfall, ECL #121):

> When this creature enters, target opponent reveals X cards from their hand, where X is the number of Goblins you control. You choose one of those cards. That player exiles it. If an instant or sorcery card is exiled this way, you may cast it for as long as you control this creature, and mana of any type can be spent to cast that spell.

## The bug

"You choose one of those cards" was modelled with `SelectionMode.ChooseUpTo`, which produces a `minSelections = 0` decision. A player clicking through — or the AI — could decline it, leaving the collection empty, so the exile step moved nothing and the whole trigger was a no-op. The clause is mandatory, so it is now `ChooseExactly`. That still resolves cleanly at the edges: an empty collection auto-selects nothing, and a count at or above the eligible set auto-selects all, so X = 0 and empty-hand cases never pause.

## The duration

The permission was granted with `MayPlayExpiry.Permanent` and a comment calling the control clause "a small simplification". It is now a real duration, `MayPlayExpiry.WhileYouControlSource`, mirroring the existing `Duration.WhileYouControlSource` field-for-field — same name, same `sourceDescription` parameter, same default.

The window closes when the granting permanent leaves the battlefield **or** when its *projected* controller stops being the grant's "you", so a Threaten-style steal ends it just as a destroy does. Per CR 611.2b it is one-way, which the engine implements in two places:

- `GrantMayPlayFromExileExecutor` refuses to create the grant at all when the source is already gone or already stolen at resolution — the rule's own Master Thief example.
- `EndedDurationExpiryCheck` *deletes* the permission once the window closes rather than gating it, so regaining control never revives it.

The exiled card stays exiled either way; only the permission ends. Phased-out sources fall out for free, since `getBattlefield()` already excludes them (CR 702.26f).

"You" is the player whose ability granted the permission, in every grouping. Under `ownerControls` the turn-keyed expiry anchor collapses to each card's owner, who never controls the source — so `expiryControllerId` is pinned for this expiry, and the window measures against the granting player rather than being deleted on the first state-based check.

The three `MayPlayPermission` lifecycle flags (`permanent`, `supersededBySameSource`, `endsWhenSourceUncontrolled`) are now derived from one exhaustive `when` over `MayPlayExpiry`. They are not independent — any expiry that is not turn-keyed must also set `permanent`, or cleanup takes the permission before its real end condition can. Deriving them separately let a new variant default silently to "expires this turn"; now it fails to compile until its cleanup behaviour is stated.

## Tests

`MayPlayWhileYouControlSourceTest` (engine, named for the mechanic) covers persistence across turns, revocation on the source leaving, revocation on a Threaten-style steal, no-revival after regaining control, no-grant-at-all when the source dies with the ability on the stack, the `ownerControls` window keying, and — as a contrast case so the two can't silently converge — a `MayPlayExpiry.Permanent` grant surviving the same destroy.

`TasterOfWaresScenarioTest` covers the card: the reveal-and-exile path asserting `minSelections == 1` (the actual regression, not just the end state), casting the exiled spell off Swamps only (which needs `withAnyManaType` to relax `{1}{R}`), the permission ending when the Taster dies while the card stays exiled, and a creature card granting no permission.

CR 611.2b was verified against the Comprehensive Rules text, including the Master Thief example.

## Gate

`:rules-engine:test :mtg-sets:test :mtg-sets:2026:tests:test` — 2,959 + 363 + 1,311 = **4,633 tests, 0 failures**, all three tasks executed fresh. That covers the engine, the corpus snapshot and JSON round-trip, and the ECL scenarios — every module this diff reaches. It says nothing about `game-server` or `web-client`, which the diff does not touch.

The ECL card snapshot is re-blessed in this branch; the regenerated diff is exactly the two intended lines, with no collateral drift elsewhere in the corpus.

`docs/card-sdk-language-reference.md` is updated in the same change.


bolav:
## Code review

Reviewed against `origin/main` (already an ancestor of this branch — no merge needed). Findings below were raised on the first two commits; the third commit (`Pin the source-keyed may-play window to the granting player`) is the fix-up, so each item records how it was resolved.

### Verdict

Behaviour is right and the SDK shape is right — `MayPlayExpiry.WhileYouControlSource` is a faithful mirror of the existing `Duration.WhileYouControlSource` rather than a new concept, and the CR 611.2b handling (never-open + delete-don't-gate) is correct and tested.

### What's good

- **The new expiry is a mirror, not an invention.** It matches `Duration.WhileYouControlSource` field-for-field, default-for-default (`Duration.kt:328-333`) — same name, same parameter, same description string, so a reader who knows one knows the other. The engine branch in `pruneEndedMayPlayPermissions` is likewise a line-for-line mirror of the `is Duration.WhileYouControlSource ->` branch at `EndedDurationExpiryCheck.kt:158-163`, including reading the *projected* controller. This is the right amount of new vocabulary: no existing `MayPlayExpiry` variant can express it, and the `condition` gate can't either, because a gate is reversible and 611.2b requires one-way.
- **CR 611.2b is cited accurately**, verified against the Comprehensive Rules text — including the Master Thief example, which really is in that rule and really does say the effect does nothing. Both halves are implemented *and* tested: the never-opens case (executor pre-check) and the never-revives case (physical deletion).
- **Phasing falls out for free.** `getBattlefield()` excludes phased-out permanents (`GameState.kt:903-921`), so CR 702.26f holds in both the executor pre-check and the SBA prune without anyone writing phasing code.
- **`endsWhenSourceUncontrolled` is guarded at construction** rather than degrading silently, and the SBA prune early-outs on `none { … }` so it never forces a projection for the states that have no such permission.
- **The contrast test is the good kind** — it pins `Permanent` surviving the same destroy that ends `WhileYouControlSource`, so the two can't silently converge later.
- **The actual bug fix is pinned, not just observed.** The scenario test asserts `minSelections == 1`, not merely that a card ended up exiled.

### Issues

**Blocking — stale ECL golden snapshot.** `CardDefinitionSnapshotTest > ECL: card trees match golden` failed: the committed golden still recorded `"type": "ChooseUpTo"` and `"expiry": "Permanent"`. Pure re-bless — the actual-vs-golden diff was exactly the two intended edits and nothing else, and the companion `cards survive a JSON round-trip` test passed, confirming the new polymorphic variant encodes and decodes correctly.
→ **Fixed:** regenerated with `-DupdateSnapshots=true`; the committed diff is those two lines only.

**Important — KDoc stated the opposite of the code for owner-scoped grants.** Both `MayPlayPermission` and `EndedDurationExpiryCheck` said an owner-scoped grant "still measures the window against the player whose ability granted it, not each card's owner". But `expiryControllerId = if (expiryAnchor != grantee) expiryAnchor else null` leaves it null under `ownerControls` (the anchor *is* the grantee), so "you" resolved to each card's **owner** — while the executor's own pre-check gated on the activating player. Not reachable by any shipped card, but a trap: such a grant would be deleted on the very next state-based check, since an owner never controls the source.
→ **Fixed in the code rather than the doc**, so the stated semantics become true: `expiryControllerId` is pinned to the activating player for this expiry in every grouping. Both KDoc paragraphs now say *how* it holds. A new test combines `ownerControls` with the source-keyed expiry — the one grouping where grantee and "you" diverge — and it was mutation-checked: reverting just that `when` turns exactly that test red and leaves the other six green.

**Minor — boolean-flag accretion on `MayPlayPermission`.** `permanent`, `supersededBySameSource` and `endsWhenSourceUncontrolled` between them re-encode four sealed `MayPlayExpiry` variants, and every one must be set in agreement or cleanup eats the permission first.
→ **Partially addressed.** The full "store the `MayPlayExpiry` on the permission" refactor was deliberately *not* done here: it reaches 12 construction sites plus every read site, and `permanent` is load-bearing for a second, unrelated job in `StackResolver` ("survives being cast from"). That is its own change, not a rider on a bug fix. What is fixed is the part that actually bites — the flags now come from one exhaustive `when` returning a named `CleanupBehavior`, so a new variant fails to compile until its cleanup behaviour is stated, instead of silently defaulting to "expires this turn".

**Minor — class KDoc gap.** `EndedDurationExpiryCheck`'s header enumerated the stores it prunes and omitted `mayPlayPermissions`.
→ **Fixed.**

### Gate

`:rules-engine:test :mtg-sets:test :mtg-sets:2026:tests:test` — **4,633 tests, 0 failures**, all three tasks executed fresh. Covers the engine, the corpus snapshot and round-trip, and the ECL scenarios. It is not evidence about `game-server` or `web-client`, neither of which this diff touches.

Issue #44, opened 2026-05-05
Heaped Harvest bug


bolav:
// When you sacrifice it, you may search for a basic land
    // (approximated as "when put into graveyard from battlefield")
    triggeredAbility {
        trigger = Triggers.PutIntoGraveyardFromBattlefield
        effect = MayEffect(
            EffectPatterns.searchLibrary(
                filter = GameObjectFilter.BasicLand,
                count = 1,
                destination = SearchDestination.BATTLEFIELD,
                entersTapped = true
            )
        )
    }

Will trigger on destroy.

Issue #10, opened 2026-02-12
When not in a game be able to see all the cards that are in the engine


Thullner:

wingedsheep:
http://magic.wingedsheep.com/deckbuilder

PR #1936, opened 2026-08-19
Add the Case mechanic (CR 719) and all twelve Murders at Karlov Manor Cases

bolav:
# Cases (CR 719) and all twelve Murders at Karlov Manor Case cards

Adds the Case card mechanic and the complete twelve-card Case cycle from Murders at Karlov Manor.

## The mechanic

A Case is an enchantment with two special abilities printed before a long dash (CR 719.3). Both lower
onto vocabulary the engine already had, so **Cases introduce no new ability kind**:

- **`toSolve(condition)`** — "To solve — [condition]" (CR 719.3a) is exactly *"At the beginning of your
  end step, if [condition] and this Case is not solved, this Case becomes solved."* It emits an ordinary
  `Triggers.YourEndStep` triggered ability whose `interveningIf` is `All(condition, Not(SourceIsSolved))`
  and whose effect is `Effects.BecomeSolved()`. Both halves are re-checked on resolution (CR 603.4).

- **`solvedStaticAbility { }` / `solvedTriggeredAbility { }` / `solvedActivatedAbility { }`** — "Solved —"
  (CR 702.169) is not one ability shape but three, each with its own gate, and the helpers apply the gate
  the rule actually specifies for each:

  | Shape | Rule | Gate applied |
  |---|---|---|
  | static | 702.169b | `condition = SourceIsSolved` |
  | triggered | 702.169c | `triggerRestriction = SourceIsSolved` |
  | activated | 702.169d | `ActivationRestriction.OnlyIfCondition(SourceIsSolved)` |

  The triggered form uses a **trigger restriction, not an intervening-if**, because 702.169c restricts
  *triggering* rather than resolution — an ability that triggered while the Case was solved still resolves
  even if the Case has left the battlefield by then. Each helper ANDs its gate with whatever
  condition/restriction the block sets itself.

The solved designation itself (CR 719.3b) is engine state, not an ability and not a copiable value:
`SolvedComponent` is stamped by `BecomeSolvedExecutor`, read through `Conditions.SourceIsSolved` /
`StatePredicate.IsSolved`, and dropped by `ZoneMovementUtils.stripBattlefieldComponents` when the
permanent leaves the battlefield. It is sticky — there is no "unsolve" — and a copy of a solved Case
enters unsolved.

## The twelve cards

Shattered Pact (#1), Gateway Express (#8), Pilfered Proof (#9), Uneaten Feast (#10), Filched Falcon (#44),
Ransacked Lab (#45), Gorgon's Kiss (#79), Stashed Skeleton (#80), Burning Masks (#113), Crimson Pulse (#114),
Locked Hothouse (#155), Trampled Garden (#156).

All twelve are MKM-only printings, so each canonical `card(...)` lives in `definitions/mkm/` with no
`Printing(...)` rows needed.

## New engine vocabulary the cards needed

- **`TurnTracker.DAMAGE_SOURCES`** (Burning Masks) — distinct sources a player controlled that dealt damage
  this turn, backed by `DamageSourcesThisTurnComponent`. Recorded on the *damage-time* controller, so a
  source that later dies or changes hands still counts. Identity is `(entity, battlefield-entry timestamp)`,
  so a permanent that deals damage repeatedly counts once while one that left and returned is a new object
  (CR 400.7).
- **`Conditions.CreaturesAttackedThisTurn`** (Gateway Express) — the player-agnostic sibling of
  `YouAttackedWithCreaturesThisTurn`, for text that names no player. Unions every player's attack record.
- **`Conditions.CreatureCardsPutIntoGraveyardsThisTurn`** (Gorgon's Kiss) — the game-wide sibling of the
  controller-scoped tracker, for "put into graveyard**s** from anywhere".
- **`CreateAdditionalToken.restrictions`** (Pilfered Proof) — a Solved static written as a replacement
  effect can't go through `solvedStaticAbility { }`, so it carries its gate here.
  `ReplacementEffectProcessor` now threads the replacement's own source into restriction evaluation, so a
  source-relative gate can resolve.
- **Graveyard-card-anchored `MayCastFromGraveyard`** (Uneaten Feast) — "creature cards in your graveyard
  gain 'You may cast this card from your graveyard'" grants to *cards*, so the affected set is fixed when
  the ability resolves (CR 611.2c) and a card arriving later that turn is not covered. Both read sites
  (`CastFromZoneEnumerator`, `CastZoneResolver`) treat a graveyard-card anchor as that one card's permission.

## Client

`CaseSolvedEvent` → `ClientEvent.CaseSolved`, `ClientCard.isSolved` → a `SOLVED` card badge, and the
card-grouping key. An **unsolved Case shows how close its criterion is** as a `current/required` progress
badge (Burning Masks counts 0/3 up to 3/3) and drops it once solved. That falls out of the generic
intervening-if badge learning to look inside the `All(criterion, Not(solved))` composite `toSolve` emits —
no Case-specific plumbing. The counts come from `ConditionEvaluator.countProgress`, which is the same
counting code the trigger runs, so a badge cannot promise progress the ability disagrees with.

## Tests

- `CaseSolveScenarioTest` — the mechanic: end-step solving, your-end-step-only, the CR 603.4 re-check,
  stickiness across cleanup, loss on zone change, non-copiability, and all three Solved ability shapes.
- `TriggerConditionProgressBadgeTest` — the progress badge across every countable clause shape.
- Nine per-card scenario tests.

## Known gaps

Three cards ship without a scenario test — Crimson Pulse, Shattered Pact, Trampled Garden — each covered
only by a manual dev scenario. Two rules the implementation is built around are also untested: the CR 400.7
re-incarnation of a blinked damage source, and the trigger-restriction-survives-resolution property that
motivates the `solvedTriggeredAbility` design. See the review comment on this PR for detail.

## Verification

`:rules-engine:test`, `:mtg-sets:2024:tests:test`, `:mtg-sets:test`, and `PoliticalTriumphScenarioTest`
(the pre-existing progress-badge consumer) all green — 4889 passed, 0 failed. `web-client` `npm run
typecheck` clean. Not run: `:game-server:test` (compile-checked only) and the other era test modules.


bolav:
# Review — Case mechanic (CR 719) + twelve MKM Cases

Reviewed with `origin/main` merged in (branch was already up to date with it, so no merge commit).

## Verdict

Behaviour is right and the SDK shape is right. The mechanic adds **no new ability kind** — `toSolve` and
the three `solved*Ability` helpers are the existing builders with a gate pre-applied, and the gate kind
matches CR 702.169b/c/d exactly per shape. Every CR number checks out, every card matches Scryfall
field-for-field, and the anti-corruption layer is fully wired. No blocking defects.

## What's good

- **The central design call is correct and non-obvious.** `solvedTriggeredAbility` uses `triggerRestriction`,
  not `interveningIf`. Verified in `TriggerMatcher.filterByTriggerCondition` and `TriggerProcessor` that a
  restriction is spent at trigger time and deliberately does *not* travel onto the stack — precisely
  CR 702.169c. Getting this backwards is the classic Solved bug and it was avoided.
- **Every cited CR number verifies** against the Comprehensive Rules: 719.3a/b/c, 702.169a–d, 111.6, 400.7,
  603.4, 601.2d, 611.2c. No drift in code, comments, or commit messages.
- **All twelve cards match Scryfall exactly** — oracle text, mana cost, type line, collector number, rarity,
  artist. All twelve are MKM-only printings, so canonical placement under `definitions/mkm/` is correct with
  zero `Printing(...)` drift.
- **`SolvedComponent` is wired at every read site**, including the ones that fail open if missed:
  `PredicateEvaluator`, `TriggerMatcher`, `AffectsFilterResolver`, the `BeginningPhaseManager` untap filter,
  and the `ZoneMovementUtils` strip.
- **The `countProgress` refactor is the right shape.** The boolean evaluators and the badge counter are now
  literally the same functions with a `cap` parameter, so a badge structurally cannot claim progress the
  trigger disagrees with. Keeping the first badge's `effectId` bare also preserves the existing
  `PoliticalTriumphScenarioTest` consumer.
- **Full client plumbing**, and the `ms-ability-case-solved` glyph genuinely exists in `mana-font` rather
  than rendering blank.
- **`CaseSolveScenarioTest` is a strong mechanic test** — stickiness, non-copiability, loss on zone change,
  the CR 603.4 re-check with the condition undone in response, your-end-step-only, and no re-trigger once
  solved.
- Doc sync in `card-sdk-language-reference.md` is thorough — every new primitive documented in the same change.

## Blocking

None.

## Important — still open

**I1 — Three of twelve cards ship with no scenario test.** `CaseOfTheCrimsonPulse`, `CaseOfTheShatteredPact`
and `CaseOfTheTrampledGarden` have only a `manual-scenarios/` JSON, which gates nothing. Trampled Garden is
the highest risk: `DistributeCountersAmongTargets(2)` with `count = 2, minCount = 1` (CR 601.2d) plus
`Triggers.YouAttack` targeting `TargetFilter.Creature.attacking()` is real wiring that could silently
misbehave. Shattered Pact needs one assertion that a gold permanent counts toward each of its colours and
the colourless Case counts for none.

**I2 — CR 400.7 source re-incarnation is implemented and cited but never tested.**
`DamageSourceIdentity.incarnation` exists solely so a blinked permanent counts as a second source; the claim
appears in a commit message, in `DamageUtils.trackDamageSourceForController`, in the `TurnTracker.DAMAGE_SOURCES`
KDoc, and in the language reference. Nothing blinks a source and re-deals damage. Suggested test in
`CaseOfTheBurningMasksScenarioTest`: creature pings, flicker it, ping again → the badge reads `2/3`, not `1/3`.
Delete the `incarnation` field today and every test still passes.

While writing that test, please settle the sibling case: `incarnation` is the *battlefield-entry* timestamp,
so it is `0L` for anything not on the battlefield. If the engine reuses the entity id when a card moves
graveyard → stack, a burn spell cast twice in one turn (flashback, or via this branch's own Case of the
Uneaten Feast) collapses to one source where CR 400.7 says two. I did not construct a repro — treat this as
a question the test should answer rather than a confirmed bug.

**I3 — The trigger-restriction-vs-intervening-if distinction is untested.** `CaseDsl.kt` and the language
reference both justify the design with *"an ability that triggered while solved still resolves even if the
Case has left the battlefield by then."* That is the whole reason for the choice and nothing pins it. It is
directly testable with the existing `triggeredCase` fixture: solve it, let the upkeep trigger go on the
stack, cast `Break the Case` in response, assert the draw still happens. Without it, a future refactor to
`interveningIf` passes the entire suite.

## Minor — addressed in `157be315`

- **M1** `DamageUtils.kt` — the new `trackDamageSourceForController` was inserted between
  `trackDamageSourceLki`'s KDoc and the function it documents, orphaning the doc. Reattached.
- **M2** `CaseOfTheUneatenFeastScenarioTest` — `setLifeTotal(p1, getLifeTotal(p1))` set life to its current
  value. Removed.
- **M3** Same file — the "amount, not the number of gains" test did five one-point gains, which is five
  events *and* five points, so it distinguished nothing; its inline comment described something else again.
  Split into three single-axis tests and added a five-life instant so one gain event actually proves the
  tracker reads life rather than gains.
- **M4** `CaseOfTheFilchedFalcon` — `addTypes = setOf("ARTIFACT")` was a no-op (`BecomeCreature` adds rather
  than replaces types, and the target filter already requires an artifact). Dropped, and the KDoc corrected:
  type retention comes from `SetCreatureSubtypes` stripping only creature subtypes, not from that argument.
  The card's own "still an artifact" assertion still passes.
- **M5** `ClientStateTransformer` — badge tooltips had silently changed for every pre-existing card
  (`condition.left.description` → the whole condition). Restored via `describeCountedClause`: a `Compare` is
  labelled with its left operand again, so Oversold Cemetery and Political Triumph read exactly as they did
  on `main`; the shapes this branch newly made countable have no left operand and describe themselves.
- **M6** `CaseSolveScenarioTest` — the CR 702.169d check matched any legal action whose `toString` mentioned
  the Case. Now matches `ActivateAbility.sourceId`/`abilityId`, and resolves the ability to confirm the 2 life.

**M7 remains open (author's discretion):** `CaseOfTheGorgonsKiss` cites the ruling that the tracker reads
the card's type line *in the graveyard* — so a creature card that was a noncreature permanent counts and an
animated noncreature does not — in both its KDoc and its `ruling(...)` metadata. Only the token half of that
pairing is tested.

## Recommendation

Merge-worthy in shape; the SDK needs no rework. Before merge, add the three missing card scenario tests (I1)
and the two tests for the rules the implementation is built around (I2, I3).

## Unrelated, noted not fixed

`DamageUtils.kt:394` emits `Condition is always 'true'`. It surfaced only because this branch forced a real
recompile of a file Gradle had been serving as `UP-TO-DATE`; the identical line is on `origin/main`, so it
predates this work and was left alone.

## Gate

`:rules-engine:test`, `:mtg-sets:2024:tests:test`, `:mtg-sets:test`, and `:mtg-sets:2026:tests:test
--tests *PoliticalTriumphScenarioTest*` — 4889 passed, 0 failed. `web-client` `npm run typecheck` clean.
Not covered: `:game-server:test` (compile-checked only), the other era test modules, and any runtime
rendering of the React changes beyond typecheck.

PR #1939, opened 2026-08-19
Add create-backlog-for-set skill

bolav:
# Add create-backlog-for-set skill

Starting work on a set means producing its `backlog/sets/<slug>/` entry, and until now that was
hand-rolled every time: paginate Scryfall, split booster cards from extras the way
`scripts/card-status` splits them, group by colour, cross-reference the Kotlin sources for what's
already implemented, and get the `cards.md` header format exactly right so `just check-backlog`
stays green. Done by hand it's slow, and it drifts from `card-status` — which is the number the
rest of the tooling treats as authoritative.

This adds a skill that does it, split so the machine does the mechanical half and the agent does
the judgement half.

## `bootstrap-set.py` — the mechanical half

Takes a set code or a name (`DRK` or `"The Dark"`; an ambiguous name exits with the candidate codes
rather than guessing) and writes:

- **`~/.cache/scryfall/_setdump-<code>.json`** — every printing in the set, `unique=prints`. It goes
  in the cache directory `scripts/card-status`, `:mtgish-tooling` and Argentum Assay already share,
  under a `_setdump-` prefix that can't collide with a set code (Assay's own entries use `_bulk-` /
  `_setlist-`). Freshness follows `card-status`: a set released more than 30 days ago is frozen, so a
  second run costs no network. Not committed.
- **`mtg-sets/<era>/…/definitions/<code>/<Name>Set.kt`** plus an empty `cards/`, when the set has no
  module yet. The era module is selected from the release year by parsing the existing
  `mtg-sets/<era>` directories, so it keeps working when a new era is appended; a year past the
  newest module is a hard error naming `settings.gradle.kts`, not a guess. `con`-style
  Windows-reserved codes map to the same alternate directory names the repo already uses.
- **`backlog/sets/<slug>/cards.md`** — the checklist, with colour sections, an `### Extras` section
  for non-booster cards, `[-] … (not planned: …)` for `coverage/card-exclusions.json` entries, and
  every already-implemented card pre-ticked by scanning `card("Name")` / `name = "Name"` out of the
  set's Kotlin sources.
- **`backlog/sets/<slug>/mechanics.md`** — a draft, marked as one.
- **`build/backlog/<code>-oracle.tsv`** — one row per card (name, rarity, cost, type, keywords,
  reprint flag, oracle text on a single line). Gitignored; it exists so the mechanics pass can read a
  whole set's text cheaply instead of 300 Scryfall lookups.

The draft/extras split and the exclusion handling deliberately mirror `scripts/card-status` rather
than reinventing a denominator, and the script imports `card_exclusions` and `set_dirs` from
`scripts/` instead of copying their logic.

## `SKILL.md` — the judgement half

The parts a script shouldn't decide are left to the agent and documented: what each mechanic means
and which CR rule defines it (verified against the local `MagicCompRules_*.txt`, never recalled),
whether the engine already models it (checked against `docs/card-sdk-language-reference.md` and the
SDK `Keyword` enum, with the primitive named), the set themes Scryfall has no keyword for, and the
three set-object fields that are a decision rather than a lookup — `block`, `basicLandsFallback`,
`sealedSupported`.

It also writes down the format rules other tooling depends on, because they're invisible failure
modes: `check-backlog-implementations.py` finds a set by the exact `# <Name> (<CODE>) - Card
Checklist` heading and matches card names verbatim, so an annotated entry like `- [ ] Foo (reprint)`
silently stops matching and can never be ticked by `--fix`.

Two guards worth calling out. Existing `cards.md` / `mechanics.md` are skipped unless `--force`, so
re-running against a curated backlog can't clobber it. And the skill explicitly says **not** to run
`scripts/gen-set-totals` to "register" a new set: it rewrites the committed 6 MB
`coverage/set-totals.json` for every set from the local Scryfall cache, so running it with a
partly-populated cache would silently empty other sets' totals.

## Verification

No Kotlin, TypeScript or Gradle-visible code changes: the diff is one markdown skill file, one
Python script under `.agents/skills/`, and two lines of `AGENTS.md`. Per `CLAUDE.local.md`'s gate
table that's a docs/tooling change, so no Kotlin suite was run — and a green Kotlin suite would say
nothing about this diff anyway.

What was actually exercised, end to end against live Scryfall:

- **`DRK` (scaffolded set, partially implemented)** — produced `46 / 119` with the per-colour
  breakdown, matching `scripts/card-status --set DRK` exactly. `scripts/check-card-counts.py
  --check` reports all 14 backlog headers in sync, and `scripts/check-backlog-implementations.py
  --check` reports no drift for the generated file (the drift it does print is pre-existing in
  `bloomburrow-commander`).
- **`S00` (unscaffolded set, by name — "Starter 2000")** — exercised name resolution, the scaffold
  path and era selection (2000 → `mtg-sets/2000-2002`), and the all-extras case (0 draft / 20
  extra). This surfaced a wrong era-module path calculation, which is fixed. Those test artifacts
  were then deleted.
- **Idempotency** — a second run re-uses the cached dump, reports `set exists`, and skips both
  backlog files.
- **Every scaffolded set (163 of them)** — the review pass checked the derived Kotlin object name
  and the "already scaffolded?" guard against each one. That found a real defect: the guard keyed on
  the derived name, but the repo shortens long titles (`LostCavernsOfIxalanSet`), so running against
  one of those 19 sets would have written a *second* `MtgSet` object into the same package for
  `MtgSetCatalog` to discover. Fixed in the second commit, along with apostrophe handling
  ("Urza's Saga" derived `UrzaSSagaSet`). Re-verified across all 163 afterwards.
- `python3 -m ast` parse and `--help` on the committed file.

Not covered: no automated test asserts the generated markdown shape. The three repo checkers above
are the real guard on it, and they run against the output.

## Deliberately not included

- **`backlog/sets/the-dark/`** — the DRK backlog produced while testing the skill is real, useful
  content, but it's set content rather than tooling, so it stays out of this PR.
- The script lives in the skill directory rather than `scripts/`. It's skill-specific bootstrap that
  nothing else calls, not a gate; if it earns wider use it can move.

### DIFF
--- a/.agents/skills/create-backlog-for-set/SKILL.md
+++ b/.agents/skills/create-backlog-for-set/SKILL.md
@@ -1,0 +1,141 @@
+---
+name: create-backlog-for-set
+description: Create the `backlog/sets/<set-name>/` entry for a Magic set — download the whole set from Scryfall (cached alongside Assay's cache), write `cards.md` as a per-colour checklist with the already-implemented cards ticked, scaffold the set's `definitions/<code>/` module if it doesn't exist, and write `mechanics.md` ordering every mechanic the set uses with a checkbox, a description, and its card list. Use when asked to "create a backlog for set X", "add set X to the backlog", "start work on set X", "scaffold set X", or when a set has no `backlog/sets/` entry yet.
+argument-hint: <SET code or name>
+---
+
+# Create a set's backlog
+
+Turn a set name or code into the three things the rest of the tooling needs: a scaffolded set
+module, `backlog/sets/<set-name>/cards.md` (the checklist `add-card`, `add-random-card`, `set-loop`
+and `just check-backlog` all read), and `backlog/sets/<slug>/mechanics.md` (the mechanic-by-mechanic
+map that says what `add-feature` work the set implies).
+
+The mechanical half is [`bootstrap-set.py`](bootstrap-set.py) — set resolution, the Scryfall
+download, the draft/extras split, colour grouping, ticking what's already implemented, counting.
+The judgement half is yours: what each mechanic *means*, whether the engine already models it, and
+the set themes Scryfall has no keyword for. **Never hand-roll the mechanical half** — the counts
+have to match `scripts/card-status` exactly or `just check-backlog` goes red.
+
+## Step 1 — run the bootstrap
+
+```bash
+python3 .agents/skills/create-backlog-for-set/bootstrap-set.py "<code or name>"
+```
+
+Accepts either form: `DRK` or `"The Dark"`. An ambiguous name exits with the matching codes — pick
+one and re-run. Useful flags:
+
+| Flag | When |
+|------|------|
+| `--slug <name>` | the backlog directory should not be the kebab-cased set name |
+| `--no-scaffold` | backlog files only; don't touch `mtg-sets/` |
+| `--refresh` | re-download the dump (a set still in spoiler season, or a suspect cache) |
+| `--force` | overwrite an existing `cards.md` / `mechanics.md` |
+| `--object-name` | the derived Kotlin object name is wrong or invalid |
+
+It prints one line per artifact. What it writes:
+
+- `~/.cache/scryfall/_setdump-<code>.json` — every printing in the set, `unique=prints`. Cached in
+  the same directory `scripts/card-status`, `:mtgish-tooling` and Argentum Assay use, under a
+  `_setdump-` prefix that can't collide with a set code. Released sets are frozen, so a second run
+  costs no network. Not committed.
+- `mtg-sets/<era>/…/definitions/<code>/<Name>Set.kt` + an empty `cards/`, if the set had no module.
+  The era comes from the release year; a year past the newest era module is an error, not a guess
+  (a new era needs `settings.gradle.kts`, which is `add-feature` work).
+- `backlog/sets/<slug>/cards.md` — the checklist.
+- `backlog/sets/<slug>/mechanics.md` — a **draft**, marked as one. Step 3 is what finishes it.
+- `build/backlog/<code>-oracle.tsv` — one row per card (name, rarity, cost, type, keywords, reprint
+  flag, oracle text on one line). Gitignored; it exists so Step 3 can read a whole set's text
+  cheaply.
+
+**`--force` overwrites a hand-maintained file.** Without it the script skips files that already
+exist, which is the right default when re-running against a set someone has been curating. Only
+force when you mean to discard what's there.
+
+## Step 2 — finish the scaffold (only if it created one)
+
+The generated `<Name>Set.kt` carries `code`, `displayName`, `releaseDate` and `incomplete = true`.
+Three fields need a human decision, so the script leaves them off:
+
+- **`block`** — the block the set belongs to, if any (`override val block = "Shadowmoor"`).
+- **`basicLandsFallback`** — only when the set prints no basics of its own. Point it at the set
+  whose lands should be registered alongside it: its block's base set, else `PortalSet`.
+  Check the dump: `grep -o '"type_line":"Basic Land[^"]*"' ~/.cache/scryfall/_setdump-<code>.json | sort -u`.
+- **`sealedSupported = false`** — for sets that can't carry a limited environment at all
+  (Commander/duel decks, planar sets, promo sets). A merely unfinished set doesn't need it;
+  `incomplete = true` already keeps it out of the picker's default list.
+
+Confirm the set is discoverable before moving on — `scripts/card-status --set <CODE>` must report
+it (it reads `code` out of the Kotlin source, so a set that doesn't show up is a scaffold problem).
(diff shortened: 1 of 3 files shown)


bolav:
## Review — `create-backlog-for-set` skill

**Verdict:** Ship it. The split is right — the script owns the parts that must agree with
`scripts/card-status` exactly, the skill owns the parts that need judgement — and the generated
`cards.md` passes all three repo checkers against a real set. One blocking defect was found during
review and fixed on the branch (`aae337d`); what's left is style-level.

No SDK types, no cards, no engine code, no CR citations in the diff, so §2/§2b/§3/§4 of the review
checklist don't apply. `origin/main` is already an ancestor of the branch — no merge commit.

### What's good

- **Reuses the repo's own definitions instead of re-deriving them.** It imports `card_exclusions`
  and `set_dirs` from `scripts/` rather than copying the exclusion manifest or hardcoding a
  definitions root, and it partitions draft/extras with the same "`booster` OR-ed across a name's
  printings" rule `card-status` uses (`bootstrap-set.py:227`) — the rule that exists because reading
  the flag off an arbitrary printing mis-filed 14 Foundations cards. The counts therefore match
  `card-status` rather than approximating it.
- **The cache placement is thought through.** `_setdump-<code>.json` shares the directory
  `card-status` / `:mtgish-tooling` / Assay already use, under a prefix that can't collide with a
  set code, with the same 30-day released-set freshness rule.
- **Judgement is not automated.** `block`, `basicLandsFallback` and `sealedSupported` are left off
  the generated set object and called out in SKILL.md Step 2; the mechanics file is emitted marked
  `DRAFT` with `_unverified_` engine-support lines rather than a guess. That's the right side of the
  line for a generator.
- **The format-rules section earns its place.** "A card line is `- [x] Name` and nothing after the
  name" is exactly the kind of invisible constraint that costs an afternoon —
  `check-backlog-implementations.py` matches names verbatim, so an annotated entry silently stops
  matching and can never be ticked by `--fix`.
- **Refusing to guess.** An ambiguous set name exits with the candidate codes; a release year past
  the newest era module is a hard error naming `settings.gradle.kts` rather than a silent placement.

### Issues

**B1 — `scaffold_set` could write a second `MtgSet` into an already-scaffolded set. FIXED in
`aae337d`.** `bootstrap-set.py` keyed its "already scaffolded?" check on the object name derived
from the Scryfall display name, but the repo shortens long titles: "The Lost Caverns of Ixalan" is
`LostCavernsOfIxalanSet`. Running the bootstrap against LCI found nothing at
`TheLostCavernsOfIxalanSet.kt` and wrote a second set object into the same package, which
`MtgSetCatalog` discovers on the classpath alongside the real one. Checked against every scaffolded
set: **19 of 163 would have hit it.** The guard now looks for any `*Set.kt` in the directory;
re-verified that LCI, DRK and USG all return `created=False` and their existing file.

**I1 — apostrophes were treated as word separators. FIXED in `aae337d`.** `re.findall(r"[A-Za-z0-9]+")`
turned "Urza's Saga" into `UrzaSSagaSet` (repo: `UrzasSagaSet`), and the same for Urza's Destiny,
Urza's Legacy, Dragon's Maze, Baldur's Gate and Marvel's Spider-Man. Apostrophes are now stripped
before splitting, and a name already ending in "Set" ("Arena Beginner Set") no longer gets a second
one appended. Six of the 19 above were this bug rather than a house-style shortening.

**M1 — section totals and `Set Size` count `[-]` cards differently.** `write_cards_md` computes
`Set Size` / `Implemented` over *countable* names (excluded-and-unimplemented cards dropped, matching
`card-status`), but the per-section table rows count every name in the section. A set with
exclusions will show section totals summing to more than the header total. `just check-backlog` only
reads the header and the checkboxes, so nothing breaks — it just reads oddly. Antiquities is the set
that would show it.

**M2 — most of `RESERVED_DIR_NAMES` would produce invalid Kotlin.** `con` → `conflux` matches the
repo. The invented fallbacks (`prn-set`, `aux-set`, `com1-set`, …) are interpolated straight into
`package com.wingedsheep.mtg.sets.definitions.{dir_name}`, and a hyphen isn't legal in a package
segment. No real set has any of those codes, so this is unreachable today; either underscore them or
drop everything but the `con` entry that's actually load-bearing.

**M3 — a scaffolded set's empty `cards/` directory can't be committed.** Git doesn't track empty
directories, so a brand-new set lands as just the `.kt` file and the directory reappears when the
first card is written. Harmless, but the SKILL.md line "plus an empty `cards/`" slightly oversells
what ends up in the commit.

### Verification

Re-ran after the fixes: `bootstrap-set.py DRK` still reports `46 / 119` matching
`scripts/card-status --set DRK`; `check-card-counts.py --check` has all 14 headers in sync;
`check-backlog-implementations.py --check` reports no drift for the generated file. The
scaffold-guard and name-derivation checks were run against all 163 scaffolded sets, not a sample.

Not covered: no automated test pins the generated markdown shape — the three repo checkers are the
only guard, and they run against output rather than the generator.

### Recommendation

Merge. M1–M3 are the author's discretion; M2 is the one worth a two-line follow-up, since an invalid
package name would fail at compile time rather than at generation time and the mapping reads as
though it works.

PR #1942, opened 2026-08-19
Make card interaction work on touch devices

bolav:
# Make card interaction work on touch devices

Playing a game on a phone surfaced four faults. They look unrelated, but all four come from
one thing: the board is wired for a mouse, and a browser fakes mouse input from a tap.

A tap fires a compatibility mouse sequence — `mouseenter`, `mousemove`, `mousedown`, `mouseup`,
`click` — after `touchend`, and never the matching `mouseleave`. A page can normally suppress that
by calling `preventDefault()` on `touchstart`, which `GameCard.handlePointerDown` already does, but
it can't take effect: React registers `touchstart`, `touchmove` and `wheel` as **passive** listeners
on its root container (`react-dom-client.development.js:19251`), so `preventDefault()` inside
`onTouchStart` is a no-op. Anything wired to both touch and mouse therefore runs twice per tap.

## What that broke

- **Tapping a card in hand opened two things.** The phantom `mouseenter` opened the hover preview
  (`--z-tooltip`, 10000) on top of the action menu the same tap opened (`--z-modal`, 1000). No
  `mouseleave` follows a tap, so it stayed there.
- **Tapping a selected attacker blinked and re-selected it.** `touchend` deselected it; the phantom
  `mousedown` immediately after re-ran the attacker-drag handler, which selects an unselected
  attacker so its arrow can be dragged, and recorded `attackerWasSelected = false` so the following
  `mouseup` had nothing to undo.
- **The preview couldn't be dismissed.** Only `touchend` cleared a long-press preview, and a long
  press is exactly the gesture a mobile browser takes the touch away for (native text selection /
  callout) — it delivers `touchcancel` instead. The full-screen overlay was also
  `pointer-events: none`, so there was nothing to tap.
- **Nothing could be read.** Without hover there is no way to see a card's rules text, and the
  action menu's own thumbnail is 120×168 on a phone.

## The fixes

**Suppressing the echo.** Hover is now driven by pointer events gated on `pointerType !== 'touch'`
(`GameCard.tsx:301`, `StackZone.tsx:206`), so touch never opens the preview by accident. A shared
guard, `utils/ghostMouse.ts`, makes every mouse-bound *drag* handler ignore a tap's echo: touch
handlers stamp a timestamp, mouse handlers bail within 700ms of it. That is deliberately time-based
rather than keyed to the hover capability, so a real mouse plugged into a touchscreen device keeps
working. `touchcancel` is now handled alongside `touchend`.

**`useHasHover()`** reports the pointer's capability (`matchMedia('(hover: hover)')`) rather than the
viewport's width. This is not `useResponsive().isMobile`: a narrow desktop window still hovers, and a
landscape tablet still doesn't. Everything below keys off it, so desktop behaviour is untouched at
every width.

**"View card".** On a device that can't hover, the action menu carries a `View card` row, and opens
for cards with nothing to do at all — an opponent's permanent (rendered `interactive={false}`, so it
used to swallow the tap), a land already played this turn, a spell on the stack. The menu stays open
*beneath* the zoomed card, so dismissing it drops the player back on the same list of actions rather
than making them re-tap the card to play what they just read. `useOpenCardMenuOnTap()` holds that
rule once for its three callers, and returns `null` — leave the tap alone — with a mouse present, or
while spectating/replaying, where `GameBoard` doesn't mount the menu at all.

**The preview itself** is now the full-screen variant on any device that can't hover, not only at
phone widths (the cursor-following one has nowhere to anchor without a cursor), takes pointer events
so a tap anywhere dismisses it, and is sized to be read — `min(280px, 78vw, (100vh - 140px) / 1.4)`
rather than a fixed 200px. It stays `pointer-events: none` where hover exists: a tap-catching
backdrop over a hovered card would swallow the board's clicks and block the `mouseleave` that clears
it.

**Stack items** get the same tap and the same touch-guarded hover. Targeting and decision selection
still own the tap whenever either is running, so counterspell targeting is unaffected; a collapsed
pile still expands on tap, and its members are readable once expanded.

**AI Insight** is hidden on phones, matching the game log it sits beside — both are bottom-left
fixed buttons that land on top of the hand, and the log already carried a comment saying exactly
that.

## Also in here

The action menu now drops a selection whose card has vanished (`ActionMenu.tsx:78`). It used to
close on its own by requiring a legal action, which a resolved spell no longer has; now that
`View card` alone can hold it open, a spell resolving off the stack would leave the menu behind.

## Deliberately not done

- The zone browsers (`ZonePiles.tsx`, `DeckBrowser.tsx`) still open the preview from raw
  `onMouseEnter`, so a tap there opens it as a side effect. That is no longer a dead end — the
  preview is dismissible now — but the handlers should get the same `pointerType` guard. Left out to
  keep this diff to the reported flows.
- Blocker/attacker mode still doesn't open the view menu on tap, to stay clear of the drag-based
  block assignment. Long-press still works there.
- No test covers any of this; see below.

## Verification

`npm run typecheck`, `npm test` (41 files, 579 tests) and `npm run build` — all green, run from
`web-client/`.

**What that does not cover: the behaviour this PR is about.** The vitest suite here is pure logic —
there are no `.test.tsx` files and no testing-library in the project, so nothing renders a card or
dispatches a touch sequence. Every claim above about what a tap now does is reasoned from the event
model and read off the code, not observed in a test, and not yet confirmed on a device by me — the
faults were all reported from a phone, and the fixes want the same phone to sign them off. Adding
component-level tests would mean introducing a rendering harness to `web-client`, which felt out of
scope for a bug fix; happy to do it as a follow-up if wanted.

Worth checking by hand: tap a land in hand (menu, no preview), tap "View card" then anywhere (back
to the menu), tap an attacker twice (selects, then deselects and stays deselected), drag an attacker
to a planeswalker (target still assigns), tap a spell on the stack, and cast a counterspell at one
(targeting still wins the tap).

No Kotlin gate was run and none applies — the diff is ten files, all under `web-client/src`.

### DIFF
--- a/web-client/src/components/game/GameBoard.tsx
+++ b/web-client/src/components/game/GameBoard.tsx
@@ -1979,7 +1979,10 @@ export function GameBoard({ spectatorMode = false, topOffset = 0 }: GameBoardPro
           that size anyway. */}
       {!spectatorMode && !responsive.isMobile && <GameLog />}
       {!spectatorMode && <ActiveYieldsPanel />}
-      {!spectatorMode && <AiInsightPanel />}
+      {/* Hidden on phones for the same reason as the log: its toggle sits in the
+          bottom-left corner, directly on top of the hand, and the expanded panel
+          is a 520px table that a phone can't show. */}
+      {!spectatorMode && !responsive.isMobile && <AiInsightPanel />}
 
       {/* Draw animations */}
       <DrawAnimations />
--- a/web-client/src/components/game/board/StackZone.tsx
+++ b/web-client/src/components/game/board/StackZone.tsx
@@ -8,6 +8,7 @@ import type { ClientAbilityIdentity, ClientCard } from '@/types/gameState'
 import { getCardImageUrl } from '@/utils/cardImages.ts'
 import { ActiveEffectBadges } from '../card/CardOverlays'
 import { AbilityText, ManaCost } from '../../ui/ManaSymbols'
+import { useOpenCardMenuOnTap } from '@/hooks/useOpenCardMenuOnTap.ts'
 import { useResponsiveContext, handleImageError } from './shared'
 import { styles } from './styles'
 import { groupStackCards, type StackGroup } from './stackGrouping'
@@ -26,6 +27,7 @@ export function StackDisplay() {
   const stackCards = useStackCards()
   const responsive = useResponsiveContext()
   const hoverCard = useGameStore((state) => state.hoverCard)
+  const openCardMenuOnTap = useOpenCardMenuOnTap()
   const targetingState = useGameStore((state) => state.targetingState)
   const addTarget = useGameStore((state) => state.addTarget)
   const removeTarget = useGameStore((state) => state.removeTarget)
@@ -99,6 +101,20 @@ export function StackDisplay() {
     } else if (isValidTarget) {
       addTarget(cardId)
     }
+  }
+
+  /**
+   * A tap/click on a stack item. Targeting and decision selection own it whenever either is
+   * running; otherwise the gesture only ever means "let me read this", which without hover has to
+   * go through the action menu's "View card" row — the same route as any other card the player
+   * can't do anything with.
+   */
+  const handleStackItemTap = (cardId: EntityId) => {
+    if (targetingState || decisionSelectionState) {
+      handleStackItemClick(cardId)
+      return
+    }
+    openCardMenuOnTap?.(cardId)
   }
 
   // A card the player can currently target/select — such a card must never be hidden inside a
@@ -187,10 +203,13 @@ export function StackDisplay() {
           ...seatBorderFor(card.controllerId),
           ...highlight,
         }}
-        onClick={opts.onClick ?? (() => handleStackItemClick(card.id))}
+        onClick={opts.onClick ?? (() => handleStackItemTap(card.id))}
         onContextMenu={(e) => openYieldMenu(card, e)}
-        onMouseEnter={(e) => hoverCard(card.id, { x: e.clientX, y: e.clientY })}
-        onMouseLeave={() => hoverCard(null)}
+        /* Pointer events with a touch guard, as on GameCard: a tap synthesizes mouseenter and
+           never the matching mouseleave, which used to open the preview on top of whatever the
+           same tap opened and leave it stranded there. */
+        onPointerEnter={(e) => { if (e.pointerType !== 'touch') hoverCard(card.id, { x: e.clientX, y: e.clientY }) }}
+        onPointerLeave={(e) => { if (e.pointerType !== 'touch') hoverCard(null) }}
       >
         {(() => {
           // A sideways-printed card (split layout, Room, battle) reads landscape only once rotated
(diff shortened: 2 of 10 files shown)


bolav:
# Review — `mobile-client` (touch card interaction)

Reviewed `origin/main...HEAD` in place; the branch already has `origin/main` merged (it was
already up to date, no merge commit). Two commits, ten files, all under `web-client/src`.

## Verdict

The diagnosis is right and the fixes are aimed at the cause rather than the symptom: React's passive
`touchstart` registration is verifiable in `node_modules/react-dom/cjs/react-dom-client.development.js:19251`,
and every fault reported follows from it. Behaviour on a device with a mouse is unchanged at every
viewport width. The gap is test coverage — there is none, and this project has no harness that could
carry any.

The SDK-elegance lens this skill leads with doesn't apply: no Kotlin, no `mtg-sdk` types, no cards,
no CR citations to verify.

## What's good

- **The gate is a capability, not a width.** `useHasHover()` (`hooks/useHasHover.ts`) draws the line
  where it actually falls. Using `responsive.isMobile` would have broken hover in a narrow desktop
  window and left a landscape tablet unfixed, and the file says so rather than leaving the next
  reader to rediscover it.
- **Two guards, each at the right grain.** Hover uses `pointerType` (exact, per-event); the drag
  handlers use a time window (`utils/ghostMouse.ts`), which is the only thing that also survives a
  mouse attached to a touchscreen. Reaching for one mechanism for both would have been worse in
  either direction.
- **`useOpenCardMenuOnTap` returning `null`** rather than a boolean puts the "leave the tap alone"
  decision in one place and makes the three call sites read as one rule. The spectator/replay case
  is handled there rather than being rediscovered per caller.
- **Comments explain the mechanism, not the code.** `utils/ghostMouse.ts` in particular will save
  the next person the hour it took to find this.
- The follow-on hazard was caught and closed: `ActionMenu.tsx:78` drops a selection whose card has
  vanished, which only became possible because `View card` can now hold the menu open.

## Issues

### Important

**I1 — Nothing tests any of this. `web-client/src/**`**
Every behavioural claim in the PR body is derived from the event model, not observed. The suite is
41 pure-logic files with no component rendering, so the regression this branch is most exposed to —
someone re-adding an `onMouseEnter` to a card, or dropping `noteTouchInteraction()` from a touch
handler — would go unnoticed. The honest options are (a) accept it as untested UI, which is what the
repo does everywhere else in `web-client`, or (b) add `@testing-library/react` and one test that
dispatches `touchstart`/`touchend` plus the compatibility mouse pair at a `GameCard` and asserts a
single toggle. (b) is a real dependency decision and belongs to the maintainer, not to a bug fix.
Flagged rather than fixed.

**I2 — `noteTouchInteraction()` is only stamped by `GameCard`. `web-client/src/utils/ghostMouse.ts:24`**
The guard reads global state but is written from one component's touch handlers. A tap on any other
element — an `ActionMenu` button, the opponent rail — produces an unmarked ghost sequence, so a
`GameCard` window-level `mouseup` listener that happened to be live at that moment would act on it.
No path today puts a drag in flight while another element is being tapped, so this isn't reachable;
it is a sharp edge for whoever wires the next touch handler. Stamping it from a single
capture-phase `touchstart` listener on `document` would make the guard self-maintaining.

### Minor

**M1 — Blocker mode can't reach `View card`. `web-client/src/components/game/card/GameCard.tsx:1109`**
The new tap branch is guarded on `!isInCombatMode`, so declaring blocks is exactly when an
opponent's attacker can't be read through the menu. That guard is defensible — blocks are assigned
by drag, and a tap that opens a menu mid-assignment would be worse — and long-press still shows the
card. Worth revisiting once the touch block flow gets attention; the PR body lists it as knowingly
out of scope.

**M2 — `100vh` in the preview sizing. `web-client/src/components/game/card/CardPreview.tsx:453`**
On iOS Safari `100vh` is the URL-bar-hidden viewport, so the card can be slightly taller than the
visible area when the bar is showing. The `- 140px` allowance absorbs it in practice. `100dvh` is
the correct unit but can't be used bare inside the `min()` without a fallback, since an unsupported
unit invalidates the whole declaration and would leave the element unsized.

**M3 — Ten-file diff, one commit.** The two halves (suppress the phantom events / add the touch
affordance) are separable in principle but not in `git`: both live in the same regions of
`GameCard.tsx`, and splitting them would have needed `git add -p`, which isn't available in this
environment. Called out so the size doesn't read as carelessness.

## Fixed during review

`hooks/useHasHover.ts` cached the media-query result in a module-level boolean that only the
`change` listener ever wrote, so a capability change landing while nothing was subscribed was never
picked up. `getSnapshot` now reads `.matches` off the singleton directly — the allocation worth
hoisting was `matchMedia()`, not the property read. Fixed in `bb20917b0c`.

## Recommendation

Merge once it has been confirmed on a phone — that is the only gate that can actually see these
fixes, and no automated one here can substitute. I2 is worth doing at the same time if the guard is
expected to grow more callers; I1 is a maintainer call about taking on a rendering harness.

PR #1958, opened 2026-08-20
Retry ready tournament pairs when an earlier match unblocks them

### LINKED ISSUE #1952: Tournament vs AI progression can stall after an early match result
## Summary

At upstream `main` (`2bc53aa1365b982615446467bb9a3d63b2e06113`), the tournament-vs-AI progression path has a reproducible liveness problem when a human finishes before the rest of the current round. The server can mark submitted AIs ready while their next scheduled match is still blocked by an earlier incomplete match, but it does not retry those already-ready AIs when that blocker finishes. The full-round completion path eventually performs another AI-ready pass, so this is not necessarily a permanent deadlock; it can nevertheless leave an otherwise-ready match unstarted, discard an early human ready action, and make the client show the next round too early.

## Deterministic reproduction

1. Create a sealed tournament with one human and seven AI players; submit all decks and start the tournament.
2. Verify that the first server `TournamentMatchStarting` event is round 1, not round 2.
3. Finish the human's round-1 match while at least one other round-1 match is still incomplete.
4. Let the server auto-ready the submitted AIs. For an AI whose next unplayed match is in a later round, `tryStartMatchForPlayer` can return because that AI or its opponent still has an incomplete earlier match.
5. Finish that earlier match while the current round is still incomplete. The result handler invokes `autoReadyAiPlayers` again, but the method returns once every submitted AI is already in the ready set; the per-AI `tryStartMatchForPlayer` call only happens on a false-to-true `markPlayerReady` transition. No new ready event is generated, so the ready pair is not reconsidered on this transition.
6. On the human client, observe that `MatchComplete` clears the current session and stores `lastRoundResults`; the overlay then displays `Round 2` and a Ready button even though other round-1 matches remain. If the human clicks Ready at this point, the server can reject the attempt through the existing earlier-match guard. When the current round finally completes, `doHandleRoundComplete` clears the ready state, so the human may have to click Ready again.

The current source path is:

- The result handler auto-readies AIs before it checks whether the whole current round is complete: [TournamentMatchHandler.kt#L115-L142](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/TournamentMatchHandler.kt#L115-L142).
- `autoReadyAiPlayers` exits if no submitted AI is newly unready and only retries a player after `markPlayerReady` returns true: [TournamentMatchHandler.kt#L803-L847](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/TournamentMatchHandler.kt#L803-L847).
- Match lookup scans all rounds, while the existing guard intentionally refuses to start a later match while either player has an earlier incomplete match: [TournamentManager.kt#L577-L618](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/tournament/TournamentManager.kt#L577-L618), [TournamentMatchHandler.kt#L381-L417](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/TournamentMatchHandler.kt#L381-L417).
- A later fallback exists only at full current-round completion, where ready state is cleared before AIs are processed again: [TournamentMatchHandler.kt#L258-L326](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/TournamentMatchHandler.kt#L258-L326).

## Separate confirmed client issue

The server's AI-ready mutation is not broadcast. The AI branch calls `markPlayerReady` and `tryStartMatchForPlayer`, whereas the human auto-ready branch calls `broadcastReadyStatus`: [TournamentMatchHandler.kt#L839-L890](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/TournamentMatchHandler.kt#L839-L890). Connected clients therefore can show stale AI-ready indicators; reconnect handling gets a fresh snapshot. This should be fixed or explicitly covered as part of the same state-transition contract.

## Round display and first-round check

The client treats any `lastRoundResults` plus no current game as an inter-round state and renders `currentRound + 1`: [lobbyHandlers.ts#L217-L238](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/web-client/src/store/slices/handlers/lobbyHandlers.ts#L217-L238), [TournamentOverlay.tsx#L106-L112](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/web-client/src/components/tournament/TournamentOverlay.tsx#L106-L112), [TournamentOverlay.tsx#L157-L160](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/web-client/src/components/tournament/TournamentOverlay.tsx#L157-L160). The waiting message is consequently hidden and the Ready action is shown: [TournamentOverlay.tsx#L205-L254](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/web-client/src/components/tournament/TournamentOverlay.tsx#L205-L254).

This is a display/state interpretation bug, not evidence that the server skips the first round. Schedule generation starts at round number 1 and the match-start message forwards the scheduled round number: [TournamentManager.kt#L166-L227](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/tournament/TournamentManager.kt#L166-L227), [TournamentMatchHandler.kt#L545-L559](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/TournamentMatchHandler.kt#L545-L559).

## Existing functionality checked

- Live Matches is already implemented end to end: the server builds active matches across rounds and broadcasts them, and the overlay renders a spectate action: [SpectatingHandler.kt#L135-L210](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/SpectatingHandler.kt#L135-L210), [TournamentOverlay.tsx#L447-L478](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/web-client/src/components/tournament/TournamentOverlay.tsx#L447-L478). This issue is not requesting a missing Live Matches feature.
- Save Deck and View Deck are present, and tournament deck state is intentionally preserved through gameplay/menu transitions: [TournamentOverlay.tsx#L182-L201](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/web-client/src/components/tournament/TournamentOverlay.tsx#L182-L201), [gameplayHandlers.ts#L541-L615](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/web-client/src/store/slices/handlers/gameplayHandlers.ts#L541-L615). Any save/view failure should be reproduced separately; no additional root cause is claimed here.
- `readyEpoch` is currently declared and incremented but has no read/use in the current server sources: [TournamentLobby.kt#L707-L713](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/lobby/TournamentLobby.kt#L707-L713), [TournamentLobby.kt#L1981-L2007](https://github.com/wingedsheep/argentum-engine/blob/2bc53aa1365b982615446467bb9a3d63b2e06113/game-server/src/main/kotlin/com/wingedsheep/gameserver/lobby/TournamentLobby.kt#L1981-L2007). This is maintenance context, not asserted as the cause.

## Expected behavior / acceptance criteria

- After an earlier match completes, every ready pair whose earlier-match guards are now clear is reconsidered without requiring a new false-to-true Ready transition.
- The existing `hasIncompleteMatchBefore` safety invariant remains intact; do not solve this by allowing unsafe cross-round starts.
- A human's Ready action remains gated behind dismissal of the game-over overlay, but it is not silently lost merely because it was clicked while other current-round matches were finishing.
- AI-ready state is broadcast consistently, or the protocol documents and tests the authoritative snapshot semantics.
- The overlay distinguishes “current round still has matches in progress” from “ready for the next round”; it must not label an early finisher as the next round solely because `lastRoundResults` is non-null.
- Add a deterministic regression test for one human plus seven AI players, including an early human finisher, a blocked later match, blocker completion, and the resulting retry.

## Related history

- [Tournament: wait for human ready before starting next AI game](https://github.com/wingedsheep/argentum-engine/commit/0754e085924a190ecb20f846cbc8bef2619aab97)
- [Fix race condition in sealed tournament AI match starting](https://github.com/wingedsheep/argentum-engine/commit/44c50f33fd26169a3d8e585053768aac9c730f72)
- [Fix tournament matchmaking bugs: stale opponent name and incomplete cross-round check](https://github.com/wingedsheep/argentum-engine/commit/2c199a9afdbe1c3577a82ef58c10834a285f7212)
- [Fix tournament games starting without Ready between rounds](https://github.com/wingedsheep/argentum-engine/commit/ba19ba99474db0f434e8950965a757b74e8e3023)
- [Fix tournament lobby not showing all active live games](https://github.com/wingedsheep/argentum-engine/commit/cc5d3673027c606963772f52e5c525e7504c863c)
- [Fix tournament deck editing after page refresh](https://github.com/wingedsheep/argentum-engine/commit/0f9d47125fc93e9fdb77414a5cb1227b2a8b64a0)

## Verification

Inspected upstream `main` at `2bc53aa1365b982615446467bb9a3d63b2e06113`; no production code or tests were changed for this report. Existing tournament tests do not cover the one-human/seven-AI early-finisher sequence.

wingedsheep:
Fixes #1952.

## The stall

A ready flag is consumed only when a match actually *starts*. So a pair that
`hasIncompleteMatchBefore` refused — because either seat still owed an earlier-round game — stays
ready, and `markPlayerReady` will never transition false→true for them again. `autoReadyAiPlayers`
returned early once every submitted AI was already in the ready set, and only called
`tryStartMatchForPlayer` on that transition, so nothing ever reconsidered the pair when the blocking
game finished. With one human and seven AI seats, the human's early finish left the entire round-2
slate idle.

The reporter noted a full-round fallback that recovers this. It doesn't always fire: `autoReadyAiPlayers`
ran *before* the round-completion check and prepares the next round itself, so when the last match of a
round completed with any AI unready, `isRoundComplete()` reported on the freshly-prepared round and
`RoundComplete` was never broadcast at all. No fallback, no clear, no retry — a permanent deadlock, and
also why the client only ever saw `MatchComplete` and had to guess the round phase from it.

## The fix

**One pure query owns the decision.** `TournamentManager.startableMatches(readyPlayerIds)` returns every
pair that could launch right now. `hasIncompleteMatchBefore` is untouched and still gates each one, so no
unsafe cross-round start is introduced. A seat's earliest unplayed match blocks its later ones through
that same guard, so no seat appears twice and the whole sweep can be launched in one pass.

**Every result re-runs the sweep.** `startReadyMatches` replaces the per-player lookup. This is the
substance of the fix: a pair goes from blocked to launchable because a *third* pair's game finished, with
no change to the ready set, so looking at ready transitions can't work.

**Round completion is decided before readiness is touched**, so `RoundComplete` is always broadcast.

**Round completion no longer wipes ready flags.** Whoever is still ready at a round boundary never had
theirs consumed — an early finisher who dismissed the overlay and clicked Ready, a BYE, an AI — and none
of them has a game-over overlay a new game could start under. Wiping them discarded a deliberate Ready
and forced a second click.

**A ready click is refused while the sender's own match is still running.** They can't have dismissed the
game-over overlay yet, and banking it consents to the match *after* this one — which the sweep would then
launch the instant this one ended. `SealedTournamentReconnectionTest` caught exactly this: a third
player's Ready arrived just after their own round-2 match started, and the new sweep dutifully began
round 3 the moment round 2 ended.

**Ready state is broadcast consistently.** The AI branch now broadcasts like the human one, and the
authoritative set is re-stated after every result — `MatchComplete`/`RoundComplete` make the client drop
its list, and readies that survive a round boundary would otherwise vanish from the counter.
`PlayerReadyForRound.playerId`/`playerName` are nullable for a plain snapshot.

**The overlay stops mislabelling an early finisher.** `MatchComplete` gains `currentRoundComplete`;
`lastRoundResults` only ever said *our* match ended. `deriveRoundPhase` reads the server's answer, keeps
naming the round in progress, keeps the waiting notice up beside the Ready button, and labels the button
"Ready for Next Match" while the round runs on. Readying early is still allowed and still honoured — only
the labelling changed.

Also restores the `readyEpoch` staleness check, which lost its only reader when this handler was split out
of `LobbyHandler` (the reporter flagged it as dead). `resumeTournament` for extra rounds is now its live case.

## Verification

- `TournamentProgressionTest` (new, pure and deterministic): one human + seven AI, asserts the schedule
  opens on round 1, that all four round-1 pairs start, that the pair stays blocked through the human's
  early finish *and* through the AI auto-ready sweep, and that finishing the blocker — changing no ready
  flag — makes exactly that pair startable. Plus the no-seat-twice and not-ready cases, and the
  mid-match/waiting distinction behind the stale-Ready guard.
- `roundPhase.test.ts` (new): the early-finisher label, the genuine between-rounds label, and the
  in-match and tournament-over cases.
- `just test-server` green (496 tests) — including `SealedTournamentReconnectionTest`'s full multi-round
  flow, which failed against an earlier draft of this change and passes now.
- `npm run typecheck` and the full vitest suite (590 tests) green.

Not covered: reconnecting mid-tournament still resets the client to round 0 because `TournamentStarted`
carries no round number. Pre-existing, unchanged here, and it leaves the new phase derivation behaving
exactly as today on that path.

### DIFF
--- a/docs/architecture-principles.md
+++ b/docs/architecture-principles.md
@@ -1409,7 +1409,11 @@ the entire round to finish. This "eager match starting" pattern means:
 4. Other players continue waiting or playing their own matches
 
 A ready flag is state, not an event: it is set by a ready click (or the AI auto-ready pass) and
-consumed only when a match actually starts — a round ending does not clear it. That matters because a
+consumed only when a match actually starts — a round ending does not clear it. Because it doesn't, the
+round-complete path opens the next round explicitly rather than as a side effect of the AI-ready pass;
+`currentRound` is what `isRoundComplete` reads, so leaving it on a finished round makes that path fire
+again on the next result. For the same reason, only a result from the current round can close it —
+later-round matches run concurrently with it and answer for their own round. That matters because a
 pair can be refused for a reason that has nothing to do with readiness: `hasIncompleteMatchBefore`
 holds a later-round match back while either seat still owes an earlier-round game, so a pair goes from
 blocked to launchable when a *third* pair's game finishes, with no change to the ready set at all.
--- a/docs/architecture-principles.md
+++ b/docs/architecture-principles.md
@@ -1401,10 +1401,19 @@ WAITING_FOR_PLAYERS → DECK_BUILDING → TOURNAMENT_ACTIVE → TOURNAMENT_COMPL
 rounds for N players), then matches start dynamically as players become ready — no waiting for
 the entire round to finish. This "eager match starting" pattern means:
 
-1. A player sends `ReadyForNextRound`
-2. The server checks if their next opponent is also ready
-3. If both are ready, the match starts immediately via a new `GameSession`
+1. A player sends `ReadyForNextRound` — refused while their own match is still running, since a ready
+   click means "I dismissed the game-over overlay"
+2. The server marks them ready and sweeps the *whole* ready set for launchable pairs
+   (`TournamentManager.startableMatches`)
+3. Each pair whose seats are both ready starts immediately via a new `GameSession`
 4. Other players continue waiting or playing their own matches
+
+A ready flag is state, not an event: it is set by a ready click (or the AI auto-ready pass) and
+consumed only when a match actually starts — a round ending does not clear it. That matters because a
+pair can be refused for a reason that has nothing to do with readiness: `hasIncompleteMatchBefore`
+holds a later-round match back while either seat still owes an earlier-round game, so a pair goes from
+blocked to launchable when a *third* pair's game finishes, with no change to the ready set at all.
+Every result therefore re-runs the sweep; looking only at ready *transitions* stalls the bracket.
 
 BYE handling (odd player counts), reconnection across matches, and spectating between rounds are
 all managed at this layer. The tournament system reuses the same `GameSession` and `ActionProcessor`
--- a/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/LobbyHandler.kt
+++ b/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/LobbyHandler.kt
@@ -211,9 +211,6 @@ class LobbyHandler(
     fun startTournament(lobby: TournamentLobby) =
         tournamentMatchHandler.startTournament(lobby)
 
-    fun startNextTournamentRound(lobbyId: String) =
-        tournamentMatchHandler.startNextTournamentRound(lobbyId)
-
     fun handleRoundComplete(lobbyId: String) =
         tournamentMatchHandler.handleRoundComplete(lobbyId)
 
@@ -1061,7 +1058,10 @@ class LobbyHandler(
                         standings = tournament.getStandingsInfo(connectedIds),
                         nextOpponentName = nextOpponentName,
                         nextRoundHasBye = nm.isBye,
-                        isTournamentComplete = false
+                        isTournamentComplete = false,
+                        // Describes `round` above — the round this player's next match sits in, which
+                        // by definition still has that match to play.
+                        roundComplete = false
                     ))
                 }
                 // Send active matches so player can watch live games while waiting
@@ -1069,10 +1069,9 @@ class LobbyHandler(
                 // Send ready status
                 val readyPlayerIds = lobby.getReadyPlayerIds()
                 if (readyPlayerIds.isNotEmpty()) {
+                    // A snapshot for the reconnecting player, not news about them — hence no playerId.
                     sender.send(session, ServerMessage.PlayerReadyForRound(
                         lobbyId = lobby.lobbyId,
-                        playerId = identity.playerId.value,
-                        playerName = identity.playerName,
                         readyPlayerIds = readyPlayerIds.map { it.value },
                         totalConnectedPlayers = connectedIds.size
                     ))
@@ -1096,7 +1095,10 @@ class LobbyHandler(
                         standings = tournament.getStandingsInfo(connectedIds),
                         nextOpponentName = nextOpponentName,
                         nextRoundHasBye = nm.isBye,
-                        isTournamentComplete = false
+                        isTournamentComplete = false,
+                        // Without this the reconnecting player is told their finished round is still
+                        // running, and the overlay names the round they already played.
+                        roundComplete = currentRound.isComplete
                     ))
                     // Send active matches so player can watch live games while waiting
                     val spectatingGameId = identity.currentSpectatingGameId
@@ -1127,10 +1129,9 @@ class LobbyHandler(
                 // Send ready status
                 val readyPlayerIds = lobby.getReadyPlayerIds()
                 if (readyPlayerIds.isNotEmpty()) {
+                    // A snapshot for the reconnecting player, not news about them — hence no playerId.
                     sender.send(session, ServerMessage.PlayerReadyForRound(
                         lobbyId = lobby.lobbyId,
-                        playerId = identity.playerId.value,
-                        playerName = identity.playerName,
                         readyPlayerIds = readyPlayerIds.map { it.value },
                         totalConnectedPlayers = connectedIds.size
                     ))
--- a/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/TournamentMatchHandler.kt
+++ b/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/TournamentMatchHandler.kt
@@ -88,33 +88,9 @@ class TournamentMatchHandler(
                 return
             }
 
-            while (true) {
-                val needsPrepare = tournament.currentRound == null ||
-                        tournament.currentRound?.isComplete == true
-                if (!needsPrepare) break
-
-                val round = tournament.startNextRound()
-                if (round == null) {
-                    completeTournament(lobbyId)
-                    return
-                }
-
-                for (match in round.matches) {
-                    if (match.isBye && match.isComplete) {
-                        val byePlayerState = lobby.players[match.player1Id]
-                        val byeWs = byePlayerState?.identity?.webSocketSession
-                        if (byeWs != null && byeWs.isOpen) {
-                            ctx.sender.send(byeWs, ServerMessage.TournamentBye(
-                                lobbyId = lobbyId,
-                                round = round.roundNumber
-                            ))
-                            spectatingHandler.sendActiveMatchesToPlayer(byePlayerState.identity, byeWs)
-                        }
-                    }
-                }
-
-                ctx.lobbyRepository.saveTournament(lobbyId, tournament)
-                logger.info("Prepared round ${round.roundNumber} for tournament $lobbyId")
+            if (!prepareRoundsIfNeeded(lobby, tournament)) {
+                completeTournament(lobbyId)
+                return
             }
 
             val wasNewlyReady = lobby.markPlayerReady(identity.playerId)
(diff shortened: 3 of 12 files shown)


wingedsheep:
## Review follow-up (eb2e26e6)

A review pass found a load-bearing side effect the first commit dropped, plus a set of smaller gaps. All fixed.

**The real one.** `clearReadyState()` at a round boundary was the only thing that made `autoReadyAiPlayers`' AI guard true there — and `startNextRound()` sat behind that guard. With ready flags surviving the boundary, the round index stopped advancing. `isRoundComplete()` reads `currentRound`, so it kept answering `true` for a round already reported, and the AI-vs-AI matches the sweep then started in the next round re-entered `doHandleRoundComplete`: a second `RoundComplete` carrying stale results, and `currentGameSessionId = null` for every player — including seats mid-game, which is exactly what reconnect and the idle-spectator list read.

Two fixes, deliberately belt-and-braces since the first one is a wiring invariant:
- `prepareRoundsIfNeeded` is now the single copy of advance-and-announce-byes (previously three divergent ones), and the round-complete path calls it explicitly after broadcasting. The advance no longer hangs off an AI heuristic.
- Only a result *from* the current round can close it. Later-round matches run concurrently with it under eager starting, so `getRoundForMatch` decides — a round is reported once even if something else leaves the index stale.

Worth recording: the reviewer's reproduction had the eager round-2 match starting *before* the boundary, which the circle method actually rules out at 4 seats (each round re-pairs across the previous one, so nobody is free early). The bug is real anyway, reached one step later — the sweep starts those round-2 matches immediately *after* the boundary, with `currentRound` still pointing at round 1.

**Also fixed**
- `handleAbandon` now sweeps: a forfeit completes every match the leaver had left in every round, which is the same unblocking event a real result is. Their ready flag is dropped too — nothing else clears it now, and a departed player left in the set inflates "n/m ready" permanently.
- Both `LobbyHandler` reconnect `MatchComplete` senders set the new field. Only the handler's own sender had it, so a player reconnecting after their round ended was told it was still running — a regression on that path specifically.
- `onTournamentBye` / `onTournamentResumed` set it as well, and the wire field is renamed `roundComplete`: it describes the message's `round`, which eager starting can make later than the tournament's current round.
- Snapshot re-broadcasts pass no `playerId` rather than naming the recipient, which the new nullable contract had made a lie.
- Deleted `tryStartMatchAfterDeckSubmit` and `startNextTournamentRound` (plus its `LobbyHandler` delegate) — both unreferenced, and each carried its own copy of the start rules; the latter started every current-round match with no regard for readiness at all, directly contradicting the invariant this PR documents.

**Not taken**
- Clearing ready on *disconnect*. A blip inside the grace window would silently drop a queued Ready, and a player who reconnects generally does want the match they asked for. Pre-existing behaviour within a round; left alone.

**Test coverage, honestly.** Two new deterministic cases pin the hazard at the `TournamentManager` level — that a finished round keeps answering `isRoundComplete()` until advanced past, and that `getRoundForMatch` attributes an eagerly-finished later-round result to its own round. That is the invariant, but it is not the wiring: no unit test drives the handler, which is why review caught this and a green suite didn't. A handler-level test needs AI seats playing real games over WebSockets (`AddAiToLobby` exists, so it's expressible) — minutes of runtime and inherently flaky, and this module's stated convention is that behaviour is covered by `e2e-scenarios/`. Flagging the gap rather than papering over it.

Re-verified: `just test-server` green, client `npm run typecheck` + 590 vitest tests green.

PR #1965, opened 2026-08-20
Assay: read "up to one/two/three/X target ..." as a quantifier table

wingedsheep:
Assay now reads **every quantifier English prints in front of the word "target"** — "Destroy **up to one** target creature.", "Exile **up to three** target creatures.", "Destroy **up to X** target artifacts.", "Return **any number of** target creatures you control to their owners' hands."

The family was the top of the tail ranking read five ways at once: `up to one …` 174 cards / 99 sole-blocked, `up to two …` 61 / 38, `Up to two …` 27 / 25, `up to three …` 23 / 17, `up to X …` 19 / 12. Five rows of one construct, which is the signal that what was missing was a *position* and not five rules.

## Six rows, because the noun disagrees in number

`Targets.quantifiers` is the table:

| printed | noun | requirement |
|---|---|---|
| `target creature` | singular | `TargetPermanent(filter)` |
| `up to one target creature` | singular | …`optional = true` |
| `two target creatures` | plural | `count = 2` |
| `up to two target creatures` | plural | `count = 2, optional = true` |
| `up to X target creatures` | plural | `optional = true, dynamicMaxCount = XValue` |
| `any number of target creatures` | plural | `unlimited = true` |

Every prefix this grammar has factored before became a **value in a slot** — the trigger join's `Prefix`, the fronted duration, `Phases`' whose-turn layer. This one cannot, and the reason is one word to its right: "up to one target **creature**" and "up to two target **creatures**" disagree in number, `Filters` keeps its singular and plural as two separately-instantiated cascades because English pluralization is a table column and not a suffix rule, and a `{filter}` slot is one phrase fixed at declaration time. A slotted quantifier would leave the noun's number undetermined, which is exactly what the round trip forbids.

So the quantifier is a **row**. What that buys is what the hand-copied shapes had already lost: "tap up to three target creatures" was written because a card needed it and "destroy up to three target creatures" was not, on a grammar that read both halves of that sentence. `quantifiedPermanentSteps` replaces `targetedPermanentStep` + `multiTargetStep`.

**`plural` is one column with two consequences**, and they are the same fact twice: a plural quantifier is exactly one that admits more than one target, so its noun comes from `Filters.plural` *and* its effect is written once per chosen target (`ForEachTargetEffect` over `ContextTarget(0)`) rather than once against the requirement. No row has those come apart, which is why the `effectOver`/`memberOf` pair that performs the wrapping lives once beside the table.

**Two families slot it**, which is the argument that it is a table: the seven one-verb permanent sentences (destroy, regenerate, exile, tap, untap, return to hand, put on top of library), and the **pump** sentence, which shares nothing with them but the noun phrase — a fronted spelling, a stat modifier, and a verb that agrees in number ("Up to one target creature **gets** +2/+0", "Up to two target creatures **each get** +2/+1").

**Two templates, because English agrees past the noun.** "Return target creature to **its owner's hand**" pluralizes to "… to **their owners' hands**", past where the `Filters` cascade can follow. That also made a third spelling cheap: Oracle prints that plural possessive *both* ways, 110 lines to 55, so the minority is `alsoSpelled` on the same rule and Scapegoat's "…to their owner's hand." reads as a `VARIANT`.

## Gates

| | before | after |
|---|---|---|
| cards fully covered | 8,111 | **8,172** (+61) |
| round-trips byte-exact | 26,282 | 26,358 (+76) |
| alternate spelling normalized | 1,438 | 1,448 (+10) |
| declined | 37,033 | 36,947 (−86) |
| MISMATCH / AMBIGUOUS / redundant | 0 | **0** |
| baked ledger `whole` | 7,897 | **7,955** (+58, none lost) |
| differential compared | 3,765 | 3,785 |
| differential DIVERGENT | 35 | 40 |

`:oracle-assay:test` green (StepsTest gains 6 tests); `:game-server:compileKotlin` green.

## The five new divergences, classified

| card | what differs | class |
|---|---|---|
| Calamitous Tide, Essence Fracture, Second Breakfast | the card unrolls its multi-target effect as a `Composite` of `BoundVariable("creature[0]")`, `…[1]`; Assay writes `ForEachTargetEffect` | second SDK spelling — majority printed, minority reported |
| Offender at Large | `EventPattern.AnyOf` vs two abilities | the entry band's standing finding, new member |
| Seize Opportunity | a stored-collection *name* (`impulseExiled` vs `exiledCards`) | pre-existing gate gap, exposed |

None is a parser bug, and the original 35 are the same 35 by name.

The first is **not** a card bug: `EffectContext.buildNamedTargets` publishes `"$id[$i]"` for every position of a multi-count requirement, so the indexed unroll resolves to nothing for a position nobody chose. It is a *minority* spelling — six occurrences across five sets, against a corpus that writes the iteration everywhere else including for "destroy two target lands" — so the grammar prints the iteration and the gate names the three cards. They are a safe mechanical cleanup, and until then the unroll is pinned to the declared count in a way the iteration is not.

Seize Opportunity is worth separating because it is **not about quantifiers at all**: the two models agree on the "up to two" mode exactly and differ on the name of a collection its *other* mode stores. A collection name is arbitrary in the way a target slot's name is, and `Differential.normalizeSlots` normalizes the second but not the first — a gap in the gate rather than in either model, and this band is the first thing to have driven a card into it.

## What the band uncovered

The five quantifier rows collapsed — `up to one …` from 174 cards to **17**, `up to two …` from 61 to 22, and `up to three …`, `up to X …`, `Up to two …` off the table. What replaced them is the payload behind the prefix, which is the product:

| tail family | cards | sole | what it needs |
|---|---|---|---|
| `any number of …` | 123 | 76 | divided damage, distributed counters, players, graveyard-zoned targets — everything the table's sixth row does *not* reach |
| `, where X …` | 59 | 50 | a trailing clause defining X for a target **count**, i.e. `dynamicMaxCount` fed from `Amounts`' existing vocabulary |
| `one other target …` | 50 | 34 | the `other` modifier, which has two SDK spellings to classify first — `TargetFilter.other()` and the `TargetOther` wrapper |
| `each of up …` | 29 | 16 | "put a +1/+1 counter on each of up to X target creatures" — the distribute sentence |
| `choose up to …` | 28 | 18 | the *modal* quantifier, which `Modal`'s KDoc already records as not a row of its header |

`, where X …` is the band's own residue and the cheapest of the five. `any number of …` is the honest one to read twice: the sixth row cost two lines and finished no card, because the corpus writes that quantifier almost entirely in sentences no verb here covers. It is in the table because the table claims to be exhaustive over *printed* quantifiers, not because it paid.

Scoping stated rather than discovered twice: `up to X` maps only the **bare** wording. "…, where X is the number of verse counters on ~" defines X from the board rather than from the cost — a different `DynamicAmount` behind a trailing clause `Amounts` owns — and it is the 59-card row above.

Written up in [`oracle-assay/README.md`](https://github.com/wingedsheep/argentum-engine/blob/worktree-assay-up-to-n/oracle-assay/README.md#the-target-quantifier); the ledger re-bake is its own commit so its diff is readable.


wingedsheep:
## Review — and the four findings, fixed in `ff251a3` + `4c760a3`

Merged `origin/main` in first (clean, brings `DeepCavernBat`). Every number below I measured rather than read off the description; the original body's figures all checked out.

### Verdict

Behaviour right, shape right. Making the quantifier a **row** rather than a slot is correct, and the argument is verifiable rather than rhetorical: `Filters.kt:101` really does carry `TypeNoun("creature or planeswalker", plural = null, …)`, so no plural rule can hand `several()` a compound filter; `Cardinals.word` really does start at two, which is what keeps the singular and plural rows disjoint in both directions. Four findings, all now fixed.

### What was already good

- **`plural` as one column really is one fact.** A quantifier admitting several targets is exactly one whose effect the SDK writes as an iteration, and putting `effectOver`/`memberOf` once beside the table is the right consequence. The inverse is exact, not approximate: `IterationSpace.Targets` is a `data object` and `List<Effect>.asBody()` returns the singleton itself, so `memberOf ∘ effectOver == id`.
- **The fail-closed seam got sharper.** Splitting `targetedFilter` (reads the noun) out of `permanentFilter` (reconstruct-and-compare) is the right cut. `Cardinals.spellable(1) == false` is quietly load-bearing — it's what stops a `count = 1` + `ForEachTargetEffect` model from printing "Destroy one target creatures."
- **The `buildNamedTargets` claim checks out** — `EffectContext.kt:496-497` publishes `"$id[$i]"` under `targets.getOrNull`, so an unchosen position is simply absent. Calamitous Tide's indexed unroll is a working minority spelling, not a card bug.
- **Independent verification:** re-baked ledger was **byte-identical** to the committed one; the ledger diff showed zero whole→declined (all 10 added `LINE_DECLINED` lines belong to cards already declining — Argivian Welcome's just moved to "perpetually gains flash"); filtering the divergence dump to cards whose *Assay-side* script carries a quantifier artifact gave exactly the 5 named plus 6 pre-existing static `ModifyStats` families.

### I1 — `XValue`/`CastX` was wrong reasoning (not wrong output)

`Targets.upToX` called them "one more two-spellings-for-one-meaning pair" and printed "the majority". They are not that. `DynamicAmount.kt:630-637`: `XValue` "resolves from the transient resolution context … and is therefore only populated while the spell itself is resolving"; `CastX` is the durable object-scoped reading. `LostInTheMaze.kt:38-43` calls the distinction "load-bearing and **silent when wrong**" — not a stylistic minority, the only correct spelling for its position.

I chased the case that worried me. Rampaging War Mammoth is newly whole with `XValue` in a **trigger** position — and it's correct: `TriggerDetector.kt:1798-1801` deliberately routes the announced `{X}` through `TriggerContext.fromEvent` for exactly this, and Valor's Flagship (hand-written, same shape, passing scenario test) reads it the same way. So the output is fine; the *argument* isn't. Left as written, it licenses the next rule to emit `XValue` in an enters-trigger position — Lost in the Maze's own declined line — where it silently resolves to zero.

Rewritten as a position rule, with the note that no `DynamicAmount` exists at all for "the X of an arbitrary activated ability", so a row wanting that position needs SDK vocabulary before it needs a template.

### I2 — the invariant wasn't established, so I established it

The payoff line said "a verb cannot have one quantifier and lack another". Only two families were slotting the table; four siblings still hand-spelled the bare quantifier, so the drift had moved one level up rather than being killed. Four more families take it now:

| family | rows | printed lines |
|---|---|---|
| keyword grant | all six | 2 |
| compound (`each get … and gain …`) | all six | 11 |
| damage | **singular only** | 11 |
| counters | **singular only** | 19 |

The last pair needed new vocabulary. A table claiming exhaustiveness has to say *which rows a sentence takes*, so `Targets.singularQuantifiers` is that declaration and the criterion is sharp: **the whole table when the plural changes only the noun; the singular rows alone when the plural is a different sentence.** Damage over several targets is "divided as you choose among …" (`DivideDamage`); counters over several is "on **each of** up to two target creatures", the distribute sentence. Giving those two the plural rows wouldn't merely win nothing — it would read a distribute model as a sentence that means something else, which is the reversible-but-wrong class the fail-closed matching exists to catch. There's a test asserting both refusals.

The compound is the one that validated the whole premise: both halves are per-target so the entire composite goes inside one iteration, and **Quickbeam, Upstart Ent matches that model byte-for-byte** — a hand-written confirmation on the first try.

### M1 / M2

`upToX` now also scopes out the bare "X target …" wording — 40 printed lines, which the corpus models with *this very requirement* (Icy Blast). It means **exactly** X where this row means at most X, and `TargetObject.minCount` is a plain `Int` that can't take a `DynamicAmount`, so it must keep declining: a rule for it would make two printed forms denote one model, the redundant-reading class the gate holds at zero. README's "three ways"/"three directions" → five (it listed five rows).

### Gates after the fix

| | before | after |
|---|---|---|
| cards fully covered | 8,172 | **8,186** (+14) |
| round-trips byte-exact | 26,358 | 26,381 |
| alternate spelling normalized | 1,438 | 1,455 |
| declined | 36,947 | 36,917 |
| MISMATCH / AMBIGUOUS / redundant | 0 | **0** |
| baked ledger `whole` | 7,955 | **7,969** (+14, none lost) |
| differential compared | 3,785 | 3,790 |
| differential DIVERGENT | 40 | 43 |

`StepsTest` 33 tests green (4 new); `:game-server:compileKotlin` green. Newly whole: Abandon Reason, Bile-Vial Boggart, Burning Sun's Fury, Coordinated Assault, Cryogen Relic, Cutthroat Maneuver, Invigorated Rampage, Karfell Kennel-Master, Press the Advantage, Quickbeam, Stress Dream, The Art of Tea, Wild Pack Squad, Windborne Charge — Invigorated Rampage picked the compound up inside a *modal bullet*.

### The three new divergences, classified

All three are cards the fix made whole, and **all three agree on the requirement the quantifier produced** — the divergence is elsewhere in the card every time, which is the evidence the rows transplanted cleanly.

| card | what differs | class |
|---|---|---|
| **Quickbeam, Upstart Ent** | trigger filter: Assay reads "another Treefolk you control" as `IsPermanent`, the card narrows to `IsCreature` | **card bug** — the bare-tribal-noun migration's own class, a leftover |
| **Stress Dream** | `order: ControllerChooses` on the bottom-of-library move, in the sentence's *other* clause | pre-existing `TopOfLibrary` gap, exposed |
| **Cryogen Relic** | `Effects.AddCounters("STUN", …)` where `CounterType.STUN` is the string `"stun"` | card-side spelling; harmless, `fromName` uppercases |

Quickbeam is the one worth acting on — narrowed to `IsCreature`, a Treefolk *artifact* entering doesn't trigger it. Not fixed here: it's a card in another set and would pull LTR's snapshot into a grammar PR, the same reason the three indexed-unroll cards are named rather than cleaned up.

Worth naming as a pattern, since Seize Opportunity is now three of a kind: **a band that finishes a card's last declining line inherits every unnormalized field in its other clauses.** The gap to close is `Differential.normalizeSlots`, which covers target slot names and covers neither collection names, ability ids, nor `order`.

### Recommendation

Ship it.

PR #1968, opened 2026-08-21
Apply death replacements to creatures dying in the same SBA batch

bolav:
# Apply death replacements to creatures dying in the same SBA batch

Reported from a live game: Head of the Hunt (The Hobbit #75) attacked, was blocked by three Human
Citizen tokens, and traded with them. All three tokens went to the graveyard and no Wolf appeared —
the card did nothing at all.

> Head of the Hunt — {2}{B}{B} Creature — Wolf, 4/3
> Flash
> If a creature an opponent controls would die, exile it instead. When you do, create a 2/2 green
> Wolf creature token.

Expected: all three tokens exiled instead of dying, and three 2/2 green Wolf tokens created.

There were two independent bugs behind that, and the first was masking the second.

## 1. Replacement effects and simultaneous state-based actions

CR 704.3 checks for state-based actions and then "performs all applicable state-based actions
**simultaneously as a single event**". CR 614.1 has replacement effects "apply continuously as
events happen — they aren't locked in ahead of time … They act like 'shields' around whatever
they're affecting." Combat damage is dealt simultaneously, so the 4/3 and the three tokens are all
destroyed by one SBA check, and Head of the Hunt is still on the battlefield right up until that
event happens. Its shield covers the creatures dying alongside it.

`LethalDamageCheck` already understood the *determination* half of this — it deliberately reads a
single projection taken before any creature moves, with a comment explaining that re-projecting off
the progressively-mutated state would make the outcome depend on battlefield iteration order. But
the destruction itself still ran one creature at a time through that mutated state, and
`ZoneMovementUtils.checkZoneChangeRedirect` scanned *that* battlefield for permanents hosting a
zone-change replacement. So iteration order silently decided the result: Head of the Hunt happened
to be moved first, its `ReplacementEffectSourceComponent` left the battlefield with it, and the
three tokens were then processed with no shield in sight.

The fix mirrors what the projection already does:

- `StateBasedActionChecker.checkOnce` passes each check the state as the pass began. This is a new
  defaulted overload on `StateBasedActionCheck`, so the other ~20 checks are untouched.
- `LethalDamageCheck` and `ZeroToughnessCheck` forward it to
  `SbaZoneMovementHelper.putCreatureInGraveyard`. Covering both means a shield dying to zero
  toughness in the same pass also still applies — the two checks are adjacent in `SbaOrder` and are
  one event as far as CR 704.3 is concerned.
- `checkZoneChangeRedirect` gains a `battlefieldSourceState` parameter, defaulting to `state` and
  used **only** for the loop that scans the battlefield for replacement sources.

The moving object's *own* replacements — finality counter, madness, commander divert, the
`SelfZoneRedirectComponent` self-redirect — deliberately keep reading live `state`. Those travel
with the card, not with a battlefield source, so freezing them would be wrong rather than merely
unnecessary.

## 2. The `CreateTokenEffect` rider was a silent no-op

`RedirectZoneChangeWithEffect` applies its rider through a small whitelist in
`applyReplacementAdditionalEffect` rather than the full effect pipeline. That whitelist held
`TakeExtraTurnEffect` (Ugin's Nexus), `AddCountersEffect` (Darigaaz Reincarnated) and
`GainLifeEffect` (The Darkness Crystal). `CreateTokenEffect` fell through to
`return state to emptyList()`.

So Head of the Hunt would never have minted a Wolf in *any* situation — not just the trade, but a
plain removal spell on an opponent's creature too. Fixing only bug 1 would have produced a correct
exile and still no token.

Rather than open-code a token mint, the rider now delegates to the real `CreateTokenExecutor`,
wired in `EngineServices` alongside the existing `ZoneTransitionService.cardRegistry` static wiring.
That matters for more than tidiness: The Hobbit registers its own Wolf art (`thob #9 — the Wolf of
Chief Warg's Company and Head of the Hunt`), and only the executor consults `TokenArtRegistry`. It
also gets the token its keywords, static abilities, and the enters-the-battlefield events ETB
triggers read. `ZoneChangeRedirectResult` carries a new `effectSourceId` so the executor has a
source to resolve that art against.

## 3. A guard the first fix made necessary

Reaching a shield that has already left the battlefield has one consequence worth calling out.
`RedirectZoneChange(linkToSource = true)` — Valgavoth, Terror Eater — tethers each card it exiles to
the source's `LinkedExileComponent`, so the source can later cast them. That link is applied *after*
the move, and the source may now be a permanent that died in the same batch. A permanent that leaves
the battlefield has had its `LinkedExileComponent` stripped, so writing one back would strand a
stale exile list on a card sitting in the graveyard that nothing cleans up (`removeFromLinkedExiles`
only walks the battlefield). If that card returned to the battlefield it would be a new object with
no memory of its previous existence (CR 400.7), yet would arrive carrying the old object's exiles.

So the link is now guarded on the source still being on the battlefield after the move. The
redirect itself is unaffected — an opponent's creature trading with Valgavoth is still exiled rather
than dying; it just isn't castable off a Valgavoth that died with it, which is the rules-correct
outcome. Before this branch the situation could not arise, because a shield that had left the
battlefield was never found in the first place.

## No new SDK vocabulary

Nothing was added to the card SDK. Head of the Hunt's existing definition is unchanged — it already
composed `RedirectZoneChangeWithEffect` correctly, and both bugs were in the engine behind it.
`docs/card-sdk-language-reference.md` is updated because it documents that rider whitelist by name,
and the list was now wrong.

## Verification

`just test-class HeadOfTheHuntScenarioTest` — **6/6 pass**:

| Case | Why it's there |
|---|---|
| Trades in combat → exiled + Wolf | the reported bug |
| Removed while the shield survives → exiled + Wolf | isolates the `CreateTokenEffect` rider |
| Two blockers → two exiles, two Wolves | a fix repairing only the *first* redirect in a batch would pass a one-blocker test |
| An opponent's token → exiled + Wolf | the reported game was tokens, and the filter is deliberately not `nontoken()` |
| Shield dies to zero toughness (Languish) → still shields | `ZeroToughnessCheck` takes the same change and cites the same rules |
| Your own creature dies → graveyard, no Wolf | the shield only covers opponents |

`just test-class ValgavothTerrorEaterScenarioTest` — **8/8 pass**, including a new case where a
deathtouch attacker trades with Valgavoth: the attacker is still exiled (the simultaneity fix
reaching a `linkToSource` shield) but is *not* linked to the dead Valgavoth (the guard).

**Mutation check.** With the simultaneity fix reverted and everything else in place:

| Test | Fixed | Simultaneity reverted |
|---|---|---|
| Trades in combat → exiled + Wolf | PASSED | **FAILED** |
| Removed while shield survives → exiled + Wolf | PASSED | PASSED |
| Your own creature dies → graveyard, no Wolf | PASSED | PASSED |

The second row staying green under the mutation is the point: it isolates the two bugs, showing the
rider works on its own and that the combat failure is purely the SBA ordering issue.

`manual-scenarios/cards/h/head-of-the-hunt.json` is the hand-play repro the bug was found with —
Head of the Hunt facing Castle Ardenvale's token making — so the trade can be replayed in the dev
scenario runner rather than only in the automated tests.

**Suites.** `:rules-engine:test` plus all seven era scenario suites green on the simultaneity +
token work. The link guard that came later was re-gated on what *it* touches:
`TheDarknessCrystalScenarioTest` (the other card whose replacement carries a rider), the full 2024
and 2026 era suites (the two modules whose tests changed), and `:rules-engine:test` — all exit 0.

Two environment notes, since both could be misread as this branch's problem:

- `just test-rules` first came back with five era modules red. The JUnit XML thread dumps showed the
  test worker still inside `CardDiscovery` reflection, class-initialising the card corpus, five
  minutes in — it had never reached game logic. `test-rules` runs the era test tasks in parallel,
  and several JVMs each loading the whole corpus starved each other past `TestHangGuard`'s hard cap;
  load average hit 22, all of it this run's own workers. Gradle then aborted before scheduling 2025
  and 2026 at all. Re-run one era at a time (`just test-scenarios <era>`), all seven returned exit 0.
- Gradle reported `resourceHashesCache.bin ... is corrupt. Discarding.` in the shared
  `/workspace/.gradle` cache during one run. It self-healed and continued; no shared state was
  cleaned. Flagging it because that cache is shared with sibling containers.

**What the gate does not cover:** nothing outside `:rules-engine` and the card scenarios ran —
`:game-server`, `:ai`, `:gym*` and the web client were not exercised. The diff touches none of them,
but `StateBasedActionCheck` gained an interface member, so any implementation of that interface
outside these modules is worth a grep before merge. No E2E or self-play run was done.

Rule numbers cited here (704.3, 614.1, 400.7) were checked against
`MagicCompRules_20260808.txt` in the workspace, not from memory.

## Deliberately left out

A sweeper is `ForEachInGroup(GroupFilter(Creature), Move(Self, GRAVEYARD, byDestruction = true))` —
Wrath of God's script — which walks the effect-executor path and moves creatures one at a time
through a mutating state. That is the same ordering dependency as bug 1, so a Head of the Hunt
caught in its own Wrath still won't shield the opponent's board. Threading a snapshot through
`ForEachInGroup` is a wider change in machinery a great many cards share, and it deserves its own
branch and its own gate rather than being folded in here.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -10704,8 +10704,12 @@ The priority groups are (CR 616.1a–f):
   appliesTo)` — like `RedirectZoneChange` but also runs `additionalEffect` when the replacement fires.
   The additional effect is applied through a small executor whitelist (not the full pipeline) —
   `TakeExtraTurnEffect` (Ugin's Nexus), `AddCountersEffect` on the redirected card (Darigaaz
-  Reincarnated), and `GainLifeEffect` (fixed amount, gained by the replacement source's controller;
-  emits `LifeChangedEvent` so life-gain triggers fire). `selfOnly = true` restricts it to the source
+  Reincarnated), `GainLifeEffect` (fixed amount, gained by the replacement source's controller;
+  emits `LifeChangedEvent` so life-gain triggers fire), and `CreateTokenEffect` (Head of the Hunt's
+  "When you do, create a 2/2 green Wolf creature token" — minted for the replacement source's
+  controller through the real `CreateTokenExecutor`, so the token gets the minting set's art, its
+  keywords and static abilities, and the enters-the-battlefield events ETB triggers read).
+  `selfOnly = true` restricts it to the source
   permanent itself; `linkToSource = true` (exile destination only) adds the redirected card to the
   source's `LinkedExileComponent` exactly like `RedirectZoneChange.linkToSource`. The Darkness Crystal's
   "If a nontoken creature an opponent controls would die, instead exile it and you gain 2 life" is
--- a/manual-scenarios/cards/h/head-of-the-hunt.json
+++ b/manual-scenarios/cards/h/head-of-the-hunt.json
@@ -1,0 +1,56 @@
+{
+  "player1Name": "Player 1",
+  "player2Name": "Player 2",
+  "player1": {
+    "lifeTotal": 20,
+    "battlefield": [
+      {
+        "name": "Head of the Hunt"
+      }
+    ],
+    "library": [
+      "Forest",
+      "Forest",
+      "Forest",
+      "Forest"
+    ]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "battlefield": [
+      {
+        "name": "Castle Ardenvale"
+      },
+      {
+        "name": "Plains"
+      },
+      {
+        "name": "Plains"
+      },
+      {
+        "name": "Plains"
+      },
+      {
+        "name": "Plains"
+      },
+      {
+        "name": "Llanowar Elves"
+      },
+      {
+        "name": "Llanowar Elves"
+      },
+      {
+        "name": "Llanowar Elves"
+      }
+    ],
+    "library": [
+      "Forest",
+      "Forest",
+      "Forest",
+      "Forest"
+    ]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "mode": "SELF"
+}
--- a/mtg-sets/2024/tests/src/test/kotlin/com/wingedsheep/engine/scenarios/ValgavothTerrorEaterScenarioTest.kt
+++ b/mtg-sets/2024/tests/src/test/kotlin/com/wingedsheep/engine/scenarios/ValgavothTerrorEaterScenarioTest.kt
@@ -70,6 +70,41 @@ class ValgavothTerrorEaterScenarioTest : FunSpec({
         linkedExileOf(driver, valgavoth) shouldContain victim
     }
 
+    test("a creature that trades with Valgavoth is still exiled, but is not linked to the dead Valgavoth") {
+        // CR 704.3 performs every applicable state-based action simultaneously as a single event,
+        // and a replacement applies as that event is about to happen (CR 614.1) — so Valgavoth
+        // still shields a creature dying alongside it. A deathtouch attacker that Valgavoth blocks
+        // is the cheap way to kill a 9/9 in the same combat-damage check that kills the attacker
+        // (it has to block rather than be blocked: Valgavoth flies, so a ground Rat can't block it).
+        //
+        // The *link* is a separate question. A permanent that has left the battlefield has had its
+        // LinkedExileComponent stripped, and re-attaching one would strand a stale exile list on a
+        // card in the graveyard that nothing cleans up — `removeFromLinkedExiles` only walks the
+        // battlefield. Coming back is a new object with no memory of its previous existence
+        // (CR 400.7), so the exile must not stay castable off a Valgavoth that died with it.
+        val driver = createDriver()
+        driver.initMirrorMatch(deck = Deck.of("Mountain" to 40), startingLife = 20)
+        driver.passPriorityUntil(Step.PRECOMBAT_MAIN)
+        val attacker = driver.activePlayer!!
+        val defender = driver.getOpponent(attacker)
+
+        val valgavoth = driver.putCreatureOnBattlefield(defender, "Valgavoth, Terror Eater")
+        val rat = driver.putCreatureOnBattlefield(attacker, "Deathtouch Rat") // 1/1 deathtouch
+        driver.removeSummoningSickness(rat)
+
+        driver.passPriorityUntil(Step.DECLARE_ATTACKERS)
+        driver.declareAttackers(attacker, listOf(rat), defender).isSuccess shouldBe true
+        driver.passPriorityUntil(Step.DECLARE_BLOCKERS)
+        driver.declareBlockers(defender, mapOf(valgavoth to listOf(rat))).isSuccess shouldBe true
+        driver.passPriorityUntil(Step.END_COMBAT)
+
+        // Deathtouch kills the 9/9 (CR 704.5h) while its 9 damage kills the 1/1 (CR 704.5g).
+        driver.getGraveyard(defender) shouldContain valgavoth
+        driver.getGraveyard(attacker) shouldNotContain rat
+        driver.getExile(attacker) shouldContain rat
+        linkedExileOf(driver, valgavoth) shouldNotContain rat
+    }
+
     test("your own creature dying is unaffected (only an opponent's graveyard)") {
         val driver = createDriver()
         driver.initMirrorMatch(deck = Deck.of("Mountain" to 40), startingLife = 20)
--- a/mtg-sets/2026/tests/src/test/kotlin/com/wingedsheep/engine/scenarios/HeadOfTheHuntScenarioTest.kt
+++ b/mtg-sets/2026/tests/src/test/kotlin/com/wingedsheep/engine/scenarios/HeadOfTheHuntScenarioTest.kt
@@ -1,0 +1,246 @@
+package com.wingedsheep.engine.scenarios
+
+import com.wingedsheep.engine.core.DeclareAttackers
+import com.wingedsheep.engine.core.DeclareBlockers
+import com.wingedsheep.engine.mechanics.layers.StateProjector
+import com.wingedsheep.engine.support.ScenarioTestBase
+import com.wingedsheep.sdk.core.Phase
+import com.wingedsheep.sdk.core.Step
+import io.kotest.assertions.withClue
+import io.kotest.matchers.collections.shouldContain
+import io.kotest.matchers.collections.shouldNotContain
+import io.kotest.matchers.shouldBe
+
+/**
+ * Scenario tests for Head of the Hunt (The Hobbit #75):
+ *   {2}{B}{B} Creature — Wolf, 4/3
+ *   Flash
+ *   If a creature an opponent controls would die, exile it instead. When you do, create a
+ *   2/2 green Wolf creature token.
+ *
+ * The interesting case is the *trade*: Head of the Hunt dying in the same combat-damage event as
+ * the creatures it is meant to exile. CR 704.3 performs every applicable state-based action
+ * simultaneously as a single event, and a replacement effect applies as that event is about to
+ * happen (CR 614.1) — so the shield is still on the battlefield and still applies. The engine
+ * moves the dying creatures one at a time, so before the pass-start snapshot went into
+ * [com.wingedsheep.engine.mechanics.sba.SbaZoneMovementHelper] this came down to battlefield
+ * iteration order: Head of the Hunt happened to move first and the creatures it had blocked went
+ * to the graveyard untouched.
+ */
+class HeadOfTheHuntScenarioTest : ScenarioTestBase() {
+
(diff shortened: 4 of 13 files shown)


bolav:
# Review — Apply death replacements to creatures dying in the same SBA batch

Self-review of `sba-simultaneous-death-replacements` against `origin/main...HEAD`. Branch is level
with `origin/main` (0 commits behind), tree clean, no merge commit was needed — reviewed in place,
no worktree created.

## Verdict

The core diagnosis and the shape of the fix are right: CR 704.3 / 614.1 genuinely require a dying
shield to protect the creatures dying alongside it, and freezing only the *battlefield scan* while
leaving the moving object's own replacements on live state is the correct seam. No new SDK
vocabulary, which is the right call — the card was already composed correctly. But the branch ships
one reachable regression I introduced (I1), and the tests do not cover the reported scenario or half
the code the diff changes (I2, I3).

## What's good

- The seam is narrow and defensible. `battlefieldSourceState` (`ZoneMovementUtils.kt:700`) is used
  by exactly one loop (`:772`), so finality counters, madness, commander divert and
  `SelfZoneRedirectComponent` keep reading live state — correct, since those travel with the card
  rather than with a battlefield source, and freezing them would have been a silent behaviour change.
- Defaulting every new parameter to the old value means the ~20 other `StateBasedActionCheck`
  implementations and the three non-SBA `applyReplacementAdditionalEffect` call sites are
  untouched, and no existing behaviour moves.
- Delegating the token rider to the real `CreateTokenExecutor` (`ZoneMovementUtils.kt:994`) rather
  than open-coding a mint is the right instinct — it is what gets the token its set art, keywords,
  static abilities and ETB events. Passing `effectSourceId` still resolves art correctly even though
  the source is in the graveyard by then, because `CardComponent` survives the zone change.
- The mutation check is real evidence, and the second test staying green under it genuinely
  isolates the two bugs rather than just asserting both were fixed.

## Issues

### I1 (Important) — a dead shield can now be re-linked, resurrecting its exile list

**Fixed in `3b36d4fb58`.**

`SbaZoneMovementHelper.kt:165` calls `linkExiledToSource(newState, entityId,
redirectResult.linkSourceId)` *after* the move. `linkSourceId` now comes from the pass-start
snapshot, so it can name a permanent that has already left the battlefield in this same batch —
which is exactly the case the branch set out to enable.

`stripBattlefieldComponents` removes `LinkedExileComponent` on the way out
(`ZoneMovementUtils.kt:522`), and `linkExiledToSource` (`:249`) then writes a fresh one back onto
the card now sitting in the graveyard. `removeFromLinkedExiles` (`:222`) only walks *battlefield*
permanents, so nothing ever cleans it up. If that card later returns to the battlefield it is a new
object (CR 400.7) but arrives carrying the previous incarnation's exile list, so a
`Creature.exiledWithSource()` target — the "cast a card exiled with ~" ability — would see cards the
old object exiled.

Reachable, not theoretical: Valgavoth, Terror Eater is a 9/9 with `linkToSource = true` and can
trade in combat with the creatures it exiles. Before this branch it could not happen, because a
shield that had left the battlefield was never found in the first place.

Fix: guard the link on the source still being on the battlefield in the *post-move* state —
`if (newState.getBattlefield().contains(redirectResult.linkSourceId))` — and add a scenario test
where a `linkToSource` shield trades with the creature it exiles. Worth deciding deliberately rather
than by accident: the card is still exiled either way, it just isn't castable off a source that died
with it, which is the rules-correct outcome.

### I2 (Important) — the reported scenario is not among the tests

**Fixed in `3b36d4fb58`.**

The bug report was three tokens dying at once to one shield, and neither of those two properties is
tested. `HeadOfTheHuntScenarioTest` uses a single non-token blocker.

- **Multiple simultaneous deaths.** With one blocker, a fix that repaired only the first redirect in
  a batch would still pass. Add a case with two blockers asserting two exiles and two Wolves.
- **Tokens.** `HeadOfTheHunt.kt:31` carries a comment specifically justifying that the filter is
  *not* `nontoken()` — an opponent's token still "would die", so it is exiled and still pays off the
  Wolf. Nothing exercises it, and it is the case the user actually hit. The scenario builder has no
  token helper, which is presumably why I skipped it; that is a reason to reach for a token-making
  card in the fixture, not a reason to leave the claim untested.

### I3 (Important) — `ZeroToughnessCheck` is changed and cites rules, but is untested

**Fixed in `3b36d4fb58`.**

`ZeroToughnessCheck.kt:19` and `:36` take the same change and `:33-34` cite CR 704.3 / 614.1, but no
test exercises a shield dying to zero toughness in the same pass as the creatures it should be
shielding. Per the "every cited rule must be exercised" bar that citation is currently decorative.
`ZERO_TOUGHNESS` (300) and `LETHAL_DAMAGE` (400) are adjacent in `SbaOrder`, so the cross-check case
— a lord dying and taking its team with it while an opponent's creature takes lethal combat damage —
is the one that justifies threading the snapshot at `checkOnce` level rather than inside
`LethalDamageCheck` alone. Without a test, the broader placement isn't earning its keep.

### M1 (Minor) — `putPermanentInGraveyard` was left on live state

`SbaZoneMovementHelper.kt:180`-ish still delegates straight to `ZoneTransitionService.moveToZone`
with redirects enabled, reading live state. Planeswalker loyalty, battle defense, legend rule, saga
sacrifice and unattached auras all die in the same simultaneous SBA event, so a permanent-filtered
"would die → exile instead" shield dying in that batch still won't cover them. No shipped card
exercises this today (Head of the Hunt and The Darkness Crystal are both Creature-filtered), which
is why I scoped it out — but the asymmetry is now a trap for whoever adds the first such card.
Either extend the snapshot there or leave a comment at the call site saying it was considered.

### M2 (Minor) — the two-arity `check` is easy to misuse

`StateBasedActionCheck.kt:31` adds `check(state, passStartState)` defaulting to `check(state)`, and
the two creature checks implement both with the 1-arg form delegating `check(state, state)`. It
works and it kept the diff small, but a future check that overrides only the 1-arg form silently
opts out of simultaneity with no diagnostic. A single `check(state, SbaPassContext)` would make the
extra state impossible to drop. Not worth churning the whole interface now; worth knowing it's a
trap.

### M3 (Minor) — the token rider swallows executor errors

`ZoneMovementUtils.kt:996`-ish does `if (result.error != null) return state to emptyList()`, so a
failed token creation vanishes silently. It matches the surrounding branches' style, but a
replacement rider failing is a real bug worth surfacing rather than dropping.

### M4 (Minor) — `tokenExecutor` extends a shaky global-mutable pattern

`ZoneMovementUtils.kt:126` is a `var` on an `object`, assigned from `EngineServices.init`. It
deliberately mirrors `ZoneTransitionService.cardRegistry`, so it is house style — but it means
concurrent `GameSession`s in one `game-server` JVM share whichever registry was wired last. That
pre-existing hazard now covers token art too. Fine to accept; flagging so it isn't a surprise later.

## Verification as reviewed

`HeadOfTheHuntScenarioTest` 6/6 and `ValgavothTerrorEaterScenarioTest` 8/8 green. The mutation check
confirms the combat test is load-bearing and, by staying green on the second case, that the
`CreateTokenEffect` rider is an independent fix rather than a consequence of the first.

`:rules-engine:test` plus all seven era scenario suites green on the simultaneity + token work; the
later link guard re-gated on `TheDarknessCrystalScenarioTest`, the 2024 and 2026 era suites and
`:rules-engine:test`, all exit 0.

Not covered: `:game-server`, `:ai`, `:gym*`, web client, E2E. `StateBasedActionCheck` gained an
interface member, so implementations outside `:rules-engine` (if any) deserve a grep before merge.

CR 704.3, 614.1 and 400.7 verified against `/workspace/MagicCompRules_20260808.txt`, not from memory.

## How the three Important findings were resolved

- **I1** — the link is now guarded on the source still being on the battlefield after the move
  (`SbaZoneMovementHelper.kt:165`), with a new `ValgavothTerrorEaterScenarioTest` case where a
  deathtouch attacker trades with Valgavoth: still exiled, no longer linked.
- **I2** — added a two-blocker case (two exiles, two Wolves) and an opponent's-token case. The token
  test routes through the opponent making and then killing their own token, which keeps it inside
  the turn where they plainly hold priority; a token minted this turn has summoning sickness and
  cannot attack, and the filter only cares who *controls* the dying creature.
- **I3** — added a Languish case: -4/-4 puts the 4/3 shield at 0/-1 and the 2/2 at -2/-2, so both
  die to one CR 704.5f check with the shield among them.

## Recommendation

Mergeable. The Minors (M1 `putPermanentInGraveyard`, M2 the two-arity `check`, M3 the swallowed
token-executor error, M4 the global-mutable `tokenExecutor`) are all pre-existing-pattern or
future-trap notes rather than defects in this diff, and are fine as follow-ups. The one thing a
second reader should weigh deliberately is the scoped-out `ForEachInGroup` sweeper path called out
in the PR body — it is the same rule, and leaving it means a Head of the Hunt caught in its own
Wrath still won't shield the opponent's board.

PR #1987, opened 2026-08-22
Add 50 cards to The Dark, taking it to 97/119

bolav:
# Add 50 cards to The Dark, taking it to 97/119

The Dark goes from 46 to 97 of 119 implemented. Fifty cards, forty-seven scenario test files, and
the handful of SDK and engine changes those cards actually needed — plus a set of playtest scenarios
for driving them by hand.

`scripts/card-status --set DRK` reports 97/119, which is what the backlog now says.

## The cards

Almost all of it is composition over primitives that already existed. A few worth calling out
because they exercise machinery rather than repeat it:

- **Mana Clash** is a do-while: `RepeatWhile` over two separate one-coin flips. Two flips rather
  than `FlipCoins(2)` because the card cares *whose* coin came up tails, and a two-coin heads count
  can't tell "I flipped heads, they flipped tails" from the reverse.
- **Angry Mob** is a characteristic-defining ability, so it goes through `dynamicStats` — a
  base-setting layer 7b effect, not a pump, which is what makes a later "becomes a 0/2" overwrite it
  instead of stacking. Both printed sentences fold into one `Conditional` amount rather than two
  abilities that could drift apart.
- **Tracker** is a fight in all but name and deliberately *not* `Effects.Fight`. CR 701.14b makes a
  fight all-or-nothing; Tracker's printed text is two ordinary sentences, so each half has to stand
  on its own.
- **Eater of the Dead**'s "if this creature is tapped" is a resolution-time check, not an activation
  restriction — a `ConditionalEffect`, not a `TimingRule`, so the ability stays visible (and the
  famous free-untap loop stays intact).
- **Safe Haven** uses a linked-exile pile (`ExileLinkedToSource` /
  `ReturnLinkedExileUnderOwnersControl`), so a second Haven never returns the first one's creatures.
- **Preacher** hands target choice to the opponent, over a "creature an opponent controls" filter.
  Legality is still measured relative to the ability's controller (CR 601.2c), so an opponent can't
  offer up something with hexproof to blank it.
- **Mana Vortex** and **Giant Shark** end themselves with state-triggered abilities (CR 603.8)
  rather than leaves-the-battlefield triggers, so they fire the moment the board condition becomes
  true — including as the Vortex's own upkeep trigger eats the last land.

Every card's `oracleText`, mana cost, type line, P/T, rarity, collector number and artist was
checked field-for-field against Scryfall. `just check-card-printing` passes for all fifty.

## New SDK vocabulary

Five additions, each a thin composition over something that already existed rather than a new
`Effect` type:

- `Conditions.NoLandsOnBattlefield` — the land sibling of `NoCreaturesOnBattlefield`, same
  `Exists(Player.Each, …, negate = true)` shape. Mana Vortex.
- `Costs.ExilePermanentsFixed(count, filter)` — the fixed-count sibling of the existing
  variable-count `ExilePermanents`, for a card that names a number and reads no X. City of Shadows.
- `TargetCreature(chooser = …)` — a passthrough of the `chooser` that `TargetObject` already
  carried, so an opponent-chosen *permanent* target is expressible. Preacher.
- `CounterType.STORAGE` / `HUNGER` — passive marker counters with no inherent rule, the same shape
  as `LOOT` and `NEST`. City of Shadows, Fasting.
- `PayOrSufferEffect.consequenceDescription` — see below.

`docs/card-sdk-language-reference.md` is updated for all of them.

## Engine changes

Four bugs the cards surfaced, and one UX change.

**A routed pay-or-suffer billed the wrong player.** `PayOrSufferExecutor` resolved the payer with
the stateless `resolvePlayerTarget` overload, which answers only the references it can read off the
context and returns null for everything else — and null there falls back to the *controller*. So
`Player.AnOpponent`, which needs turn order to resolve, had Scarwood Bandits' own controller paying
the ransom and buying back their own theft.

**A source-relative cost filter silently skipped the prompt.** The same executor built its candidate
sets with no `sourceId` in the `PredicateContext`. Filters like `attachedToBySource()` (Curse
Artifact's "sacrifice **that** artifact") return false outright without one, so the cost read as
unpayable and the suffer effect ran *without ever asking*. The player lost a choice the card gives
them, which is the worst shape this bug can take — and the obvious test still passes, because taking
the damage is the outcome either way. The test now asserts the decision exists before answering it.

**A routed question was asked in the wrong person's words.** An effect description is an imperative
fragment addressed to the ability's controller, so `GainControlEffect`'s "gain control of target"
put to an *opponent* offered them the theft they were the subject of. `consequenceDescription` lets
the card write the clause out; write it whenever `player` isn't the controller. Both prompts of the
choice-of-costs path honour it, including the follow-up rebuilt across a continuation.

**"Until end of combat" never expired at end of combat.** CR 511.2: those effects expire at the end
of the combat *phase*, which per CR 511.3 is once the end of combat step has ended. Nothing swept
them — `CleanupPhaseManager` dropped them at end of turn with a "should already be removed" note,
and no step ever removed them, so Murk Dwellers' +2/+0 stayed on through the postcombat main phase.
The sweep now lives in `CombatManager.endCombat`, the single hook all three combat-ending call sites
already share, which was already discarding end-of-combat mana at the same point.

**Optional triggers now prompt with the card's own text.** A generated description is assembled
bottom-up from the effect tree, so a composed effect reads as its own plumbing: Safe Haven's upkeep
trigger asked *"You may sacrifice this creature. If you do, look at cards exiled by this permanent.
Put those cards onto the battlefield"* — wrong noun for a land, a gather described as a look, and no
mention of whose control the cards return under. Lowering `optional = true` now passes the authored
`description` into the gate.

**This is the widest change in the branch and it is worth a reviewer's attention.** It alters the
yes/no prompt for **157 optional triggered abilities across 153 card files**, corpus-wide, and the
new prompt is the whole printed sentence — trigger clause and ability word included. The re-blessed
goldens show what that reads like as a question: *"Imprint — When this creature enters, you may
exile target nontoken creature."* It is a clear improvement on the plumbing text, but stripping the
leading trigger clause would be better still, and that is not done here. One card (Safe Haven)
asserts the new behaviour.

## Client

- Landwalk renders as an on-card badge. It decides whether an attacker can be blocked at all, and
  Scarwood Hag both grants and strips it — without a badge, "gains forestwalk" and "loses
  forestwalk" look identical to the player.
- The card preview gained a full counter-inventory panel. The battlefield badges are an allowlist,
  and the stats box is gated on the card having power/toughness, so a land's counters (City of
  Shadows' storage) were unreachable in the UI even though the server had been sending them.
- Eleven counter types had drifted out of the hand-written client mirror. `CounterTypeClientMirrorTest`
  reads `enums.ts` from the engine side and now fails when the two diverge.
- Defense counters are deliberately still badge-less: per CR 310.4c they are the battle analogue of
  loyalty — a number the permanent is defined by, not a marker sitting on it — so they belong with a
  loyalty-style display rather than in the marker allowlist. They do show in the new preview panel.

## Playtest scenarios

Twenty hand-driven scenarios under `manual-scenarios/sets/drk/`, five cards per batch, split by
colour. They start cards on the battlefield rather than in hand so a permanent's ability can be
activated straight away, and stock the boards with the subjects those abilities need.

## Rule citations

Verified against `MagicCompRules_20260808.txt` in the repo, not from memory: 202.3, 310.4c, 511.2,
511.3, 601.2b, 601.2c, 603.8, 605.1a, 608.2b, 611.2c, 614.12a, 701.14b, 702.14, 810.9a.

## Verification

| Gate | Result |
|---|---|
| `scripts/gradle-locked :mtg-sdk:test :rules-engine:test :mtg-sets:test :mtg-sets:1993-1999:tests:test` | **BUILD SUCCESSFUL** — 3921 passed, 0 failed |
| `just check-card-printing` on every new/moved card | all exit 0 |
| Scryfall field verification, all 50 cards | 0 discrepancies |
| `npx vitest run` over the changed client areas | 133 passed |
| `npm run typecheck` | clean |

The three engine fixes were mutation-checked: reverting each one turns exactly the test that pins it
red, and leaves the pre-existing cases green.

**What this does not cover.** `:game-server:test`, `:ai:test`, and the other eras' scenario suites
did not run. The only corpus-wide *behavioural* change is the end-of-combat expiry, and
`Duration.EndOfCombat` is used by exactly two cards — Murk Dwellers and Antiquities' Battering Ram —
both in the 1993-1999 era that did run, alongside `CamelDesertEndOfCombatScenarioTest`. The
corpus-wide *prompt* change is covered by the re-blessed goldens in `:mtg-sets:test`; I grepped the
other eras' tests for prompt-string assertions on optional triggers and found none, but that is a
grep, not a run. A full `just test-scenarios` would close both gaps.

## Deliberately left out

- **Three cards ship without a scenario test:** Goblins of the Flarg, Word of Binding, Wormwood
  Treefolk. The other forty-seven each have one. Goblins of the Flarg's state trigger and Wormwood
  Treefolk's two landwalk grants are both real behaviour, and Wormwood Treefolk is the only card in
  the batch that grants swampwalk — the keyword the client badge change was written for. These
  should be written before merge; they are called out in the review rather than hidden.
- **The "may" prompt still leads with the trigger clause** (see above). Stripping it is a separate
  change with its own blast radius.
- **Fasting's middle clause is modelled one step early.** The printed text is a replacement on
  *beginning* the draw step; here it is an optional upkeep trigger that arms `SkipNextDrawStep` and
  gains the 2 life. Nothing can happen between the two points that changes the decision, but the
  divergence is real and is written into the card's KDoc rather than hidden. Modelling it properly
  needs the draw step itself to be able to stop and ask, which is a draw-phase feature, not a card.
- **`descriptionOverride` is stored twice** on an optional trigger — once on the gate, once on the
  ability. Both are read, by layers that cannot reach each other, and the two strings are never
  equal (the gate's own fallback carries no trigger clause), so neither is derivable from the other.
  Every removal either changes client-facing text corpus-wide or risks a nested `MayEffect`
  inheriting the whole trigger's sentence. The lowering site now documents the split instead.


bolav:
# Review — `drk-no-engine` vs `origin/main`

Reviewed at `origin/main...HEAD` (78 commits, 171 files). The branch is current with
`origin/main` (`8ba8b5b02a`, 0 behind), so no merge was needed and the diff is post-merge reality.

## Verdict

Behaviour is right and the SDK shape is right: fifty cards, no new `Effect` type, no card-specific
engine code. Five SDK additions, each a thin composition over something that already existed. The
engine changes are genuine bugs the cards exposed, each with a test that fails without the fix.

Ten findings were raised on an earlier pass; eight are fixed in this branch and two are open. Both
open ones are called out in the PR body rather than left silent.

## Fixed during review

Anchored at the commit that fixed them:

- `23e1449a53` — **"until end of combat" never expired at end of combat.** `Duration.EndOfCombat`
  was swept only at the turn boundary and at cleanup (`CleanupPhaseManager.kt:518`, whose comment
  already said "should already be removed"), so Murk Dwellers' +2/+0 survived into the postcombat
  main phase. Fixed in `CombatManager.endCombat`, and the assertion deleted by `4fab097395` is
  restored. That commit's reason — the fix "touches rules-engine and so is not on this branch" —
  did not hold: the branch carries four rules-engine files.
- `e623ee6572` — **`consequenceDescription` was dead wiring on the choice-of-costs path.**
  `PayOrSufferExecutor` built `PayOrSufferChoiceContinuation` without the field, so the resumer's
  read was always null and the rebuilt effect fell back to generated text — precisely what the
  field's KDoc says must not happen. The choice list had the same gap one step earlier.
- `3d227d769e` — **Psychic Allergy counted the wrong battlefield.**
  `DynamicAmount.AggregateZone(zone = BATTLEFIELD)` is owner-keyed and evaluates against an empty
  projection; the printed clause is "permanents … they control". Swapped to `DynamicAmount.Count`,
  the variant that special-cases the battlefield. It was the only card in the corpus pairing
  `AggregateZone` with `BATTLEFIELD`.
- `3d227d769e` — Wormwood Treefolk declared `colorIdentity = "G"` despite a `{B}{B}` ability
  (Scryfall: BG); Tracker cited CR 701.12c for fight's all-or-nothing rule (it is **701.14b**;
  701.12 is Exchange in all three local rulebooks); Martyr's Cry lost a flavour-text newline; War
  Barge's description dropped its printed no-regeneration clause.
- `2b89ceb03c` — Mana Clash was missing its 8ED `Printing(...)` row (8ED is scaffolded — Blood Moon
  got one in the same sweep). Fellwar Stone's c16/clb/msc rows were pre-existing drift the branch
  inherited by moving the canonical to DRK.
- `9c81d817f8` — two comments pointed at a `counterTypeParity.test.ts` that does not exist; `hunger`
  was undocumented in the language reference while `storage` was; the backlog table lost its
  preceding blank line; SPORE was in the client enum with no badge.

## Open findings

### 1. Three cards ship with no scenario test — Blocking

`mtg-sets/1993-1999/src/main/kotlin/.../drk/cards/GoblinsOfTheFlarg.kt:1`,
`WordOfBinding.kt:1`, `WormwoodTreefolk.kt:1`.

Forty-seven of the fifty cards have a `<CardName>ScenarioTest.kt`; these three have nothing.
AGENTS.md's one-card-one-test-file rule is about which *file* a card's tests live in, but the
premise is that each card has tests at all.

Two of the three are not trivial:

- **Goblins of the Flarg** has a state-triggered ability (CR 603.8) that self-sacrifices when you
  control a Dwarf. State triggers fire on the condition becoming true, so it also fires on a Dwarf
  you gain control of and re-fires if you lose and regain one — none of that is exercised.
- **Wormwood Treefolk** has two landwalk-granting abilities and is the **only** card in the batch
  that grants swampwalk. The branch's own client change adds swampwalk badges; nothing proves the
  engine grants the keyword.

Word of Binding is the mildest of the three, but see finding 3.

### 2. The "may" prompt change is `add-feature` scale, riding in a card batch — Important

`mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/dsl/CardBuilder.kt:1592` and
`rules-engine/src/main/kotlin/com/wingedsheep/engine/event/TriggerProcessor.kt:478`.

Passing the authored `description` into the gate changes the yes/no prompt for **157 optional
triggered abilities across 153 card files**, corpus-wide. The direction is plainly right — the old
pipeline-derived text was unreadable — but the new prompt is the entire printed sentence, trigger
clause and ability word included. The re-blessed goldens are the evidence: *"Imprint — When this
creature enters, you may exile target nontoken creature."* is now a question put to a player.

The fix is to strip the leading trigger clause — everything up to the first comma after a
`When/Whenever/At` opener — or to give the DSL an explicit prompt field distinct from the catalog
text. Either is a separate change; what this branch should not do is ship the blast radius silently.
The PR body now names it, and one card (Safe Haven) asserts the new behaviour. That is the minimum,
not the target.

Worth noting the two are entangled: the same call site is why `descriptionOverride` is now stored
twice per affected trigger. Splitting the prompt from the catalog text would resolve both at once.

## Notes, not findings

- **Fasting's documented divergence** (`Fasting.kt:26`) is handled the right way — the printed
  replacement on beginning the draw step is modelled as an optional upkeep trigger, and the KDoc
  says so plainly instead of hiding it. Worth revisiting when a second card wants the same shape.
- **`CounterTypeClientMirrorTest`** reads `web-client/src/types/enums.ts` from a Kotlin test by
  walking up from `user.dir`. It is a design smell, the KDoc justifies it, and it fails loudly
  rather than silently if the file moves. Fine as it stands; if mtg-sdk is ever built without the
  client checkout present it will need a skip.
- **Scarwood Bandits and Curse Artifact's tests** are the model for the rest of the batch: each
  states why the obvious assertion was insufficient before adding the one that actually pins the
  bug. `CurseArtifactScenarioTest.kt:88` in particular explains how the card shipped broken past a
  passing test.

## Verification I ran

`scripts/gradle-locked :mtg-sdk:test :rules-engine:test :mtg-sets:test :mtg-sets:1993-1999:tests:test`
— **BUILD SUCCESSFUL**, 3921 passed, 0 failed. `just check-card-printing` clean on every new and
moved card. Scryfall field verification of all 50 cards: 0 discrepancies. Client: `vitest` 133
passed, `npm run typecheck` clean.

Each of the three engine fixes was mutation-checked — reverting it turns exactly the test that pins
it red and leaves the pre-existing cases green.

Not run: `:game-server:test`, `:ai:test`, other eras' scenario suites. See the PR body's Verification
section for why the residual risk is small and what would close it.

PR #1992, opened 2026-08-23
Complete The Dark (119/119)

bolav:
# Complete The Dark (119/119)

Finishes the card corpus for The Dark. `backlog/sets/the-dark/cards.md` goes from
97/119 to 119/119 — 22 new cards, one commit each, each with its own scenario test
file — plus the engine and SDK vocabulary they needed, the playtest scenarios that
cover them, and the client-side follow-ups the playtesting turned up.

Every one of the 22 is verified field-by-field against Scryfall: name, mana cost,
type line, collector number, P/T, artist, rarity and oracle text all match, the last
of those character-for-character.

## The 22 cards

Blood of the Martyr · Brainwash · Cleansing · Dance of Many · Dark Sphere ·
Deep Water · Frankenstein's Monster · Goblin Rock Sled · Leviathan · Lurker ·
Nameless Race · Necropolis · Reflecting Mirror · Runesword · Season of the Witch ·
Sorrow's Path · Spitting Slug · Tangle Kelp · The Fallen · Wand of Ith ·
Whippoorwill · Worms of the Earth

## The engine bugs these cards found

The interesting part isn't the cards — it's that several of them turned up genuine
engine bugs whose failure mode was **silence**, not an error. Each was caught by a
scenario test, not by review.

| Card | Bug it exposed |
|---|---|
| Nameless Race | `OnEnterRunEffect` never ran for permanents *cast* as spells, so "as this permanent enters …" was inert on every non-land user. Its two existing users are lands, which is why nothing had caught it — and `PermanentEntryReplacements`' own KDoc already listed the omission as a real gap. |
| Frankenstein's Monster | `ModalEffect.allowRepeat` was honoured at cast time but ignored at resolution time, silently turning "choose X, you may repeat" into "choose X different" and dropping picks once they outnumbered the modes. The resolution path also forced the floor to 0, offering "Don't choose a mode" on a mandatory modal. And the spell's chosen X never reached the as-enters context, so `DynamicAmount.XValue` evaluated to 0 there. |
| Runesword | Death riders modelled as a triggered ability resolve *after* state-based actions have already binned the dying creature (CR 704.3), so "exile it instead" changed nothing. Reworked as a damage-time static. |
| Dance of Many | `StatePredicate.CreatedBySource` sat in `TriggerMatcher`'s fail-open bucket, so "when **the token** leaves the battlefield" fired for every permanent that left — a playtest log shows it triggering twice off two sacrificed Islands and again off the enchantment's own death. |
| Season of the Witch | `CouldNotHaveAttackedThisTurn` asked only the per-creature reasons and missed the two broadest ones: that only the active player declares attackers (CR 508.1a), and that a skipped combat phase means nobody could have been declared at all. |
| Lurker | Making `TargetValidator.validateTargets`' `targetingSourceType` required — deliberate, so every call site states whether a spell or an ability is targeting — missed one test call site, and `:rules-engine:compileTestKotlin` had been broken since. |
| Frankenstein's counters | `CounterTypeClientMirrorTest` went red on rebase: `PLUS_TWO_PLUS_ZERO` / `PLUS_ZERO_PLUS_TWO` had no entry in the client's hand-written `enums.ts` mirror. Neither branch could have caught it alone — the test and the counters arrived on opposite sides of the rebase. |

## New engine/SDK vocabulary

`docs/card-sdk-language-reference.md` is updated in the same commit as every SDK
addition.

**Damage** — `halvePreventedDamage` on the next-instance shield (Dark Sphere);
`CreaturesDamagedBySourceAreDoomed` (Runesword); `creaturesOnly` and `optional`
redirect shields plus the `OptionalDamageRedirect` pre-pass (Blood of the Martyr);
`DealtDamageToThisGameComponent` + `EachDamagedBySourceThisGame` (The Fallen);
`DamageUnpreventableThisTurnComponent` (Whippoorwill).

**Combat** — `Triggers.BlocksOrBecomesBlocked` (CR 509.3a/509.3c, once per combat)
and `blockingOrBlockedBySource` (Spitting Slug); `CantAttackUnlessSacrifice`, the
first non-mana attack cost (Leviathan); `SwapBlockingAssignmentsEffect`
(Sorrow's Path); `CouldNotHaveAttackedThisTurn` + `AttackersDeclaredThisTurnComponent`
(Season of the Witch); `AttackedLastTurn` + `PlayerAttackersLastTurnComponent`
(Goblin Rock Sled, Tangle Kelp); `BlockedThisTurnComponent` and a spells-only
targeting restriction (Lurker).

**Other** — `PlayersCantPlayLands` + `LandsCantEnterTheBattlefield` (Worms of the
Earth); fixed-colour and *granted* `ReplaceLandManaColor` (Deep Water);
`stampCreator` on the token-copy path and `CreatedBySource` on the zone-change
trigger gate (Dance of Many); `ValueChosenAsEntered` (Nameless Race); `+2/+0` and
`+0/+2` counters (Frankenstein's Monster); two `ChangeTargetEffect` restrictions
(Reflecting Mirror).

### Why Blood of the Martyr needed an engine pre-pass

"Until end of turn, if damage would be dealt to any creature, you **may** have that
damage dealt to you instead" is the one card here that couldn't compose. Damage is
applied by pure code deep inside loops, and a whole combat damage step is one
simultaneous batch (CR 510.2) with no decision point inside it — so the card had
originally shipped as a *mandatory* redirect, documented as a divergence.

`OptionalDamageRedirect` asks the question where the rules put it: a replacement
effect applies as the event *would* happen (CR 614.1), so the whole batch is settled
question by question before any of it is dealt. One question per damage instance,
keyed by shield + source + recipient, so a sweeper hitting four creatures asks four
times and the caster can soak the damage aimed at their own blocker while letting an
opponent's creature take theirs. An instance that was never asked about counts as
**declined**, so a damage path that skips the pre-pass deals its damage normally
rather than redirecting onto a player who was never asked; and declining an optional
shield falls through to the next shield covering the same instance, so it can't shut
off a mandatory Glarecaster underneath it.

## Playtest scenarios and client follow-ups

Five new `manual-scenarios/sets/drk/` batches (21–25) cover the 22 cards no existing
batch touched; all 119 DRK cards now appear in at least one batch. Playing them by
hand is what surfaced the Dance of Many trigger bug, and three client-side gaps:

- a token that copies a *real* card carries that card's full image, so on the
  battlefield it was indistinguishable from a nontoken permanent — now gets a TOKEN
  chip that counter-rotates with a tapped card;
- Dark Sphere's shield left nothing on screen (it sacrifices itself paying for the
  ability, so there is no card left to carry the badge) — single-instance
  chosen-source shields now render as a player badge, which also fixes the Circle of
  Protection family;
- Deep Water's granted ability read "If a you control land is tapped for mana …";
  the mana statics now render through `describeObjectForEvent`.

## Review round

`fd3be7a746` is a self-review pass over the whole branch, folded in before opening:

- **Sorrow's Path was not restricted to an opponent's blockers.** The printed line is
  "two target blocking creatures controlled by the same **opponent**"; the target
  filter was a bare `Creature.blocking()` and the executor re-checked only that the
  two targets shared *a* controller. On the opponent's turn — when the Sorrow's Path
  player is the one blocking — you could target and swap your own two blockers. Both
  existing scenarios used opponent-controlled blockers, so nothing caught it. Fixed on
  both layers, with a third scenario that pins the targeting half specifically:
  reverting only the filter turns it red while the other two stay green.
- **That executor also read control off the base `ControllerComponent`** rather than
  projected state — the same divergence the earlier round fixed at five sites in
  `AffectsFilterResolver` / `PredicateEvaluator`. The two answers can't actually differ
  today (CR 506.4 removes a creature from combat when its controller changes, so the
  `BlockingComponent` reads above have already refused any such target), so this is a
  consistency fix with no behaviour change and no test of its own — the KDoc says so.
- **`CR 508.1e–g` was the wrong citation for Leviathan's attack cost,** at five sites.
  508.1e is banding, 508.1f is tapping the chosen attackers, 508.1g is *optional*
  costs. A required "can't attack unless you sacrifice two Islands" is a restriction
  checked at **CR 508.1c** whose cost is determined and paid at **CR 508.1h–j**, with
  **508.1d** the reason an unpayable one keeps the creature home rather than making the
  declaration illegal. Verified against `MagicCompRules_20260808.txt`.
- **`RedirectNextDamageEffect` had lost its KDoc** — `SwapBlockingAssignmentsEffect`
  was inserted between the doc block and the class, so only the second of two
  consecutive comments attached. Moved back.
- **Two oracle strings drifted from Scryfall** and are now exact: Frankenstein's
  Monster had a newline where Scryfall keeps one paragraph, and Leviathan dropped its
  reminder text "(This cost is paid as attackers are declared.)". DRK's snapshot
  golden is re-blessed for those two and for the new target filter.

## Documented divergences

Two cards ship with a deliberate, documented gap rather than a silent approximation.
Each says so in its KDoc and names what would fix it:

- **Reflecting Mirror** — X is chosen and validated at resolution rather than
  computed at activation, because ability cost calculation can't see the chosen
  targets. The redirect can never be had below its printed price; underpaying just
  wastes the activation. Proper fix: thread chosen targets into ability cost
  calculation.
- **Frankenstein's Monster** — "if you can't, put it into the graveyard instead" is a
  true entry replacement, but `OnEnterRunEffect` runs just *after* entry, so the
  Monster is modelled as entering and then being put into its owner's graveyard. An
  "enters the battlefield" trigger sees it, and so does a "whenever a creature dies"
  one. Proper fix: a pre-entry replacement hook.

Blood of the Martyr's divergence (see above) was closed on this branch rather than
shipped.

## Verification

Run serially, one Gradle invocation at a time, against this branch's tip
(`0f7c2d1bb9`) on a rebased-onto-`origin/main` base:

| Command | Result |
|---|---|
| `scripts/gradle-locked :rules-engine:test` | **BUILD SUCCESSFUL** — 521 classes / 3056 tests, 0 failures, 0 errors |
| `scripts/gradle-locked :mtg-sdk:test :mtg-sets:test` | **BUILD SUCCESSFUL** — 59 classes / 871 tests, 0 failures (`CardDefinitionSnapshotTest`, `CardLintTest`, `FacadeBoundaryTest`, `MtgSetCatalogTest`, `CounterTypeClientMirrorTest`) |
| `scripts/gradle-locked :mtg-sets:<era>:tests:test`, all nine eras, one invocation each | **BUILD SUCCESSFUL** every time — 3173 classes / 9258 tests, 0 failures, 0 errors |

Per era: 1993-1999 587 tests (includes all 22 new `<Card>ScenarioTest` files),
2000-2002 602, 2003-2007 669, 2008-2016 473, 2017-2022 729, 2023 1269, 2024 2017,
2025 1595, 2026 1317. Everything above was re-run after the review-fix commit
(`fd3be7a746`), not carried over from before it.

All nine era suites were run — not just the one holding the new cards — because this
branch's engine changes reach the damage pipeline, the trigger matcher, cleanup, and
declare-attackers. Those are exactly the shared paths that broke other eras earlier in
this set's work (`NivMizzetGuildpactScenarioTest` / `DragonfireBladeTest` in 2024/2025
and `WillbenderTest` in 2003-2007, both from this branch's own Lurker commit, both
since fixed).

**What these gates do not cover:**

- **The web-client TypeScript is unverified by any tool.** `node_modules` is absent
  and the network is firewalled here, so neither `tsc` nor `vitest` could run. The
  Kotlin green says nothing about the six `.ts`/`.tsx` files. `CounterTypeClientMirrorTest`
  reads `enums.ts` off disk and passes, which proves the counter mirror is in sync —
  and nothing else about the client.
- `:game-server:test`, `:ai:test`, `:oracle-assay:test` and the gym modules were not
  re-run in this session; the diff does not reach them, with the one exception below.
- No self-play / gym pass, and no Playwright e2e.

### One known red, deliberately left out

`SetCoverageServiceTest > "a set that reaches 100% is no longer flagged incomplete"`
fails on this branch. It is not a regression in anything here: DRK now compiles
119/119 while `TheDarkSet.kt:24` still carries `override val incomplete = true`, and
that flag gates `fullyImplemented`, which is what keeps a finished set out of the
lobby's set picker. Antiquities and Foundations both sat stale the same way, which is
why the guard exists.

Dropping that line is a one-word change, but it is a *set-completion* action rather
than a card change, and it is being made in its own follow-up commit alongside the
rest of the finishing-a-set checklist. Nothing else in the game-server suite is red.

## Before merging

This branch is already rebased onto `origin/main` and the gates above were run on
that base, not on a stale one. If it sits long enough to fall behind again, the
rebase invalidates the gate — re-run with `--no-build-cache --rerun-tasks` before
calling it green.


bolav:
# Review: drk-complete

**Verdict: ready to merge.** Reviewed `origin/main...HEAD` — 40 commits, 161 files,
22 new cards each with its own scenario test, against the SDK-elegance and projection
rules in `AGENTS.md`. Four findings; all four are fixed on the branch in
`fd3be7a746`, and the gate was re-run in full afterwards.

The composition discipline holds up. Of the 22 cards, only three needed new engine
vocabulary that couldn't be expressed by composing primitives — Blood of the Martyr's
optional redirect, Leviathan's non-mana attack cost, and Sorrow's Path's block swap —
and each of those has a KDoc that argues why. The rest reuse existing facades. The
engine bugs the set exposed are all real, all previously silent, and all landed with a
regression test rather than a comment.

Field verification independently confirmed: all 22 cards match Scryfall on mana cost,
type line, collector number, P/T, artist, rarity and — after finding 5 below — oracle
text, character-for-character.

---

## Findings, and how they were fixed

### 1. Sorrow's Path could swap *your own* blockers — fixed in `fd3be7a746`

The printed line is "Choose two target blocking creatures controlled by the same
**opponent**." Neither layer enforced the opponent part:

- the target filter was a bare `GameObjectFilter.Creature.blocking()` with no
  controller restriction;
- the executor re-checked only that both targets shared *a* controller — while its own
  comment claimed to be implementing "the printed 'controlled by the same opponent'".

Failure scenario: on the opponent's turn you are the defending player. You block with
two of your own creatures, then tap Sorrow's Path (a land, untapped, no timing
restriction on the ability), target your own two blockers, and swap them. Both
existing scenarios used opponent-controlled blockers, so nothing caught it.

Fixed on both layers — `.opponentControls()` on the filter, and an
`if (firstController == context.controllerId) return` in the executor. Both are needed:
a filter can say "an opponent's", but only the executor can relate the two targets to
each other and only it sees the state targets are re-validated against (CR 608.2b).

The new scenario, *"refuses to swap the activating player's own blockers"*, pins the
targeting half specifically: reverting only the filter (leaving the executor guard in
place) turns it red while the other two stay green — mutation-checked, not assumed.

### 2. Same executor read base controller, not projected — fixed in `fd3be7a746`

```kotlin
val firstController  = state.getEntity(first)?.get<ControllerComponent>()?.playerId   // was
val firstController  = state.projectedState.getController(first) ?: …                 // now
```

`AGENTS.md`'s load-bearing rule: control of battlefield permanents comes from projected
state, because base state can't see continuous effects. This is the same divergence the
branch's earlier round fixed at five sites in `AffectsFilterResolver` and
`PredicateEvaluator` (finding M5), and the value is handed straight to `canBlockAll` as
the blocking player, so a wrong answer would propagate into the legality check.

Worth being precise about the impact, because the fix ships without a test of its own:
**the two answers cannot currently differ.** CR 506.4 removes a creature from combat
when its controller changes, and that is implemented as a state-based action
(`ControlChangedRemovesFromCombatCheck`), so any target whose control changed has
already been refused by the `BlockingComponent` reads above. This is a
consistency fix against the next control-change effect that doesn't remove from combat,
not a behaviour change — and the KDoc now says exactly that rather than implying a
live bug.

### 3. `CR 508.1e–g` was the wrong citation, at five sites — fixed in `fd3be7a746`

`AttackPhaseManager.kt`, `AttackSacrificeCosts.kt`, `CombatStaticAbilities.kt`,
`Leviathan.kt` and `docs/card-sdk-language-reference.md` all cited **CR 508.1e–g** for
"can't attack unless you sacrifice two Islands". Checked against
`MagicCompRules_20260808.txt`:

- **508.1e** is banding ("If any of the chosen creatures have banding …")
- **508.1f** is tapping the chosen attackers
- **508.1g** is *optional* costs — "costs a player may pay 'as' a creature attacks"

Leviathan's cost is not optional; the clause is a restriction. Corrected to **508.1c**
(restrictions — "effects that say a creature … can't attack unless some condition is
met"), **508.1h–j** (determining and paying the total cost), and **508.1d** for the
"that player is not required to pay that cost" clause that makes an unpayable one keep
the creature home rather than voiding the declaration. `AGENTS.md` makes rule-number
accuracy a hard rule, and this is the 508.1a/508.1c class of slip the same branch had
already corrected once.

### 4. `RedirectNextDamageEffect` had lost its KDoc — fixed in `fd3be7a746`

`SwapBlockingAssignmentsEffect` was inserted between `RedirectNextDamageEffect`'s doc
comment and the class it documents, leaving two consecutive KDoc blocks — only the
second attaches. The redirect doc (still describing `protectedTargets`, `redirectTo`,
`amount`, `scope`) dangled above the swap effect, and the type this PR *extends* with
`creaturesOnly` and `optional` had no class-level doc at all. Moved back below the
swap effect.

### 5. Two oracle strings drifted from Scryfall — fixed in `fd3be7a746`

`FrankensteinsMonster.kt` had a `\n` between its two sentences where Scryfall keeps
them one paragraph, and `Leviathan.kt` dropped the reminder text "(This cost is paid as
attackers are declared.)". Both corrected; DRK's snapshot golden re-blessed for those
two plus the new Sorrow's Path target filter, and the diff to `DRK.json` is exactly
those three lines.

---

## Observations, not asks

- **`OptionalDamageRedirect.check` keys settled instances on the `Instance` data
  class** — (source, target, amount). Two *distinct* simultaneous damage instances
  sharing all three would collapse into one question, since they also share a
  `choiceKey`. No current damage path produces that shape: combat pairs one source with
  one recipient, and the divided/per-entity executors spread across distinct
  recipients. Left alone; worth knowing when the first such path is written.
- **`PlayerAttackersLastTurnComponent`'s team-turn edge** (`CleanupPhaseManager.kt:900`):
  the rollover is keyed to `activePlayerId`, so on a shared team turn (CR 805.10b) only
  one teammate's record rolls. Goblin Rock Sled and Tangle Kelp are unaffected in every
  format we test, and the comment already explains why rolling for *everyone* would be
  worse.
- **`abe1f3ccdc` invites its own removal** — its message says "drop this commit if the
  older boards were the ones worth keeping". It is dev-UI re-saves of four playtest
  scenarios; keeping it should be a decision rather than an oversight. `0f7c2d1bb9` is
  likewise an outside-session edit committed as found.

## What's good

- **The Blood of the Martyr rework is the strongest thing here.** Turning a documented
  mandatory-redirect divergence into a real per-instance choice, placed before the
  simultaneous batch (CR 510.2) because a replacement applies as the event *would*
  happen (CR 614.1), is the right shape rather than the cheap one. The two fail-safe
  defaults — an unasked instance counts as declined, and a declined optional shield
  falls through to a mandatory one underneath — are exactly the invariants that keep
  this from silently changing unrelated damage.
- **Load-bearing negative tests.** The `I1` pair added in `f285aafef7` is the model: the
  first attempt passed vacuously because `passUntilPhase` bailed at DRAW, and the paired
  control caught it. Both now assert the marker component directly, so a sweep that
  fails to run can't pass them. The new Sorrow's Path test was mutation-checked the same
  way.
- **`AttackersDeclaredThisTurnComponent` is the right primitive.** "Declared nothing" vs.
  "never got to declare" is a distinction the existing per-combat marker couldn't
  express, and stamping it unconditionally at declare-attackers — rather than inferring
  it from whose turn it is — is what makes False Peace and Fatespinner behave.
- **Divergences are documented rather than approximated.** Reflecting Mirror and
  Frankenstein's Monster each name the engine hook that would close the gap; the
  Frankenstein gate in particular is asked *before* anything moves.
- **`docs/card-sdk-language-reference.md` is in sync** — all 16 new vocabulary items are
  present, and the fixes above updated it in the same commit.
- **One card, one test file** — 22 new cards, 22 new `<Card>ScenarioTest.kt`. No
  batching.

## Gate

Re-run in full **after** the fix commit, not carried over from before it:

| Command | Result |
|---|---|
| `:rules-engine:test` | 521 classes / 3056 tests, 0 failures |
| `:mtg-sdk:test` + `:mtg-sets:test` | 59 classes / 871 tests, 0 failures |
| `:mtg-sets:<era>:tests:test`, all nine eras, one invocation each | 3173 classes / 9258 tests, 0 failures |

Every era's result XML was confirmed freshly written rather than served from an earlier
run — worth checking here, because a sibling container had clobbered this workspace's
results once already during the review.

The only red anywhere is `SetCoverageServiceTest`'s stale-`incomplete` guard, which the
body calls out and defers to the set-completion follow-up. That is the correct call: the
guard is telling the truth, and the fix belongs with the rest of the finishing checklist.

The web-client TypeScript is unverified by any tool in this environment, and the body
says so plainly rather than borrowing the Kotlin green. That's the right disclosure.

PR #1993, opened 2026-08-23
Verify The Dark end to end, and archive its backlog

bolav:
# Verify The Dark end to end, and archive its backlog

The Dark's backlog has read 119 / 119 for a while. That is a claim about *names*, not about the
cards. This change runs the `verify-set` pass over it and turns the claim into evidence: every card
checked field for field against Scryfall, every card's script read against its oracle text, tokens
checked, then the backlog closed out and archived.

Two of the eleven field discrepancies were things no existing gate in the repo could have caught, and
the per-card review turned up a corpus-wide engine bug that has nothing to do with The Dark except
that The Dark happens to contain both halves of it.

## The report

```
Set:              DRK — The Dark
Completeness:     119/119 cards, 0 missing, 0 reprint rows, 0 basic lands   (scripts/card-status)
Field fidelity:   119 cards × 13 fields vs Scryfall — 0 discrepancies, 0 waivers
                  11 fixed on the way: 1 dynamic-P/T rendering, 1 missing oracle_text,
                  5 missing flavor_text, 4 flavor_text line-break
Behaviour:        70 scenario tests; 25 manual playtest batches covering all 119 cards
DSL fidelity:     119 cards reviewed in 12 batches — 8 gameplay claims raised, 6 confirmed under
                  adversarial verification, 2 refuted; 0 are card-shaped defects in DRK's scripts
                  assay-differential: 3 divergent of 45 read whole (72 declined, not covered) —
                  2 folds, 1 Assay parser bug, 0 card bugs
Tokens:           1 token site (Dance of Many, a copy token) — token-art-gaps lists DRK 0 times
Waived:           none in the field check. 10 engine/SDK divergences documented and routed —
                  backlog/archived/sets/the-dark/the-dark-engine-gaps.md
Gate:             just test — 2 unrelated timeout reds, both green on serial re-run (see below)
```

## What the field check found

`CardDefinitionSnapshotTest` pins what we compiled against what we compiled last time, and
`CardLintTest` checks internal hygiene. Neither compares a card to the printed card, which is what
this stage exists for. Eleven rows, all now fixed:

- **Angry Mob rendered as `*` / `*` instead of `2+*` / `2+*`.** Display only — the +2 really was in
  the computed value, and the existing test proves 5/5 against three opponent Swamps. But it was
  folded *into* the amount, which builds a plain `CharacteristicValue.Dynamic` whose description is a
  hardcoded `"*"`. The constant is the printed *offset*, so it now rides in
  `powerOffset`/`toughnessOffset` and renders as `*+2` — the same shape Tarmogoyf uses. Identical
  computed values; the snapshot diff shows `Dynamic` → `DynamicWithOffset`.
- **Bog Imp had no `oracleText` at all.** Gameplay was fine — the card declares
  `keywords(Keyword.FLYING)` — but the file was an untouched `mtgish-tooling` render that never
  assigned the text. Added, and the stale generator banner dropped.
- **Five cards were missing their flavor text**, and **four had `" —Author"` where Scryfall has
  `"\n—Author"`**. The line break is a real difference in the snapshot and in the client.

## What the per-card review found

Twelve reviewers, batched by mechanic rather than alphabetically so a whole cycle sat in one context.
Every gameplay-severity claim then went to an adversarial verifier whose brief was to refute it.

**Nothing card-shaped in DRK's scripts turned out to be wrong.** The Goblin Caves / Goblin Shrine
cycle renders consistently; Knights of Thorn and Pikemen declare banding identically with
byte-identical reminder text; Blood Moon's own layer-4/layer-6 implementation is correct.

Six confirmed divergences are engine- or SDK-level and are written up in full, with file:line and the
exact fix, in `the-dark-engine-gaps.md`. They are **not fixed here** — each is its own `add-feature`
unit, and a set-verification pass is the wrong place to change a shared combat rule. The two biggest:

- **Landwalk reads base type lines, not projected subtypes** (`BlockEvasionRules.kt:148-157`). Under
  Blood Moon a Tropical Island still stops a forestwalker and still fails to stop a mountainwalker.
  The tell is that the *same two lambdas* already read the projection for control and only the type
  read was left on base state — precisely the hazard `AGENTS.md` names. Corpus-wide: 64 card files
  across ~20 sets carry landwalk, against six different set-land-type effects. `ai/CombatMath.kt`
  does the same check correctly, so the AI and the authoritative rule currently disagree. No test
  covers it: `LandwalkEvasionTest` is all printed Forests and Islands, and `BloodMoonScenarioTest`
  asserts only the projection — the two never meet.
- **Granted zone-change replacements aren't scoped to the granted entity**
  (`ZoneMovementUtils.kt:845-858`). Whippoorwill exiles *every* creature that dies that turn, both
  sides, instead of the one it targeted. Blast radius is exactly this card — the other two users of
  the shape are intentionally global — and the fix is two lines, gated on the existing but
  unconsulted `selfOnly` flag.

Also confirmed and documented: Worms of the Earth's unpayable sacrifice mode is a free escape from
its own lock (and it's the *common* case, since the card is what makes players landless); Fasting's
draw-step replacement is modelled as an upkeep trigger and double-pays life with two copies out;
Reflecting Mirror can be aimed at abilities, where its X gate collapses to 0; Word of Binding has no
exact-X target floor, a shape shared by 12 cards.

Two claims were **refuted** by verification, which is the point of running it: Spitting Slug's
"otherwise" clause is implemented and tested (only a narrow keyword-removal side effect survives),
and Reflecting Mirror's single-target restriction *is* enforced, just at resolution.

## Also corrected

- **`mechanics.md` was stale.** It listed "coin flips with an unbounded repeat" as unsupported with a
  ❌ and a note that the engine "needs a repeat-while primitive". `Effects.RepeatWhile` exists and
  Mana Clash uses it correctly. The entry is now ticked and describes what's actually there — so all
  25 of the set's mechanics are checked.
- **`TheDarkSet` no longer declares `override val incomplete = true`**, and its KDoc says why: 119/119
  implemented and field-verified, no reprints, no basic lands of its own.

## Not addressed here, deliberately

- `just check-backlog-implementations` fails on 72 entries across four `bloomburrow-commander` files.
  Untouched by this change and pre-existing — reported, not fixed.
- Regenerating `backlog/token-art-gaps.md` produces 14 gaps across nine sets (EMN, MH3, PC2, CMD,
  DKA, JMP, MRD, NCC, ODY) where the committed file says "**None.**". DRK has no rows in either
  version, so it doesn't affect this set's verdict, and the file is left as committed rather than
  shipping other sets' churn in a DRK change. Worth a look on its own.

## The gate

`just test` finished with two failures, neither in DRK and neither an assertion:

```
AccursedCentaurTest > ... when only the Centaur is on battlefield   TimeoutCancellationException
AbuJafarTest        > ... destroys the attacker it blocked          TimeoutCancellationException
```

Both are `TimeoutCancellationException`, the contention signature — the box's load average was 10.19
across that window, with sibling containers competing for the same cores. Other cases in both classes
passed in the same run. Re-run serially, both classes are green:

```
just test-class AbuJafarTest         BUILD SUCCESSFUL in 30s
just test-class AccursedCentaurTest  BUILD SUCCESSFUL in 20s
```

Every corpus-wide gate this change could actually move is green in the full run:
`CardDefinitionSnapshotTest` 324 passed (this includes the re-blessed DRK golden), `CardLintTest`,
`FacadeBoundaryTest`, `TokenArtCoverageTest`, `BasicLandArtOrderTest`, `MtgSetCatalogTest`.

## What the gate does not cover

`just test` covers the Kotlin corpus. The field-fidelity number means: against a Scryfall dump taken
2026-08-23, on those thirteen fields, for the cards in `set.cards`. It says nothing about rulings,
legalities, or whether a card plays correctly. The DSL pass is twelve reviewers' judgement plus six
adversarial verifications, not a proof — and `assay-differential` read only 45 of 119 cards whole,
so 72 cards got no signal from it at all. No web-client code changed, so nothing here is evidence
about the client.

The throwaway `DrkCardFieldVerificationTest` and the Scryfall dump it reads are deliberately not
committed, per the skill.


bolav:
# Self-review — verify The Dark, archive its backlog

Reviewing the two commits on `drk-finish` that this pass added, not the 40 that preceded them.

## What I'd want a reviewer to check hardest

**Angry Mob's rewrite is the only change with gameplay surface.** Everything else in the first commit
is a string. The claim is that these two spellings compute identically:

```kotlin
// before — constant inside the amount
Conditional(IsYourTurn, ifTrue = Add(Fixed(2), swampCount), ifFalse = Fixed(2))

// after — constant as the printed offset
Conditional(IsYourTurn, ifTrue = swampCount, ifFalse = Fixed(0)), powerOffset = 2, toughnessOffset = 2
```

On your turn: `swamps + 2` either way. Off your turn: `2` vs `0 + 2`. `AngryMobScenarioTest` passed
unchanged in the full suite, and it asserts 5/5 against three opponent Swamps, which is the case that
would break if the offset weren't applied. The snapshot diff shows exactly `Dynamic` →
`DynamicWithOffset` with `offset: 2` on both power and toughness. I did **not** add a new test for the
off-turn branch; the existing test covers the on-turn one and the snapshot pins the shape.

**The snapshot re-bless is DRK-only.** `just rebless-cards` moved one file, `DRK.json`, 23 insertions
/ 30 deletions, and every hunk maps to one of the eleven fixes. An unrelated card moving would have
meant I'd changed shared SDK behaviour; none did.

## Where I was wrong along the way

- My first flavor-text patch used `re.sub` with the replacement containing `\n`, which `re.sub`
  interprets — so four files got a literal newline inside a Kotlin string literal and the module
  stopped compiling. Caught by the very next build, reverted those four files with `git checkout` and
  redid them with a lambda replacement. Worth noting only because the four files were briefly broken;
  the committed state compiles and the full suite ran against it.
- One reviewer reported Angry Mob's `2+*` as a possible P/T bug and another reported Spitting Slug's
  "otherwise" clause as missing. Both were wrong, and both were caught by sending the claim to a
  second agent briefed to refute it rather than acting on the first read. That's two card edits not
  made. I'd have shipped both as "fixes" without the verification step.

## Judgement calls a reviewer might disagree with

- **I did not fix the six confirmed engine/SDK divergences.** Three of them are genuinely small and
  card-shaped — Sorrow's Path is two lines using a `sameController` flag that already exists, Worms of
  the Earth is about five, Reflecting Mirror is a new `Targets` value plus a one-line card edit. I
  left all of them because each is a behaviour change that wants its own scenario test and its own
  review, and bundling five of them into a verification pass makes the diff unreviewable. They are
  written up with file:line and the exact fix in `the-dark-engine-gaps.md`. Reasonable to push back
  and ask for the three card-shaped ones as a follow-up unit.
- **I reverted `backlog/token-art-gaps.md`.** Regenerating it produces 14 gaps across nine sets where
  the committed file says "**None.**" DRK appears in neither version, so it doesn't change this set's
  verdict, and shipping other sets' churn inside a DRK change seemed wrong. But it does mean the
  committed file may be stale and I've left it that way — flagged in the PR body.
- **I archived the backlog even though `check-backlog-implementations` is red.** All 72 of its
  entries are in four `bloomburrow-commander` files, untouched by this change and failing before it.
  Per AGENTS.md that's someone else's work to fix, not mine to touch.

## Gate honesty

`just test` went red on `AbuJafarTest` and `AccursedCentaurTest`, both
`TimeoutCancellationException` rather than assertion failures, with the box's load average at 10.19.
Both pass on serial re-run (30s and 20s). I did **not** re-run the whole suite afterwards — under
this box's contention that reliably manufactures a different unrelated red, so the evidence I'm
offering is: every corpus-wide gate the change can move passed in the full run
(`CardDefinitionSnapshotTest` 324, `CardLintTest`, `FacadeBoundaryTest`, `TokenArtCoverageTest`,
`BasicLandArtOrderTest`, `MtgSetCatalogTest`), and the two reds are green in isolation.

No web-client code changed, so nothing here is evidence about TypeScript.

## Coverage I'm claiming, precisely

- Field fidelity: 119 cards × 13 fields against a dump taken 2026-08-23, 0 discrepancies, `WAIVED`
  empty. Says nothing about rulings or legalities.
- DSL fidelity: 12 reviewers' judgement over 119 cards, plus 6 adversarial verifications. Not a proof.
- `assay-differential` read 45 of 119 cards whole; **72 declined and got no signal at all.** A green
  differential here is a floor, not a pass.
- Tokens: `token-art-gaps` lists DRK zero times and `TokenArtCoverageTest` passes. That means no DRK
  token falls back to generic art or a Scryfall name-guess. The set's only token site is Dance of
  Many's copy token, which takes the copied creature's art.
- Behaviour: 70 of 119 cards have a scenario test. The 49 without are mostly vanilla or
  single-keyword; the ones I'd actually want covered are listed in the gaps file, with **Maze of Ith**
  first because no test anywhere in the repo exercises `PreventCombatDamageToAndBy`.

PR #1999, opened 2026-08-23
Implement Fallen Empires (102/102) and verify it against Scryfall

bolav:
# Implement Fallen Empires (102/102) and verify it against Scryfall

Fallen Empires, complete and verified. 102 canonical `CardDefinition`s, the SDK and engine
vocabulary the set turned out to need, and a `verify-set` pass that proved the result card by card
and fixed what it found.

The set is 1994, so almost every card here is its own earliest printing and the canonical belongs in
`definitions/fem/`. Four later sets that reprint a FEM card get `Printing(...)` rows instead
(8ED Orcish Spy, C13 Night Soil, C14 Havenwood Battleground, M12 Goblin Grenade).

## What the set needed from the SDK

Fallen Empires has no named keyword mechanics. Its identity is five factions expressed as unnamed
themes, and what they cost is mostly *shapes* rather than keywords:

- **Asymmetric stat counters.** The set prints +1/+2, +2/+2 and -2/-2. CR 122.1a defines a +X/+Y
  counter generally, but `CounterType` enumerated only the six ±1 kinds. Added the three the set
  needs, summed by `EffectApplicator` and round-tripped by `CounterTypeResolver`.
- **Tide counters** (Homarid, Tidal Influence) — a counter whose *exact* count switches a static on
  and off, then resets at four or more. That is an equality condition, not a threshold: three
  counters is the -1/-1 and specifically **not** also the +1/+1.
- **Per-turn activation counting** (Farrelite Priest, Initiates of the Ebon Hand) —
  `ActivatedAbility.trackActivations` opts into the bookkeeping without being limited by it, and
  `Conditions.ThisAbilityActivatedThisTurnAtLeast` reads it back, scoped per ability per permanent.
- **`AbilityFlag.ASSIGNS_NO_COMBAT_DAMAGE`** for the set's signature "attacks and isn't blocked …
  it assigns no combat damage this turn" rider. Deliberately *not* modelled as prevention: the
  creature assigns nothing at all, so no damage event happens and nothing downstream sees one.
- **Routing a per-creature payment to that creature's own controller** —
  `Player.ControllerOfIterationEntity`, with `StatePredicate.IsBlockingIterationEntity` for Tidal
  Flats' "creatures you control blocking that creature". Heroism bills each attacking red creature's
  controller separately; a single controller reference cannot express that.
- **`AbilityCost.AttachedPermanentManaCost`** (Merseine) — an ability whose cost is the *enchanted*
  permanent's mana cost.

Each is a small generic primitive rather than a card-specific type, and each is catalogued in
[`docs/card-sdk-language-reference.md`](docs/card-sdk-language-reference.md) as the same-change rule
requires.

## What verification found

`cards.md` reading 102/102 is a claim. The `verify-set` pass turned it into evidence, and the
evidence was not clean.

**Field fidelity.** All 102 compiled `CardDefinition`s were compared against a fresh Scryfall dump on
thirteen fields. Four discrepancies, all real, all fixed: Heroism's and Thelon's Curse's colour
identities were wrong (W instead of WR, G instead of GU — both cards put a mana symbol in *rules
text*, which counts toward colour identity under CR 903.4 and which the cost-derived heuristic cannot
see), Goblin Flotilla had no flavour text, and Goblin Grenade's attribution sat on the wrong line.
The harness itself is not committed — it is scaffolding that asserts against a snapshot of a moving
source.

**Four gameplay bugs**, each confirmed by an independent verifier briefed to *refute* it:

- **Deep Spawn** — "sacrifice this creature unless you mill two cards" is the only printed
  pay-or-suffer cost in the corpus that mills, and the branch was never written:
  `PayOrSufferExecutor` returned `EffectResult.error("… not yet implemented")`. The ability had
  already left the stack by then, so the upkeep trigger silently did *nothing* — the Spawn neither
  milled nor died. Implemented. Affordability is CR 701.17b: a player can't pay a cost that mills
  more cards than their library holds, so a short library can't pay and the sacrifice happens with no
  prompt.
- **Farrel's Mantle** — its trigger could never fire, making it a `{2}{W}` Aura that did nothing.
  `BecomesUnblockedEvent` was matched for the SELF binding only, and `AttachmentTriggerDetector` had
  no route for it either. Fixed the way the sibling case already was: `BlocksOrBecomesBlockedByEvent`
  takes a documented exception to "ATTACHED belongs to the attachment detector" because it needs the
  whole block map; "isn't blocked" needs it more strongly still, being a *negative* over that map.
  Also made the may-then-target trigger path honour `decisionMaker`, so "its controller may" asks the
  enchanted creature's controller — a different player whenever the Mantle sits on an opponent's
  creature, which is most of the reason to play it.
- **Orcish Spy** — "look at the top three cards" silently *reordered* them. `GatherCardsEffect` never
  removes the cards, so the move putting them "back" on top was a real relocation, and top placement
  prepends one card at a time: [c1,c2,c3] came back [c3,c2,c1]. The gather alone is the whole ability.
- **Dwarven Soldier** — "blocked by **one or more** Orcs" fired once per Orc.
  `Triggers.BlocksOrBecomesBlockedBy` had no way to ask for the collapsed form, so it gained an
  `oncePerCombat` parameter (default false — no existing card changes).

**Behaviour.** 26 of 102 cards carry a scenario test; the rest are vanilla creatures, cycle members
covered by a sibling, or lands. A self-play pass drove 13 games across 8 decklists, up to 41 turns,
exercising 62 of the 102 cards, and produced **no** gameplay finding — no 5xx, no missing legal
action, no failed trigger, no illegal board state. It positively confirmed the Thallid spore cycle,
Homarid's four-beat tide reset, Deep Spawn's mill, Ebon Praetor's counter arithmetic, Icatian Town's
four tokens and Goblin Warrens.

**Tokens.** Five tokens from eight cards. Fallen Empires predates token *cards* — there is no `tfem`
set to sync from — so four borrow the closest matching art via `MtgSet.tokenArt`, and the Camarid,
which has never been printed as a token anywhere, resolves through the engine-wide table to this
set's own Homarid art. No token effect bakes an `imageUri`.

## Two findings worth reading even if you skim the rest

**A passing test that proved nothing.** The Dwarven Soldier fix initially passed its own mutation
check. With two Orc blockers the projected P/T was *already* 2/3, because two identical `+0/+2`
continuous effects collapse into one — the defect was visible only as two copies of the ability on
the stack. Asserting the projection would have recorded a false fix. The test asserts stack size for
that reason, and the trap is written up in the gaps file: for stat-modifying abilities, the
projection can read correctly for the wrong reason.

**A wrong rule in a verification report.** The verifier that confirmed Deep Spawn also proposed a fix
citing CR 701.13b, saying a short library should still be able to pay. CR 701.17b says the opposite.
Implemented per 701.17b, which also matches what the card's own docblock had always claimed.

## Deliberately not fixed

Recorded in
[`backlog/archived/sets/fallen-empires/fallen-empires-engine-gaps.md`](backlog/archived/sets/fallen-empires/fallen-empires-engine-gaps.md),
because each would change shared behaviour well outside this set:

- **Vodalian War Machine's** dies-trigger is created during each ability's *resolution*, so a Merfolk
  tapped to pay is spared if removal resolves while that activation is still on the stack. Confirmed,
  but the fix is infrastructure — payers must be recorded at cost-payment time. A second claim about
  this card, that N activations produce N wrong triggers, was **refuted**: each one-shot trigger
  destroys its own baked payer, so the same set dies.
- **`MoveCollectionEffect` to the top of a library reverses the batch** — the engine defect behind the
  Orcish Spy bug. The prompting `ControllerChooses` path writes the batch in one go and preserves
  order; the non-prompting path loops and prepends. Reconciling them touches every multi-card
  move-to-top in the corpus.
- **The SDK filter-description composer** emits duplicated, misordered adjectives
  (`.withColor(WHITE).untapped().youControl()` renders as "untapped you control untapped creature
  whites you control") and leaks raw `AbilityFlag` names. Worked around by writing out five FEM
  ability descriptions; the composer renders every card in the corpus that omits one.
- Rainbow Vale's `Player.AnOpponent` auto-picks the first opponent — multiplayer only, already
  tracked.

Also out of scope: **Brushwagg** (MIR) has the same per-partner trigger bug as Dwarven Soldier and is
now a one-line fix, but it is another set's card and another set's snapshot.

## Verification

```
just test-rules    — green, 9m24s, 8400 scenario tests, 0 failures  (final run)
just test          — 2 failures, both TimeoutCancellationException, zero assertion failures
just test-scenarios 2003-2007          — green (the era that timed out, re-run serially)
:ai:test --rerun-tasks                 — green, 5m46s (the other module that timed out)
npm run typecheck                      — clean
npx vitest selectors + passiveCounters — green, 127 tests
just test-class CardLintTest           — green
just test-class FacadeBoundaryTest     — green
just test-class TokenArtCoverageTest   — green
just assay-gate --set FEM              — 102 assayed, 0 non-invertible normalizations
just assay-differential --set FEM      — 41 compared, 39 confirmed, 2 divergent
```

The two `just test` failures are the documented shared-box contention pattern
(`CLAUDE.local.md` → "Gate selection"): both are 120-second coroutine timeouts, not assertion
failures, and **both modules pass when re-run in isolation** — the `:ai` re-run used
`--rerun-tasks` so it cannot be a cached green. A pre-PR review then found several real defects
(worst: a counter payment bypassing the placement chokepoint); those fixes are the final commit and
were re-gated with the `just test-rules` run above. The client changes are covered only by the
TypeScript gates, and the Kotlin greens say nothing about them. All 25 CR numbers cited in this branch's code and
comments were checked against `MagicCompRules_20260808.txt`.

**What this does not cover.** The 0-discrepancy field result means: against that dump, on those
thirteen fields, for the cards in `set.cards`. It says nothing about rulings or legalities. The
`assay-differential` number is thin coverage by construction — 60 of 102 cards declined, so the
parser read only 41 whole; neither divergence was a card bug (one encoding fold, one parser
limitation on Hand of Justice, whose `excludeSelf` is the rules-correct reading and was independently
confirmed by a second reviewer). The DSL pass is 11 reviewers' judgement across 102 cards, not a
proof. On tokens, the honest statement is narrower than "no gaps": regenerating
`token-art-gaps` *does* list one FEM row — the Camarid, which has no art anywhere on Scryfall and
resolves through the engine-wide table to this set's own Homarid. That regeneration also churned
rows for nine unrelated sets, so it is not committed here; the file in the tree is the stale one
from July. `TokenArtCoverageTest` is green, which proves only that nothing renders via a Scryfall
name-guess, and none of this certifies the four borrowed images are what a 1994 printing would have
shown. Nothing here
exercises the web client beyond the enum mirroring.


bolav:
# Review — `fem-complete`

Reviewed `origin/main...HEAD` (already contains `origin/main`; the merge was a no-op). Three
reviewers: trigger/cost engine, SDK vocabulary, printings + client. Every CR number cited anywhere in
the branch — 25 of them — was checked against `MagicCompRules_20260808.txt`; all correct.

**Verdict.** The engine work is sound and the trigger change is safe — the `BecomesUnblockedEvent` +
`ATTACHED` path was traced end to end and does **not** double-fire (the new branch is an `else if` in
a chain already gated by `matchesTrigger`, `AttachmentTriggerDetector.getRelevantEntityIds` returns
empty for `BlockersDeclaredEvent`, and `TriggerIndex` dedupes per entity). Printing placement is
clean across all 15 cards checked. The review found one real gameplay bug and several places where
the SDK grew vocabulary it did not need; the blocking items are fixed, the design items are flagged
below for a decision.

## Fixed in this branch after review

- **Counter payments bypassed the placement chokepoint.** `SacrificeAndPayContinuationResumer`'s
  put-counters payment (Tourach's Chant, Thelon's Chant) wrote straight to `CountersComponent` and
  emitted a bare `CountersAddedEvent`, skipping all four steps every other counter path runs: the
  "can't have counters put on it" gate (Solemnity), the placement replacements (Hardened Scales,
  Doubling Season), the first-placement-this-turn marker, and a placer-bearing event. A
  placer-less event makes every placer-restricted `CountersPlacedEvent` trigger silently decline.
  Now mirrors `CostHandler`'s `PutCountersOnSelf` exactly. **Genuine gameplay bug, and the most
  valuable thing this review found.**
- **The batched may-question asked the wrong player.** My `decisionMaker` change covered the single
  trigger but not the batch: `BatchKey` and `BatchYesNoDecision` still keyed on the ability's
  controller, so two Farrel's Mantles — one on your creature, one on an opponent's — would batch and
  fan one player's answer onto the other's decision. It also split the auto-answer store between the
  two paths. Both now key on the asked player.
- **A third hand-rolled `EffectTarget → player` resolver.** My `askedPlayerId` `when` silently
  answered "the controller" for `TargetController`, `TriggeringPlayer` and everything else — the same
  class of bug it was fixing. Replaced with the shared `TargetResolutionUtils.resolvePlayerTarget`,
  which already handles every shape. That also removed the need for the `private → internal`
  widening of `controllerOf`, now reverted: the widening was what *enabled* the duplicate ladder.
- **`GatherCardsEffect.excludeCollection` duplicated an existing primitive.**
  `CollectionFilter.ExcludeOtherCollection` already expresses set difference, and
  `ForEachPlayerCollectingEffect`'s own KDoc instructs authors to pair with it. Adding an optional
  parameter to an existing effect to cover a variation is the named anti-pattern in the principles
  doc, and it split one concept across two vocabularies. Parameter, executor arm and doc paragraph
  removed; Raiding Party composes the primitive instead, and its scenario test still passes.
- **Two documentation defects that the same-change rule should have caught.**
  `StatePredicate.ControllerControls` (new in this branch, Seasinger) was undocumented — a hard-rule
  violation. And the reference documented `Costs.pay.Exile(filter, zone = GRAVEYARD, count)` as
  "growing two flags", but that facade takes no flags and defaults to `HAND`; the flags live on
  `CostAtom.ExileFrom` and are reached only through `Costs.ExileFromSingleGraveyard`. An author
  following the docs would have written a cost that silently pays from their own graveyard.
- **A false claim about `CounterTypeFilter`.** The reference said every stat counter has a matching
  filter case; the five asymmetric kinds have none. Corrected to name the split.
- **Seven counter kinds were mirrored to the client by name only.** They reached `enums.ts` but none
  of the three tables a battlefield badge needs, so an Ebon Praetor accumulating `-2/-2` or a Homarid
  stacking tide counters rendered nothing while a `+2/+0` on the next creature rendered fine. Tide is
  the worst case: Homarid's entire clock *is* the exact count. All seven now carry allowlist entry,
  icon class and palette.
- **A storage land with no counters glowed as clickable** — `hasXCost` alone, with no check that X
  could exceed zero, so the picker opened at `maxX: 0` and confirming tapped the land for nothing.
- Minor: a KDoc comment displaced onto the wrong function in `ObjectFilter.kt`, and the "Mill cards"
  button that dropped its count.

## Open — design decisions, not defects

These are the SDK reviewer's over-specialisation findings. Each is defensible as written and none
changes behaviour; they are recorded because the *next* card in each shape forces another type.

1. **`CounterType` stat kinds are a hardcoded enumeration.** Adding one size now costs seven
   coordinated hand edits (`CounterType`, `Counters`, `STAT_COUNTERS`, two `CounterTypeResolver`
   functions, `EffectApplicator`'s literal arithmetic, and the client enum) — and this branch made
   that worse by adding three. The (X,Y) is derivable from the printed name; the coverage test's own
   regex proves it. A `statDelta` on the enum, summed generically, would reduce it to one entry.
2. **`Conditions.ThisAbilityActivatedThisTurnAtLeast`** bakes the comparison into the type where the
   principles table calls for `Compare` over a tracked `DynamicAmount`. The tally is already
   first-class; only its exposure is wrong.
3. **`Conditions.ExiledAsCostHadSubtype`** is subtype-only where `Conditions.EntityMatches(entity,
   filter)` is the documented shape — and this branch adds `EffectTarget.TappedAsCost` right next
   door, so the missing piece is one sibling `EffectTarget.ExiledAsCost`.
4. **`Duration.WhileYouControlSourceAndSourceTapped`** is the second hand-fused conjunction duration;
   a `Duration.All(listOf(...))` combinator would absorb both and the third.
5. **`StatePredicate.IsBlockingIterationEntity`** is the third hardcoded-referent blocking predicate,
   byte-identical to `IsBlockingSource` with one field swapped.
6. **`AbilityCost.AttachedPermanentManaCost`** bakes an entity role into a type where an
   `EffectTarget` parameter would generalise it.

I'd take (1) before the next old set — it is the one with a growing per-addition cost. The rest are
one-card-each today.

## Also open

- **`Costs.pay.Exile` cannot express the single-graveyard shape at all.** The flags exist only on the
  atom, so a `PayOrSuffer` version of Night Soil is currently unwritable. Documented as a limitation
  rather than papered over.
- **Affordability vs mill replacement.** The mill cost checks affordability against the unmodified
  count and pays the `MillAmountModifier`-adjusted one, so a mill-doubler lets a player accept a
  "mill 2" they can afford and mill 4 from a 3-card library, truncated. This matches
  `CostPaymentService`'s existing shape, so it is consistency rather than regression — but it is a
  real corner and CR 701.17b arguably wants the check on the modified number.
- **`ASSIGNS_NO_COMBAT_DAMAGE` gets no battlefield chip.** Consistent with
  `ASSIGNS_COMBAT_DAMAGE_AS_TOUGHNESS`, but that one is printed on its own card while this is granted
  onto someone else's creature mid-combat, so the defending player sees damage vanish with no trace.
- **Brushwagg (MIR)** has the same per-partner trigger bug Dwarven Soldier had, and is now a one-line
  fix using the `oncePerCombat` parameter this branch adds. Left alone as another set's card.

## Verification of the review fixes

`just test-rules` and `just test-scenarios 1993-1999` re-run after the changes above;
`just test-class RaidingPartyScenarioTest` and `FarrelsMantleScenarioTest` green individually. The
client changes are covered by `npm run typecheck` (clean) and `vitest` over `selectors.test.ts` +
`passiveCounters.test.ts` (127 tests, green) — a Kotlin green says nothing about them, and vice
versa. The card snapshot was re-blessed for the Raiding Party rewrite; only `FEM.json` moved.

PR #2008, opened 2026-08-24
Evaluate dynamic mana amounts during auto-tap

bolav:
# Evaluate dynamic mana amounts during auto-tap

Mana sources whose output depends on the current game state were previously treated as producing one mana by the auto-tap solver. That made sources such as Elvish Archdruid, Gaea's Cradle, Tolarian Academy, Marwyn the Nurturer, and Viridian Joiner disagree with manual activation: payable spells could be hidden, and automatic payment could credit the wrong amount.

Evaluate existing `DynamicAmount` values against the current state when building `ManaSource` entries and when collecting additional leaves from composite mana effects. The solver now reports and spends the amount the ability would actually produce.

An ability whose dynamic amount currently evaluates to zero is omitted as a source. In particular, a land such as Gaea's Cradle with no creatures no longer falls through to the generic colorless-land fallback and offers phantom mana. Composite abilities remain available when another mana-producing leaf still has a positive amount.

This uses the existing dynamic-amount evaluator and introduces no new SDK vocabulary.

## Verification

- `just test-class DynamicManaAmountAutoTapTest` — passed all six regression tests, covering source discovery, available-mana counting, solving, end-to-end auto-payment, zero-output exclusion, and restoration after the count becomes positive.
- `just test-rules` — all observed engine and scenario tasks passed except `:mtg-sets:2008-2016:tests:test`, where the unrelated `AbattoirGhoulScenarioTest` exceeded its 120-second timeout. That test and its combat/life-gain behavior are outside this branch's diff.

The targeted gate directly covers this change but does not independently prove every other rules-engine or card scenario. The broader gate attempted that coverage but did not complete green because of the unrelated timeout above. This branch contains no server or web-client changes, and these commands do not cover those layers.

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/mechanics/mana/ManaSolver.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/mechanics/mana/ManaSolver.kt
@@ -82,7 +82,11 @@ data class ManaSource(
     val painAmount: Int = 0,
     /** Whether this creature can attack (no summoning sickness or has haste) */
     val canAttack: Boolean = false,
-    /** Amount of mana this source produces per tap (e.g., 3 for Elvish Aberration) */
+    /**
+     * Amount of mana this source produces per tap (e.g., 3 for Elvish Aberration). A *dynamic*
+     * amount — Elvish Archdruid's "Add {G} for each Elf you control", Gaea's Cradle, Marwyn — is
+     * evaluated against the current board, so this is what the tap would actually yield right now.
+     */
     val manaAmount: Int = 1,
     /** Extra mana produced per tap from auras like Elvish Guidance */
     val bonusManaPerTap: Int = 0,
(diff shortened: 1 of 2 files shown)


bolav:
# Self-review: dynamic mana auto-tap

Verdict: No findings. The implementation is correct, composes existing engine primitives, and is ready for review with the unrelated broad-suite timeout disclosed in the PR body.

The change evaluates the same `DynamicAmount` representation already used by mana resolution, so it adds no SDK surface or card-specific execution path. The zero-output guard excludes dry abilities before their colors and cost characteristics enter the combined source, while preserving composite abilities with another positive mana leaf. The regression class exercises both positive and zero board-dependent amounts through source enumeration, affordability solving, and actual automatic payment.

Verification:

- `git diff --check origin/main...HEAD` passed.
- `just test-class DynamicManaAmountAutoTapTest` passed all six tests.
- `just test-rules` was not fully green: the unrelated `AbattoirGhoulScenarioTest` timed out after 120 seconds in `:mtg-sets:2008-2016:tests:test`; no changed file or behavior in this branch concerns that test.

Recommendation: Merge after normal CI confirms the broader suite without the environmental timeout.

PR #2009, opened 2026-08-24
Fire "activates an ability" triggers off mana abilities

bolav:
# Fire "activates an ability" triggers off mana abilities

Elrond, Moon-Reader reads "Whenever you activate an ability of a creature, draw a card."
There is no *"that isn't a mana ability"* clause on it, and a mana ability is still an
activated ability — CR 605.3 says activating one "follows the rules for activating any
other activated ability", it just resolves without using the stack (CR 605.3b). So
tapping a Llanowar Elves for `{G}` has to trigger it. The card's own Scryfall ruling says
so in as many words:

> An activated ability of a creature that's also a mana ability (such as "{T}: Add {G}")
> can cause Elrond's first ability to trigger.

We had the opposite behaviour codified in three places: the trigger spec, the card's
KDoc, and a scenario test asserting the mana ability does *not* fire it. Fixing it took
two independent changes, because two separate things stopped the event from ever reaching
the matcher.

## The trigger vocabulary

`EventPattern.AbilityActivatedEvent` gains `includeManaAbilities`, which drops the default
"isn't a mana ability" gate. It is deliberately a separate field from the existing
`excludeManaAbilities` rather than a reuse of it: that one *subtracts* mana abilities from
the exhaust semantic (Pit Automaton), this one *adds* them to the default semantic. Same
axis, opposite polarity — folding them into one flag makes both unreadable at the call
site. `Triggers.activatesAbilityOf` takes it through, and the pattern's rendered
`description` now omits the qualifying clause entirely when nothing narrows the trigger
beyond its source, so Elrond reads as "you activate a creature's ability" rather than
gaining a contradictory tail. Every existing combination renders byte-identically.

A trigger that sets the flag must not itself be able to produce mana, or CR 605.1b would
make it a mana ability and it would resolve off the stack. Elrond draws a card, so it uses
the stack normally — and per CR 603.3 it waits for priority, which means the draw goes on
the stack *above* the spell whose cost the mana paid for.

No new effect, executor, or SDK type: this is a flag on an existing event pattern and a
parameter on an existing trigger factory.

## Gap 1 — the manual activation path

`ActivateAbilityHandler` emitted `AbilityActivatedEvent` for a mana ability only when its
cost lacked `{T}` — the carve-out that serves the Antiquities "activates an ability without
{T} in its activation cost" template (Haunting Wind, Powerleech, Artifact Possession).
Tapping Llanowar Elves emitted nothing at all, so there was no event for any matcher to
accept or reject.

It now emits for every mana ability, with `costsTap` set from the cost, so consumers pick
their semantic off the flags rather than off which events happen to exist: the default
wording rejects `isManaAbility`, the Antiquities wording rejects `costsTap`, and the
unqualified wording accepts both.

The emission also **moved to the top of the mana branch**. That branch has two pause exits
— an any-color mana effect (Birds of Paradise pausing for a color decision) and an
any-color tap bonus (Fertile Ground) — and the ability is already activated by the time we
reach it, its costs paid. Building the event up front and adding it to the carried event
list gets it out through those exits; the pause path's deferred-trigger detection names it
too.

## Gap 2 — the auto-tap fast path

When the engine auto-taps sources to pay a cost, the mana solver taps them directly and
credits the mana, bypassing the activated-ability flow entirely. No activation event was
emitted anywhere. That path is a UI shortcut for the player activating the source's mana
ability, not a different game action, so it now emits the same event — otherwise Elrond
would miss every mana dork tapped to pay for a spell, which in real games is most of them.

`ManaAbilitySideEffectExecutor.tapSourcesWithSideEffects` is the shared choke point for ten
call sites (cast, cycle, plot, foretell, suspend, typecycle, morph, room unlock, cost
payment, mana payment window), so one emission covers all of them.
`ActivateAbilityHandler.autoTapForManaCost` taps inline rather than through that executor
and needed the same treatment; the executor is now a constructor dependency, wired through
the single `create(services)` factory.

## Knock-on effects, both intended

**`YouActivateExhaustAbility` now works as documented.** It is described as counting an
exhaust *mana* ability too, but never could, because a `{T}`-costed one emitted no event.
This is a live behaviour change for Aetherdrift cards beyond Elrond, and the most likely
thing to surprise a reviewer.

**Mana-ability activations stay out of the client game log.** `ClientEvent` drops them.
They have never appeared there, and the mapping has no ability description to render, so
un-gating emission would otherwise have put an empty-description line in the log on every
land tap. `ManaAddedEvent` already reports what the tap produced.

## Deliberately not fixed here

`LandTappedForManaEvent` still doesn't fire on auto-pay, so "whenever a player taps a land
for mana" (Mana Flare, Overabundance) continues to under-fire there. I had expected this to
come for free at the same emission site and it does not: the solver already pre-accounts
for mana-adding tap statics via `bonusManaSpentByColor` / `remainingBonusMana`, so firing
those triggers on auto-pay would double-add the mana. It needs that accounting reworked
first and belongs in its own change. `AbilityActivatedEvent` has no such conflict — nothing
consumes it to produce mana.

## Verification

`just test-rules` and `just test` on this box both surface killed test JVMs
(`finished with non-zero exit value 13`) and coroutine spec timeouts whenever several test
JVMs run concurrently, so the modules were also gated **serially**, and every result below
was confirmed fresh by reading the JUnit XML timestamps rather than trusting
`BUILD SUCCESSFUL`:

**14,470 tests, 0 failures, 0 errors** across every module, each run against source at or
newer than the last edit it depends on.

A fresh PR-readiness pass also ran the focused gates on the final branch:
`just test-class ManaAbilityActivationTriggerScenarioTest` (3/3 green) and
`just test-class ElrondMoonReaderScenarioTest` (6/6 green).

Three failures during the parallel runs were investigated rather than retried away:

- `CardDefinitionSnapshotTest` (HOB golden) — **real, and mine.** Re-blessed; the diff is
  exactly `includeManaAbilities: true` on Elrond and nothing else in the corpus moved,
  which also confirms no other card's trigger spec or rendered description shifted.
- `ARealmRebornTest` and `AbhorrentOculusScenarioTest` — 120-second coroutine spec
  timeouts. Both are contention, not a slowdown from the extra event: run serially those
  same tests take **20.2s** and **16.8s**, and their modules complete 1,595 and 1,921 tests
  in about a minute each, which leaves no room for a spec anywhere near 120s.
- `FreeForAllLobbyTest` — websocket `eventually(10s)` timeout at load average 9.5; 14/14
  green on an idle box.

Both halves are mutation-checked. Disabling only the auto-tap emission turns
`auto-tapping a creature for mana while casting a spell` red with the manual-path tests
still green; disabling only `autoTapForManaCost`'s emission turns
`auto-tapping a land to pay an ability cost fires it too` red with the other two engine
cases still green. So each tier is independently load-bearing and each test exercises the
code it names.

**What this does not cover.** No TypeScript changed, so nothing here is evidence about the
web client beyond the fact that `AbilityActivatedEvent` for a mana ability simply stops
arriving; no client code branches on it today (the type is declared but has no consumer).
No E2E run. And no self-play pass — the auto-tap emission touches the mana-payment hot
path used by every game, and while the unit and scenario suites are green, a
`gym`-driven self-play sweep would exercise it far harder than they do.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -5368,13 +5368,21 @@ Named sugar for the common type-primitive cases; reach for `youCastSpell(...)` p
   with `Effects.DealDamage(n, EffectTarget.PlayerRef(Player.TriggeringPlayer))` to punish the activator (Flamescroll
   Celebrant). Backed by `EventPattern.AbilityActivatedEvent(player)`.
 - `YouActivateAbility` — you activate an ability that isn't a mana ability (the `Player.You` form of the above).
-- `activatesAbilityOf(sourceFilter, player?)` — the source-scoped form of the above: an ability that isn't a mana
-  ability, activated from a permanent matching `sourceFilter`. Elrond, Moon-Reader's "whenever you activate an
-  ability of a creature" is `activatesAbilityOf(GameObjectFilter.Creature)`; pair it with `oncePerTurn = true` on
-  the triggered ability for the "this ability triggers only once each turn" clause. Backed by
-  `EventPattern.AbilityActivatedEvent(player, sourceFilter)`. Unlike `activatesAbilityWithoutTap` (below), which
-  keys on the literal `{T}`-in-cost wording, this keeps the "isn't a mana ability" gate — so a creature's
-  tap-for-mana never fires it while a `{T}`-costed non-mana ability does.
+- `activatesAbilityOf(sourceFilter, player?, includeManaAbilities?)` — the source-scoped form of the above: an
+  ability activated from a permanent matching `sourceFilter`. By default it keeps the "isn't a mana ability" gate;
+  set `includeManaAbilities = true` for the unqualified Oracle wording, where mana abilities count too. Elrond,
+  Moon-Reader's "whenever you activate an ability of a creature" is
+  `activatesAbilityOf(GameObjectFilter.Creature, includeManaAbilities = true)` — its ruling confirms a creature's
+  `{T}: Add {G}` triggers it, since a mana ability is still an activated ability (CR 605.3). Pair it with
+  `oncePerTurn = true` on the triggered ability for the "this ability triggers only once each turn" clause. Backed
+  by `EventPattern.AbilityActivatedEvent(player, sourceFilter, includeManaAbilities)`. Unlike
+  `activatesAbilityWithoutTap` (below), which keys on the literal `{T}`-in-cost wording, this ignores the tap cost
+  entirely. A trigger that sets `includeManaAbilities` must not itself be able to add mana, or CR 605.1b would make
+  it a mana ability (and it would resolve off the stack).
+
+  **Engine note.** The engine emits `AbilityActivatedEvent` for *every* activated mana ability, on the manual
+  activation path and on the auto-tap fast path used to pay for spells and abilities, with `isManaAbility = true`
+  and `costsTap` set from the cost. Mana-ability activations are deliberately kept out of the client game log.
 - `YouActivateExhaustAbility` — you activate an ability marked `isExhaust`. Backed by
   `EventPattern.AbilityActivatedEvent(player = Player.You, requireExhaust = true)` and the activation event's
   `isExhaust` flag. The event is emitted as soon as the exhaust ability is put on the stack, so the triggered
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/dsl/Triggers.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/dsl/Triggers.kt
@@ -1275,20 +1275,28 @@ object Triggers {
     )
 
     /**
-     * Whenever [player] activates an ability *of a permanent matching [sourceFilter]* that isn't a
-     * mana ability — Elrond, Moon-Reader's "whenever you activate an ability of a creature".
+     * Whenever [player] activates an ability *of a permanent matching [sourceFilter]* — the
+     * source-scoped sibling of [YouActivateAbility].
      *
-     * The source-scoped sibling of [YouActivateAbility]: same "isn't a mana ability" gate (CR 605 /
-     * 606), plus the activated ability's source permanent must match. Unlike
-     * [activatesAbilityWithoutTap], which keys on the literal "{T} in its activation cost" wording,
-     * this leaves the tap cost alone — a {T} ability of a creature fires it just as a costless one
-     * does.
+     * By default this keeps that trigger's "isn't a mana ability" gate (CR 605 / 606) and only adds
+     * the requirement that the activated ability's source permanent match. Set
+     * [includeManaAbilities] for the unqualified Oracle wording — Elrond, Moon-Reader's "whenever
+     * you activate an ability of a creature", whose ruling confirms a creature's "{T}: Add {G}"
+     * triggers it, since a mana ability is still an activated ability (CR 605.3).
+     *
+     * Unlike [activatesAbilityWithoutTap], which keys on the literal "{T} in its activation cost"
+     * wording, this leaves the tap cost alone — a {T} ability fires it just as a costless one does.
      */
     fun activatesAbilityOf(
         sourceFilter: GameObjectFilter,
-        player: Player = Player.You
+        player: Player = Player.You,
+        includeManaAbilities: Boolean = false
     ): TriggerSpec = TriggerSpec(
-        event = AbilityActivatedEvent(player = player, sourceFilter = sourceFilter),
+        event = AbilityActivatedEvent(
+            player = player,
+            sourceFilter = sourceFilter,
+            includeManaAbilities = includeManaAbilities
+        ),
         binding = TriggerBinding.ANY
     )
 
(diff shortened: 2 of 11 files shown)


bolav:
## Review — `mana-ability-activation-triggers`

**Verdict.** Behaviour is right and matches the card's ruling and CR 605.3; the SDK shape is
right — a flag on an existing event pattern plus a parameter on an existing trigger factory,
no new `Effect`, executor, or type. Two findings were raised against the first commit and
fixed in `3836ed3f6c`; what remains is one pre-existing engine hole the change now touches
(M1) and three notes. Nothing blocking.

Branch already has `origin/main` merged (it was already up to date, no merge commit).
`just check-card-printing` is not applicable — no `card(...)` or `Printing(...)` row is added
or moved; Elrond's canonical stays in HOB, its only printing.

### What's good

- The flag is genuinely parameterized rather than card-shaped. `includeManaAbilities` sits on
  `EventPattern.AbilityActivatedEvent` next to the flags for the other three Oracle wordings,
  so the next unqualified "whenever you activate an ability of a ~" card is a one-argument
  change rather than a new trigger type.
- Keeping it distinct from `excludeManaAbilities` is the right call and the KDoc at
  `mtg-sdk/.../EventPattern.kt:2114` says why: opposite polarity on the same axis, one
  subtracting from the exhaust semantic and one adding to the default. Merging them would
  have been the tempting, worse move.
- Emitting for *every* mana ability and letting consumers discriminate on `isManaAbility` /
  `costsTap` is better than the previous shape, where the Antiquities template's needs
  determined which events existed at all. Semantics now live in the matcher, not in the
  emitter's coverage.
- `tapSourcesWithSideEffects` was the right choke point to find — ten auto-tap call sites
  covered by one emission.
- The `description` rewrite at `EventPattern.kt:2135` is careful: the qualifier is computed
  first and appended only when non-empty, so every pre-existing combination renders
  byte-identically and only the new wording changes. The corpus-wide snapshot re-bless
  touching exactly one line confirms it.
- Both halves are mutation-checked, not just asserted green.

### Fixed during review — `3836ed3f6c`

- **I1 (fixed).** `ManaAbilitySideEffectExecutor` resolved the matching mana ability twice per
  auto-tapped source — once in `activationEvent`, once in `runSideEffects` — each a registry
  lookup plus a filter over `activatedAbilities`, on a loop that runs for every source of
  every payment. Now resolved once in `tapSourcesWithSideEffects` and passed to both private
  overloads (`:109`/`:118`, `:149`/`:158`), with the public forms kept as wrappers for callers
  holding only a produced color.
- **I2 (fixed).** The engine rule was pinned only by `ElrondMoonReaderScenarioTest` — a card
  test in `mtg-sets/2026`, for engine vocabulary. Added
  `rules-engine/.../ManaAbilityActivationTriggerScenarioTest.kt` in the shape of the existing
  `ArtifactTapActivateTriggerScenarioTest`, with inline cards and no dependency on Elrond or
  HOB. Its third case is the one that matters: it covers
  `ActivateAbilityHandler.autoTapForManaCost`, which taps inline instead of through the shared
  executor and is therefore unreached by every other test on the branch.

### Minor

- **M1.** `ActivateAbilityHandler.kt:1769` — `if (bonusResult.isPaused) return bonusResult`
  returns before the branch's `detectTriggers`, so if an any-color *tap bonus* (Fertile Ground
  on a mana creature) pauses for a color decision, the activation event rides out in the event
  stream but nothing detects a trigger on it: `SubmitDecisionHandler` re-detects over the
  resume's own events, not the ones already handed back with the paused result. Elrond would
  miss that activation. This is a pre-existing hole of the same shape — cost-side triggers are
  dropped there today for the same reason — and the *other* pause exit (an any-color mana
  effect, the common Birds of Paradise case) is handled at `:1596`. Narrow enough to leave,
  but it should be a known limitation rather than an unnoticed one, and it wants fixing with
  the cost-trigger case rather than separately.
- **M2.** The rendered description reads "you activates a creature's ability" —
  `Player.You.description` is `"you"` and the builder appends `" activates "`. Pre-existing
  (it previously produced "you activates an ability that isn't a mana ability"), untouched by
  this change, and out of scope here; noting it because this diff edits that exact
  `buildString`.
- **M3.** `runSideEffects`' public 4-arg form (`:149`) has no callers anywhere in the repo —
  pre-existing dead API, now with one more wrapper layer in front of it. Worth deleting in a
  cleanup, not here.
- **M4.** Mana-ability activation events carry `abilityEntityId = null`, so a `targetMatch`
  trigger can never match one. That is correct — a mana ability can't target (CR 605.1a) — but
  it is load-bearing and undocumented at the emission sites.

### Recommendation

Merge as-is. Flag two things to reviewers who own adjacent areas: `YouActivateExhaustAbility`
starts firing for `{T}`-costed exhaust mana abilities, which is what its documentation always
claimed but never did, so Aetherdrift cards change behaviour; and M1 should be folded into
whoever next touches the mana-branch pause paths.

The gate is green serially across every module (14,470 tests, 0 failures), but it has **not**
been through a self-play pass, and the auto-tap emission sits on the mana-payment path every
game uses. A `gym`-driven sweep before or shortly after merge would exercise it harder than
the scenario suites do.

The final PR-readiness pass additionally re-ran
`just test-class ManaAbilityActivationTriggerScenarioTest` (3/3 green) and
`just test-class ElrondMoonReaderScenarioTest` (6/6 green).

PR #2010, opened 2026-08-24
Add modal double-faced lands, and the whole ten-card Pathway cycle

bolav:
# Add modal double-faced lands, and the whole ten-card Pathway cycle

CR 712.12, verified against `MagicCompRules_20260808.txt`:

> A player playing a modal double-faced card or a copy of a modal double-faced card as
> a land chooses one of its faces that's a land before putting it onto the battlefield.
> It enters the battlefield with that face up.

Nothing in the engine modelled that. `PlayLandEnumerator` had no notion of faces at
all, and the only MDFC-land path that existed was the *opposite* shape — a land front
with a castable spell back (the FIN Towns adventure), handled over in
`CastSpellEnumerator`. A Pathway could only ever have been its front face, so the whole
ten-card cycle was unimplementable.

## Engine

- **`PlayLand.asBackFace`** — the face choice, taken at play time. Defaulted `false`,
  so every existing land play and every serialized action is unchanged. There is no
  mode prompt after the fact; the choice *is* which action you take.
- **`ModalDfcCasts.landFace`** — the land-play mirror of `castFace`, in the object that
  is already the single source of truth for "which face of this modal DFC, and may I".
  The two are now disjoint by construction: `castFace` excludes a land back face, which
  also stops a future spell-front // land-back card (Emeria's Call // Emeria, Shattered
  Skyclave) from being offered as an illegal cast. That was a latent bug — a land is a
  permanent, so the old `isPermanent` filter would have surfaced "Cast Emeria,
  Shattered Skyclave".
- **`PlayLandEnumerator`** now emits one action per land *face*, each described by the
  face it plays, from hand, from a Muldrotha-style graveyard permission, and from a
  linked exile. The front offer still keys on the card's own type line, which off the
  battlefield is the front face's (CR 712.8a) — so a spell-front card correctly offers
  only its back.
- **`PlayLandHandler`** validates against the face being played (CR 712.11c's
  play-a-land analogue: only the face being played is evaluated) and swaps to it
  *before* the permanent enters — ahead of the ETB-by-type record, the
  static/replacement registration, the enters-tapped and as-enters replacements, and
  the entry event's name, all of which must read the played face (CR 712.8f). The
  existing code already anticipated this seam in a comment; this fills it in.
- **`setDfcFace`, extracted from `flipDfcInPlace`.** Choosing a face on the way in is
  *not* a transform — CR 712.9 excludes modal DFCs from transforming at all — so it
  must fire no transform trigger and must not be stoppable by a "can't transform"
  effect (Bound by Moonsilver). The extracted half does the state work; the flip
  wrapper adds back the CR 701.27b restriction check and the `TransformedEvent`. Both
  callers then share the `CardComponent` swap, the CR 712.8a front-face save, the
  static/replacement re-registration and the self-redirect rebuild, rather than the
  play path growing a second copy of them.

## SDK

`CardDefinition.modalDoubleFacedLand(front, back)` — a sibling of
`modalDoubleFacedPermanent` for exactly the case its "the back face must carry its own
mana cost, because it is castable (CR 712.11b)" requirement excludes. Both faces must
be lands with no mana cost and no colour indicator.

## Client

Two `PlayLand` actions for one card broke two client assumptions that were a `find`
and a boolean where they needed to be a `filter` and a count:

- `buildActionOptions` took `legalActions.find(type === 'PlayLand')` and would have
  silently dropped the back face, dumping the player onto whichever face the server
  happened to list first.
- `hasMultipleCastingOptions` counted "has a play-land action" as one option, so two
  land faces read as one way to play the card.

Both now count faces. The menu labels each entry with the server's description
("Play Lavaglide Pathway") rather than `cardInfo.name`, which is the *front* face's
name for both entries and so can't tell them apart. The hand card's flip preview and
the battlefield art needed nothing: `ClientStateTransformer.dfcBackFace` already falls
back to `cardDef.backFace` off the battlefield, and `buildCardComponentForDfcFace`
already prefers the played face's own image.

## The cards

All ten Pathways. The cycle is **not** a Zendikar Rising cycle, which is what an earlier
draft of this branch assumed: ZNR printed six and Kaldheim printed the other four, and
between them they cover all ten colour pairs exactly once.

| Set | # | Card | Front // Back |
|-----|---|------|---------------|
| ZNR | 258 | Branchloft Pathway // Boulderloft Pathway | `{G}` // `{W}` |
| ZNR | 259 | Brightclimb Pathway // Grimclimb Pathway | `{W}` // `{B}` |
| ZNR | 260 | Clearwater Pathway // Murkwater Pathway | `{U}` // `{B}` |
| ZNR | 261 | Cragcrown Pathway // Timbercrown Pathway | `{R}` // `{G}` |
| ZNR | 263 | Needleverge Pathway // Pillarverge Pathway | `{R}` // `{W}` |
| ZNR | 264 | Riverglide Pathway // Lavaglide Pathway | `{U}` // `{R}` |
| KHM | 251 | Barkchannel Pathway // Tidechannel Pathway | `{G}` // `{U}` |
| KHM | 252 | Blightstep Pathway // Searstep Pathway | `{B}` // `{R}` |
| KHM | 254 | Darkbore Pathway // Slitherbore Pathway | `{B}` // `{G}` |
| KHM | 260 | Hengegate Pathway // Mistgate Pathway | `{W}` // `{U}` |

None is a dual land and none is a land that transforms: each is one card offering two
different land plays, chosen on the way in and final. The nine added after the engine
work needed no new vocabulary — every one is
`CardDefinition.modalDoubleFacedLand(front, back)` over two single-mana-ability land
faces, which is the point of the SDK facade above.

Each card gets its own scenario test, per the repo's one-card-one-test-file rule. Those
nine are deliberately narrower than `RiverglidePathwayScenarioTest`: that one pins the
cycle's rules work once (CR 712.12, 712.9, 712.8a, the shared land drop), so the
per-card tests assert only what is card-specific — that the single card offers both
named land plays, and that each face taps for the colour it prints.

## Verification

- `./scripts/gradle-locked :rules-engine:test` — **0 failures**, re-run after the
  enumerator hot-path fix below. This is the gate that matters for the
  `flipDfcInPlace` extraction: the whole transform / daybound / disturb / DFC suite
  still passes against the refactored helper.
- `./scripts/gradle-locked :mtg-sets:test` — green after re-blessing, re-run with all
  ten cards in the corpus. Covers the golden snapshots, `FacadeBoundaryTest` and the set
  catalog. Only additions in `ZNR.json` and `KHM.json`; no existing card moved.
- `just test-class "*PathwayScenarioTest"` — **33/33 across all ten cards.** Riverglide's
  six pin the cycle's rules work: both faces offered and named; each face's identity and
  mana ability once played (asserted on the mana pool); either face spending the same
  single land drop; the face being final; and CR 712.8a putting **Riverglide** Pathway
  into the graveyard when the *Lavaglide* permanent dies — the test that proves the
  front-face save survives `ZoneTransitionService`. The other nine cards contribute three
  each: both named land plays offered, and each face taps for the colour it prints.
- `:oracle-assay`, `:game-server`, `:ai` all compile against the changed
  `PlayLand` shape.
- `just check-card-printing` on all ten — clean. ZNR is the earliest real printing for
  six, KHM for four; PZNR/PKHM/PRM/SLU/PLST are unimplemented sets, so no reprint rows
  are owed.
- `just assay-differential --set ZNR` — 9/9 confirmed, 0 divergences. Run when the branch
  held one Pathway; not re-run since, and not run for KHM.
- **Web client:** `npx tsc --noEmit` clean, and `vitest` on the two touched suites at
  45 tests (up from 42). Both client changes were mutation-checked — reverting the
  option list to `find` reddens the two-face label test, and reverting the counter to a
  boolean reddens the `hasMultipleCastingOptions` test — so neither is an untested
  edit.

**What this does not cover:**

- **No E2E run.** The drag-to-play menu path is exercised only by the unit tests above,
  so "two entries appear and each plays the right face" is verified at the option-list
  level, not in a real browser. If you want a Playwright pass before merge, say so and
  I'll add one.
- **The era scenario suite did not complete.** `just test-scenarios 2017-2022` ran 762
  tests with **zero test failures** — the 33 Pathway tests among them — and then the
  Gradle build itself died writing HTML report entries into
  `/workspace/.gradle/9.6.1/fileHashes/fileHashes.bin` (`EOFException`, "Could not add
  entry … to cache fileHashes.bin"). That is shared-cache corruption in this sandbox, not
  anything in the diff, and the house rule here is to report it rather than `clean`
  shared state. So the era-wide green is **claimed only up to those 762 passing tests**,
  not as a completed gate. `:mtg-sets:test` and the targeted `*PathwayScenarioTest` run
  both completed cleanly on their own.
- **No TypeScript re-run for the nine new cards** — they touch no client code, and the
  `tsc`/`vitest` results above are from the engine/client half of the branch.

## One thing caught on review

The first cut of the enumerator dropped the outer `typeLine.isLand` guard, so the
per-face helper ran for every card in hand and did a card-registry lookup before
concluding "not a land, no faces". Legal-action enumeration runs on every priority pass
and every AI/MCTS node, so that is a lookup per hand card per node. Fixed in `3c48304`
with an early-out on two booleans already on the `CardComponent`; the registry is now
consulted only for cards that could actually have a second face.

## Known limits, stated rather than hidden

- **The AI rollout policy is face-blind.** `PlayoutPolicy` takes
  `candidates.firstOrNull { it.actionType == "PlayLand" }`, so a rollout always picks a
  Pathway's front face. Real MCTS search sees both actions and can pick either; this
  only biases rollouts. Left alone rather than reaching into `:ai` from a rules PR.
- **The Mayhem graveyard loop** in `PlayLandEnumerator` still offers only the front
  face. It is gated on the card carrying Mayhem, which no Pathway does, so it is
  unreachable for this cycle — extending it would have been untestable speculation.
- **The dev scenario covers one card, not ten.** `manual-scenarios/` gets the Riverglide
  scenario only. The client path is name-agnostic — it keys on the server's action
  descriptions — so nine more near-identical JSON files would exercise nothing new.

## Try it

```bash
  -H 'Content-Type: application/json' -d @manual-scenarios/cards/r/riverglide-pathway.json
```

Deliberately no other lands: the hand holds one `{R}` spell and one `{U}` spell and you get
one land drop, so the face choice *is* the turn. Dragging the Pathway out should open the play
menu with two entries — this is the client half of the change, and the part no unit test
drives (see I2 in the review).

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -285,6 +285,15 @@ carries the other side for the hover preview's flip toggle.
     (unlike CR 712.8e for nonmodal DFCs, where it stays the front's). The Marvel Super Heroes hero cycle —
     Jennifer Walters // The Sensational She-Hulk, Bruce Banner // The Incredible Hulk, King T'Challa, Tony Stark,
     Monica Rambeau.
+  - **Land back** — a full `CardDefinition` in `backFace`, built with
+    `CardDefinition.modalDoubleFacedLand(front, back)`. Neither face is ever *cast*: CR 712.12 makes this a
+    **play-a-land** choice — *"A player playing a modal double-faced card as a land chooses one of its faces
+    that's a land before putting it onto the battlefield. It enters the battlefield with that face up."* So the
+    card shows up as **two land plays** in hand (`PlayLandEnumerator` emits one per land face, each named for the
+    face it plays) and `PlayLand.asBackFace` says which was taken. Neither face carries a mana cost or a color
+    indicator. Once down the permanent has only the played face's characteristics (CR 712.8f) and can never turn
+    over (CR 712.9 excludes modal DFCs from transforming); off the battlefield the card is its front face again
+    (CR 712.8a). The Zendikar Rising Pathway cycle — Riverglide Pathway // Lavaglide Pathway.
 - `PREPARE` — primary characteristics are the creature face, `cardFaces[0]` is the **prepare spell** (an
   instant/sorcery) (Secrets of Strixhaven). The card is only ever cast as the creature; the prepare spell is never
   cast from hand. A creature that carries `Keyword.PREPARED` ("This creature enters prepared") becomes prepared on
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -293,7 +293,8 @@ carries the other side for the hover preview's flip toggle.
     face it plays) and `PlayLand.asBackFace` says which was taken. Neither face carries a mana cost or a color
     indicator. Once down the permanent has only the played face's characteristics (CR 712.8f) and can never turn
     over (CR 712.9 excludes modal DFCs from transforming); off the battlefield the card is its front face again
-    (CR 712.8a). The Zendikar Rising Pathway cycle — Riverglide Pathway // Lavaglide Pathway.
+    (CR 712.8a). The ten-card Pathway cycle, split across Zendikar Rising (six) and Kaldheim (four) —
+    Riverglide Pathway // Lavaglide Pathway, Hengegate Pathway // Mistgate Pathway, and the rest.
 - `PREPARE` — primary characteristics are the creature face, `cardFaces[0]` is the **prepare spell** (an
   instant/sorcery) (Secrets of Strixhaven). The card is only ever cast as the creature; the prepare spell is never
   cast from hand. A creature that carries `Keyword.PREPARED` ("This creature enters prepared") becomes prepared on
--- a/docs/engine-server-interface.md
+++ b/docs/engine-server-interface.md
@@ -166,11 +166,11 @@ Used to perform the special action of playing a land.
 | `cardId`      | `EntityId` | The land card to play.                                                                          |
 | `asBackFace`  | `Boolean`  | Play a modal double-faced card as its **back** face (CR 712.12). Defaults to `false`. See below. |
 
-A modal double-faced land (the Zendikar Rising Pathway cycle) is one card with two land faces, and
-CR 712.12 makes the face a choice taken *before* the permanent enters. The server therefore
-enumerates **one `PlayLand` per land face**, each `description` naming the face it plays, and
-`asBackFace` is what distinguishes the two otherwise-identical actions. Clients must not collapse
-them: filtering `PlayLand` actions down to the first one silently drops a face.
+A modal double-faced land (the Pathway cycle, from Zendikar Rising and Kaldheim) is one card with
+two land faces, and CR 712.12 makes the face a choice taken *before* the permanent enters. The
+server therefore enumerates **one `PlayLand` per land face**, each `description` naming the face it
+plays, and `asBackFace` is what distinguishes the two otherwise-identical actions. Clients must not
+collapse them: filtering `PlayLand` actions down to the first one silently drops a face.
 
 #### D. `PassPriority`
 
--- a/docs/engine-server-interface.md
+++ b/docs/engine-server-interface.md
@@ -161,9 +161,16 @@ Used to use an activated ability of a permanent (or card in hand/graveyard).
 
 Used to perform the special action of playing a land.
 
-| Field    | Type       | Description            |
-|----------|------------|------------------------|
-| `cardId` | `EntityId` | The land card to play. |
+| Field         | Type       | Description                                                                                     |
+|---------------|------------|-------------------------------------------------------------------------------------------------|
+| `cardId`      | `EntityId` | The land card to play.                                                                          |
+| `asBackFace`  | `Boolean`  | Play a modal double-faced card as its **back** face (CR 712.12). Defaults to `false`. See below. |
+
+A modal double-faced land (the Zendikar Rising Pathway cycle) is one card with two land faces, and
+CR 712.12 makes the face a choice taken *before* the permanent enters. The server therefore
+enumerates **one `PlayLand` per land face**, each `description` naming the face it plays, and
+`asBackFace` is what distinguishes the two otherwise-identical actions. Clients must not collapse
+them: filtering `PlayLand` actions down to the first one silently drops a face.
 
 #### D. `PassPriority`
 
--- a/manual-scenarios/cards/r/riverglide-pathway.json
+++ b/manual-scenarios/cards/r/riverglide-pathway.json
@@ -1,0 +1,31 @@
+{
+  "player1Name": "You",
+  "player2Name": "AI",
+  "mode": "AI",
+  "aiPlayer": 2,
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Riverglide Pathway", "Shock", "Sleight of Hand"],
+    "battlefield": [],
+    "graveyard": [],
+    "library": ["Grizzly Bears", "Hill Giant", "Mountain", "Craw Wurm"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": ["Grizzly Bears", "Mountain"],
+    "battlefield": [
+      {"name": "Grizzly Bears"},
+      {"name": "Mountain"},
+      {"name": "Mountain"}
+    ],
+    "graveyard": [],
+    "library": ["Hill Giant", "Mountain", "Grizzly Bears"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": []
+}
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/model/CardDefinition.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/model/CardDefinition.kt
@@ -795,6 +795,43 @@ data class CardDefinition(
         }
 
         /**
+         * Creates a **modal double-faced land** — a modal DFC (CR 712.3) both of whose faces are
+         * lands, the Zendikar Rising Pathway cycle (Riverglide Pathway // Lavaglide Pathway and its
+         * nine siblings).
+         *
+         * The rule that makes this its own shape is CR 712.12: *"A player playing a modal
+         * double-faced card or a copy of a modal double-faced card as a land chooses one of its
+         * faces that's a land before putting it onto the battlefield. It enters the battlefield
+         * with that face up."* So neither face is ever cast — the card is **played**, and the
+         * choice is made on the way in — which is exactly why it can't use
+         * [modalDoubleFacedPermanent]: that factory requires the back face to carry its own mana
+         * cost because it is castable (CR 712.11b), and a land has none.
+         *
+         * Once on the battlefield the permanent has only the played face's characteristics
+         * (CR 712.8f) and can never turn over — CR 712.9 excludes modal DFCs from transforming.
+         *
+         * @param frontFace The front face (must be a land with no mana cost).
+         * @param backFace The back face (must be a land with no mana cost).
+         */
+        fun modalDoubleFacedLand(
+            frontFace: CardDefinition,
+            backFace: CardDefinition
+        ): CardDefinition {
+            for (face in listOf(frontFace, backFace)) {
+                require(face.typeLine.isLand) {
+                    "Modal double-faced land face '${face.name}' must be a land (CR 712.12)"
+                }
+                require(face.manaCost.isEmpty()) {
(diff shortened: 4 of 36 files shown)


bolav:
## Review — `mdfc-land-faces`

Reviewed `origin/main...mdfc-land-faces`. Local `main` is level with `origin/main`, so no
merge was needed. This is the largest of the four branches and the only one crossing
SDK → engine → client.

**Re-reviewed after `16c2126`**, which adds the other nine Pathways and completes the
cycle. That commit is pure content — nine `cardDef`s, nine scenario tests, two snapshot
goldens and two doc lines — so nothing in the engine assessment below changes. The
issues it does move are M3 (now resolved) and the verification section.

### Verdict

The seam is chosen well: the face choice is an *action shape*, not a decision prompt,
which is what CR 712.12 describes and what keeps the engine from growing a
pause/resume for something the player already answered by clicking. The
`flipDfcInPlace` extraction is the right refactor and `:rules-engine:test` staying at 0
failures is the evidence. One perf regression was introduced and fixed on the branch
(I1); the rest is scope notes.

The nine-card follow-up is the best possible evidence for the SDK facade: every one of
them is `modalDoubleFacedLand(front, back)` over two one-ability land faces, with no new
effect, executor or DSL surface. A cycle that lands as pure data is what "composition
before new types" is supposed to look like.

### What's good

- **Choosing a face is deliberately *not* a transform, and the code says why.**
  CR 712.9 excludes modal DFCs from transforming, so the play path must fire no
  transform trigger and must not be stoppable by a "can't transform" effect. Splitting
  `setDfcFace` out of `flipDfcInPlace` — state work in the shared half, the CR 701.27b
  check and `TransformedEvent` in the wrapper — is exactly the right cut, and it means
  the play path inherits the `CardComponent` swap, the CR 712.8a front-face save, the
  static re-registration and the self-redirect rebuild instead of growing copies.
  `TransformEffectExecutor.kt:127-196`.
- **The face swap is placed before everything that reads characteristics.**
  `PlayLandHandler.kt:~215-232` swaps ahead of `BattlefieldEntry.place`, the
  static/replacement registration, the enters-tapped branches and the entry event's
  name, then rebinds `cardComponent` from the post-swap state at `:~340`. Doing it after
  entry would have been the natural mistake and would have mis-typed the ETB record for
  a spell-front // land-back card. The existing code even had a comment predicting this
  seam; the diff fills it in rather than inventing a new one.
- **`castFace` was hardened at the same time.** `ModalDfcCasts.kt:33-38` now excludes a
  land back face, which closes a latent bug: a land is a permanent, so the old
  `isPermanent` filter would have offered "Cast Emeria, Shattered Skyclave" for a future
  spell-front // land-back card. Fixing the mirror while adding the new half is the
  right instinct.
- **The client fix is the interesting half and it's tested.** Two `PlayLand` actions for
  one card broke a `find` and a boolean that had never had to distinguish faces. Both
  were mutation-checked, and the option list is labelled from the *server's* description
  because `cardInfo.name` is the front face's name for both entries — the detail that
  makes the menu legible instead of showing "Play Riverglide Pathway" twice.
- **The scenario test proves the round trip, not just the play.** "off the battlefield it
  is the front face again" is the case that actually exercises the `frontFaceCard` save
  surviving `ZoneTransitionService` — much stronger than asserting the permanent's name.

### Issues

**I1 — (fixed in `3c48304`) the enumerator put a card-registry lookup in a hot path.**
Removing the outer `if (cardComponent.typeLine.isLand)` guard meant `landPlays` ran for
*every* card in hand, and its first act was `context.cardRegistry.getCard(...)` to ask
about a back face. Legal-action enumeration runs on every priority pass and every
AI/MCTS node, so that's a lookup per hand card per node — exactly the
"cheap-looking work in a primitive multiplies" case `add-feature` Step 5 warns about.
Fixed with an early-out on two booleans already present on `CardComponent`
(`PlayLandEnumerator.kt:~121`): a card that is neither a land nor double-faced can offer
no land play, and the registry is only consulted for the handful that could.

**I2 — no E2E coverage for the two-entry drag-to-play menu.** The client change is
verified at the option-list level (`actionOptions.test.ts`, `shared.test.ts`) but nothing
drives the real menu. "Two entries appear and each one plays the face it names" is the
user-visible contract of this whole branch, and it is currently asserted only against a
pure function. `add-feature` Step 7 asks for an E2E test when a change introduces a new
visual/UX mechanic, and a second entry in the play menu qualifies. The PR body discloses
this rather than claiming coverage it doesn't have, which is the right call — but I'd
want `e2e-scenarios/tests/znr/riverglide-pathway.spec.ts` before this is called done.

**M1 — the Mayhem graveyard loop still offers only the front face.**
`PlayLandEnumerator.kt:~57-77` was left alone while hand / graveyard-permission / linked-exile
gained per-face offers. It is gated on the card carrying Mayhem, which no Pathway does,
so it is unreachable today and extending it would have been untestable speculation. The
PR body says so. Fine — but it is now the one land-play path with different face
semantics from the others, which is the kind of asymmetry that bites later. A `// no
per-face offer here: see PR` comment at the loop would be cheap insurance.

**M2 — the AI rollout policy always takes the front face.**
`ai/.../PlayoutPolicy.kt:103` does `candidates.firstOrNull { it.actionType == "PlayLand" }`.
Real MCTS search sees both actions, so this only biases rollouts, and reaching into `:ai`
from a rules PR would be scope creep. Correctly disclosed. Worth a follow-up issue rather
than a silent limitation, since "the AI always plays the blue Pathway" will read as a bug
to whoever notices it first.

**M3 — (resolved in `16c2126`) only one of the ten Pathways was added.** The other nine
are now in, and the cycle is complete. Two things worth noting about how:

- The commit also corrects a factual error the earlier commits and two doc lines carried:
  this is not a Zendikar Rising cycle. ZNR printed six Pathways and Kaldheim printed the
  other four; together they cover all ten colour pairs exactly once. Cheap to have got
  wrong, and it would have quietly mis-scoped any later "finish the ZNR cycle" work.
- The nine new scenario tests are each their own file — the repo's one-card-one-test-file
  rule — but each is only three tests, not a copy of Riverglide's six. That is the right
  split: the CR clauses are properties of the *mechanic* and are pinned once, so
  duplicating them nine times would add maintenance without adding coverage, while the
  face names and mana colours genuinely are per-card and are the thing a typo would break.
  Worth checking on any future MDFC cycle that this restraint doesn't drift into skipping
  the per-card assertions altogether.

From the engine/client half of the branch:

- `:rules-engine:test` — 0 failures, re-run *after* the I1 fix. This is the gate that
  matters for the `flipDfcInPlace` extraction: the whole transform / daybound / disturb /
  DFC suite passes against the refactored helper.
- `:oracle-assay`, `:game-server`, `:ai` all compile against the changed `PlayLand` shape
  — worth having checked, since the action is serialized across the API boundary.
- `npx tsc --noEmit` clean; `vitest` 45 tests across the two touched suites; both client
  mutations reproduced their expected reds.
- CR 712.8a, 712.8f, 712.9, 712.11c and 712.12 all checked against
  `MagicCompRules_20260808.txt`; every citation in the code, commits and PR body matches
  the rule it names.

Re-run for the nine-card commit (`16c2126`):

- `just test-class "*PathwayScenarioTest"` — 33/33 across all ten cards, BUILD SUCCESSFUL.
- `:mtg-sets:test` — BUILD SUCCESSFUL with all ten in the corpus. Snapshot diff is pure
  insertion in `ZNR.json` and `KHM.json`; no existing card's golden moved.
- `scripts/check-card-printing.py` on each of the nine — every canonical sits in its
  set's earliest real printing, and the PZNR/PKHM/PRM/SLU/PLST reprints belong to
  unimplemented sets, so no `Printing` rows are owed. This is the half of card work that
  usually gets forgotten, and it is clean here.

**Not covered — read this before calling the branch green:**

- **No E2E run** (see I2).
- **`just test-scenarios 2017-2022` did not complete.** It ran 762 tests with zero test
  failures, then the Gradle build died writing HTML report entries into the shared
  `/workspace/.gradle/9.6.1/fileHashes/fileHashes.bin` (`EOFException`). That is
  sandbox cache corruption, not a defect in the diff — and per this workspace's rules it
  gets reported, not `clean`ed. So the era suite is **inconclusive as a gate**: 762
  passing tests is real evidence, a completed era run is not something this branch has.
  Worth re-running once the shared cache is healthy.
- **No gym/RL gate** — `PlayLand` gained a defaulted field, so no action-space encoding
  changed, but `:gym:test` was not run to prove it.
- **No Assay differential for KHM.** `just assay-differential --set ZNR` was 9/9 with 0
  divergences, but that ran when the branch held a single Pathway; it has not been re-run
  since, and KHM was never covered.

### Recommendation

Add the E2E spec from I2, then merge. M1's one-line comment is cheap and worth taking at
the same time; M2 should become a follow-up issue against `:ai` rather than growing this
PR — and note that M2 stings a little more now that the cycle is complete: "the AI always
plays the front face" is ten cards' worth of rollout bias rather than one.

Re-run `just test-scenarios 2017-2022` before merge if the shared Gradle cache has been
repaired, so the era gate is an actual green rather than 762 tests and an infrastructure
crash.

PR #2011, opened 2026-08-24
Add three red cards: Reckless Rage, Redcap Melee, Experimental Synthesizer

bolav:
# Add three red cards: Reckless Rage, Redcap Melee, Experimental Synthesizer

Three cards from the Izzet Prowess shell, each built entirely from existing
`Effects.*` / `Patterns.*` — no new SDK vocabulary, so they share one branch and one
PR per `CONTRIBUTING.md`. Each is canonical in its own earliest real printing
(`just check-card-printing` clean on all three).

## The cards

**Reckless Rage** — RIX #110, `{R}` Instant.
"Reckless Rage deals 4 damage to target creature you don't control and 2 damage to
target creature you control." Two required targets with different amounts, composed
as `DealDamage(4, theirs) then DealDamage(2, yours)`. The card's real drawback — it
is uncastable unless you have a creature to hurt — falls out of the targeting rules
for free: a required `target(...)` with no legal object makes the spell illegal to
cast (CR 601.2c), so there is nothing to model. "Creature you don't control" is
`Targets.CreatureOpponentControls`; every permanent is controlled by some player and
this engine has no teams, so the two readings pick out the same creatures.

**Redcap Melee** — ELD #135, `{R}` Instant.
"Redcap Melee deals 4 damage to target creature or planeswalker. If a nonred
permanent is dealt damage this way, you sacrifice a land." The colour test is a
`Conditions.TargetMatchesFilter` over the same chosen target, so it reads **projected**
state: a creature painted red by a continuous effect correctly spares the land, and one
whose red was stripped correctly doesn't. The sacrifice is `Effects.SacrificeOwn` — the
bare imperative "you sacrifice a land", not a named player. Ordering matters and is
load-bearing: both sub-effects sit in one `Effects.Composite`, which resolves
sequentially with no interleaved state-based-action pass, so a lethally damaged
creature has not been put into the graveyard yet and its projected colour is still
readable when the condition runs.

**Experimental Synthesizer** — NEO #138, `{R}` Artifact.
"When this artifact enters or leaves the battlefield, exile the top card of your
library. Until end of turn, you may play that card." plus
"`{2}{R}`, Sacrifice this artifact: Create a 2/2 white Samurai creature token with
vigilance. Activate only as a sorcery." Both trigger halves are
`Patterns.Exile.impulse(1)`. "Enters **or** leaves" is written as two triggered
abilities: the two events can never happen at once, so one ability with a two-event
pattern and two abilities are the same card in play, and two abilities is the shape
the corpus already uses for this wording (Cryogen Relic). The card's own activated
ability sacrifices it, so activating it collects the leave trigger too — which is the
whole design.

## Deliberately not modelled

Redcap Melee's printed trigger is "*is dealt* damage this way", so damage that is
prevented or replaced down to nothing should leave your lands alone. This models the
clause as "the target is a nonred permanent", which gives the same answer in every
case except an active damage-prevention shield — there this card sacrifices a land
where the printed card would not. Fixing it properly needs a "was this target dealt
damage during this resolution" condition the SDK does not have; adding one would have
ejected the card from the batch. The deviation is written into the card's KDoc rather
than left silent.

## Verification

- `./scripts/gradle-locked :mtg-sets:test` — green after re-blessing. The only
  failures on the first run were the three expected `CardDefinitionSnapshotTest`
  mismatches (RIX/ELD/NEO); `just rebless-cards` moved **only** the three new cards in
  the goldens (pure additions, no existing card's tree changed), which is the check
  that no shared SDK behaviour shifted.
- `./scripts/gradle-locked :mtg-sets:2017-2022:compileKotlin` — green.
- `just check-card-printing` — clean for all three; each canonical is in its earliest
  real printing.
- `just assay-differential --set RIX / ELD / NEO` — RIX 0 divergences, ELD's single
  divergence is the pre-existing Keeper of Fables row (untouched by this branch), and
  NEO's row is **Experimental Synthesizer matching Assay's independent reading
  exactly**, differing only in ability ids, the token's name/art, and the ability
  description — all known folds. That is the one check here that isn't me re-reading
  my own work.

- `just test-class ExperimentalSynthesizerScenarioTest` — 2/2. Reckless Rage and Redcap
  Melee add no engine behaviour and are covered by the snapshot and lint nets, per
  `add-card` step 5. Experimental Synthesizer got a test anyway, for one reason: its
  most common line sacrifices it to its own ability, paying the sacrifice as a **cost**,
  so the "or leaves the battlefield" trigger has to fire off a battlefield exit during
  *activation* rather than during resolution — a shape this engine has got wrong before.
  It behaves correctly; the test is there so it stays that way.

**What this does not cover:** nothing here touches the web client, the server, or the
AI, so no TypeScript or server gate was run and none would say anything about this
diff.

## Try it

```bash
  -H 'Content-Type: application/json' -d @manual-scenarios/mechanics/izzet-prowess-red-cards.json
```

All three cards in hand with six Mountains — enough for every cast plus the Synthesizer's
`{2}{R}` sacrifice activation. The opponent's board makes each branch reachable in one turn:
Craw Wurm is a nonred 6/4 that *survives* Redcap Melee (so the land sacrifice is visible with
the creature still standing), and Raging Regisaur is red (so aiming Redcap Melee there
sacrifices nothing).

### DIFF
--- a/manual-scenarios/mechanics/izzet-prowess-red-cards.json
+++ b/manual-scenarios/mechanics/izzet-prowess-red-cards.json
@@ -1,0 +1,40 @@
+{
+  "player1Name": "You",
+  "player2Name": "AI",
+  "mode": "AI",
+  "aiPlayer": 2,
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Reckless Rage", "Redcap Melee", "Experimental Synthesizer"],
+    "battlefield": [
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Centaur Courser"}
+    ],
+    "graveyard": [],
+    "library": ["Lightning Bolt", "Hill Giant", "Mountain", "Grizzly Bears", "Mountain"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": ["Grizzly Bears", "Mountain"],
+    "battlefield": [
+      {"name": "Craw Wurm"},
+      {"name": "Raging Regisaur"},
+      {"name": "Mountain"},
+      {"name": "Mountain"}
+    ],
+    "graveyard": [],
+    "library": ["Hill Giant", "Mountain", "Grizzly Bears"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": []
+}
--- a/mtg-sets/2017-2022/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/eld/cards/RedcapMelee.kt
+++ b/mtg-sets/2017-2022/src/main/kotlin/com/wingedsheep/mtg/sets/definitions/eld/cards/RedcapMelee.kt
@@ -1,0 +1,63 @@
+package com.wingedsheep.mtg.sets.definitions.eld.cards
+
+import com.wingedsheep.sdk.core.Color
+import com.wingedsheep.sdk.dsl.Conditions
+import com.wingedsheep.sdk.dsl.Effects
+import com.wingedsheep.sdk.dsl.Filters
+import com.wingedsheep.sdk.dsl.Targets
+import com.wingedsheep.sdk.dsl.card
+import com.wingedsheep.sdk.model.Rarity
+import com.wingedsheep.sdk.scripting.GameObjectFilter
+import com.wingedsheep.sdk.scripting.effects.ConditionalEffect
+
+/**
+ * Redcap Melee — Throne of Eldraine #135 (canonical printing)
+ * {R} · Instant
+ *
+ * Redcap Melee deals 4 damage to target creature or planeswalker. If a nonred permanent is dealt
+ * damage this way, you sacrifice a land.
+ *
+ * Four damage for one mana, with a land sacrifice unless the thing you hit is red — which is why
+ * the drawback clause reads on the *permanent's colour*, not on the spell's. The colour test is a
+ * [Conditions.TargetMatchesFilter] over the same chosen target, so it reads **projected** state:
+ * a creature painted red by a continuous effect (Painter's Servant, Whim of Volrath) correctly
+ * spares the land, and one whose red was stripped correctly doesn't.
+ *
+ * The sacrifice is [Effects.SacrificeOwn] — the bare imperative "you sacrifice a land", where the
+ * spell's controller chooses, not a named player.
+ *
+ * Ordering matters: the two sub-effects run in one [Effects.Composite], so the colour is read
+ * after the damage, while the damaged permanent is still on the battlefield (a composite resolves
+ * sequentially with no interleaved state-based-action pass, so a lethally damaged creature has not
+ * been put into the graveyard yet and its projected colour is still readable).
+ *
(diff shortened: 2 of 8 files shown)


bolav:
## Review — `izzet-prowess-red-cards`

Reviewed `origin/main...izzet-prowess-red-cards`. Local `main` is level with
`origin/main` (`git rev-list --count main..origin/main` = 0), so no merge was needed
and the diff below is post-merge reality.

### Verdict

Ship it. Three cards, no new SDK types, every one composed from primitives that
already had a second caller. The one behavioural deviation is real and is documented
in the card rather than hidden.

### What's good

- **Reckless Rage leans on the targeting rules instead of re-implementing them.** The
  card's drawback ("you need a creature of your own to hurt") is enforced by two
  required `target(...)` declarations and CR 601.2c, with nothing card-specific. That's
  the right instinct.
- **Redcap Melee's colour test reads projected state.** `Conditions.TargetMatchesFilter`
  lowers to `EntityMatches(ContextTarget(0), …)`, which the evaluator matches against
  projection — so a creature painted red by a continuous effect spares the land. This is
  the engine's most-repeated bug class and the diff gets it right by construction.
- **The composite-ordering reasoning is written down.** `RedcapMelee.kt:29-32` explains
  *why* the colour can still be read after lethal damage (no interleaved SBA pass inside
  a `CompositeEffect`). That's the non-obvious fact a future editor would otherwise have
  to rediscover.
- **The Synthesizer's riskiest interaction is now pinned by a test.** See I1.

### Issues

**I1 — (fixed in `c38056e`) the sacrifice-as-cost leave trigger was untested.**
`ExperimentalSynthesizer.kt:52-56` sacrifices the artifact as an *activation cost*,
which is paid before the ability reaches the stack — so the "or leaves the battlefield"
trigger has to fire off an exit during activation, not during resolution. That is the
card's most common line and a shape this engine has got wrong before. It was shipped on
the strength of "the primitives exist", with no test. Verified and correct;
`ExperimentalSynthesizerScenarioTest` now pins both the enter half and the
sacrifice-as-cost leave half. No behaviour change was needed.

**I2 — Redcap Melee ignores damage redirection, not just prevention.** The PR body
discloses the prevention case ("if a nonred permanent *is dealt* damage this way"), but
there is a second instance of the same gap: if the damage is redirected to a different
permanent (`RedirectDamage`, Harsh Judgment / Martyrs of Korlis), the printed card asks
about the colour of the permanent that *was dealt* the damage, while
`RedcapMelee.kt:50-53` asks about the colour of the original target. Same root cause —
the SDK has no "was this target dealt damage during this resolution" condition — and
same right answer for now, but the KDoc's deviation note only names prevention. Worth
one more sentence there so the next reader doesn't think redirection was considered and
ruled out.

**M1 — "creature you don't control" vs "creature an opponent controls".**
`RecklessRage.kt:36` uses `Targets.CreatureOpponentControls` and the KDoc explains the
two readings coincide in this engine. Correct today. If teams ever land (CR 805 shared
turns), this and every other card using that target become wrong together, so it's a
one-place-to-fix problem rather than a card bug — no action, just noting the coupling.

**M2 — the Samurai token has no `PredefinedTokens` entry.** `ExperimentalSynthesizer.kt:58-66`
inlines the 2/2 white Samurai with vigilance and its NEO art. Fine for one card, but NEO
prints several Samurai-token makers; if a second lands, promote it rather than copying
the literal.

### Verification I re-ran

- `just test-class ExperimentalSynthesizerScenarioTest` — 2/2 (the new file).
- The author's `:mtg-sets:test` + rebless is confirmed by the golden diff: pure
  additions, only the three new cards moved.
- `just check-card-printing` on all three — clean.

Not covered by anything here: no client, server or AI code is touched, so no TypeScript
or `:game-server` gate was run and neither would be evidence about this diff.

### Recommendation

Merge. Optionally add the one-sentence redirection note from I2 to `RedcapMelee.kt`'s
deviation paragraph first — it costs nothing and it's the kind of gap that reads as an
oversight later.

PR #2012, opened 2026-08-24
Add sneaky snacker

AndreyItMe:


wingedsheep:
Closing — Sneaky Snacker already ships on `main`, merged as #1950. This branch predates that merge, so what it renders as is a *modification* of the shipped card rather than an addition, and the modification regresses it on two counts:

- it swaps the `Effects.PutOntoBattlefieldFromGraveyard(Self, tapped = true)` facade for a raw `MoveToZoneEffect(...)`, against the facades-not-raw-constructors rule;
- it widens `triggerZones` from `{GRAVEYARD}` to `{BATTLEFIELD, GRAVEYARD}`. The ability functions only in the graveyard (CR 113.6m) — a Snacker already on the battlefield must not see your third draw.

The stale base is also why CI went red: the re-bless commits rewrote snapshot goldens for MH2, SHM, MRD and MKM against a corpus that had moved on.

See #2017 for the Cryptic Serpent half of this stack, redone on a current base.

PR #2013, opened 2026-08-24
Add cryptic serpent

AndreyItMe:


wingedsheep:
Superseded by #2017, which carries the Cryptic Serpent definition from this branch on a base cut from current `main`.

This branch was stacked on #2012, so it inherited that PR's stale base. The re-bless commits deleted 11 scenario test files (the Pathway cycle, Experimental Synthesizer, Riverglide) and rewrote 10 snapshot goldens — about 1,000 lines of already-merged work — which is what turned the `content` and `backend` CI jobs red. The card definition itself was correct and is carried over unchanged apart from KDoc; #2017 adds the JMP and J22 `Printing` rows and a scenario test alongside it.

PR #2023, opened 2026-08-24
Add the Learn keyword action, and Academic Dispute

bolav:
# Add the Learn keyword action, and Academic Dispute

Learn (CR 701.48) had no SDK spelling. Both halves of the mechanic already existed as
primitives — that's why this is a `Patterns.Mechanic` recipe rather than a new effect
type — but with nothing composing them, Strixhaven's Lesson-fetchers were
unimplementable.

## Learn is sequential, not a choose-one

CR 701.48a, verified against `MagicCompRules_20260808.txt`:

> "Learn" means "You may discard a card. If you do, draw a card. If you didn't discard
> a card, you may reveal a Lesson card you own from outside the game and put it into
> your hand."

The order is load-bearing and easy to get wrong, because the printed *reminder* text
lists the two options the other way round ("You may reveal a Lesson card ... or
discard a card to draw a card"). The rules text governs: discarding **forecloses** the
Lesson. A player who discards never gets the sideboard option, and modelling it as a
two-mode choice would let them take the Lesson and then still discard.

`Patterns.Mechanic.learn()` follows 701.48a literally:

```
GatherCards(hand) -> SelectFromCollection(ChooseUpTo 1)
-> MoveCollection(-> graveyard, MoveType.Discard)
-> ConditionalOnCollection(ifNotEmpty = DrawCards(1),
                           ifEmpty    = Patterns.Sideboard.wish(Lesson))
```

Both printed "may"s are `ChooseUpTo(1)` declines rather than yes/no prompts. That is a
UX decision as much as a modelling one: an empty hand or a Lesson-less sideboard
auto-resolves to none and raises **no decision at all**, so the player is never asked
a question that cannot matter. The discard is a real discard (`MoveType.Discard`), so
madness and "whenever you discard a card" payoffs see it. Collection names are
`learn_`-prefixed so a Learn nested inside another pipeline can't collide with that
pipeline's own `hand` / `discarded`.

`Patterns.Sideboard.wish` already models "outside the game" as the private
`Zone.SIDEBOARD` (CR 100.4 / 400.11a) with reveal-on and no shuffle, so the Lesson
half needed only a subtype-restricted filter.

## Academic Dispute

STX #91, `{R}` Instant — "Target creature blocks this turn if able. You may have it
gain reach until end of turn. Learn."

`Effects.MarkMustBlockThisTurn` + an optional `Effects.GrantKeyword(REACH)` on the
same target + `learn()`. The reach rider is a real `MayEffect` and lands on the
targeted creature, because its whole point is letting the forced blocker block a
flier. "Blocks this turn if able" is a *requirement*, not a guarantee (CR 509.1c): a
tapped creature, or one whose every possible block would be illegal, simply doesn't
block. The three clauses resolve in printed order, so the Learn happens whether or not
the creature ends up blocking.

## Client

None needed, and that is the point rather than an omission. Both Learn prompts are
ordinary `SelectCards` decisions — the sideboard one is the same wish flow Burning
Wish has used since Judgment — and the reach rider is a yes/no. Nothing new reaches
`web-client/src/components/decisions/`.

## Verification

- `just test-class AcademicDisputeScenarioTest` — 6/6. Covers **both branches of
  701.48a** (discard-then-draw with the Lesson never offered; decline-then-fetch with
  only Lesson cards offered), declining both halves, an empty hand skipping straight to
  the Lesson, the block requirement being a requirement rather than a guarantee, and
  the reach rider.
- **Mutation-checked.** Widening the Lesson filter to `GameObjectFilter.Any` reddens
  exactly the "only Lesson cards are offered" test; restored and verified identical.
- `./scripts/gradle-locked :mtg-sets:test` — green after re-blessing; the only failure
  was the expected STX snapshot, and only Academic Dispute moved in the golden, which
  is the check that the new `MechanicPatterns` entry changed no existing card.
- `just check-card-printing "Academic Dispute"` — clean; STX is the only printing.
- `just assay-differential --set STX` — 6/6 confirmed, 0 divergences.
- `:oracle-assay:compileKotlin` — green.

**What this does not cover:** no TypeScript gate was run, because the diff contains no
`.ts`; a green Kotlin suite would say nothing about the client either way. No E2E run
— the decisions Learn raises are existing components on an existing route.

## Note for the reviewer

The Lesson fixture in the scenario test is Boomerang Basics (TLA), a real
`Sorcery — Lesson` already in the corpus, so no card was added just to have something
to fetch. Strixhaven's own Lessons remain unimplemented; Learn is reachable today
through the Avatar Lessons.

## Try it

```bash
  -H 'Content-Type: application/json' -d @manual-scenarios/cards/a/academic-dispute.json
```

Built around the card's actual combo: attack with the Air Elemental, then Dispute the
opponent's ground Craw Wurm — forcing it to block a flier only does anything if you also take
the optional reach rider, so both combat clauses are exercised in one attack. **Learn is only
half-testable there**: the scenario API's `PlayerConfig` has no sideboard field, so the
"reveal a Lesson from outside the game" branch can't be set up by hand and is covered by the
scenario test instead.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -3179,6 +3179,18 @@ one-off pipeline belongs inline in the card file via `Effects.Pipeline { }` (§5
 - `rummage(count?)` — discard then draw.
 - `connive(target?)` — draw 1, discard 1, then put a +1/+1 counter on `target` (default Self) if the discard was a nonland (CR 701.50). Also exposed as `Effects.Connive(target)`. Returns the pipeline wrapped in `ConniveEffect(subject = target, body = …)`: the wrapper names the keyword action and its subject, which is what lets `ModifyKeywordAction` replace it and what makes it emit `PermanentConnivedEvent` (CR 701.50f). The pipeline itself is unchanged.
 - `conniveTargeting(requirement, storeAs?)` — connive whose +1/+1 counter lands on a *reflexively chosen* target: "draw a card, then discard a card. When you discard a nonland card this way, put a +1/+1 counter on target creature you control" (Teo, Spirited Glider). The recipient is selected at resolution via `SelectTargetEffect` *inside* the nonland gate — so the player never chooses up front or when the discard is a land. Pass the recipient's `TargetRequirement` (e.g. `Targets.CreatureYouControl`); do **not** also declare it as a cast-time `target(...)`. Exposed as `Effects.ConniveTargeting(requirement)`. Deliberately **not** wrapped in `ConniveEffect` — Teo's printed text spells the looting out and never says "connive", so it is not the keyword action: it fires no connive triggers and is not touched by "if a creature you control would connive" replacements.
+- `Patterns.Mechanic.learn()` — **Learn** (CR 701.48, Strixhaven), the keyword action printed as a
+  bare "Learn." with reminder text. CR 701.48a is sequential, not a choose-one: *"You may discard a
+  card. If you do, draw a card. If you didn't discard a card, you may reveal a Lesson card you own
+  from outside the game and put it into your hand."* — discarding **forecloses** the Lesson.
+  Composed as `GatherCards(FromZone(HAND, You))` → `SelectFromCollection(ChooseUpTo 1)` →
+  `MoveCollection(→ graveyard, MoveType.Discard)` →
+  `ConditionalOnCollection(ifNotEmpty = DrawCards(1), ifEmpty = Patterns.Sideboard.wish(Lesson))`.
+  Both printed "may"s are `ChooseUpTo(1)` declines rather than yes/no prompts, so an empty hand or a
+  Lesson-less sideboard raises no decision at all. The discard is a real discard
+  (`MoveType.Discard`), so madness and "whenever you discard a card" payoffs see it. Collection
+  names are `learn_`-prefixed so a nested Learn can't collide with its host pipeline's `hand` /
+  `discarded`. Academic Dispute.
 - `Patterns.Mechanic.recruit()` — **Recruit** (The Hobbit): "draw a card, then discard a card. If you
   discarded a nonland card, create a 1/1 white Human Soldier creature token." A keyword *action* with
   fixed reminder text, not a keyword ability, so there is no `Keyword.RECRUIT` — it is connive's pipeline
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -1887,6 +1887,12 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
 - `CreateEldraziSpawn(count?, controller?, imageUri?)` — 0/1 colorless Eldrazi Spawn creature tokens
   ("Sacrifice this creature: Add {C}."). `count` accepts an `Int` (default 1) or a `DynamicAmount`
   evaluated at resolution, such as `DynamicAmount.XValue` for Kozilek's Command.
+- `CreatePest(count?, controller?)` — 1/1 **black and green** Pest creature tokens with "When this
+  creature dies, you gain 1 life." (`PredefinedTokens.Pest`) — Strixhaven's Witherbloom token
+  (Hunt for Specimens, Pest Summoning, Sedgemoor Witch, …). Predefined rather than inline because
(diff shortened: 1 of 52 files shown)


bolav:
## Review — `learn-keyword-action`

Reviewed `origin/main...learn-keyword-action`. Local `main` is level with `origin/main`,
so no merge was needed.

### Verdict

Correct and correctly placed. Learn is a keyword *action* composed entirely from existing
primitives, so a `Patterns.Mechanic` recipe is exactly the right home — no new `Effect`,
no new executor, no new condition. The sequencing that most implementations get wrong is
the thing this one is built around.

### What's good

- **It implements CR 701.48a, not the reminder text.** The printed reminder lists the
  Lesson option first; the rules text is discard-first, and discarding *forecloses* the
  Lesson. `MechanicPatterns.kt:398-424` follows the rule, and the KDoc quotes 701.48a and
  says explicitly that the order is load-bearing. Modelling this as a two-mode choice —
  the obvious shortcut, and what the reminder text invites — would have let a player take
  the Lesson and then still loot. Getting this right is the whole review.
- **`ChooseUpTo(1)` for both "may"s is a UX decision as much as a modelling one.** An
  empty hand or a Lesson-less sideboard auto-resolves and raises *no* prompt, so the
  player is never asked a question that can't matter. A `MayEffect` yes/no would have
  asked twice, every time. It also sidesteps a real trap: `SuccessCriterion.Auto` can't
  see through a `GatedEffect`, so `IfYouDoEffect(action = MayEffect(discard), …)` would
  have been rejected at card load — the chosen shape avoids that rather than working
  around it.
- **Collection names are namespaced.** `learn_hand` / `learn_discarded` / `learn_lessons`
  instead of the `HandPatterns` defaults (`hand`, `discarded`), so a Learn nested inside
  another pipeline can't clobber its host's collections. That's the failure mode that
  would have been near-impossible to debug from a bug report.
- **The test covers both branches of the rule and the decline path.** Six cases, including
  "discarding means the Lesson is never offered" — which is the assertion that actually
  pins the sequencing — and the empty-hand case that proves the auto-resolve. The Lesson
  filter was mutation-checked.
- **No client work, correctly.** Both prompts are existing `SelectCards` decisions on the
  existing wish route; the PR says so rather than leaving it ambiguous.

### Issues

**M1 — the discard prompt is offered before the player can see whether a Lesson is
available.** Sequential is what CR 701.48a requires, so this is not a bug — but a player
who declines the discard hoping for a Lesson, and then finds their sideboard has none,
has spent a decision for nothing. Nothing to do in the engine; if the client ever wants
to help, the prompt text (`"Learn — you may discard a card to draw a card"`,
`MechanicPatterns.kt:415`) is where a "(you have N Lessons available)" hint would go.
Noting it so it isn't later mistaken for a modelling error.

**M2 — Academic Dispute's reach rider and the Learn share a resolution with no ordering
note.** `AcademicDispute.kt:41-46` chains `MarkMustBlockThisTurn then MayEffect(reach)
then learn()`. Printed order, which is right. The KDoc explains the reach rider's
*purpose* (letting the forced blocker block a flier) but not that the Learn happens
whether or not the creature ends up blocking — which the code guarantees and a reader
might wonder about. One clause would settle it.

**M3 — `learn()` currently has exactly one caller.** `sdk-design-principles.md` warns
against single-use `Patterns.*` recipes. I don't think it applies: Learn is a named
keyword action with ~20 printed users in STX, and the recipe is what makes the rest of
them a pure `add-card` batch. Recording the judgement rather than the rule.

- `just test-class AcademicDisputeScenarioTest` — 6/6, and the filter mutation reproduced
  the expected single red.
- `:mtg-sets:test` + rebless — only Academic Dispute moved in the golden, which is the
  check that the new `MechanicPatterns` entry didn't disturb an existing card.
- `just check-card-printing "Academic Dispute"` — clean.
- `just assay-differential --set STX` — 6/6 confirmed, 0 divergences.
- CR 701.48a verified against `MagicCompRules_20260808.txt` — the quoted text in the
  KDoc and PR body matches the rule verbatim. CR 509.1c likewise for the block
  requirement.

Not covered: no TypeScript gate (the diff has no `.ts`), no E2E.

### Recommendation

Merge as-is. M2 is a one-line comment if the author wants it; M1 and M3 are notes, not
requests.

PR #2024, opened 2026-08-24
Replace damage with counters on the damaged permanent, and add Soul-Scar Mage

bolav:
# Replace damage with counters on the damaged permanent, and add Soul-Scar Mage

Soul-Scar Mage (AKH #148) needs one thing the SDK didn't have: a damage replacement
whose counters land on the **damaged** permanent rather than on the permanent
carrying the ability.

## The SDK change: parameterize, don't clone

`ReplaceDamageWithCounters` hardcoded the counter recipient as its own host — "put
that many depletion counters on **this enchantment** instead" (Force Bubble), and the
self-damage case where host and damaged permanent coincide (Anti-Venom). Soul-Scar
Mage says the other thing: "put that many -1/-1 counters on **that creature**
instead".

That is a constant baked into a type, which `sdk-design-principles.md` says to
parameterize rather than clone, so the type gains
`counterRecipient: DamageCounterRecipient` — `ReplacementHost` (the default, so every
existing printing is untouched) or `DamagedPermanent`. A near-duplicate replacement
type would have doubled the engine's damage-replacement surface for one field.

Nothing else was needed at the card layer: every clause of the printed text is
carried by the event pattern the SDK already had —
`DamageEvent(recipient = CreatureOpponentControls, source = YouControl, damageType = NonCombat)`.

## The engine change: honour the pattern that was already there

`DamageUtils.applyReplaceDamageWithCounters` was ignoring most of its own
`appliesTo` pattern. That was invisible while the only users left `source` and
`damageType` at `Any`, and immediately wrong for a noncombat-only, source-scoped
replacement — it would have swallowed combat damage and an opponent's burn spell
alike. It now honours:

- **damage type** (`Any` / `Combat` / `NonCombat`), with the two combat call sites in
  `CombatDamageManager` passing `isCombatDamage = true`;
- **source filter** (`Any` / `Self` / `YouControl`), *declining* any other filter
  rather than guessing at it — answering an unmodelled filter wrong would silently
  widen a card's reach;
- **`RecipientFilter.CreatureOpponentControls`**, read off projected state for both
  the type and the controller, so a stolen creature is judged by its current
  controller and a creature that is only a creature because of a continuous effect is
  judged correctly.

Both are no-ops for Force Bubble and Anti-Venom. The card snapshot golden is the
evidence: re-blessing moved **only** Soul-Scar Mage's tree.

### A latent bug fixed on the way

The helper parsed the printed counter name with a homegrown transform
(`"+1/+1"` → uppercase → `'+'`→`'P'` → `'-'`→`'M'` → `'/'`→`'_'` → `"P1_P1"`), which
`CounterType.valueOf` always threw on, and the `catch` fell back to
`PLUS_ONE_PLUS_ONE`. Anti-Venom therefore worked *by accident*, and a `"-1/-1"`
replacement would have silently placed **+1/+1** counters — i.e. Soul-Scar Mage would
have pumped the creature it was meant to shrink. Replaced with the shared
`CounterType.fromName`, which knows the symbolic stat names. Same answer for both
existing cards, correct answer for the new one.

## The second engine change: answer "a source you control" with last-known information

Playtesting the dev scenario surfaced a bug the first commit's tests could not see. The
Mage replaced Prodigal Sorcerer's ping but *not* Fanatical Firebrand's. Both are
noncombat damage from a source their controller controls; the only difference is that
the Firebrand sacrifices itself to pay its own activation cost, so it deals its damage
from the graveyard.

`controllerOfObject` asked the projection and then the base `ControllerComponent`, and a
permanent that has left the battlefield has neither — `stripBattlefieldComponents` takes
the `ControllerComponent` on the way out. The lookup returned null, `SourceFilter.YouControl`
declined, and the damage stayed ordinary damage. **CR 113.7a** and **CR 608.2h** both say
the source's last known information answers questions about a source that is no longer in
the zone it was expected to be in, so the lookup now falls back to
`LastKnownPermanentComponent`.

**This was never Soul-Scar-Mage-specific.** `damageSourceMatches` — shared by the
prevention, doubling and flat/dynamic-modification scans — carried its own copy of the same
two-step lookup and so had the same hole. Every implemented card with the same source
clause was equally blind to a self-sacrificing source:

| Card | Set | Clause |
|---|---|---|
| Hawkeye, Young Avenger | MSH | "…deals that much damage plus X" |
| Fated Firepower | TLA | source-scoped damage modification |
| Far Fortune, End Boss | DFT | source-scoped damage modification |

Both now route through `damageSourceController`, which takes the projection as a parameter
so a mid-projection caller can pass its in-flight one instead of `state.projectedState`.

`applyReplaceDamageWithCounters` held a *third* copy of the source-filter `when` — which is
how the two drifted apart in the first place — and now calls the shared
`damageSourceMatches`. That also widens it from `Any`/`Self`/`YouControl` to every filter
the other scans already support, rather than declining the rest. The "declines rather than
guesses" property noted in the review is preserved: `damageSourceMatches` has the same
`else -> false` floor.

## Rules

- **CR 614.1a** — effects that use the word "instead" are replacement effects. The
  damage is never dealt, so nothing keying on damage being dealt sees it.
- **CR 614.5** — a replacement gets one opportunity at an event, which is why the
  first matching effect wins and the scan stops.
- The card's own 2017-04-18 rulings, carried into `metadata { ruling(...) }`: fight
  damage is noncombat damage and *is* replaced; the effect is not a prevention effect,
  so it still applies to damage that can't be prevented; and when several
  prevention/replacement effects compete, the damaged creature's controller orders
  them.

Rule numbers checked against `MagicCompRules_20260808.txt`, not from memory.

## Verification

The branch is two units of work with two gates. Everything below was re-run after the
last commit unless marked otherwise.

**The last-known-information fix** (engine-wide damage code, so the gate is scoped to the
modules the diff can reach):

- `scripts/gradle-locked :rules-engine:test` — **BUILD SUCCESSFUL**. This is the gate that
  matters: it covers Force Bubble, Anti-Venom, wither, and the rest of the damage paths the
  shared helper feeds.
- `just test-scenarios` for **2017-2022, 2025, 2026 and 2003-2007** — 786 + 1595 + 1320 +
  670 = **4,371 tests, 0 failures, 0 errors** (counted from the JUnit XML, not from a
  "BUILD SUCCESSFUL" line). Those four eras hold every card that reaches the changed code:
  Soul-Scar Mage, Anti-Venom, Force Bubble, Hawkeye, Fated Firepower, Far Fortune.
- `just test-class SoulScarMageScenarioTest` — **10/10**, up from 8 with two new cases, one
  per side of the difference the bug turned on: Prodigal Sorcerer (source stays on the
  battlefield) as the regression guard, Fanatical Firebrand (source sacrificed to its own
  cost) for the bug.
- **Mutation-checked.** Dropping the `LastKnownPermanentComponent` fallback reddens exactly
  the Firebrand test and leaves the other nine green. Restored and re-run green.

**The SDK + pattern-honouring change** (from the first commit, not re-run since — nothing
after it touched the SDK or the goldens):

- `:mtg-sets:test` — green after re-blessing; only Soul-Scar Mage moved in the golden.
- Both new filters mutation-checked: forcing the damage-type branch to `true` reddens
  exactly "combat damage is untouched"; forcing the source branch to `true` reddens exactly
  "damage from a source your opponent controls is untouched".
- `just check-card-printing "Soul-Scar Mage"` — clean, AKH is the earliest real printing.
- `just assay-differential --set AKH` — 16/16 confirmed, 0 divergences.
- `:oracle-assay:compileKotlin` / `:compileTestKotlin` — green.

Card oracle text re-checked against Scryfall: still "a **creature** an opponent controls",
not "permanent", so `RecipientFilter.CreatureOpponentControls` is right and no errata has
landed. Rule numbers checked against `MagicCompRules_20260808.txt`, not from memory.

**What this does not cover:** nothing here touches the web client or the server, so no
TypeScript or `:game-server` gate was run and neither would say anything about this diff.
No E2E run — the change surfaces no new decision or UI affordance; -1/-1 counters already
render. The scenario eras *not* listed above were not re-run; the grep for
`SourceFilter.YouControl` and `ReplaceDamageWithCounters` across the corpus found no card
outside those four.

## Not done, on purpose

**CR 616.1 ordering is not modelled**, and this card is the first to make the deviation
observable in the affected player's disfavour. When this replacement competes with another
replacement or prevention effect, 616.1 gives the ordering choice to the affected object's
controller — the opponent — and the card's own 2017-04-18 ruling spells out both directions
they would use it in. The engine has no 616.1 prompt at all: it amplifies before it
replaces and places counters ahead of a shield counter, in both cases the order the
affected player would not choose. That is engine-wide behaviour rather than anything this
branch introduced, and building 616.1 ordering is plainly out of scope. Since the card's
`metadata` ships the ordering ruling verbatim, the deviation is now named in the card's
KDoc so the two don't read as a contradiction.

`ReplaceDamageWithCounters` has no entry in the mtgish bridge — the generator can't
predict or draft *any* card using it, before or after this change. Adding one would be
about the pre-existing type rather than about this parameter, so it is left as a
standing gap rather than folded in here (`add-feature` Step 8b gates on introducing a
new IR tag, and this introduces none).

## Try it

```bash
  -H 'Content-Type: application/json' -d @manual-scenarios/cards/s/soul-scar-mage.json
```

Burn an opponent's creature to see the counters and the permanent shrink; tap Prodigal
Sorcerer and then the hasty Fanatical Firebrand to confirm the clause covers an *ability*
source, whether or not that source survives paying its own cost — those two side by side
are the bug this branch's second commit fixes; attack with the Mage next turn to confirm
combat damage stays ordinary damage.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -11460,6 +11460,28 @@ The priority groups are (CR 616.1a–f):
   Martyrs of Korlis uses `Conditions.SourceIsUntapped` for "As long as this creature is untapped, all
   damage that would be dealt to you by artifacts is dealt to this creature instead" (`redirectTo =
   EffectTarget.Self`, `source = SourceFilter.Matching(GameObjectFilter.Artifact)`).
+- `ReplaceDamageWithCounters(counterType, sacrificeThreshold = null, appliesTo = DamageEvent(recipient =
+  You), counterRecipient = DamageCounterRecipient.ReplacementHost)` — replace matching damage
+  (CR 614.1a, an "instead" effect: the damage is never dealt, so nothing that keys on damage being
+  dealt sees it) with `counterType` counters. Not a prevention effect, so it still applies to damage
+  that can't be prevented. `appliesTo` filters recipient, source, damage type and amount;
+  `counterRecipient` says where the counters land:
+  - `ReplacementHost` (default) — on the permanent that has the ability. **Force Bubble**: "If damage
+    would be dealt to you, put that many depletion counters on this enchantment instead"
+    (`DamageEvent(recipient = You)`, `sacrificeThreshold = 4` for the paired "sacrifice it when it has
+    four or more"). **Anti-Venom, Horrifying Healer** is the self-damage case where host and damaged
+    permanent coincide (`recipient = RecipientFilter.Self`).
+  - `DamagedPermanent` — on the permanent that would have been dealt the damage. **Soul-Scar Mage**:
+    "If a source you control would deal noncombat damage to a creature an opponent controls, put that
+    many -1/-1 counters on that creature instead" — `DamageEvent(recipient =
+    RecipientFilter.CreatureOpponentControls, source = SourceFilter.YouControl, damageType =
+    DamageType.NonCombat)`. A player recipient has nowhere to put counters, so the replacement
+    declines rather than eating the damage.
+
+  Wired in both damage paths (`DamageUtils.applyReplaceDamageWithCounters` for the general path,
+  `CombatDamageManager` for combat); combat callers pass `isCombatDamage = true` so a noncombat-only
+  pattern isn't applied to combat damage. Supported source filters are `Any`, `Self` and `YouControl`
+  — any other `SourceFilter` declines rather than guessing.
 - `ReplaceDamageWithMill(appliesTo = DamageEvent(recipient = Opponent))` — replace matching damage
   (CR 615, neither dealt nor prevented): each opponent of the replacement's controller mills that many
   cards instead. The Mindskinner: `DamageEvent(recipient = RecipientFilter.Opponent, source =
--- a/manual-scenarios/cards/s/soul-scar-mage.json
+++ b/manual-scenarios/cards/s/soul-scar-mage.json
@@ -11,6 +11,7 @@
       {"name": "Mountain"},
       {"name": "Mountain"},
       {"name": "Mountain"},
+      {"name": "Prodigal Sorcerer"},
       {"name": "Fanatical Firebrand"}
     ],
     "graveyard": [],
--- a/manual-scenarios/cards/s/soul-scar-mage.json
+++ b/manual-scenarios/cards/s/soul-scar-mage.json
@@ -1,0 +1,38 @@
+{
+  "player1Name": "You",
+  "player2Name": "AI",
+  "mode": "AI",
+  "aiPlayer": 2,
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Soul-Scar Mage", "Lightning Bolt", "Shock"],
+    "battlefield": [
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Fanatical Firebrand"}
+    ],
+    "graveyard": [],
+    "library": ["Hill Giant", "Mountain", "Grizzly Bears", "Mountain"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": ["Grizzly Bears", "Mountain"],
+    "battlefield": [
+      {"name": "Craw Wurm"},
+      {"name": "Serra Angel"},
+      {"name": "Mountain"},
+      {"name": "Mountain"}
+    ],
+    "graveyard": [],
+    "library": ["Hill Giant", "Mountain", "Grizzly Bears"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": ["DECLARE_ATTACKERS", "DECLARE_BLOCKERS"],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": []
+}
--- a/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/ReplacementEffect.kt
+++ b/mtg-sdk/src/main/kotlin/com/wingedsheep/sdk/scripting/ReplacementEffect.kt
@@ -1971,13 +1971,50 @@ data class EntersWithDevour(
 // =============================================================================
 
 /**
- * Replace damage dealt to a player with counters on this permanent.
- * Example: Force Bubble — "If damage would be dealt to you, put that many
- * depletion counters on this enchantment instead."
+ * Where the counters land when a [ReplaceDamageWithCounters] applies.
  *
- * @param counterType The type of counter to add (e.g., "depletion")
+ * The recipient of the *damage* is already named by the replacement's `appliesTo` pattern; this
+ * names the recipient of the *counters*, which is a separate question the printed cards answer
+ * two different ways.
+ */
+@Serializable
+enum class DamageCounterRecipient {
+    /**
+     * The permanent that has the replacement effect — "put that many depletion counters on
+     * **this enchantment** instead" (Force Bubble), and the self-damage case where the host and
+     * the damaged permanent happen to be the same object (Anti-Venom).
+     */
+    ReplacementHost,
+
+    /**
+     * The permanent that would have been dealt the damage — "put that many -1/-1 counters on
+     * **that creature** instead" (Soul-Scar Mage). Only meaningful when the pattern's recipient
+     * is a permanent; a player recipient has nowhere to put them, and the replacement declines.
+     */
+    DamagedPermanent
+}
+
+/**
+ * Replace damage with counters (CR 614.1a — an "instead" effect, so the damage is never dealt and
+ * nothing that keys on damage being dealt sees it; notably *not* a prevention effect, so it still
+ * applies to damage that can't be prevented).
+ *
+ * Which damage is replaced comes from [appliesTo] — recipient, source, damage type and amount are
+ * all filterable there. Where the counters go comes from [counterRecipient]:
+ *
+ * - Force Bubble — "If damage would be dealt to you, put that many depletion counters on this
+ *   enchantment instead": `DamageEvent(recipient = You)`, counters on the [ReplacementHost].
+ * - Soul-Scar Mage — "If a source you control would deal noncombat damage to a creature an
+ *   opponent controls, put that many -1/-1 counters on that creature instead":
+ *   `DamageEvent(recipient = CreatureOpponentControls, source = YouControl,
+ *   damageType = NonCombat)`, counters on the [DamagedPermanent].
+ *
+ * @param counterType The type of counter to add (e.g., "depletion", "-1/-1")
  * @param sacrificeThreshold If non-null, sacrifice this permanent when it has
- *        this many or more counters of the specified type (state-triggered ability)
+ *        this many or more counters of the specified type (state-triggered ability).
+ *        Only meaningful together with [DamageCounterRecipient.ReplacementHost].
+ * @param counterRecipient Which permanent receives the counters. Defaults to the replacement's
+ *        own host, which is what every "on this permanent" printing says.
  */
 @SerialName("ReplaceDamageWithCounters")
 @Serializable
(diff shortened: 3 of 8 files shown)


bolav:
## Review — `soul-scar-mage`

Reviewed `origin/main...soul-scar-mage`. Local `main` is level with `origin/main`, so no
merge was needed.

### Verdict

The SDK shape is right — this parameterizes a constant that was baked into an existing
replacement rather than cloning the type, which is what `sdk-design-principles.md` asks
for. The engine change is a genuine correctness fix to a shared helper, and the card
snapshot golden is good evidence it didn't move anything else. One rules-fidelity limit
worth stating on the PR before merge (I1).

### What's good

- **Parameterizing beat cloning, and the default proves it.** `counterRecipient` defaults
  to `ReplacementHost`, so Force Bubble and Anti-Venom are untouched at the data level —
  and the re-blessed golden shows only Soul-Scar Mage's tree moved. That's the strongest
  available evidence that a shared-type change was behaviour-preserving, and it's the
  right check to have run.
- **The source filter declines rather than guesses.**
  `DamageUtils.kt:2617-2624` handles `Any` / `Self` / `YouControl` and returns `false` for
  everything else. Silently matching an unmodelled `SourceFilter` would widen a card's
  reach in a way no test would catch; failing closed is correct and the comment says why.
- **The counter-name parse fix is the real find here.** The old inline transform mapped
  `"+1/+1"` to `"P1_P1"`, always threw, and worked only via the `catch`'s
  `PLUS_ONE_PLUS_ONE` fallback — meaning a `"-1/-1"` replacement would have placed **+1/+1**
  counters, i.e. this exact card would have pumped what it meant to shrink. Routing to the
  shared `CounterType.fromName` is right, and the PR body says so plainly instead of
  burying it.
- **The test suite is built around over-application, not the happy path.** Combat damage,
  your own creatures, players, and an opponent-controlled source each get their own case,
  and both new filters were mutation-checked. That is the correct shape for a change whose
  failure mode is "applies too widely".

### Issues

**I1 — CR 616.1 ordering is fixed, and this card is the first to make it observable in
the player's disfavour.** `DamageUtils.kt:~276` applies
`applyStaticDamageAmplification` *before* the damage-to-counters replacement. CR 616.1
gives the choice to "the affected object's controller", and Soul-Scar Mage's own
2017-04-18 ruling spells out both directions: an opponent "could choose to convert the
damage to -1/-1 counters before Insult could double it, and then Insult's effect won't
apply anymore", and could choose Djeru's Resolve's prevention over the counters. The
engine has no 616.1 ordering prompt at all, so it always doubles first (2x counters) and
always applies the counters ahead of a shield counter — in both cases the order the
affected player would *not* choose. That is pre-existing engine-wide behaviour, not
something this branch introduced, and building 616.1 ordering is plainly out of scope
here. But the card's `metadata { ruling(...) }` block ships the ordering ruling as if it
were honoured, which is the part that misleads. Either add a sentence to the card KDoc
naming the fixed order as a known deviation, or add it to the PR body's "not done"
section. My preference is the KDoc — that's where the next reader looks.

**M1 — the placed counters bypass `applyCounterPlacementModifiers`.**
`DamageUtils.kt:~2646` writes `CountersComponent` directly, so a
`ModifyCounterPlacement` / `DoubleCounterPlacement` replacement doesn't see these
counters. Pre-existing for the host path (Force Bubble's depletion counters have the same
gap) and it doesn't bite any real card today, since the modifiers in the corpus are
+1/+1-scoped. Flagging so it isn't rediscovered as a Soul-Scar-specific bug.

**M2 — `damageSourceController` and `replacementHostController` are the same one-liner.**
`DamageUtils.kt:135-152` gives one implementation two named aliases. I think that's the
right call — the names carry which question is being asked, at zero cost, and renaming
the 13 existing call sites would have been pure churn — but a reviewer who sees two
identical bodies will ask. The KDocs already explain it; no change wanted, just
pre-empting the question.

- `:rules-engine:test` — 0 failures. This is the gate that matters: it covers Force
  Bubble, Anti-Venom and wither, all of which feed the modified helper.
- `:mtg-sets:test` + rebless — only Soul-Scar Mage moved in the golden.
- `just test-class SoulScarMageScenarioTest` — 8/8, and both mutations reproduced the
  expected single red.
- `just assay-differential --set AKH` — 16/16 confirmed, 0 divergences.
- CR 614.1a, 614.5 and 616.1 checked against `MagicCompRules_20260808.txt`; all three
  match the text they're cited for.

Not covered: no web-client or `:game-server` gate, and neither would be evidence about
this diff. No E2E — the change adds no decision or UI affordance.

### Recommendation

Merge after adding the I1 note. Everything else is discretionary.

---

## Round 2 — reviewing the two commits added after the first pass

Reviewed `9bf4477f73` (*Answer "a source you control" with last-known information*) and
`8dca74aab1` (the CR 616.1 KDoc note) on top of the branch as first reviewed.

### Verdict

The LKI fix is the right one, and it is *wider* than the card that surfaced it — which is
the part worth saying on the PR, because the title still reads as a single-card change.
**I1 is now addressed.** No new issues; two of my earlier observations need amending
against the new code.

### What's good

- **The diagnosis is falsifiable and was falsified.** Prodigal Sorcerer and Fanatical
  Firebrand differ in exactly one thing — whether the source is still on the battlefield
  when the damage is dealt — and the branch now has a test for each. The mutation check
  (drop the `LastKnownPermanentComponent` fallback → exactly the Firebrand test reddens,
  the other nine stay green) is the evidence that the new test tests the new code and not
  something incidental.
- **Fixed at the shared helper, not at the card.** The tempting fix was a special case in
  `applyReplaceDamageWithCounters`. Putting it in `controllerOfObject` instead means
  Hawkeye, Fated Firepower and Far Fortune — which had the identical hole through
  `damageSourceMatches` and no test between them — are fixed by the same three lines. The
  PR body names all three rather than leaving the blast radius implicit.
- **The third copy of the source-filter `when` is gone.** `applyReplaceDamageWithCounters`
  now calls `damageSourceMatches` (`DamageUtils.kt:2636`, joining the three existing call
  sites at 2014/2093/2162). Duplicated matching logic is exactly how the doubling scan came
  to silently ignore `YouControl` in the first place, per that function's own KDoc — one
  fewer copy is one fewer chance to drift.
- **Passing the projection as a parameter rather than reading `state.projectedState`.**
  `controllerOfObject` (`:170`) defaults to `state.projectedState` but lets a mid-projection
  caller pass its in-flight one, which is what the root `AGENTS.md` asks for. Free
  correctness for a caller that doesn't exist yet, at the cost of one defaulted parameter.
- **Rule citations check out.** CR 113.7a ("if the source is no longer in the zone it's
  expected to be in at that time, its last known information is used"), CR 608.2h and CR
  616.1 all verified against `MagicCompRules_20260808.txt`, and the Scryfall oracle re-check
  confirms the card still reads "creature", not "permanent".

### Amendments to the first-round notes

- **I1 — resolved.** The KDoc now names the fixed-ordering deviation and says why it sits
  next to a `ruling(...)` block that describes the choice being honoured. That was my stated
  preference and it is where the next reader looks. Nothing further wanted.
- **M2 — no longer quite true, and better for it.** `damageSourceController` and
  `replacementHostController` are no longer byte-identical one-liners:
  `damageSourceController` (`:146`) now carries the projection parameter that the source
  side needs and the host side never will. The names still document which question is being
  asked; the signatures now differ for a reason.
- **The "declines rather than guesses" property survives the consolidation.** My first-round
  praise pointed at the inline `when` that this commit deleted. `damageSourceMatches` has
  the same `else -> false` floor (`:1750` onward), so an unmodelled `SourceFilter` still
  fails closed — it just now supports `EnchantedCreature`/`EquippedCreature`/`Matching`
  first, which is a widening toward the filters the other three scans already honoured, not
  toward guessing.
- **M1 stands unchanged.** The placed counters still bypass
  `applyCounterPlacementModifiers`; nothing in these commits touches that path.

- `:rules-engine:test` — BUILD SUCCESSFUL.
- `just test-scenarios` for 2017-2022 / 2025 / 2026 / 2003-2007 — 4,371 tests, 0 failures,
  0 errors, counted from the JUnit XML rather than trusting the build line. Those four eras
  hold every card reaching the changed paths; I confirmed that by grepping the corpus for
  `SourceFilter.YouControl` and `ReplaceDamageWithCounters` and finding no card outside
  them.
- `just test-class SoulScarMageScenarioTest` — 10/10, plus the mutation reproducing a single
  expected red.
- The second commit is comment-only, so no gate was run for it and none would be evidence
  about it.

Not covered, unchanged from round 1: no web-client, `:game-server` or E2E gate, and none
would say anything about this diff.

### Recommendation

Merge. The first round's one blocking item is closed and the second commit strictly widens
the fix.

PR #2029, opened 2026-08-25
Import a pasted decklist's sideboard, and carry it into the game

bolav:
# Import a pasted decklist's sideboard, and carry it into the game

Pasting a full Arena/Moxfield decklist into a lobby's **Paste** tab produced a deck with an
empty sideboard, however faithfully the parser had read the list. Fixing that turned out to
need three separate things, one per commit.

## The picker was discarding it (`b03443a`)

`parseArenaDeckList` has always split a `Sideboard` section (and the per-line `SB:` prefix)
into its own list — the deckbuilder's import modal uses it. The lobby picker did not: it kept
only `parsed.entries` and hardcoded the sideboard to `{}` for every tab but Saved.

```ts
const currentSideboard = useMemo(() => {
  if (tab === 'saved') return saved?.sideboard ?? {}
  return {}          // <- a pasted Sideboard section died here
}, [...])
```

The paste box's text↔deck conversion moves into `web-client/src/components/ui/deckPasteText.ts`
so it can be tested without mounting the picker, and keeps both boards. The Paste tab feeds its
parsed sideboard into `currentSideboard`, which was *already* wired through `onDeckChange` →
`submitLobbyDeck` → the `sideboard` wire field; "Save deck" persists it; and `formatDeckText`
emits a `Sideboard` header so editing a saved deck that has one round-trips instead of silently
dropping it.

Parse errors now carry the section they were read under, so the picker's "bare card name = one
copy" shorthand rescues a line onto the board it was written under rather than always the main
deck.

Cloud saves dropped the sideboard too — the same hole in the deckbuilder, where a signed-in user
lost their sideboard on every save. `SharedDeck` gains a `sideboard`, carried in
`savedDeckToShared` and `buildSharedDeck`, read back in `detailToUnified` / `applySharedDeck`.
The account controller stores the deck body verbatim, so this needed **no server change**.

## Quick Game had no sideboard at all (`1bdd463`)

That still didn't put a sideboard in a game, because the Quick Game lobby drops it at every
layer: no `sideboard` on `SubmitQuickGameLobbyDeck`, none on the lobby seat, and the seating loop
calls `addPlayer(session, deck, commanderCardName = ...)` — leaving the `sideboard` parameter on
its `emptyMap()` default. **A Quick Game has therefore always started with an empty sideboard, for
saved decks as much as pasted ones.** The Premade Decks lobby was already wired end to end, which
is the only reason the difference between the two screens was invisible.

The field now goes through the wire message, `QuickGameLobbyPlayer`, and the seating loop (empty
for Momir Basic and for a random pool, which has no constructed sideboard — a Limited sideboard is
derived as pool − maindeck, CR 100.4b). The submit handler's resubmit no-op compares the sideboard
too, so editing only the sideboard is still a real change.

### `SideboardSanitizer` — the half that isn't optional

A submitted *deck* is validated and rejected when it names a card this engine hasn't implemented.
A sideboard is not: it reaches `GameInitializer` unchecked, where `cardRegistry.requireCard` throws
on the first unknown name, failing a game start whose decks were all perfectly legal. Pasted Arena
sideboards routinely name cards the corpus hasn't reached yet, so this became reachable the moment
the picker started delivering them — the first commit created the exposure that the second closes.

Rejecting the whole submission would be the wrong trade: one unimplemented card shouldn't lock a
player out of a game they could otherwise play. Unknown names are dropped and logged instead — a
wish simply won't find them, which is the truth of it — on the premade path as well as the new one.

## Rule citations (`d6cd32d`)

Verified against the local `MagicCompRules_20260808.txt`, which is how the mistake surfaced:

- **CR 400.11a** — "Cards in a player's sideboard are outside the game." This is the rule the
  comments wanted; every citation added on this branch originally said 100.4a, and was corrected.
- **CR 100.4a** — the fifteen-card cap, and the four-of limit applying to deck and sideboard
  *combined*. A different claim, cited nowhere in this branch's code.
- **CR 100.4b** — the Limited sideboard is the pool minus the maindeck. Pre-existing in
  `SideboardDerivation`, referenced here for the random-pool case.

The same 100.4a/400.11a mispairing predates this branch in `deckLibrary.ts`, `DeckbuilderPage.tsx`
and `messages.ts`. Those are deliberately left alone rather than swept into this change.

## No new SDK vocabulary

Nothing here is engine or SDK work. `SideboardSanitizer` is a `game-server` object, which is where
it belongs: the server is the anti-corruption layer between untrusted client input and an engine
that is entitled to assume a well-formed `Deck`. Keeping the filter out of `GameInitializer` leaves
the engine strict — it still throws on a malformed deck, which is what you want from it.

## Review fixes (`0d34bfa`)

A review pass over the branch found four defects, all fixed before opening this:

- The bare-name rescue read the *un-preprocessed* line, so `SB: Counterspell` became a card named
  `"SB: Counterspell"` — counted in the paste box's sideboard total, then silently dropped by the
  server as unknown. Introduced by this branch; two regression tests added.
- A page refresh or deep link to `/deckbuilder/<id>` rehydrated the deck without its sideboard, and
  the next Save wrote that emptiness back over the stored one. Pre-existing, but this branch is what
  made a lost sideboard matter.
- Loading an example kept the previous deck's sideboard attached.
- `GamePlayHandler`'s create/join paths handed the engine an unsanitized sideboard — the same
  `requireCard` throw the sanitizer exists to prevent, on the two call sites the first pass missed.

The full review, including one open item, is in `sideboard-paste_review.md`.

## Verification

- `just test-server` — **BUILD SUCCESSFUL, 507 passed, 0 failed**, including four new
  `SideboardSanitizerTest` cases. Re-run in full after the review fixes; an earlier targeted
  re-run (`just test-class SideboardSanitizerTest`) covered a comment-only edit, so no green here
  is claimed from a stale tree.
- `web-client` — `npx tsc --noEmit` clean; `npx vitest run` **45 files, 658 tests passed**,
  including eight new `deckPickerPaste` tests covering the paste split, the `SB:` prefix (both as a
  counted entry and as a rescued bare name), Moxfield decorations, the shorthand rescue, and the
  textarea round-trip.
- The parser was checked against the exact 20-line/11-name decklist that prompted this: it splits
  **60 main / 15 sideboard**.

**What the gates do not cover:**

- **No engine gate was run**, and none is warranted — nothing in `rules-engine` changed. A green
  `game-server` suite says nothing about engine behaviour, and isn't claimed to.
- **No e2e or integration coverage of the actual round trip.** Paste → lobby → `submitDeck` →
  `addPlayer` → `Zone.SIDEBOARD` is verified by reading the code and by unit tests at each end, not
  by exercising the path. There is no e2e harness for lobby deck submission here.
- **The cloud save/load round-trip is unverified by any test** — it relies on the account
  controller storing `SharedDeck` JSON verbatim, which is true by inspection but not asserted.

## Deliberately out of scope

- **Share links still drop the sideboard.** The v2 share code is a compact delimited format with no
  field for one; adding it needs a version bump and a decode path. Documented on the new
  `SharedDeck.sideboard` property so it doesn't read as supported.
- **A sideboard remains invisible during play.** `ClientStateTransformer` excludes it from client
  state as outside the game (CR 400.11a) and no component renders that zone, so it is observable
  only through a wish effect's search. Giving it a UI, or sideboarding between games of a match,
  would each be their own change.
- **Neither CR 100.4a limit is enforced on the paste path** — not the four-of limit across deck and
  sideboard combined, nor the fifteen-card cap. `DeckValidator` has no sideboard handling at all,
  while the deckbuilder's *manual* add path does enforce the combined cap, so the two contradict
  each other. Pre-existing, but this branch makes the lobby the shortest route to an illegal
  sideboard, so it's worth a decision — see the open item in `sideboard-paste_review.md`. Left out
  because it's a deck-legality feature and a truncate-or-reject product call, not part of making
  paste work.

### DIFF
--- a/game-server/src/main/kotlin/com/wingedsheep/gameserver/deck/SideboardSanitizer.kt
+++ b/game-server/src/main/kotlin/com/wingedsheep/gameserver/deck/SideboardSanitizer.kt
@@ -1,0 +1,45 @@
+package com.wingedsheep.gameserver.deck
+
+import com.wingedsheep.engine.registry.CardRegistry
+
+/**
+ * Drops sideboard entries the card registry doesn't know, at the server boundary.
+ *
+ * A submitted *deck* is validated and rejected outright when it names a card this engine hasn't
+ * implemented. A sideboard isn't: it lives "outside the game" (CR 100.4a) and reaches
+ * `GameInitializer` unchecked, where `cardRegistry.requireCard` throws on the first unknown name —
+ * so a single unimplemented sideboard card would fail game *initialization*, taking down a lobby
+ * whose decks were all perfectly legal.
+ *
+ * Rejecting the submission instead would be worse: real decklists are pasted from Arena or
+ * Moxfield with full 15-card sideboards, and one card the corpus hasn't reached yet would lock the
+ * player out of a game they could otherwise play. So unknown names are dropped and reported to the
+ * caller for logging — a wish simply won't find them, which is the truth of it.
+ *
+ * Non-positive counts are dropped too; they can't produce a card and would otherwise sit in the
+ * lobby state as noise.
+ */
+object SideboardSanitizer {
+
+    /**
+     * @param kept entries the registry can resolve, safe to hand to the engine
+     * @param dropped names that were discarded, for the caller's log line
+     */
+    data class Result(val kept: Map<String, Int>, val dropped: List<String>) {
+        val hasDrops: Boolean get() = dropped.isNotEmpty()
+    }
+
+    fun sanitize(sideboard: Map<String, Int>, registry: CardRegistry): Result {
+        if (sideboard.isEmpty()) return Result(emptyMap(), emptyList())
+        val kept = LinkedHashMap<String, Int>(sideboard.size)
+        val dropped = mutableListOf<String>()
+        for ((name, count) in sideboard) {
+            when {
+                count <= 0 -> Unit
+                registry.hasCard(name) -> kept[name] = count
+                else -> dropped += name
+            }
+        }
+        return Result(kept, dropped)
+    }
+}
--- a/game-server/src/main/kotlin/com/wingedsheep/gameserver/deck/SideboardSanitizer.kt
+++ b/game-server/src/main/kotlin/com/wingedsheep/gameserver/deck/SideboardSanitizer.kt
@@ -6,7 +6,7 @@ import com.wingedsheep.engine.registry.CardRegistry
  * Drops sideboard entries the card registry doesn't know, at the server boundary.
  *
  * A submitted *deck* is validated and rejected outright when it names a card this engine hasn't
- * implemented. A sideboard isn't: it lives "outside the game" (CR 100.4a) and reaches
+ * implemented. A sideboard isn't: it lives "outside the game" (CR 400.11a) and reaches
  * `GameInitializer` unchecked, where `cardRegistry.requireCard` throws on the first unknown name —
  * so a single unimplemented sideboard card would fail game *initialization*, taking down a lobby
  * whose decks were all perfectly legal.
--- a/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/GamePlayHandler.kt
+++ b/game-server/src/main/kotlin/com/wingedsheep/gameserver/handler/GamePlayHandler.kt
@@ -2,6 +2,7 @@ package com.wingedsheep.gameserver.handler
 
 import com.wingedsheep.gameserver.ai.AiGameManager
 import com.wingedsheep.gameserver.ai.AiWebSocketSession
+import com.wingedsheep.gameserver.deck.SideboardSanitizer
 import com.wingedsheep.ai.engine.SealedDeckGenerator
 import com.wingedsheep.gameserver.protocol.ClientMessage
 import com.wingedsheep.gameserver.protocol.ErrorCode
@@ -51,6 +52,21 @@ class GamePlayHandler(
     private val deckProfiler: com.wingedsheep.gameserver.stats.DeckProfiler
 ) {
     private val logger = LoggerFactory.getLogger(GamePlayHandler::class.java)
+
+    /**
+     * Drop sideboard cards this engine hasn't implemented before they reach the engine, which
+     * would throw on the first one and fail the whole game start. See [SideboardSanitizer] for
+     * why unknown names are dropped rather than the submission rejected.
+     */
+    private fun sanitizeSideboard(sideboard: Map<String, Int>?, playerName: String): Map<String, Int> =
+        SideboardSanitizer.sanitize(sideboard.orEmpty(), cardRegistry).also { result ->
+            if (result.hasDrops) {
+                logger.info(
+                    "Dropped ${result.dropped.size} unknown sideboard card(s) from $playerName's " +
+                        "submission: ${result.dropped}",
+                )
+            }
+        }.kept
 
     @Volatile
     var waitingGameSession: GameSession? = null
@@ -138,7 +154,13 @@ class GamePlayHandler(
         )
         gameSession.quickGameSetCode = quickGameSetCode
         // A generated random/quick deck has no sideboard; only a real submitted deck carries one.
-        val sideboard = if (quickGameSetCode != null) emptyMap() else message.sideboard
+        // Sanitize whatever the client sent: an unimplemented name would throw out of
+        // GameInitializer and fail the game creation outright (see SideboardSanitizer).
+        val sideboard = if (quickGameSetCode != null) {
+            emptyMap()
+        } else {
+            sanitizeSideboard(message.sideboard, playerSession.playerName)
+        }
         gameSession.addPlayer(playerSession, deckList, sideboard = sideboard)
 
         // Store player info for persistence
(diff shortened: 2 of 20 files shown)


bolav:
## Review — `sideboard-paste`

**Verdict:** behaviour is right and the shape is right. No new SDK vocabulary, and the one new
type (`SideboardSanitizer`) is a `game-server` object, which is where a filter over untrusted
client input belongs — putting it in `GameInitializer` would have made the engine lenient about
malformed decks, which is the wrong trade. Reviewed in place on the branch; `origin/main` is
already fully contained in `HEAD`, so no merge was needed and no worktree was created.

A review pass found five defects. **Four are fixed in `0d34bfa`**; the fifth is left open
deliberately and is the one thing needing a decision before merge.

### What's good

- The sanitizer's "drop and log, don't reject" call is the right one and the KDoc argues for it
  rather than just asserting it — one unimplemented card shouldn't lock a player out of a game.
- `deckPasteText.ts` being extracted from the component is what made the paste logic testable at
  all; the six tests there are real behaviour tests, not smoke.
- The commit that corrects `100.4a` → `400.11a` scopes itself to the citations this branch
  introduced and explicitly leaves the identical pre-existing mispairing elsewhere alone.

### Fixed in `0d34bfa`

- **B1 — the bare-name rescue manufactured garbage card names.**
  `web-client/src/components/ui/deckPasteText.ts:41` read `err.raw`, the line *before*
  `preprocessLine` strips the `SB:` prefix and Moxfield decorations. Verified empirically:
  `parseDeckText('4 Lightning Bolt\nSB: Counterspell')` produced
  `sideboard: { 'SB: Counterspell': 1 }`. Silent in the worst way — the paste box counted it in
  the "Sideboard: N cards" readout, then the server dropped it as unknown. Introduced by this
  branch's section-aware rescue. `ParseResult.errors` now carries `cleaned` alongside `raw`;
  two regression tests pin the `SB:` and `#tag` forms.
- **B2 — a page refresh destroyed a saved sideboard.**
  `web-client/src/components/deckbuilder/DeckbuilderPage.tsx:322` (the `deckId` hydration effect,
  taken on refresh or a deep link) restored everything except `sideboardCards`, and `saveDeck`
  replaces the stored record rather than merging — so the next Save wrote the emptiness back.
  Pre-existing; the branch is what made it load-bearing.
- **I1 — loading an example kept the previous deck's sideboard.**
  `DeckbuilderPage.tsx:1025` replaced every other piece of deck state.
  `handleImport` and `handleNew` both already reset it; this was the odd one out.
- **I2 — the sanitizer was two spot-fixes, not a boundary.**
  `GamePlayHandler.handleCreateGame` / `handleJoinGame` passed `message.sideboard` straight into
  `addPlayer`, i.e. into the `requireCard` throw the sanitizer exists to prevent. Latent (no
  current client populates those fields) but reachable from a hand-built frame. Both now route
  through the same helper.

### Open — needs a decision (I3)

**Nothing enforces CR 100.4a on any paste path.** Verified against
`MagicCompRules_20260808.txt:206`: "a sideboard may contain no more than fifteen cards. The
four-card limit applies to the combined deck and sideboard."

- Lobby paste (`deckPasteText.ts:29`) aggregates raw — `parseDeckText('Sideboard\n20 Lightning Bolt')`
  returns 20 copies.
- Deckbuilder import calls `resolveAgainstCatalog` separately per board, each capping at 4
  independently, so 4 + 4 = 8 combined.
- `DeckValidator` has no sideboard handling at all (zero references), and `SideboardSanitizer`
  only checks name-known and `count > 0`.

The sharp edge is that the deckbuilder's *manual* add path **does** enforce the combined cap
(`DeckbuilderPage.tsx:746`, with the CR citation in the comment above it) — so the paste paths
contradict the UI sitting next to them, and a wish fetching from a 20-copy sideboard is a real
gameplay illegality that the lobby is now the shortest route to.

Recommended shape: enforce server-side, where the client isn't authoritative — cap each name at
`4 - deckCount[name]` (basics exempt) and truncate past 15, reporting both alongside `dropped`.
Worth noting `SideboardSanitizer` only ever sees *constructed* sideboards (Limited derives
pool − maindeck inside `TournamentLobby`, and can legitimately exceed 15), so the cap is safe to
put there. Left out of this branch because it's a deck-legality feature and a truncate-or-reject
product decision, not part of making paste work.

### Checked and fine

Worth recording, since these were the plausible failure modes: no resubmit loop between the
debounced client emit and the server's sanitize (the lobby view carries no sideboard, so the
client has nothing to read back, and `sanitize` is idempotent); no render/emit loop from
`currentSideboard`'s new `parsedPaste` dependency (every dep is referentially stable);
`serializeDeck` is sorted and filtered, so the dedupe key is order-independent; Momir Basic and
random-pool paths correctly force an empty sideboard; `LinkedHashMap` preserves submission order
in `kept`, and `hasCard` is the same lookup `requireCard` uses, so there are no false keeps.

One theoretical gap not worth fixing now: `SideboardSanitizer` calls `hasCard` on the raw map key,
while `GameSession.startGame` splits `Name#SET-CN` into name + printing before the engine resolves
it. All four current producers emit plain names, so a printing-qualified sideboard key can't occur
today — but if one ever does, it fails silently in the drop direction.

### Verification

`just test-server` — BUILD SUCCESSFUL, 507 passed, 0 failed (the 19m wall time is box contention,
not the diff). `tsc --noEmit` clean; vitest 45 files, 658 tests. **Not covered:** no engine gate
(nothing in `rules-engine` changed); no e2e for the paste → lobby → `addPlayer` → `Zone.SIDEBOARD`
round trip; the cloud save/load round trip is asserted by inspection only; and the two deckbuilder
fixes (B2, I1) have no test — those paths need a mounted component and there's no harness here.

PR #2030, opened 2026-08-25
Check state-based actions after a land is played

bolav:
# Check state-based actions after a land is played

Playing a second copy of a legendary land you already control left **both** copies on the
battlefield, permanently. No prompt, nothing put into the graveyard, and no self-correction on any
later priority pass.

## What was wrong

Playing a land is a special action (CR 116.2a): it uses no stack, so it is the one route onto the
battlefield that never passes through spell resolution. `PlayLandHandler` ran no state-based-action
check, and none of the other five SBA call sites cover a land drop — `TurnManager`,
`StepActionHelper`, `ConcedeHandler`, `SubmitDecisionHandler`, `PassPriorityHandler`.

The visible consequence was the legend rule. Note the half that was already correct: **playing the
second copy is legal.** CR 704.5j is a state-based action, not a play restriction —

> **704.5j** If two or more legendary permanents with the same name are controlled by the same
> player, that player chooses one of them, and the rest are put into their owners' graveyards. This
> is called the "legend rule."

— so the land enters and the game *then* makes you choose. What was missing is that second half.

Casting a second legendary **spell** was never affected, because `PassPriorityHandler` checks SBAs
after resolution. That is why `LegendRuleTest` passed the whole time and why this went unnoticed: the
existing coverage exercises the cast path exclusively.

## Scope

Not specific to any recent card. **22 legendary lands already in the corpus** had this — Tolarian
Academy, Gaea's Cradle, Eiganjo Castle, Shizo, Okina, Phyrexian Tower and the rest of the Kamigawa
and Urza's cycles.

## The change

`PlayLandHandler.execute` had several success return paths, so rather than patching each, the
existing body becomes `executePlay` and `execute` wraps it: on success, run
`sbaChecker.checkAndApply`, propagating a pause so the legend rule's "which one do you keep?"
decision resolves against the real board with both copies on it. Nothing else about the action
changes, and no new SDK vocabulary is introduced.

## Rules cited

Verified against the local `MagicCompRules_20260808.txt` (effective 2026-08-07):

- **CR 704.5j** — the legend rule, quoted above, and a state-based action rather than a restriction.
- **CR 704.3** — state-based actions are checked whenever a player would receive priority.
- **CR 116.2a** — playing a land is a special action that doesn't use the stack.

## Verification

`scripts/gradle-locked` run **serially, one module per invocation** — `just test` fans the per-era
test modules out in parallel and starves them into `TestHangGuard` timeouts on this box. All 14
modules green:

```
sdk · rules-engine · mtg-sets · game-server · ai
era-1993-1999 · 2000-2002 · 2003-2007 · 2008-2016 · 2017-2022 · 2023 · 2024 · 2025 · 2026
```

The full suite matters more than usual here: every land drop in the codebase now runs an SBA check,
so the blast radius is wide even though the diff is 35 lines.

`LegendRuleLandDropTest` adds three cases — the cull itself, a duplicate *nonlegendary* land raising
no prompt (the check must not make ordinary land drops noisier), and two players each controlling one
copy (the rule is per-player). Mutation-checked: short-circuiting the new call reddens only the
legend-rule case and leaves both controls green. The four pre-existing `LegendRuleTest` cases and
`GhaltaStampedeTyrantLegendRuleTest` still pass.

The fixture is Eiganjo Castle (CHK), deliberately an existing corpus card, so this branch is
independent of any in-flight card work.

**What this does not cover.** No TypeScript or web-client tooling was run — the client side is
unverified by any tool. `manual-scenarios/bugs/legend-rule-land-drop.json` is validated only as JSON
with its card names resolved against the registry; it was not booted against a running server.

### DIFF
--- a/manual-scenarios/bugs/legend-rule-land-drop.json
+++ b/manual-scenarios/bugs/legend-rule-land-drop.json
@@ -1,0 +1,36 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Eiganjo Castle", "Plains"],
+    "battlefield": [
+      {"name": "Eiganjo Castle"},
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Mirri, Cat Warrior"}
+    ],
+    "graveyard": [],
+    "library": ["Plains", "Grizzly Bears", "Plains"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Eiganjo Castle"},
+      {"name": "Centaur Courser"},
+      {"name": "Forest"},
+      {"name": "Forest"}
+    ],
+    "graveyard": [],
+    "library": ["Forest", "Hill Giant", "Forest"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": ["DECLARE_ATTACKERS", "DECLARE_BLOCKERS"],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "AI"
+}
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/actions/land/PlayLandHandler.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/actions/land/PlayLandHandler.kt
@@ -172,6 +172,11 @@ class PlayLandHandler(
         val played = executePlay(state, action)
         if (!played.isSuccess) return played
 
+        // Ordering: the land's own ETB triggers (landfall) are already detected and on the stack
+        // by this point — `executePlay` does that before returning — which is the order CR 603.10
+        // wants and the one PassPriorityHandler arranges explicitly for the resolution path. Moving
+        // this check earlier would drop a landfall trigger whose permanent the legend rule then
+        // removes, and no test here would notice.
         val sbaResult = sbaChecker.checkAndApply(played.state)
         // An SBA that needs input (the legend rule's "which one do you keep?") pauses the action;
         // the land is already on the battlefield, so the decision resolves against the real board.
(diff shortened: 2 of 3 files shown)


bolav:
## Review — `legend-rule-on-land-drop`

**Verdict.** Correct fix in the right place, with a genuinely discriminating test. No new SDK
vocabulary, so the elegance lens has nothing to object to. **No blocking or important findings.**

Reviewed in place on the branch; `origin/main` is already an ancestor, so no merge was needed and
there is no merge commit.

### What's good

- **The diagnosis is the interesting part, and it's right.** The obvious reading — "the engine let me
  play a second legendary land" — is not the bug; CR 704.5j is a state-based action, not a play
  restriction, so that was correct behaviour. The actual defect is one layer down and easy to miss.
  Both the commit message and the PR body lead with that distinction instead of the surface symptom.
- **The wrap is the right shape.** `execute` has several success return paths; wrapping as
  `executePlay` catches all of them, where patching each would have been four chances to miss one.
- **Non-regression is tested, not asserted.** Two of the three cases exist purely to prove the new
  check doesn't make ordinary land drops noisier or break the per-player scope of the rule. Under the
  mutation those two stay green while only the legend-rule case reddens, so the test binds to this
  fix specifically.
- **The fixture choice keeps the branch independent.** Eiganjo Castle is a corpus card, so this
  reviews and merges without reference to any in-flight card branch.
- **The gate matches the blast radius.** 35 lines, but every land drop in the codebase now runs an
  SBA check — the full 14-module serial suite is the proportionate response, not overkill.

### Findings

None blocking, none important.

**M1 — `sbaChecker` is a sixth constructor parameter.** `PlayLandHandler.kt:61`. The handler now
takes six collaborators and the list is getting long. Not worth churn in this PR — the codebase
threads `EngineServices` fields into handlers positionally throughout, and `PassPriorityHandler` does
exactly the same thing with the same dependency — but if handler construction is ever tidied,
these are the call sites.

**M2 (fixed).** The ordering question was decided by silence. `PassPriorityHandler` deliberately detects
triggers *before* SBAs and queues them beneath the SBA continuation (CR 603.10), with a comment
explaining why. `PlayLandHandler` already processes the land's own ETB triggers before the new SBA
call, so the ordering happens to be equivalent — but nothing in the file says so, and a future reader
moving the `sbaChecker` call earlier would break landfall-into-legend-rule without any test noticing.
Fixed before opening: a comment at `PlayLandHandler.kt:175` now records that triggers are
intentionally already on the stack by that point.

### Checks run

Full serial gate green across all 14 modules. Mutation check confirms the new test fails without the
fix. Pre-existing `LegendRuleTest` (4 cases) and `GhaltaStampedeTyrantLegendRuleTest` unaffected. No
card definitions touched, so `check-card-printing` and the snapshot goldens are not in scope.

### Recommendation

Merge as-is. M1 is a note for whenever handler wiring gets revisited, not for this PR.

PR #2031, opened 2026-08-25
Add the five NEO channel lands and Divide by Zero

bolav:
# Add the five NEO channel lands and Divide by Zero

Six cards, plus the two engine gaps that were blocking them. Both gaps are the same shape:
a modifier that existed on one code path but was never wired into the parallel one.

## The cards

**The Kamigawa: Neon Dynasty channel land cycle** — Boseiju, Who Endures; Eiganjo, Seat of the
Empire; Otawara, Soaring City; Sokenzan, Crucible of Defiance; Takenuma, Abandoned Mire. All five
are rare legendary lands sharing one shape: tap for their colour, plus an ability activated by
discarding the land *from hand*, costing {1} less per legendary creature you control. NEO is the
earliest real printing for all five.

**Divide by Zero** (STX #41) — `{2}{U}` instant, bounce a spell or permanent with mana value 1 or
greater, then Learn. STX is the earliest real printing.

Channel is an **ability word** (CR 207.2c: ability words "have no special rules meaning and no
individual entries in the Comprehensive Rules"), so it is not modelled as a keyword. It lives in
`oracleText` and nowhere else, exactly as Landfall and Magecraft do on their cards — the repo has no
ability-word mechanism and this change doesn't add one. Each channel ability is an ordinary
`activatedAbility` with `activateFromZone = Zone.HAND` and `Costs.DiscardSelf`, the shape Steel
Wrecking Ball already uses.

Everything else composes existing primitives: `Effects.DealDamage`, `Effects.ReturnToHand`,
`Patterns.Library.mill` plus the Gather → Select → Move pipeline, `Effects.CreateToken` with the
haste grant iterating `CREATED_TOKENS`, and the Assassin's Trophy shape (`ForEachPlayer` +
`Player.ControllerOf`) for Boseiju's consolation fetch.

## Engine gap 1 — cost reduction never reached zone-activated abilities

`ActivatedAbility.genericCostReduction` was honoured in exactly two places: the battlefield
enumerator and `ActivateAbilityHandler`. `ZoneActivatedAbilityEnumerator` — the one that surfaces
abilities activated from hand and graveyard — gated affordability on the **unreduced** cost. With two
legendary creatures out and `{1}{U}` available, Otawara's channel was never offered as a legal
action, even though the handler would have charged the reduced price. Unreachable in a real game.

The battlefield enumerator's three private helpers are extracted into a shared
`AbilityCostReduction` and both enumerators route through it. The zone enumerator also built its
displayed cost and auto-tap preview from the printed `ability.cost`, ignoring `effectiveCost`
entirely; both now describe what the player will actually pay.

## Engine gap 2 — `TargetSpellOrPermanent` couldn't filter its stack half

The requirement carried only a `permanentFilter`; every spell on the stack was a legal target
regardless of what the card said. That is right for the 22 printed cards in this shape that don't
restrict the stack side (the lace effects, Artificial Evolution, Venser, Shaper Savant), and wrong
for the two that do — Divide by Zero, and Aether Gust ("target spell or permanent that's red or
green"), each applying one restriction to both halves.

A `spellFilter` is added alongside, honoured in the three places the requirement is read for
legality: `TargetValidator`, `TargetFinder`, `TargetEnumerationUtils`. Null keeps today's behaviour,
so no existing card's snapshot moved.

## New SDK vocabulary

Four additions, each a named recipe over existing primitives rather than a new type:

- `TargetSpellOrPermanent.spellFilter` — genuinely new capability; the stack half had no filter at
  all, so composition couldn't express it.
- `TargetSpellOrPermanent.descriptionOverride` — the escape hatch `TargetPermanentOrPlayer` already
  carries. A both-halves restriction is printed once but passed to each filter separately, so the
  generated text otherwise repeats it per side.
- `DynamicAmounts.legendaryCreaturesYouControl()` — `battlefield(You, Creature.legendary()).count()`,
  named because five cards in this PR alone wanted it. Sits beside `creaturesYouControl()`.
- `GameObjectFilter.LandWithBasicLandType` — the five-way basic-type union, named because "a land
  card with a basic land type" is not the same set as `BasicLand` (a shockland qualifies).

`docs/card-sdk-language-reference.md` is updated for all four.

## Rules cited

Verified against the local `MagicCompRules_20260808.txt` (effective 2026-08-07):

- **CR 207.2c** — ability words, listing `channel` among them, with no rules meaning.
- **CR 202.3b** — a spell's mana value on the stack uses the chosen value for `{X}`. See below.

## Verification

`scripts/gradle-locked` run **serially, one module per invocation** — `just test` fans the per-era
test modules out in parallel and starves them into `TestHangGuard` timeouts on this box. All 14
modules green on the final tree, after the review fixes:

```
sdk · rules-engine · mtg-sets · game-server · ai
era-1993-1999 · 2000-2002 · 2003-2007 · 2008-2016 · 2017-2022 · 2023 · 2024 · 2025 · 2026
```

Also run: `just rebless-cards` (only NEO and STX moved, pure additions plus the one
`descriptionOverride` line — no existing card shifted); `just check-card-printing` green for all six;
`just assay-differential --set STX` 0 divergent, `--set NEO` 1 divergent which is the pre-existing
Experimental Synthesizer, not from this branch.

Both engine fixes were mutation-checked: reverting each reddens exactly its own test and leaves the
controls green.

**What this does not cover.** No TypeScript or web-client tooling was run — `node_modules` is absent
and the network is firewalled here, so the client side of this change is unverified by any tool. The
client reasoning (the ability's menu label comes from `action.description`, which now carries the
reduced cost) is from reading `ActionMenu.tsx`, not from running it. No scenario was booted against a
running game server; the six dev scenarios under `manual-scenarios/` are validated only as JSON with
every card name resolved against the registry.

## Deliberately left out

- **`{X}` spells read mana value 0 on the stack.** `CardPredicate.ManaValueAtLeast` reads the card's
  mana value, not the value chosen for `{X}` (CR 202.3b), so a spell whose cost is purely `{X}` is
  wrongly untargetable by Divide by Zero. Every other `{X}` spell has a non-X pip and clears "1 or
  greater" either way. Documented in the card's KDoc; fixing it means threading the stack object's
  chosen X into the predicate evaluator, which is its own change.
- **`ReselectTargetRandomlyExecutor`** still ignores both filters on this requirement. It already
  ignored `permanentFilter` before this branch; narrowing only the new half would be inconsistent.
- **Aether Gust** is now expressible but not implemented — it belongs in its own card PR.
- **Otawara's scenario has no planeswalker** on the opposing board, so it exercises three of the four
  target types.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -3767,6 +3767,10 @@ Each side carries its own optional `GameObjectFilter`, and either may be left nu
 | `permanentFilter` | the battlefield half | `GameObjectFilter.Creature` → "target spell or creature" (Swat Away) |
 | `spellFilter` | the stack half | `GameObjectFilter.Any.manaValueAtLeast(1)` → the stack half of Divide by Zero |
 
+`descriptionOverride` supplies the printed wording when the generated one reads badly — the same
+escape hatch `TargetPermanentOrPlayer` carries. A restriction covering *both* halves is printed once
+but has to be passed to each filter separately, so the generated text otherwise repeats it per side.
+
 Most printed cards in this shape restrict neither side or only the permanent one. A card whose
 single restriction applies to *both* halves spells it by passing the same filter to each:
 
@@ -3776,7 +3780,8 @@ val t = target(
     "target spell or permanent with mana value 1 or greater",
     TargetSpellOrPermanent(
         permanentFilter = GameObjectFilter.Permanent.manaValueAtLeast(1),
-        spellFilter = GameObjectFilter.Any.manaValueAtLeast(1)
+        spellFilter = GameObjectFilter.Any.manaValueAtLeast(1),
+        descriptionOverride = "target spell or permanent with mana value 1 or greater"
     )
 )
 effect = Effects.ReturnSpellOrPermanentToOwnersHand(t) then Patterns.Mechanic.learn()
@@ -4110,6 +4115,10 @@ This is the player-arm prerequisite for the planned composable mixed `TargetUnio
   the land base (`GameObjectFilter.Land.nonbasic()`), or use the named constant `GameObjectFilter.NonbasicLand`
   / `TargetFilter.NonbasicLand` (Rocket Volley, Shivan Harvest, Encroaching Wastes). `TargetFilter.Land.nonbasic()`
   is the target-side passthrough.
+- `GameObjectFilter.LandWithBasicLandType` — a land with one of the five basic land types
+  (CR 205.3i): "a land card with a basic land type" (Boseiju, Who Endures). Deliberately *not*
+  `GameObjectFilter.BasicLand` — a shockland (`Land — Forest Island`) or Dryad Arbor qualifies, and a
+  basic is only the common case.
 - `.withChosenColor()` — `CardPredicate.HasChosenColor`: matches the color chosen during the current
   effect's resolution (read from `EffectContext.chosenColor`, set by `Effects.ChooseColorThen`). Use with
   `AggregateBattlefield(Player.Each, …)` for "for each permanent of that color" (Coalition Dragon cycle).
@@ -10549,6 +10558,16 @@ both spellings, and the ability its bare-noun line grants says "Regenerate this 
   as a `ForEachInGroup` over `Creature.youControl().powerGreaterThanBase()` reading the amount off
   `EntityReference.IterationEntity`. Not projected — always the printed base.
 - `CardNumericProperty(card, property)` — generic numeric property accessor.
+
+### Board-count shortcuts (`DynamicAmounts.*` facades)
+
+`DynamicAmounts.creaturesYouControl()`, `.landsYouControl()`, `.otherCreaturesYouControl()`,
+`.attackingCreaturesYouControl()`, `.equipmentYouControl()` — each a named `battlefield(…).count()`.
+
+`DynamicAmounts.legendaryCreaturesYouControl()` is the same shape over
+`GameObjectFilter.Creature.legendary()`: the Kamigawa channel lands' "costs {1} less to activate for
+each legendary creature you control", fed to `ActivatedAbility.genericCostReduction`. Counts creatures
+only — a legendary land or artifact does not qualify.
 
 ### Triggering-entity shortcuts (`DynamicAmounts.*` facades)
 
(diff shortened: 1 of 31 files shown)


bolav:
## Review — `channel-lands-divide-by-zero`

**Verdict.** Behaviour is right and the SDK shape is right: no new `Effect`, executor, condition or
keyword, and the two engine changes both close a gap where an existing modifier reached one code
path but not its sibling. Four findings from this pass were fixed in `976246e954` before opening;
they are recorded below with what changed. Nothing outstanding is blocking.

Reviewed in place on the branch; `origin/main` is already an ancestor (0 commits behind), so no merge
was needed and there is no merge commit.

### What's good

- **The two engine fixes are deduplications, not additions.** `AbilityCostReduction` is the
  battlefield enumerator's own logic hoisted, so the zone path can't drift from it again — the
  category of bug being fixed. Same for `spellFilter`: it mirrors the `permanentFilter` that was
  already there rather than inventing a parallel mechanism.
- **Channel is modelled as what it is.** Treating an ability word as an `oracleText`-only concern,
  matching Landfall and Magecraft, avoids a keyword flag that CR 207.2c says has no rules meaning —
  and keeps the reduced cost visible in the menu, which a static `descriptionOverride` would have
  frozen.
- **Sokenzan doesn't take the easy approximation.** Baking `Keyword.HASTE` into the token would test
  green and be wrong on any later turn the token changes control; iterating `CREATED_TOKENS` grants
  it until end of turn to exactly the two tokens made.
- **The tests discriminate.** `ZoneAbilityCostReductionEnumerationTest` includes a no-blue-source
  case, so it can't pass by accident, and both engine fixes were mutation-checked.

### Fixed during review (all in `976246e954`)

**I1 — generated target description repeated the restriction per half.**
`TargetRequirement.kt:398`. With both filters set, Divide by Zero rendered as *"target with mana
value 1 or greater spell or permanent with mana value 1 or greater"* — the string the player reads
when choosing a target. Fixed by adding the `descriptionOverride` that `TargetPermanentOrPlayer`
already carries twenty lines above (`TargetRequirement.kt:304`), and giving the card its printed
wording. The generated fallback also now reads `"spell <filter>"` rather than `"<filter> spell"`.

**I2 — five cards repeated the same dynamic amount.**
`battlefield(Player.You, GameObjectFilter.Creature.legendary()).count()` appeared verbatim in all
five lands. Five call sites in a single PR is past the threshold for a named recipe; extracted to
`DynamicAmounts.legendaryCreaturesYouControl()` (`DynamicAmounts.kt:175`), beside the existing
`creaturesYouControl()` / `landsYouControl()`.

**M1 — stringly-typed basic land types.** `BoseijuWhoEndures.kt` spelled the five-way union with
string literals while `Subtype.PLAINS`… exist. Hoisted to `GameObjectFilter.LandWithBasicLandType`,
built from the constants and documented against `BasicLand`, which it deliberately is not — a
shockland or Dryad Arbor qualifies.

**M2 — this branch introduced its own duplication.** The "rebuild the label only if the cost moved
and there's no custom text" rule ended up in both enumerators. Moved to
`AbilityCostReduction.describe`. Note the subtlety that caught a would-be regression: the battlefield
enumerator compares against its *text-replaced* cost, not the printed one, so the baseline is a
parameter rather than hardcoded to `ability.cost`.

### Remaining, non-blocking

**M3 — `{X}` spells read mana value 0 on the stack.** `PredicateEvaluator.kt:624` reads
`card.manaValue`, so CR 202.3b's "use the value chosen for X" doesn't apply to a stack spell. A spell
costing purely `{X}` is wrongly untargetable by Divide by Zero. Narrow (every other `{X}` spell has a
non-X pip), documented in the card's KDoc, and fixing it means threading the chosen X into the
predicate evaluator — its own change, not this one.

**M4 — `ReselectTargetRandomlyExecutor.kt:179`** ignores both `permanentFilter` and `spellFilter`.
Pre-existing for the former; called out here so it isn't mistaken for an oversight of this branch.

### Checks run

`just check-card-printing` green for all six cards; canonical placement is the earliest real printing
in both sets. Goldens re-blessed with only NEO and STX moving and no existing card shifting.
`just assay-differential` — STX 0 divergent, NEO's 1 divergence is the pre-existing Experimental
Synthesizer. Full serial gate green across all 14 modules on the final tree.

### Recommendation

Merge as-is. If the two engine fixes want to land independently of six cards, they split cleanly:
`dfdea0669b` (cost reduction) and `4020037e10` (`spellFilter`) touch disjoint files from each other
and from the cards.

PR #2035, opened 2026-08-25
Stop the game log naming face-down objects

bolav:
# Stop the game log naming face-down objects

A creature cast with disguise announced itself the moment it resolved. The cast line was
masked; the four lines around it were not:

```
Opponent cast Face-down creature
Aurelia's Vindicator resolved                                       <- leak
Opponent's Aurelia's Vindicator entered the battlefield             <- leak
You cast Igneous Inspiration targeting Aurelia's Vindicator         <- leak
Opponent's Aurelia's Vindicator triggered: ... unless ... pays {2}  <- leak
```

That transcript is from a real playtest game. By the time the Vindicator's ward countered
the removal spell, the log had named the card four times.

A face-down permanent has no name (CR 708.2a) and neither does a face-down spell
(CR 708.4), so none of those lines may carry the printed name of the card underneath.

## Why the fix is where the text is built

`ClientEventTransformer` is where log text is assembled, and it is already per-viewer — but it
maps engine events to strings with no `GameState` in hand, so it cannot tell a face-down object
from a face-up one. Every decision therefore has to be made where the text is built, which is what
the pre-existing cast-line mask already did.

Those scattered decisions now route through one predicate in `state/FaceDownNaming.kt`. That
matters more than it looks, because **all three public zones answer differently**:

- **Stack** — a spell cast face down carries no `FaceDownComponent`; that is stamped as the spell
  resolves. Its status rides `SpellOnStackComponent.castFaceDown` instead. My first attempt checked
  only the component and left the stack case leaking; the test caught it.
- **Battlefield** — `FaceDownComponent`, but checked against the *physical* zone.
  `GameState.getBattlefield()` deliberately hides phased-out permanents, and phasing never moves a
  permanent out of the zone (CR 702.26d), so a phased-out face-down permanent must stay masked.
  The memoized accessor still handles the fast path; the unmemoized physical scan is reached only
  for phased-out entities.
- **Exile** — `FaceDownComponent` again, but the object is not a creature, so it reads
  **"Face-down card"** — the name its own card view already uses.

Anything that has left all three keeps its real name: its owner reveals it to every player as it
goes (CR 708.9).

## Who the mask applies to

**Text with no audience** — game-log lines, decision summaries — masks for both players, as the
cast line always has. The object genuinely has no name, and a controller who wants to know what
their own face-down permanent is looks at the card (CR 708.5).

**Text with an audience** uses `nameVisibleTo`, which lets a player read their own face-down object
and treats a spectator as never entitled to. This is the gate `ClientStateTransformer` already
applies to card views, so a label and the card it labels always agree — and it keeps modal spells
consistent with non-modal ones, whose targets ship as bare entity ids and are named client-side off
that same per-viewer card view. Threading it through cost one extra parameter on
`transformAbilityOnStack` and `buildPerModeTargetGroups`.

## Sites

| File | Text it fixes |
|---|---|
| `StackResolver` `targetNames` (permanent, spell, card) | `cast X targeting <face-down object>` |
| `StackResolver` `ResolvedEvent` | `Aurelia's Vindicator resolved` |
| `StackResolver` `ZoneChangeEvent` | `Opponent's ... entered the battlefield` |
| `StackResolver` `putTriggeredAbility` | `Opponent's ... triggered: ...` |
| `TriggerDetector` `descriptionOverride` ×2 | names interpolated into Evoke / Decayed trigger text |
| `SubmitDecisionHandler` ×2 | `Targeting X`, `Distributed: N to X` |
| `ClientStateTransformer.resolveTargetDisplayName` | stack-view per-mode target labels |
| `Grant{Hexproof,Protection}FromChosenColorExecutor` | keyword-granted event target names |

The two `TriggerDetector` sites are worth calling out: they interpolate the card name *into the
description string*, so masking `AbilityTriggeredEvent.sourceName` downstream would have left the
name sitting in the text beside it. Evoke can't apply to a face-down cast; Decayed can.

No new SDK vocabulary — this is one internal predicate plus a shared constant, no `Effect`,
trigger, or DSL surface.

## Blast radius

Every field touched is display-only. `CommitCrimeEvent`'s matcher
(`TriggerMatcher.kt:377`) reads only `playerId`, never `sourceName`, and both
`CommitCrimeEvent` and `TargetsChosenEvent` map to `null` in `ClientEventTransformer` — so
those two edits change nothing observable today and are defensive only. `SpellCastEvent
.targetNames` has no consumer outside the log. No trigger, matcher, or state read touches
any of it.

The one client-side check keyed on the literal `'Face-down creature'`
(`web-client/src/store/slices/handlers/gameplayHandlers.ts:238`) is gated on
`type === 'spellCast'` — the one line this PR does not change.

## Rules cited

Verified against `MagicCompRules_20260808.txt`, not from memory:

- **708.2a** — a face-down permanent is a 2/2 with "no text, no name, no subtypes, and no
  mana cost". This is the rule for *no name*; 708.2 is the general characteristics rule.
- **708.4** — objects cast face down are turned face down before going on the stack, so
  effects see only the face-down spell's characteristics.
- **708.5** — you may look at a face-down permanent you control.
- **708.9** — a face-down permanent moving off the battlefield is revealed to all players.
- **702.168a / 701.58a** — disguise and cloak both list ward {2} among the face-down
  characteristics. This is why the Vindicator's ward fired at all: the printed ward {2} is
  a face-*up* keyword and does not apply. `TriggerAbilityResolver.getWardTriggeredAbilities`
  already took the right branch; that behavior was correct before this PR and is unchanged.

## Verification

`scripts/gradle-locked :rules-engine:test` — **3116 test cases, 0 failures, 1m 37s.** Confirmed the
suite actually ran rather than being served up-to-date: all 532 result XMLs rewritten at the run
timestamp.

Gate chosen by what the diff reaches: all nine files are under `rules-engine/`. `test-rules` would
add the `:mtg-sets:scenarioTest` fan-out, which nothing here can affect — of the 19 scenario tests
that use `castFaceDown`, not one references `ClientEvent` or `ClientEventTransformer`; they assert
on state, not log strings.

Nine tests in `FaceDownGameLogMaskingTest` cover: the four original log lines, a counterspell
naming a face-down spell, the decision summary, the per-viewer split (controller / opponent /
spectator), and three zone-boundary cases — battlefield, phased-out battlefield, exile, and a
face-down card in hand that must *keep* its name so the mask can't drift into over-masking.

**What this does not cover:**

- **No TypeScript ran.** `node_modules` is absent and the network is firewalled, so the web-client
  is unverified by any tool. The reasoning below about `gameplayHandlers.ts` is a code read, not a
  test result.
- **No game-server run.** `ClientStateTransformer` is used by `:game-server`, and this PR changes
  two of its private signatures (adding `isSpectator`). Both are private with call sites inside the
  same file, so nothing outside the class can break — but `:game-server:test` was not run.

## Deliberately left out

Nothing from the review round remains open. The earlier draft of this PR deferred face-down cards
in exile, the three duplicated string literals, and the helper's placement; all three are now in.

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/event/TriggerDetector.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/event/TriggerDetector.kt
@@ -25,6 +25,7 @@ import com.wingedsheep.engine.handlers.PredicateContext
 import com.wingedsheep.engine.handlers.PredicateEvaluator
 import com.wingedsheep.engine.registry.CardRegistry
 import com.wingedsheep.engine.state.GameState
+import com.wingedsheep.engine.state.nameVisibleToAll
 import com.wingedsheep.engine.state.components.battlefield.AttachedToComponent
 import com.wingedsheep.engine.state.components.battlefield.ClassLevelComponent
 import com.wingedsheep.engine.state.components.battlefield.CountersComponent
@@ -670,7 +671,11 @@ class TriggerDetector(
                 effect = com.wingedsheep.sdk.dsl.Effects.SacrificeTarget(
                     com.wingedsheep.sdk.scripting.targets.EffectTarget.Self
                 ),
-                descriptionOverride = "Evoke — Sacrifice ${cardComponent.name}"
+                // The name is interpolated into the text, so the mask has to be applied here —
+                // masking AbilityTriggeredEvent.sourceName downstream would leave it in the
+                // description. Evoke can't apply to a face-down cast today; Decayed below can.
+                descriptionOverride =
+                    "Evoke — Sacrifice ${nameVisibleToAll(state, event.entityId, cardComponent.name)}"
             )
             triggers.add(
                 PendingTrigger(
@@ -717,7 +722,8 @@ class TriggerDetector(
                         com.wingedsheep.sdk.scripting.targets.EffectTarget.Self
                     )
                 ),
-                descriptionOverride = "Decayed — When ${cardComponent.name} attacks, " +
+                descriptionOverride = "Decayed — When " +
+                    "${nameVisibleToAll(state, attackerId, cardComponent.name)} attacks, " +
                     "sacrifice it at end of combat."
             )
             triggers.add(
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/actions/decision/SubmitDecisionHandler.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/actions/decision/SubmitDecisionHandler.kt
@@ -8,6 +8,7 @@ import com.wingedsheep.engine.core.EngineServices
 import com.wingedsheep.engine.handlers.actions.ActionHandler
 import com.wingedsheep.engine.mechanics.StateBasedActionChecker
 import com.wingedsheep.engine.state.GameState
+import com.wingedsheep.engine.state.nameVisibleToAll
 import com.wingedsheep.engine.state.components.identity.CardComponent
 import com.wingedsheep.sdk.core.Step
 import kotlin.reflect.KClass
@@ -359,8 +360,11 @@ class SubmitDecisionHandler(
             }
 
             pending is ChooseTargetsDecision && response is TargetsResponse -> {
+                // A face-down permanent or spell has no name (CR 708.2a / 708.4) — naming it here
+                // would leak a morphed or disguised card the moment somebody targeted it.
                 val targetNames = response.selectedTargets.values.flatten().mapNotNull { targetId ->
                     state.getEntity(targetId)?.get<CardComponent>()?.name
+                        ?.let { nameVisibleToAll(state, targetId, it) }
                         ?: if (state.turnOrder.contains(targetId)) "player" else null
                 }
                 if (targetNames.isNotEmpty()) {
@@ -371,6 +375,7 @@ class SubmitDecisionHandler(
             pending is DistributeDecision && response is DistributionResponse -> {
                 val parts = response.distribution.mapNotNull { (targetId, amount) ->
                     val name = state.getEntity(targetId)?.get<CardComponent>()?.name
+                        ?.let { nameVisibleToAll(state, targetId, it) }
                         ?: if (state.turnOrder.contains(targetId)) "player" else null
                     name?.let { "$amount to $it" }
                 }
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/effects/permanent/protection/GrantHexproofFromChosenColorExecutor.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/handlers/effects/permanent/protection/GrantHexproofFromChosenColorExecutor.kt
@@ -8,8 +8,8 @@ import com.wingedsheep.engine.mechanics.layers.Layer
 import com.wingedsheep.engine.mechanics.layers.SerializableModification
 import com.wingedsheep.engine.mechanics.layers.addFloatingEffect
 import com.wingedsheep.engine.state.GameState
+import com.wingedsheep.engine.state.nameVisibleToAll
 import com.wingedsheep.engine.state.components.identity.CardComponent
-import com.wingedsheep.engine.state.components.identity.FaceDownComponent
 import com.wingedsheep.sdk.scripting.effects.GrantHexproofFromChosenColorEffect
 import kotlin.reflect.KClass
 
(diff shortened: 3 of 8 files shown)


bolav:
## Review — `facedown-log-masking`

Reviewed `origin/main...HEAD` in place (branch already current with `origin/main`; the merge was a
no-op, no merge commit). Two rounds; every finding from round one is fixed on the branch and this
file records both the findings and where they landed, so the PR carries the audit trail.

**Verdict:** behavior is right and the shape is right — one predicate owning a decision that was
previously implicit at every call site, which is the correct response to a bug that surfaced at
four sites simultaneously and then at four more under scrutiny. No open issues.

**Standing caveat:** this reviews my own change. The findings below are real, but a second pair of
eyes on the per-viewer decision in particular would be worth having.

### What's good

- **The predicate is the deliverable, not the edits.** `FaceDownNaming.kt` encodes the genuinely
  non-obvious part — that all three public zones store and describe face-down-ness differently —
  and its KDoc says outright not to test `FaceDownComponent` directly because it is "neither
  necessary nor sufficient". That sentence is what stops this recurring.
- **The tests assert an absence.** `log.forEach { it shouldNotContain "Disguised Angel" }` caught
  three sites that a careful code read had missed: the cast-targeting line, the face-down spell on
  the stack, and (indirectly) the description-interpolation class. A test written as "these lines
  should say Face-down creature" would have passed while still leaking.
- **The boundary is pinned in both directions.** A face-down card in hand must *keep* its name;
  a phased-out permanent must *lose* it. Masking bugs get "fixed" into over-masking easily.
- **Rule citations were verified, and one was wrong before it shipped.** A draft of the phasing
  test cited CR 702.25a, which is Flanking; phasing is 702.26, and the "never changes zones" clause
  is 702.26d. Checked against `MagicCompRules_20260808.txt`, as were 708.2a, 708.4, 708.5, 708.9,
  702.168a, 701.58a.

### Round-one findings — all fixed on this branch

**I1 — a phased-out face-down permanent was not masked.** The helper gated on
`state.getBattlefield()`, which deliberately filters `PhasedOutComponent` (`GameState.kt:927–936`);
the accessor whose own KDoc names client rendering as its consumer is `allBattlefieldEntities()`.
Fixed in `FaceDownNaming.kt` with the memoized accessor still on the fast path, so the unmemoized
scan is reached only for phased-out entities. Covered by "a phased-out face-down permanent is still
masked", which asserts the entity is absent from `getBattlefield()` *and* still masked.

**I2 — the mask covered `sourceName` but not trigger descriptions.** `AbilityTriggeredEvent` carries
both, and `TriggerDetector` interpolates the card name straight into two `descriptionOverride`
strings (Evoke, Decayed). Masking the name field downstream would have left the name in the text
beside it. Both sites now build their text through the helper. Decayed is the reachable one — a
face-down permanent can carry a decayed counter; evoke cannot apply to a face-down cast.

**I3 — a card face down in exile was named as a target.** `ChosenTarget.Card` was unmasked in both
`StackResolver` and `resolveTargetDisplayName`, while `ClientStateTransformer` was already drawing
such a card as "Face-down card". Fixed with a second constant rather than by stretching the
creature name over a non-creature object.

**I4 — the helper lived in a pure-data package.** Moved out of `state.components.identity` (where
every other declaration is a `@Serializable` data class) into `state/FaceDownNaming.kt`.

**M1 — modal and non-modal stack labels disagreed for the controller.** Non-modal targets ship as
bare entity ids and are named client-side off the per-viewer card view, so a controller sees their
own face-down permanent's real name; modal targets shipped a flat string masked for everyone.
`nameVisibleTo` now applies the same gate the card view uses, and `isSpectator` is threaded through
`transformAbilityOnStack` → `buildPerModeTargetGroups` → `resolveTargetDisplayName`. The compile
error this produced confirmed the plumbing the round-one review predicted would be needed.

**M2 — four sites hardcoded the string the constant owns.** All converted. The two
`Grant*FromChosenColorExecutor`s were the interesting ones: they tested `container.has
<FaceDownComponent>()` by hand, which round one called "probably correct today". That was too
generous — the hand-rolled check would have named a face-down *spell* wrongly. Routing them through
the helper fixes that incidentally.

**M4 — the decision summaries had no end-to-end test.** Now covered. Getting there took a detour
worth recording: a non-modal spell declares its targets at cast time and never prompts, so the
first attempt found `pendingDecision == null`. A targeted ETB trigger is the reachable path to a
`ChooseTargetsDecision`.

**M3 (round one, not acted on) — `entityId in state.getBattlefield()` is a linear scan**, run once
per triggered ability and once per target per cast. Still true, still negligible at real board
sizes. Left alone deliberately; if a reverse zone index ever exists, this is a caller for it.

### Recommendation

Merge. No follow-up issue needed — the three items round one deferred are all in this PR.

Issue #20, opened 2026-04-17
Discord server for communication?


dustinlacewell:
Hey, do you have any interest in starting a Discord server for discussion of the project?

dustinlacewell:
Vincent I know you can hear me 😅

wingedsheep:
Hey! I added a discord link to the README :)

Issue #19, opened 2026-04-15
Plan for mechanic and card coverage


dustinlacewell:
Hello, awesome project, I'm very curious about it. I was curious what your plan is regarding coverage of the many mechanics and cards? I know that we are vibe-coding here, so I suspect LLMs will be involved in some way. I was just wondering if you already had thoughts about some way to systematize the process somehow.

Happy to theory-craft with you. BTW I messaged you on Twitter 👋

vlad2135:
I'd like to chip in on card coverage. Do you have any plans to automate card addition (if they use existing mechanics, obviously)? I see you are adding one card per commit, it seems slow. I believe these JSONs can be automatically converted to Kotlin code - https://mtgjson.com/downloads/all-sets/ , no?

wingedsheep:
@dustinlacewell 

Thank you! Regarding coverage. I recently started adding a bit more coverage by running kover and seeing where coverage is missing, then asking claude code to look up all the rules regarding this particular effect and then implement tests for those effects.

Not very structured yet. I just created a discord, in case you want to chip in!

I missed your message on twitter, so maybe it went to spam somehow :/

@vlad2135 

 I see you are adding one card per commit, it seems slow. I believe these JSONs can be automatically converted to Kotlin code -> Adding cards requires needs engine coverage of the mechanics. That takes a while to build and get right. Some cards can be added without any new effects, but many still need extra coverage in the engine.

PR #2041, opened 2026-08-26
Add the Mirrodin Besieged Zenith cycle

bolav:
# Add the Mirrodin Besieged Zenith cycle

All five Zeniths, canonical in MBS (their earliest real printing — the PMH-style promos that predate
the set by days are excluded), plus the one piece of card-script vocabulary the cycle needed.

| Card | | |
|---|---|---|
| Green Sun's Zenith | `{X}{G}` Sorcery | #81 |
| Blue Sun's Zenith | `{X}{U}{U}{U}` Instant | #20 |
| White Sun's Zenith | `{X}{W}{W}{W}` Instant | #19 |
| Black Sun's Zenith | `{X}{B}{B}` Sorcery | #39 |
| Red Sun's Zenith | `{X}{R}` Sorcery | #74 |

Green Sun's Zenith is the card that motivated this; the other four are the rest of its cycle and cost
almost nothing once the shared clause works.

## The new vocabulary, and why composition wasn't enough

All five print the same trailing clause — *"Shuffle &lt;card name&gt; into its owner's library."* —
and there was no way to say it.

The tempting composition is `Effects.ShuffleIntoLibrary(EffectTarget.Self)` in the spell effect. It
does not work. `StackResolver` moves the spell off the stack unconditionally after resolution, so the
card would be moved twice and end up in the graveyard anyway. Nor does the existing
`AfterResolveDestination` enum fit: it has `EXILE` and `BOTTOM_OF_LIBRARY` but deliberately not a
shuffle, and it is a *cast-this-way rider* that another effect stamps onto a spell it is casting, not
something a card declares about itself.

The right seam is the one next door. `CardScript.selfExileOnResolve` already models
*"Exile &lt;card name&gt;."* by replacing the destination of **CR 608.2n** — *"As the final part of an
instant or sorcery spell's resolution, the spell is put into its owner's graveyard."* This PR adds its
sibling, `selfShuffleIntoLibraryOnResolve`, reached from the DSL as
`spell { selfShuffleIntoLibrary() }`. That follows the shape `CardScript` already uses for
card-intrinsic destinations — one field per mechanic (`selfExileOnResolve`, `paradigm`,
`returnTransformedFromGraveyardOnResolve`), all feeding one `when` — rather than inventing a fourth
shape.

The two are two spellings of one slot, so setting both is rejected at card-construction time with the
offending card named — the winner would otherwise be decided silently by the order `StackResolver`
checks them in. No existing card trips that guard.

Two things about the wiring are worth a reviewer's attention.

**The destination seam is duplicated.** `StackResolver` decides where a spell goes once on the
full-resolve path and again on the paused-resolve path, and each then places the card at its own site
— four edits, not two. That is load-bearing rather than academic here: Green Sun's Zenith pauses for
its library-search decision and therefore takes the *paused* path, which the other four Zeniths never
reach. Wiring one and not the other fails silently, with the card simply turning up in the graveyard.
A mutation check confirmed the coverage: unwiring only the paused path fails exactly one test and
nothing else.

**The clause outranks the cast-this-way rider**, uniquely among the card-intrinsic destinations, so it
sits above it in both `when` blocks. Those riders read *"if that spell **would be put into a
graveyard**, [somewhere] instead"* — Kylox's Voltstrider, which is implemented in this corpus — and a
spell that shuffles itself into its owner's library never would be, so the rider has nothing to
replace. Without this the interaction is reachable and wrong: a Zenith in a graveyard (put there by
being countered) and then cast off Voltstrider would land on the bottom of the library unshuffled.
On the countered and fizzled paths, where the card really *is* put into a graveyard, the rider still
wins — and it wins for free, because those paths never read this flag.

That last point is also the rules answer the whole cycle turns on. Every Zenith carries the ruling
*"If this spell doesn't resolve, none of its effects occur. In particular, it will go to the graveyard
rather than to its owner's library."* The clause is part of the spell's **effect**, so a spell that is
countered (**CR 701.5a**) or fizzles (**CR 608.2b**) never performs it. That is the exact opposite of
how the cast-this-way rider answers the same two questions, which is why the two mechanisms get
separate test files.

Rule numbers above were checked against `MagicCompRules_20260808.txt` in this repo, not from memory.

## The cards

Four of the five are stock compositions:

- **Green Sun's Zenith** — `Patterns.Library.searchLibrary` pointed at the battlefield. `count = 1` is
  how many cards are found; the `{X}` paid lives in the *filter*, as `manaValueAtMostX()`.
- **Blue Sun's Zenith** — `Effects.DrawCards(DynamicAmount.XValue, …)` on a targeted **player**, not
  "you"; the card is as happy decking an opponent as refilling your own hand.
- **White Sun's Zenith** — the dynamic-count `Effects.CreateToken` overload. **2/2** Cats, not 1/1.
  MBS printed no Cat token of its own, so the art is the one Scryfall links from this printing via
  `all_parts`.
- **Black Sun's Zenith** — Darkness Descends' shape (`ForEachInGroup` over `GroupFilter.AllCreatures`,
  `EffectTarget.Self` rebound per creature) with the fixed count swapped for the dynamic one.
- **Red Sun's Zenith** — Carbonize's shape exactly: `AnyTarget()` plus `MarkExileOnDeathEffect`, the
  CR 614 death replacement lasting the turn. The mark is applied *before* the damage, as on Carbonize,
  so a creature killed by this very spell is exiled rather than only one that survives and dies later.

**Seven `Printing(...)` rows** come with them — Blue in C13/C15, White in C14/C17/CMR, Black in
C14/TLE. `just check-card-printing` fails on a missing reprint row in a scaffolded set exactly as it
fails on a misplaced canonical, and all five now report clean. Scryfall genuinely reports the TLE
printing's artist as `"Viacom"` (a licensed-art set); that was verified against the API rather than
assumed to be a data error.

## Dev scenarios

One `manual-scenarios/cards/…` JSON per card, for exercising them by hand in the client. Each board
is built so the card's own question is the interesting one — Blue's opponent has a 3-card library so
targeting *them* decks them, Black's board has deliberately staggered toughness so X = 2 kills the
caster's own Birds of Paradise, Red has two copies so both a 2/2 and a 6/4 can be checked for landing
in exile rather than the graveyard.

## Verification

- `just test-class SelfShuffleIntoLibraryOnResolveScenarioTest` — **8/8 pass.** Full-resolve path;
  paused-resolve path; shuffled-in rather than bottomed (asserted via `LibraryShuffledEvent`, since
  the resulting position is random); countered → graveyard; fizzled → graveyard; the rider-precedence
  case; the both-clauses-set guard; and a no-clause control.
- **Mutation check** — unwiring only the paused-resolve path failed exactly the paused test and left
  every other test in the file green, which is what proves that path is really covered.
- `just test-class GreenSunsZenithScenarioTest` — **3/3 pass.** Finds a creature within X and returns
  to the library; the X = 0 / Dryad Arbor line; and nothing-within-X, where the search finds nothing
  (CR 701.23b) and the spell still shuffles itself in.
- `just check-card-printing` — clean for all five.
- `just rebless-cards` — **263 insertions, 0 deletions, `MBS.json` only.** `CardSerialization` encodes
  with `encodeDefaults = false`, so the new field is omitted wherever it is false and does not churn
  the rest of the corpus; printings are not snapshotted, so the seven reprint rows produce no golden
  change.
- `just assay-differential --set MBS` — 0 divergences. **This says nothing about the five new cards**:
  only 4 of MBS's 12 hand-written cards are covered by the grammar, and all five Zeniths are among the
  8 declines (Assay has no rule for "Shuffle ~ into its owner's library"). It confirms the four
  pre-existing MBS cards were not disturbed. A decline is not a pass.
- `just test` — run twice. Every suite the diff can reach was green both times: all of `rules-engine`,
  the snapshot and round-trip tests across every set, and the 2008-2016 era where these cards live.

**What the gate does not cover, and one caveat on the full runs.** Neither full `just test` run came
back green overall, but no test has ever failed on an assertion — every red was a timeout or a killed
worker, in a *different* module each run, on a box showing load average 9.4 and a run stretching from
19 to 56 minutes. Run 1: `AbuJafarTest` (1993-1999), 120 s coroutine timeout. Run 2: `AIPlayerTest`
(`:ai`), 120 s timeout, plus `:mtg-sets:2003-2007:tests` failing with *"Test process encountered an
unexpected problem… exit value 13"* and **zero** recorded test failures — a killed worker JVM, not a
failing test. All three were re-run serially and pass: `AbuJafarTest` 3/3 in 25 s, `AIPlayerTest` 14/14
in 30 s, and the 2003-2007 era suite 670 test cases / 0 failures in 28 s.

There is also an argument from the diff itself. The change adds one `CardScript` field defaulting to
`false` and one `StackResolver` branch gated on it. No card outside these five sets that flag, so for
every other card in the corpus both `when` blocks evaluate exactly as before — including the
precedence reorder, which only fires when the flag is true.

Nothing here touches the web client, so no TypeScript was run and none is claimed.

## Deliberately not included

- **The mtgish emitter entry** (`add-feature` step 8b). It does not apply: the mtgish corpus has no IR
  tag for this clause — only `Shuffle`, `ShuffleGraveyardCardIntoLibrary`, and `ExileSpell`, which maps
  to `ExileTargetSpell`. This is a spell-envelope flag, and its sibling `selfExile()` likewise has no
  bridge entry; it is lowered by a set-specific recognizer. Registering a guessed tag would be a
  fabrication, so it is skipped rather than approximated.
- **An Assay grammar band.** Per `docs/oracle-assay.md` a rule ships as its own measured band with its
  own probe and PR, so this doesn't grow into one. Worth naming as newly reachable, though: Assay
  declines all five Zeniths today on the "Shuffle ~ into its owner's library" clause, and that is now
  expressible in the SDK.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -9131,6 +9131,23 @@ composite abilities).
     `manaCost == ManaCost.ZERO` — a printed `{0}` parses to a non-empty one-symbol cost and stays normally castable,
     and plenty of engine test fixtures construct `ManaCost.ZERO` directly as shorthand for "free," which must not
     trip this gate.
+- `selfShuffleIntoLibrary()` — `spell { effect = …; selfShuffleIntoLibrary() }` for a card that prints
+  "Shuffle <card name> into its owner's library." (the Mirrodin Besieged Zenith cycle: Green Sun's Zenith,
+  Blue/White/Black/Red Sun's Zenith). Sets `CardScript.selfShuffleIntoLibraryOnResolve`, the sibling of
+  `selfExileOnResolve`: both replace the destination of **CR 608.2n** ("as the final part of an instant or
+  sorcery spell's resolution, the spell is put into its owner's graveyard"), and `StackResolver` reads them
+  at the same seam on both the full-resolve and paused-resolve paths. The card is added to its owner's
+  library and the library is then shuffled, emitting `LibraryShuffledEvent` — the same tail the Omen face
+  uses. A card prints one clause or the other, so don't set both (this one wins if you do — it is checked
+  first). Deliberately **not** a zone-change replacement (`RedirectZoneChange` is that, and applies to any
+  card heading to a graveyard from anywhere), and **not** `AfterResolveDestination.BOTTOM_OF_LIBRARY` (which
+  also lands in `Zone.LIBRARY` but does not shuffle). It is also **not** the cast-this-way rider
+  `AfterResolveDestination` — and it *outranks* one, uniquely among the card-intrinsic destinations: those
+  riders read "if that spell would be put into a graveyard, [somewhere] instead" (Kylox's Voltstrider), and a
+  spell that shuffles itself in never would be, so the rider has nothing to replace. On the countered and
+  fizzled paths, where the card really is put into a graveyard, the rider still wins. Because it is read at resolution-destination
+  time it is correctly inert when the spell is countered or fizzles — those paths never reach CR 608.2n, so
+  the card goes to its owner's graveyard as usual.
 - `Paradigm` (Secrets of Strixhaven) — `spell { effect = …; paradigm() }` on a Lesson spell. An **exile-zone
   recurrence** mechanic, modelled exactly like Suspend (a marker the engine reads off an exiled card), differing
   only in that it casts a **copy** rather than the card itself, so the original recurs forever. Oracle: "[effect]
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -9138,16 +9138,22 @@ composite abilities).
   sorcery spell's resolution, the spell is put into its owner's graveyard"), and `StackResolver` reads them
   at the same seam on both the full-resolve and paused-resolve paths. The card is added to its owner's
   library and the library is then shuffled, emitting `LibraryShuffledEvent` — the same tail the Omen face
-  uses. A card prints one clause or the other, so don't set both (this one wins if you do — it is checked
-  first). Deliberately **not** a zone-change replacement (`RedirectZoneChange` is that, and applies to any
+  uses. A card prints one clause or the other, and setting both is **rejected at card-construction time**
+  (in the DSL, with a message naming the card, and again in `CardScript`'s own `init` for scripts built
+  directly). Deliberately **not** a zone-change replacement (`RedirectZoneChange` is that, and applies to any
   card heading to a graveyard from anywhere), and **not** `AfterResolveDestination.BOTTOM_OF_LIBRARY` (which
   also lands in `Zone.LIBRARY` but does not shuffle). It is also **not** the cast-this-way rider
-  `AfterResolveDestination` — and it *outranks* one, uniquely among the card-intrinsic destinations: those
-  riders read "if that spell would be put into a graveyard, [somewhere] instead" (Kylox's Voltstrider), and a
-  spell that shuffles itself in never would be, so the rider has nothing to replace. On the countered and
-  fizzled paths, where the card really is put into a graveyard, the rider still wins. Because it is read at resolution-destination
+  `AfterResolveDestination` — and it *outranks* one: those riders read "if that spell would be put into a
+  graveyard, [somewhere] instead" (Kylox's Voltstrider), and a spell that shuffles itself in never would be,
+  so the rider has nothing to replace. On the countered and fizzled paths, where the card really is put into
+  a graveyard, the rider still wins. Because it is read at resolution-destination
   time it is correctly inert when the spell is countered or fizzles — those paths never reach CR 608.2n, so
   the card goes to its owner's graveyard as usual.
+  **It does not outrank flashback (CR 702.34a) or harmonize (CR 702.180a)**, printed or granted. Every other
+  clause at this seam — the rider, rebound, Adventure, Omen — is worded "instead of putting it into its
+  owner's *graveyard*", which is why the printed clause beats them; those two are worded "exile this card
+  instead of putting it *anywhere else* any time it would leave the stack", which covers the library move,
+  so they still apply. A Blue Sun's Zenith flashbacked off Snapcaster Mage is exiled, not shuffled in.
 - `Paradigm` (Secrets of Strixhaven) — `spell { effect = …; paradigm() }` on a Lesson spell. An **exile-zone
   recurrence** mechanic, modelled exactly like Suspend (a marker the engine reads off an exiled card), differing
   only in that it casts a **copy** rather than the card itself, so the original recurs forever. Oracle: "[effect]
--- a/manual-scenarios/cards/b/black-suns-zenith.json
+++ b/manual-scenarios/cards/b/black-suns-zenith.json
@@ -1,0 +1,41 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Black Sun's Zenith"],
+    "battlefield": [
+      {"name": "Birds of Paradise"},
+      {"name": "Craw Wurm"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"},
+      {"name": "Swamp"}
+    ],
+    "graveyard": [],
+    "library": ["Swamp", "Grizzly Bears", "Swamp", "Swamp"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Llanowar Elves"},
+      {"name": "Grizzly Bears"},
+      {"name": "Hill Giant"},
+      {"name": "Elder Gargaroth"},
+      {"name": "Forest"},
+      {"name": "Forest"}
+    ],
+    "graveyard": [],
+    "library": ["Forest", "Grizzly Bears", "Forest"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": []
+}
--- a/manual-scenarios/cards/b/blue-suns-zenith.json
+++ b/manual-scenarios/cards/b/blue-suns-zenith.json
@@ -1,0 +1,37 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Blue Sun's Zenith"],
+    "battlefield": [
+      {"name": "Island"},
+      {"name": "Island"},
+      {"name": "Island"},
+      {"name": "Island"},
+      {"name": "Island"},
+      {"name": "Island"},
+      {"name": "Island"}
+    ],
+    "graveyard": [],
+    "library": ["Air Elemental", "Island", "Island", "Grizzly Bears", "Island", "Island", "Island"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Mountain"},
+      {"name": "Mountain"}
+    ],
+    "graveyard": [],
+    "library": ["Mountain", "Hill Giant", "Mountain"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "AI"
+}
--- a/manual-scenarios/cards/g/green-suns-zenith.json
+++ b/manual-scenarios/cards/g/green-suns-zenith.json
@@ -1,0 +1,37 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Green Sun's Zenith", "Green Sun's Zenith"],
+    "battlefield": [
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"},
+      {"name": "Forest"}
+    ],
+    "graveyard": [],
+    "library": ["Dryad Arbor", "Llanowar Elves", "Grizzly Bears", "Elder Gargaroth", "Craw Wurm", "Forest", "Forest"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Mountain"},
+      {"name": "Mountain"}
+    ],
+    "graveyard": [],
+    "library": ["Mountain", "Mountain", "Hill Giant", "Mountain"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "AI"
+}
--- a/manual-scenarios/cards/r/red-suns-zenith.json
+++ b/manual-scenarios/cards/r/red-suns-zenith.json
@@ -1,0 +1,39 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["Red Sun's Zenith", "Red Sun's Zenith"],
+    "battlefield": [
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"}
+    ],
+    "graveyard": [],
+    "library": ["Hill Giant", "Mountain", "Mountain", "Mountain"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Grizzly Bears"},
+      {"name": "Craw Wurm"},
+      {"name": "Forest"},
+      {"name": "Forest"}
+    ],
+    "graveyard": [],
+    "library": ["Forest", "Llanowar Elves", "Forest"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
+  "priorityPlayer": 1,
+  "player1StopAtSteps": [],
+  "player2StopAtSteps": [],
+  "player1OpponentStopAtSteps": [],
+  "player2OpponentStopAtSteps": [],
+  "mode": "AI"
+}
--- a/manual-scenarios/cards/w/white-suns-zenith.json
+++ b/manual-scenarios/cards/w/white-suns-zenith.json
@@ -1,0 +1,40 @@
+{
+  "player1Name": "Alice",
+  "player2Name": "Bob",
+  "player1": {
+    "lifeTotal": 20,
+    "hand": ["White Sun's Zenith"],
+    "battlefield": [
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Plains"},
+      {"name": "Plains"}
+    ],
+    "graveyard": [],
+    "library": ["Serra Angel", "Plains", "Plains", "Plains"]
+  },
+  "player2": {
+    "lifeTotal": 20,
+    "hand": [],
+    "battlefield": [
+      {"name": "Craw Wurm"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"},
+      {"name": "Mountain"}
+    ],
+    "graveyard": [],
+    "library": ["Mountain", "Hill Giant", "Mountain"]
+  },
+  "phase": "PRECOMBAT_MAIN",
+  "activePlayer": 1,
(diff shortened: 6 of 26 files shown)


bolav:
# Review — `mbs-sun-zenith-cycle`

Reviewed `origin/main...HEAD` in place; the branch sits exactly on `origin/main`
(`git merge-base` == `origin/main`), so no merge was needed and no merge commit was produced.

## Verdict

Behaviour is right and the SDK shape is right. The new vocabulary earns its place — composition was
tried and genuinely doesn't reach it — and it is placed at the seam the codebase already uses for
this exact question rather than as a new mechanism. **No blocking issues.** One finding was fixed on
the branch before this review was written; the rest are minor.

## What's good

- **The seam choice.** Three plausible homes existed — a new `AfterResolveDestination` enum member, a
  parameter bolted onto an existing effect, or a `CardScript` field beside `selfExileOnResolve`. The
  third is correct and the reasoning is recorded: the enum is a *cast-this-way rider* another effect
  stamps on a spell it is casting, and `CardScript` already models card-intrinsic destinations as one
  field per mechanic feeding one `when`. Following the local pattern beat inventing a fourth shape.

- **The precedence fix** (`StackResolver.kt:2400`, `:2237`). Putting the printed clause *above* the
  rider is the non-obvious part of this diff and it is correct: the riders replace "would be put into
  a graveyard", and a spell that shuffles itself in never would be. Kylox's Voltstrider is
  implemented in this corpus, so this was reachable and wrong, not theoretical. Good catch, correctly
  justified, and it falls out for free on the countered/fizzled paths.

- **The mutation check.** Unwiring only the paused-resolve path failed exactly one test and left the
  rest green. That is the difference between "there is a test" and "the test would catch it", and the
  duplicated seam is precisely where a silent half-wiring would have shipped.

- **Negative coverage.** Countered (CR 701.5a) and fizzled (CR 608.2b) both assert *graveyard*, which
  is what every Zenith's ruling says and the opposite of the rider's answer. Cheap to have omitted,
  and the omission would not have failed anything.

- **Reprint rows.** Seven of them, in the sets that actually needed them. This is the half of card
  work that routinely gets forgotten, and `just check-card-printing` was run per card rather than
  assumed.

## Issues

### Fixed on the branch

**F1 — `CardBuilder.kt` — a card could print both destination clauses.**
`selfExile()` and `selfShuffleIntoLibrary()` are two spellings of one slot, and nothing rejected
setting both; the winner was decided silently by `when`-clause order and documented only in prose.
Fixed in `260a1db6b2` with an order-independent `require` at both `CardScript` construction sites
(`CardBuilder.kt:928`, `:2148`), which also catches `paradigm()` — it implies `selfExile()` — paired
with `selfShuffleIntoLibrary()` in either order. Names the offending card, since it fires at
card-construction time for the whole corpus. A corpus-wide load confirms no existing card trips it,
and a test asserts it fires.

### Minor

**M1 — only Green Sun's Zenith has a card-level scenario test.**
This is defensible: `add-card` says a card built purely from existing primitives is covered by the
snapshot and lint nets, the shared clause has thorough engine coverage, and Green Sun's Zenith is the
one card taking the paused-resolve path. But two spots would repay a cheap test — Black Sun's Zenith
at X = 0 ("put 0 -1/-1 counters on each creature", where the interesting question is that it is a
no-op rather than an error), and Red Sun's Zenith's exile rider, which is the only card in the cycle
with behaviour beyond "do a thing, then shuffle". Author's call.

**M2 — `RedSunsZenith.kt:35` — the "no-op" comment is right but elides a difference.**
The comment says marking a player or planeswalker is a no-op. True, but by two different mechanisms,
and this comment is the only place either is recorded: for a **player** the executor early-returns
(`MarkExileOnDeathExecutor.kt:35`), whereas for a **planeswalker** the marker *is* applied and simply
never consulted, because the loyalty SBA routes through `SbaZoneMovementHelper.putPermanentInGraveyard`
(`:197`), which — unlike `putCreatureInGraveyard` — does not check `ExileOnDeath`. Worth one clause,
because the planeswalker half is load-bearing on a helper the comment doesn't name.

**M3 — pre-existing, out of scope: `MarkExileOnDeathEffect` scoping lives in the consumer.**
`MarkExileOnDeathExecutor` applies the marker to any non-player entity rather than checking
creature-ness, and it is only creature-scoped because `putCreatureInGraveyard` is the sole consumer.
Harmless today and shared with Carbonize, Scorching Dragonfire, and Obliterating Bolt, so this branch
inherits it rather than introducing it. Noting it only so it isn't mistaken for something this PR
did.

## On the verification claims

Worth a reviewer's attention rather than a finding, because the PR body states it plainly: **neither
full `just test` run came back green overall.** I checked the claim that this is environmental and it
holds up — no test failed on an assertion at any point; every red was a 120 s watchdog timeout or a
killed worker JVM (`exit value 13` with *zero* recorded test failures), in a **different module each
run**, on a box at load average 9.4 with the run stretching 19 → 56 minutes. All three implicated
suites pass serially. The structural argument is also sound: the new field defaults to `false` and
no card outside these five sets it, so both `when` blocks evaluate exactly as before for the entire
rest of the corpus.

The `assay-differential` result is reported honestly — 0 divergences, but only 4 of 12 MBS cards are
grammar-covered and all five Zeniths are among the declines, so it says nothing about the new cards.
Good that the body says so rather than banking the green.

## Recommendation

Merge as-is. M1 and M2 are optional polish; M3 is somebody else's ticket.

bolav:
# Review round 2 — `mbs-sun-zenith-cycle`

Second pass, focused on the `StackResolver` destination change. Reviewed `origin/main...HEAD` in
place; the branch sits exactly on `origin/main` (`git rev-list --count HEAD..origin/main` = 0), so no
merge was needed and no merge commit was produced. Round 1's findings (`_review.md`) are not repeated
here — this round only adds what it found, including one correction to round 1.

**Everything below has been fixed on the branch.** The findings are kept in full because the
reasoning is the load-bearing part: the precedence at this seam is not a total order, and the next
person to add a clause to it needs to know why.

## Verdict

The card work is clean and the seam choice is right, but the **precedence decision that round 1
signed off on was half wrong**, and the half that was wrong was reachable with two cards this corpus
already ships. Putting `selfShuffleIntoLibrary` at the top of the `when` is correct against the
cast-this-way rider, rebound, Adventure and Omen — every one of those replaces *"instead of putting
it into its owner's graveyard"* — but **flashback and harmonize are worded "instead of putting it
anywhere else any time it would leave the stack"**, and those two still have to win.

## What's good (beyond round 1)

- **Reprint coverage is exactly right, and I checked it against Scryfall rather than the script.**
  Pulled `unique=prints` for all five cards: MBS (2011-02-04, `expansion`) is the earliest real
  printing of every one, so all five canonicals are correctly placed. Of the 22 other printings,
  exactly six live in sets scaffolded under `definitions/` — C13, C15 (Blue), C14, TLE (Black), C14,
  C17, CMR (White) — and all seven rows are present. The other sixteen (A25, CMA, MKC, ECC, TD2, EMA,
  2X2, V13, SPG, ONC, SCD, PLST, …) are unscaffolded, so correctly absent. No drift either way, which
  is unusual for a seven-row diff.

- **Owner, not caster, throughout.** `ownerId = cardComponent?.ownerId ?: casterId` feeds both the
  destination `ZoneKey` and `shuffleOwnerLibrary`/`LibraryShuffledEvent`, so the Knowledge Pool ruling
  quoted in `GreenSunsZenith.kt` ("that opponent then shuffles Green Sun's Zenith into *your*
  library … you won't shuffle any library") falls out for free on both paths.

- **Copies are already handled, on both paths.** `StackResolver` removes a `CopyOfComponent` spell
  before the destination block runs, so a copied Zenith ceases to exist per CR 112.3b instead of
  adding a card to a library. "The copy shuffles a real card in" is the obvious way this feature goes
  wrong, and it can't happen.

- **The X-spell filter placement in Green Sun's Zenith.** `count = 1` for how many cards are found,
  with the {X} living in the *filter* (`manaValueAtMostX()`), is the non-obvious right call, and the
  KDoc says why. Dryad Arbor is the correct corner to have picked.

## Issues, all fixed on the branch

### B1 (blocking) — the new clause also outranked flashback and harmonize

`StackResolver.kt`, both the full-resolve and paused-resolve `when` blocks. The justification in the
code comment — the riders replace *"would be put into a graveyard"*, and a spell that shuffles itself
in never would be — is correct, and holds for every clause at that seam **except two**:

| Clause | CR | Wording it replaces | Printed clause above it? |
|---|---|---|---|
| `exileAfterResolveComp` (Kylox's Voltstrider, Lilah) | card text | "would be put into a graveyard" | ✅ correct |
| `reboundExile` | 702.88a | "instead of putting it into your graveyard as it resolves" | ✅ correct |
| `adventureFaceExile` | 715.3d | "Instead of putting … into its owner's graveyard as it resolves" | ✅ correct |
| `omenFaceShuffle` | 720.3d | "instead of putting it into its owner's graveyard as it resolves" | ✅ correct (moot) |
| `flashbackExile` — flashback | **702.34a** | **"exile this card instead of putting it anywhere else any time it would leave the stack"** | ❌ **was wrong** |
| `flashbackExile` — harmonize | **702.180a** | **"exile this card instead of putting it anywhere else any time it would leave the stack"** | ❌ **was wrong** |

(Rule numbers grepped out of `MagicCompRules_20260808.txt`, not recalled.)

Reachable in the most ordinary way possible: **Snapcaster Mage** (`isd/`) grants flashback to target
instant or sorcery in your graveyard, and Blue Sun's Zenith is an **instant**. Iroh, Grand Lotus
(`tla/`, via `GraveyardCardsHaveFlashback`) and Archmage's Newt (`otj/`) reach all five.
`FlashbackGrants` routes printed and granted flashback through one read site, so the grant behaves
like a printed one — including here. The consequence was a repeatable value loop: flashback Blue
Sun's Zenith, draw X, shuffle it back into your library instead of exiling it, repeat.

Verified rather than reasoned, before the fix — a throwaway test (a `{1}` sorcery with
`selfShuffleIntoLibrary()` plus printed flashback), cast from the graveyard with
`useAlternativeCost = true`:

```
REPRO life=22            <- it resolved
REPRO inExile=false      <- CR 702.34a says this must be true
REPRO inLibrary=true     <- what the branch did
REPRO inGraveyard=false
```

**The fix, and why it isn't a plain reorder.** The three replacements are not totally ordered — the
rider loses to the printed clause, the printed clause loses to flashback, and flashback loses to the
rider (that last one is pre-existing, and left exactly as it was). A `when` chain can't express a
cycle, so what breaks it is a guard that says *what each replacement is worded to replace*:

```kotlin
val intendedDestination = when {
    exileAfterResolveComp != null && !selfShuffleIntoLibrary -> exileAfterResolveComp.zone
    flashbackExile -> Zone.EXILE          // CR 702.34a / 702.180a — "anywhere else"
    selfShuffleIntoLibrary -> Zone.LIBRARY
    selfExile || adventureFaceExile || reboundExile -> Zone.EXILE
    omenFaceShuffle -> Zone.LIBRARY
    else -> Zone.GRAVEYARD
}
```

and the mirror on the paused path. For any card without the new flag the guard is vacuous and the
chain evaluates exactly as before, so the rest of the corpus is provably untouched. The shuffle tails
were already gated on the *final* destination, so they went quiet on their own — no second edit.

Tests added: `flashback outranks the printed clause — CR 702.34a's "anywhere else"` and `flashback
outranks it on the paused-resolve path too` (engine), plus `flashbacked off Snapcaster Mage it is
exiled, not shuffled in — CR 702.34a` in a new `BlueSunsZenithScenarioTest` — two real cards, no
synthetic fixture. **Mutation-checked:** restoring the old ordering fails exactly those two engine
tests and nothing else, so each path is independently covered.

Round 1 called this precedence choice "correct … reachable and wrong, not theoretical" and praised
it. That is right about the rider and wrong about flashback; this finding supersedes that bullet.

### I1 (important) — Red Sun's Zenith marked the exile rider even when no damage was dealt

`MarkExileOnDeathEffect(t) then DealDamageEffect(X, t)` marked the target unconditionally, but the
card says "If **a creature dealt damage this way** would die this turn, exile it instead". Cast for
**X = 0** — a legal, `{R}`-cheap mode of an {X} spell — and CR 120.8 applies: "If a source would deal
0 damage, it does not deal damage at all." No creature was dealt damage this way, so nothing should
be exiled; the branch marked the target anyway, and a creature that died to anything else that turn
was wrongly exiled.

Not round 1's M3 (the executor marking any non-player entity, shared with Carbonize). That one is
about *what* can be marked; this is about *whether* the mark should exist at all, and Carbonize can't
reach it — its 3 damage is fixed, so only prevention gets there, whereas X = 0 is a first-class mode
of this card.

Fixed by composition rather than new vocabulary — the existing `Compare` condition over two
`DynamicAmount`s, which reads the spell's own X through the same evaluator the damage already uses:

```kotlin
effect = ConditionalEffect(
    condition = Conditions.CompareAmounts(DynamicAmount.XValue, ComparisonOperator.GT, DynamicAmount.Fixed(0)),
    effect = Effects.MarkExileOnDeath(t),
) then Effects.DealDamage(DynamicAmount.XValue, t)
```

New `RedSunsZenithScenarioTest` covers X = 2 (killed creature exiled), X = 0 (creature bolted later
the same turn goes to the **graveyard**), and the shuffle tail. **Mutation-checked:** dropping the
gate fails the X = 0 test alone. Note the deeper case — X ≥ 1 but the damage is *prevented* — is
still wrong, and is genuinely pre-existing and shared with every `MarkExileOnDeath` card; fixing that
needs a "was damage actually dealt" condition the SDK doesn't have, so it stays round 1's M3.

### M1 (minor) — stale "checked first" prose, in two places including the canonical catalog

`CardScript.kt` ("this one wins if a card somehow sets both, because `StackResolver` checks it
first") and `docs/card-sdk-language-reference.md` ("this one wins if you do — it is checked first")
both described behaviour that commit `260a1db6b2` made unreachable. Both now say the pair is rejected
at card-construction time, and both gained the flashback/harmonize carve-out from B1 — as written
they told the next card author the clause beats everything, which is what produced B1.

### M2 (minor) — the `require` guarded the DSL, not the type

`validateResolutionDestination` runs in `CardBuilder.build()`, but `CardScript` is a `data class` with
both booleans public and defaulted, so `CardScript(selfExileOnResolve = true,
selfShuffleIntoLibraryOnResolve = true)` still constructed — and test fixtures and the Assay compiler
build `CardScript`s directly. `CardScript` now has an `init` block rejecting the pair; the DSL guard
stays, purely so its message can name the offending card. Covered by `nor can a CardScript built
directly set both`.

**Deliberately not done:** collapsing the two booleans into a `SelfResolveDestination?` enum, which
is the shape that makes the invalid state unrepresentable rather than merely unconstructable. It
would rename a serialized field, and re-blessing that means touching 15 other sets' card goldens (25
occurrences of `selfExileOnResolve`) from a five-card PR. The `init` guard closes the actual hole; the
enum is worth its own change, and worth a line in the PR body as the known follow-up, because a third
destination clause would otherwise add a third boolean and a second `require`.

### M3 (minor) — untested justifications

The shuffle-tail comment gives two reasons for gating on the final destination:
`AfterResolveDestination.BOTTOM_OF_LIBRARY` (was covered) and "a `RedirectZoneChange` may have sent
the card elsewhere" (was not). Added `a redirect away from the library suppresses the shuffle` — a
Progenitus-shaped card-intrinsic self-replacement pointed at the library, asserting the card lands in
exile and *no* `LibraryShuffledEvent` fires. Also added `the printed clause beats the rider on the
paused-resolve path too`, since the rider case previously only exercised the full-resolve site.

### M4 (minor) — the paused path moves the card early

`StackResolver` puts the spell in the library and shuffles while the decision is still pending, so
any continuation of that same resolution sees the card already in the library. Harmless for Green
Sun's Zenith — a search's candidate list is built during effect execution, before the move, so the
Zenith can never find itself — but it is a sharper edge than the same early move is for
`selfExile()`, because a library card can be drawn or searched, and Blue Sun's Zenith's own
2018-03-16 ruling depends on the ordering. Recorded in the engine test file's header, where the next
person adding a pausing self-shuffling card will read it.

### M5 (minor) — raw effect constructors where facades exist

`RedSunsZenith.kt` used `MarkExileOnDeathEffect` / `DealDamageEffect` and `BlackSunsZenith.kt` used
`AddDynamicCountersEffect`, though `Effects.MarkExileOnDeath`, `Effects.DealDamage(DynamicAmount, …)`
and `Effects.AddDynamicCounters` all exist. `FacadeBoundaryTest`'s forbidden list doesn't cover these
three and Carbonize does the same, so it was inherited style rather than a new violation — but
AGENTS.md's "Facades, not raw constructors" is unqualified. Both cards now use the facades.
`ConditionalEffect` is left as a direct call: it has no `Effects.*` facade and 38 cards in this era
already construct it directly.

## Verification

| What | Result |
|---|---|
| `SelfShuffleIntoLibraryOnResolveScenarioTest` | 13/13 PASSED (5 new) |
| `RedSunsZenithScenarioTest` (new) | 3/3 PASSED |
| `BlueSunsZenithScenarioTest` (new) | 3/3 PASSED |
| `AfterResolveDestinationScenarioTest` | 6/6 PASSED — the rider path is unchanged |
| `ParadigmMechanicTest` | 8/8 PASSED — the other `selfExileOnResolve` consumer |
| `just test-scenarios 2008-2016` | 508 tests / 184 classes, 0 failures |
| `:mtg-sdk:test` | 511 tests, 0 failures (ran after the `CardScript` edit) |
| `FacadeBoundaryTest` | PASSED |
| `just rebless-cards` | only `MBS.json` moved, and only Red Sun's Zenith's tree — the whole corpus loads, so no existing card trips the new `init` guard |
| Mutation checks | old ordering → exactly the 2 flashback tests red; ungated marker → exactly the X = 0 test red |
| Scryfall `unique=prints`, all five cards | canonical + reprint placement verified card by card |
| CR 608.2n, 120.8, 702.34a, 702.88a, 702.180a, 715.3d, 720.3d | grepped from `MagicCompRules_20260808.txt` |

**Not run:** the full `just test`. The diff reaches `mtg-sdk`, `rules-engine` and `mtg-sets` but not
`game-server`, `ai`, `gym` or `web-client`, and the two engine edits are provably inert for every card
without the new flag (both guards are vacuous when `selfShuffleIntoLibrary` is false). I take round
1's analysis of the two red full runs at face value — no assertion failures, watchdog timeouts and
killed workers in a different module each time, green serially. Note B1 would not have been caught by
a green full gate either: nothing crossed flashback with a self-shuffling spell, because until this
branch no such card existed.

## Recommendation

Merge. Every finding above is fixed and mutation-checked on the branch. The one open item is M2's
enum, deliberately deferred with its reason; round 1's M3 (damage-prevention scoping of
`MarkExileOnDeath`) remains somebody else's ticket, and Black and White Sun's Zenith still have no
card-level scenario test of their own — round 1 judged that acceptable and nothing here changes it.

PR #2117, opened 2026-08-30
Preserve source metadata for countered abilities

huxinq:
When an activated or triggered ability is countered, `AbilityCounteredEvent` currently retains only the ability entity ID and description. Once the stack object is removed, downstream engine consumers can no longer recover the ability's source or controller reliably.

Capture the ability's source ID, source name, and controller from the authoritative stack metadata before removal and carry them on `AbilityCounteredEvent`.

The new fields are nullable and defaulted so previously serialized events continue to decode. Countering behavior itself is unchanged.

Coverage includes:

- activated and triggered abilities;
- a source that has already left the battlefield;
- a source whose controller has changed after the ability was put on the stack;
- event serialization and legacy defaults;
- client projection, confirming the additional internal metadata is not exposed through the existing `ClientEvent`.

Verification:
- `just test-class CounteredAbilityEventTest`
- `just test-class CounteredAbilityEventProjectionTest`
- `just test-rules`

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/core/GameEvent.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/core/GameEvent.kt
@@ -783,7 +783,11 @@ data class SpellCounteredEvent(
 @SerialName("AbilityCounteredEvent")
 data class AbilityCounteredEvent(
     val abilityEntityId: EntityId,
-    val description: String
+    val description: String,
+    /** Last-known source metadata captured from the ability while it is still on the stack. */
+    val sourceId: EntityId? = null,
+    val sourceName: String? = null,
+    val controllerId: EntityId? = null,
 ) : GameEvent
 
 /**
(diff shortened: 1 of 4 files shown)


wingedsheep:
Thanks for this — the capture point you picked was right, and it surfaced a real gap next to it.

`counterAbility` was calling `removeFromStack`, which is only `copy(stack = stack - entityId)`, so the ability entity stayed in `GameState.entities` with its stack component intact. Every countered ability lingered there for the rest of the game and rode along in every serialized state, while all the resolution paths already called `removeEntity`. That also meant source and controller were still recoverable from `event.abilityEntityId`, so the new fields had no reader yet.

#2119 destroys the ability object on the counter path too (CR 701.6a, matching CR 608.2n on resolution), which is what makes your fields load-bearing — they're merged as you wrote them, with the KDoc extended to say why. The two tests didn't come along: no `GameEvent` is persisted anywhere (replays store `actions`), and the projection test pinned the fields as unused, which is the opposite of the intent. `CounteredAbilityTest` covers the behaviour instead.

Closing as superseded by #2119. Nice catch.

PR #2118, opened 2026-08-30
Don't offer a "you may" whose payment can't be completed

bolav:
# Don't offer a "you may" whose payment can't be completed

## The bug

From a playtest log: Emeritus of Ideation attacked with **one** card in its controller's graveyard.
The engine offered "you may exile eight cards from your graveyard", the controller said yes, and the
lone card was exiled — for nothing, since the payment could never be completed.

The selection filled with as many cards as it could reach (one), the `CollectionNonEmpty(min = 8)`
gate correctly scored the payment as failed, so the creature did *not* become prepared — but the
card was gone anyway.

CR 609.3 ("if an effect attempts to do something impossible, it does only as much as possible") is
what produced the partial exile, but it governs *mandatory* instructions. The rule that applies to an
optional payment is **CR 608.2d**: "The player can't choose an option that's illegal or impossible."
The prompt has to be absent, not offered and then scored. (Both quoted from
`MagicCompRules_20260808.txt`.)

Three other cards share the idiom and had the same hole: **Skywarp Skaab** (exile two creature cards
→ draw), **Aegis Sculptor** (exile two → +1/+1 counter), **Heated Argument** (exile one → 2 damage).

## The fix

No new SDK vocabulary — this is an engine-side gate decision, in the one place that already owns
"don't offer an impossible yes".

`GatedEffectExecutor`'s `Gate.MayDecide` branch now skips its prompt when the effect inside the gate
is a `Gate.DoAction` scored by `SuccessCriterion.CollectionNonEmpty(name, min)` whose named
collection is filled by a `SelectFromCollectionEffect(ChooseExactly)` over a `GatherCardsEffect`
pool holding fewer than `min` cards. No answer to the yes/no could clear that bar, so the option
isn't a legal choice at all. It sits with the branch's existing skips — an unaffordable `Gate.MayPay`
cost, a `ChooseActionEffect` with no feasible option, an unreachable collect-evidence threshold, and
a declared `FeasibilityCheck`.

Two properties worth stating, because both are load-bearing:

- **The pool is read at resolution**, not when the ability triggered — a graveyard emptied or shrunk
  in response takes the option away.
- **The check runs before the remembered-answer (auto-answer yield) path**, so a stored "yes" can't
  perform the impossible either.

The count is deliberately generous: a selection's `restrictions` could narrow it further and aren't
applied, so the prompt is only ever withheld when the payment is impossible outright. Every shape
that isn't this idiom prompts exactly as before.

`docs/card-sdk-language-reference.md` gains the corresponding paragraph under `Gate.MayDecide`.

## What this does not change

Mandatory instructions keep CR 609.3's do-as-much-as-possible behaviour — `SelectionMode.ChooseExactly`
still takes as many as it can reach when the instruction isn't an option. That's the correct reading
for "discard two cards" with one in hand, and changing it globally would have broken it.

## Verification

- `scripts/gradle-locked :rules-engine:test` — **BUILD SUCCESSFUL** (full engine suite, 2m14s).
- `just test-class EmeritusCycleScenarioTest` — green, including two new tests:
  - *fewer than eight cards in the graveyard: nothing is exiled and it stays unprepared*
  - *a graveyard shrunk in response takes the option away* (eight cards at declare-attackers, one
    eaten by Withered Wretch while the trigger is on the stack, seven survivors untouched)
- `just test-class SkywarpSkaabScenarioTest` — green, including a new test for the
  one-creature-card case (no prompt raised, the card stays in the graveyard, no draw).
- `just test-class HeatedArgumentScenarioTest`, `just test-class AegisSculptorScenarioTest` — green
  (the other two cards that reach the new branch).
- **Mutation check:** with the new branch disabled, both new Emeritus tests fail — the pre-fix engine
  exiles the lone card, and exiles all seven survivors in the in-response case.

**Not covered:** the full per-era card scenario suite (`just test-scenarios`) was not run. A grep of
`mtg-sets/*/src/main` shows the four cards above are the only ones that construct this shape, and no
`*Patterns` helper emits it, so no other card can reach the new branch; all four have green scenario
tests. Nothing outside Kotlin is touched — no client, server, or SDK code in the diff.

## Left out

The playtest report also said the creature became prepared. That half does not reproduce: with one
card in the graveyard the gate already scored the payment as failed and `BecomePrepared` never ran
(the new test asserts it). The likely explanation is that it was still prepared from entering the
battlefield — it enters prepared and stays so until the Ancestral Recall copy is cast — or a stale
server build. Nothing was changed on that basis.

### DIFF
--- a/docs/card-sdk-language-reference.md
+++ b/docs/card-sdk-language-reference.md
@@ -2548,6 +2548,13 @@ Atomic effect factories. For library/zone manipulation, prefer the pipelines in 
     player "doesn't": the prompt is skipped and `otherwise` runs directly. Lets "you may sacrifice an
     artifact. If you don't, …" apply its else automatically when the controller has no artifact (the
     no-target analogue of a targeted "may" with no legal targets falling to its else branch).
+    The executor skips the prompt on its own in one more case, no authoring needed: a `then` that is
+    a `Gate.DoAction` scored by `SuccessCriterion.CollectionNonEmpty(name, min)` whose collection is
+    filled by a `SelectFromCollectionEffect(ChooseExactly)` over a `GatherCardsEffect` pool holding
+    fewer than `min` cards. No answer could clear that bar, so the option isn't a legal choice
+    (CR 608.2d) — **Emeritus of Ideation** with seven cards in the graveyard isn't asked whether to
+    exile eight, and the seven aren't spent for nothing. (CR 609.3's do-as-much-as-possible governs
+    *mandatory* instructions; it doesn't turn an unavailable option into a partial payment.)
   - `Gate.MayPay(cost)` — "You may [cost]. If you do, [then]." `cost` is a cost **effect**
     (`PayManaCostEffect`, `PayDynamicManaCostEffect`, `PayLifeEffect`, `SacrificeEffect`, or a
     `CompositeEffect` of them). An unaffordable cost (fixed mana, dynamic mana, and life are
--- a/mtg-sets/2017-2022/tests/src/test/kotlin/com/wingedsheep/engine/scenarios/SkywarpSkaabScenarioTest.kt
+++ b/mtg-sets/2017-2022/tests/src/test/kotlin/com/wingedsheep/engine/scenarios/SkywarpSkaabScenarioTest.kt
@@ -79,6 +79,34 @@ class SkywarpSkaabScenarioTest : ScenarioTestBase() {
                 }
             }
 
+            test("one creature card in the graveyard: the may is never offered and nothing is exiled") {
+                // CR 608.2d — exiling two creature cards is impossible with one there, so the
+                // option isn't offered at all. The lone card must not be exiled for nothing.
+                val game = scenario()
+                    .withPlayers("Player1", "Player2")
+                    .withCardInHand(1, "Skywarp Skaab")
+                    .withCardInGraveyard(1, "Grizzly Bears")
+                    .withCardInLibrary(1, "Plains")
+                    .withLandsOnBattlefield(1, "Island", 5)
+                    .withActivePlayer(1)
+                    .inPhase(Phase.PRECOMBAT_MAIN, Step.PRECOMBAT_MAIN)
+                    .build()
+
+                game.castSpell(1, "Skywarp Skaab").error shouldBe null
+                game.resolveStack()
+
+                withClue("no yes/no is raised for an exile that can't be completed") {
+                    game.hasPendingDecision() shouldBe false
+                }
+                withClue("the lone creature card stays in the graveyard") {
+                    game.isInGraveyard(1, "Grizzly Bears") shouldBe true
+                    game.isInExile(1, "Grizzly Bears") shouldBe false
+                }
+                withClue("no card was drawn") {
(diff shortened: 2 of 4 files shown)


bolav:
# Self-review — `impossible-may-option`

Reviewed `git diff origin/main...HEAD` (4 files, +181). The branch sits directly on `origin/main`
(0 commits behind), so no merge was needed and the diff is post-merge reality.

## Verdict

Behaviour is right and the shape is right: no new SDK type, no card edits, one more case in the
executor branch that already owns "don't offer an impossible yes" (`Gate.MayPay` affordability,
infeasible `ChooseActionEffect`, unreachable collect-evidence, declared `FeasibilityCheck`). Both
cited rules were checked against `MagicCompRules_20260808.txt` — 608.2d at line 2799, 609.3 at 2851 —
and each is exercised by a test rather than being decorative.

## What's good

- **The fix is where the decision is, not where the damage showed up.** Making
  `SelectionMode.ChooseExactly` all-or-nothing would have been the tempting one-liner and would have
  broken CR 609.3 for every mandatory "discard two cards" in the corpus. Gating the *option* keeps
  mandatory partial performance intact; the PR body says so explicitly.
- **Placement inside `execute` is deliberate.** The check runs inside the `Gate.MayDecide` block,
  ahead of the auto-answer yield, so a remembered "yes" can't perform the impossible. That ordering
  is easy to get wrong and is covered by the surrounding comments.
- **The in-response test earns its keep.** Withered Wretch eating one of eight cards mid-stack proves
  the pool is read at resolution, and it fails pre-fix by exiling all seven survivors — a Tormod's
  Crypt variant that empties the graveyard entirely would have passed before *and* after, proving
  nothing. Both new Emeritus tests were mutation-checked with the branch disabled.
- **Fails open by construction.** Every unrecognized shape returns `true`, so the blast radius is the
  four cards that build this idiom, all of which have green scenario tests.

## Issues

### Minor

**M1 — the probe re-executes the gather** (`GatedEffectExecutor.kt:714`). `effectExecutor(state,
gather, context)` runs the real gather to size the pool, then the action runs it again moments later.
That is safe (gathering only stores a collection — no state change, no events, and the KDoc says so)
and it is the only way to size the pool without duplicating `GatherCardsExecutor`'s source-resolution
logic. Worth knowing it's a deliberate second evaluation rather than an oversight, and worth
revisiting if a `CardSource` ever gains a side effect — at which point the honest move is a shared
`resolveSource` helper rather than an executor call.

**M2 — only top-level pipeline steps are inspected** (`GatedEffectExecutor.kt:708-713`). `steps` is
the `CompositeEffect`'s immediate children, so an idiom that nests the gather or the selection one
level deeper isn't recognized and prompts as before. Correct-but-incomplete, and cheap to extend if a
future card needs it; flagged so the limit is on the record rather than discovered later.

**M3 — selection `restrictions` aren't applied to the pool count.** The count can therefore
over-estimate what's actually selectable, so a prompt could still be offered for a payment that a
restriction makes impossible. Conservative in the right direction (it only ever *fails to* skip) and
now stated in the KDoc.

No Blocking or Important findings.

## Recommendation

Merge as-is. M1–M3 are notes for the record, not changes to make; none of them alters behaviour for
any card in the corpus today.

PR #2143, opened 2026-08-30
Add explicit hidden-world materialization

huxinq:
Adds a narrow rules-engine primitive for constructing hypothetical complete states from caller-supplied hidden-card assignments.

Argentum’s rules engine always operates on a complete `GameState`, even when an AI or simulation caller is reasoning from partial information. The existing AI determinizer handles one version of that problem inside the AI layer, but it also owns sampling policy and fallback behavior. There is currently no reusable engine-level contract for the narrower operation: “apply these hypothetical hidden identities to this state if doing so is structurally safe.”

`HiddenWorldMaterializer` adds that boundary.

A caller supplies:

- an explicit `EntityId -> CardDefinition` assignment for hand/library slots;
- a caller-selected RNG for future simulated game events.

The materializer does not choose those identities. It answers only whether the supplied assignment can be installed coherently into the engine state.

For example, a search caller can take a position with three unresolved opponent hand/library slots and construct one hypothetical world in which those slots contain Mountain, Shock, and Grizzly Bears, then another world with a different assignment, while using the ordinary rules engine to simulate both. The search or belief layer remains free to decide why those worlds were chosen and how much weight to give them.

The same separation applies to randomness. A caller that wants each hypothetical world to have an independent future shuffle/draw stream can supply a separately derived `futureRng`. A caller that deliberately wants to preserve the source stream can explicitly pass the source RNG instead. The materializer does not derive either policy implicitly.

For supported hand/library slots, materialization preserves:

- the existing `EntityId`;
- owner and controller-component state;
- zone membership;
- exact library ordering;
- per-card reveal state;
- unrelated game state.

Definition-derived components are rebuilt through `CardEntityFactory`.

Thus if `card-1001` is the second card in a library, assigning a different hypothetical definition does not replace or move `card-1001`. The opaque slot remains the same object from the perspective of existing entity references; only the current card definition and its definition-derived component state change.

Some hidden-zone objects cannot safely be treated as interchangeable. A card can retain runtime state from an earlier incarnation, or the engine can have an in-flight stack/decision/continuation that depends on current object relationships. In those cases the materializer returns a typed unsupported result rather than producing a superficially complete but internally inconsistent state.

One regression exercises a real battlefield-to-hand transition. The card retains `LastKnownPermanentComponent`; assigning a different hypothetical identity at that point would attach historical battlefield information to the wrong card, so materialization is refused. After a subsequent zone transition clears that runtime state, the same slot becomes materializable again.

The materializer likewise refuses identity rewrites while stack, pending-decision, or continuation state makes generic substitution unsafe.

There is also a DFC-specific validity check. `CardRegistry` intentionally registers back-face definitions so transform machinery can resolve them, but those definitions are not legal standalone card identities in hand or library. A request to materialize `Test DFC Back` directly into a hidden hand slot is therefore rejected as an invalid assignment rather than accepted merely because the definition is registered.

This primitive deliberately has no viewer, deck model, or probability semantics. It does not inspect visibility, infer which slots are unknown, constrain assignments to a deck, sample worlds, or assign probabilities. `GameState` itself contains complete identities, so the materializer also cannot detect that a caller has omitted a slot that is epistemically unresolved from that caller’s perspective. A caller intending to construct a complete hidden-world hypothesis must provide the appropriate assignments itself.

The existing Phase 8 AI `Determinizer` is not migrated by this change and retains its current sampling and fallback behavior. This PR establishes only the generic engine-state materialization seam.

Coverage includes:

- explicit hand/library assignments while preserving entity IDs and zone ordering;
- reconstruction of definition-derived components such as Madness;
- caller-controlled reproducible assignment generation;
- source-state and source-RNG immutability;
- preservation of unrelated tracker and reveal state;
- typed refusal of battlefield last-known runtime state;
- successful materialization after that runtime state is cleared;
- typed refusal during a real stack state;
- rejection of unregistered definitions;
- rejection of DFC back-face identities in hand/library;
- ordinary `ActionProcessor` execution from a successfully materialized world.

Verification:

- `just test-class HiddenWorldMaterializerTest`
- `just test-rules`
- `just test-ai`
- `just test-gym`
- `just test-gym-trainer`
- `git diff --check upstream/main..HEAD`

### DIFF
--- a/docs/ai/architecture.md
+++ b/docs/ai/architecture.md
@@ -176,6 +176,21 @@ Phase 8 starts with one shared determinization per search. More worlds compete d
 count, so raising K requires an arena result showing it buys more strength at the same wall-clock
 budget.
 
+For callers that already own an assignment, the generic engine boundary separate from Phase 8's
+sampling policy is `HiddenWorldMaterializer` in `rules-engine`. It accepts an explicit
+`EntityId → CardDefinition` assignment and a caller-selected RNG for future simulated game events.
+It preserves the opaque entity slots and ordered zones, rebuilds only definition-derived card
+components, and returns a typed unsupported result when an identity rewrite would cross in-flight
+references or runtime-bearing card state it cannot preserve safely. It has no viewer, deck model,
+or probability semantics: choosing candidate identities remains the caller's job. It also does
+not accept a `ClientGameState` or Gym observation; those are lossy views, while the materializer
+operates on a trusted complete engine-state substrate. The assignment map explicitly names the
+slots to rewrite; unmentioned slots remain unchanged because `GameState` itself cannot say which
+identities are epistemically unresolved. Callers that require hypothetical future randomness to be
+independent of an authoritative game must supply an independently derived `futureRng`; deliberately
+reusing the source state's RNG remains an explicit caller choice. Phase 8's existing
+`Determinizer` remains a separate AI policy path and is not migrated by this primitive.
+
 ---
 
 ## Scores are in raw evaluator units, everywhere above the leaf
--- a/docs/architecture-principles.md
+++ b/docs/architecture-principles.md
@@ -469,6 +469,29 @@ combat state, attachments — leaving only the immutable identity (`CardComponen
   the immutable `GameState` — modifying a component means creating a new container, which means
   creating a new entity map, which means creating a new `GameState`.
 
+#### Hypothetical hidden worlds
+
+`GameState` is always a complete engine state; the masked client and Gym DTOs are not alternate
+`GameState` representations and cannot be expanded back into one. A simulation caller that has
+already chosen hypothetical identities for hidden hand or library slots can apply that explicit
+assignment through `HiddenWorldMaterializer` in `rules-engine/hidden`.
+
+The materializer preserves each slot's `EntityId`, owner, controller-component state, zone
+membership, zone ordering, and every unrelated part of the state. It rebuilds the slot's
+definition-derived components through `CardEntityFactory`, and requires the caller to provide the
+RNG for future events in the hypothetical world. It does not choose hidden slots, inspect
+visibility, infer a deck, or define a sampling distribution. Identity rewrites are refused with a
+typed unsupported result when stack objects, pending decisions, continuation frames, or runtime
+state on the assigned hidden object make substitution unsafe.
+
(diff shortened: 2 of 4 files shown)


wingedsheep:
Thanks for this — the design here is right, and it's the design that shipped. Closing as superseded by #2153, which landed on `main` as `402c0001f1` while this was open and contains the same `HiddenWorldMaterializer` API plus a shared `HiddenSlotRewrite`. GitHub already reports this branch as conflicting; merging `main` into it produces add/add conflicts on both source files and content conflicts on both docs.

What carried over verbatim:

- the request as an explicit `EntityId -> CardDefinition` map with no sampling policy;
- the mandatory caller-supplied `futureRng`, so a hypothetical world can't silently inherit the authoritative stream;
- the typed `UnsupportedHiddenWorld` result instead of a superficially complete but incoherent state;
- and the best idea here — deriving the safe-component set from `CardEntityFactory.create(currentDefinition, ownerId)` rather than hand-listing it. That derivation is what let #2153 delete the AI `Determinizer`'s 12-component allowlist, so the two callers can't drift into two notions of "safe".

The battlefield -> hand -> graveyard regression around `LastKnownPermanentComponent` was a well-chosen test and survives unchanged.

Three behavioural differences in the merged version, for the record:

1. **A reveal is now a blocker, not something preserved.** This branch excluded `RevealedToComponent` from the runtime-state check and copied it onto the replacement, so a slot an opponent had already been shown could be rewritten underneath them — leaving that player holding a pointer to a card that was never there, which is the incoherence the class exists to prevent.
2. **A live stack pins only the slots its chosen targets name.** `hasInFlightReferences` here refused on any non-empty stack, which rules out a search root taken with a spell on the stack — the normal MCTS position. Targets are the one in-flight reference shape with a uniform representation (`TargetsComponent`), so they're audited per slot; `pendingDecision` and `continuationStack` stay whole-state refusals because their entity references have no common visitor.
3. **An empty assignment is gated too.** `slotAssignments.isNotEmpty() && hasInFlightReferences(state)` let an empty request through to `copy(rng = futureRng)`, rewriting the future random stream of a mid-decision state — the same hazard, minus the card swap.

If you want to keep going on this, the open seam is the one both versions name and neither closes: a visitor over continuation frames would turn that whole-state refusal into a per-slot pin like the stack check. That's the next real increment.

PR #2146, opened 2026-08-30
Unify card identity visibility

huxinq:
Argentum already has an engine-level `Visibility` service, but several consumers still answer a narrower question independently: “does this viewer know this card’s identity?” Those local approximations disagree in real states.

The clearest example is Gym. Its observation builder historically treated visibility mostly as a whole-zone property: your hand was visible, an opponent’s hand was hidden, and libraries were hidden. That works for the ordinary case, but Magic allows partial knowledge. If one card in an opponent’s two-card hand has been individually revealed to you, the hand is still a hidden zone, but you know one of its identities. Likewise, an effect may let only its controller look at the top card of a library, while another effect may reveal that top card publicly. The engine already represents these distinctions, but Gym collapsed all of them to “zone hidden” and therefore either omitted legally known cards or needed its own special cases.

The opposite problem appears with face-down objects. A battlefield or stack is normally public, but that does not make the underlying identity of a face-down permanent or face-down spell public. A Gym observation could therefore expose `"Craw Wurm"` or `"Hill Giant"` simply because the object lived in an otherwise public zone, even though an opponent should see only the face-down object’s public characteristics. The controller, meanwhile, is normally entitled to know what their own face-down object is. These are object-level identity rules, not zone-level rules.

This change makes that distinction explicit with `Visibility.isCardIdentityVisibleTo`. The query composes the identity rules Argentum already models: whole-zone access, individual `RevealedToComponent` knowledge, public and private top-of-library access, who may look at a face-down object, effect-granted face-down access, playable face-down cards in exile, and spectator restrictions.

Consumers still choose their own representation. For example, Gym now represents an opponent hand with one revealed card as a hidden zone of size two whose `cards` list contains only the one identity the perspective actually knows. It does not expose the unknown slot. A face-down battlefield object still exposes its public projected creature characteristics, such as its face-down P/T, but its `cardDefinitionId`, printed name, mana cost, and oracle text stay hidden from an opponent. `revealAll` remains an explicit debug escape hatch.

The same semantic question had also accumulated nearby ad hoc answers outside Gym. AI determinization separately excluded public top cards, individually revealed cards, and visible hands when deciding which hidden identities it could rewrite. Client modal-target presentation used owner/zone shortcuts for hidden card names. The deck tracker had its own approximation of whether the viewer knew which physical copy a card was. Combat-decision presentation masked face-down names using controller identity alone, so an explicit reveal could remain hidden in the decision UI even though the engine knew the viewer had learned the card.

Those paths now delegate identity entitlement to the same `Visibility` query. This removes cases where one subsystem knows a card while another hides it, or where one presentation leaks an identity another correctly masks.

The tests exercise the distinctions rather than just the helper itself. For example:

- with two cards in Player 2’s hand, revealing only the Mountain to Player 1 gives Player 1 exactly that identity while a third-player bystander still sees none;
- a private top-of-library effect reveals the top card only to its controller, while a public reveal exposes it to both players;
- a face-down `Craw Wurm` on the battlefield and a face-down `Hill Giant` on the stack expose their underlying identities to their controller but serialize no such names for the opponent;
- after a face-down combat object is explicitly revealed, server decision presentation stops masking its name;
- a modal spell targeting a hidden hand card initially shows only “a card in … hand”, then shows the real card name once that specific card is revealed;
- AI determinization now asks the same per-card visibility question before deciding whether an identity is safe to rewrite.

This deliberately does not create a shared client/Gym DTO, change Gym input-authority semantics, or merge presentation policy into the rules engine. `Visibility` owns only the semantic answer to whether a perspective may know an identity; Gym, client, server, and AI continue to own what they do with that answer.

Verification:

- `just test-class ObservationVisibilityTest`
- `just test-class ModalSpellStackVisibilityTest`
- `just test-class DeckListClientViewTest`
- `just test-class CombatDamageMaskingEnricherTest`
- `just test-class DeterminizerInvariantsTest`
- `just test-rules`
- `just test-ai`
- `just test-gym`
- `just test-server`

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
@@ -117,15 +117,10 @@ class Determinizer(
         viewerId: EntityId,
     ): List<EntityId> {
         val handKey = ZoneKey(opponentId, Zone.HAND)
-        val handHidden = !visibility.isZoneVisibleTo(state, handKey, viewerId)
-        val candidates = buildList {
-            addAll(state.getLibrary(opponentId))
-            if (handHidden) addAll(state.getHand(opponentId))
-        }
-        val visibleTop = state.getLibrary(opponentId).firstOrNull()?.takeIf {
-            visibility.revealsTopOfLibraryPublicly(state, opponentId) ||
-                (opponentId == viewerId && visibility.hasLookAtTopOfLibrary(state, viewerId))
-        }
+        val libraryKey = ZoneKey(opponentId, Zone.LIBRARY)
+        val library = state.getLibrary(opponentId)
+        val hand = state.getHand(opponentId)
+        val candidates = library + hand
         val referencedByStack = state.stack.flatMapTo(mutableSetOf()) { stackId ->
             state.getEntity(stackId)?.get<TargetsComponent>()?.targets.orEmpty().mapNotNull {
                 when (it) {
@@ -137,13 +132,13 @@ class Determinizer(
             }
         }
         return candidates.filter { id ->
-            id != visibleTop &&
+            val zoneKey = if (id in library) libraryKey else handKey
+            !visibility.isCardIdentityVisibleTo(state, zoneKey, id, viewerId) &&
                 id !in referencedByStack &&
                 // Continuation frames carry entity references in several different shapes.
                 // Quiet search roots normally have none; pinning while one exists is the safe
                 // fallback until those shapes share a common reference visitor.
                 state.continuationStack.isEmpty() &&
-                !visibility.isCardRevealedTo(state, id, viewerId) &&
                 isSafeToRewrite(state.getEntity(id))
         }
     }
--- a/docs/ai/architecture.md
+++ b/docs/ai/architecture.md
@@ -165,12 +165,13 @@ every candidate. Sampling before the one-ply simulation matters because pending 
 by that simulation can inspect hidden zones too. Sharing the world is essential: independently
 sampling candidates would make hidden-information variance look like move quality.
 
-`Visibility` in `rules-engine/view` is the common oracle for both client masking and AI
-determinization. The determinizer rewrites identities, never entities: entity IDs, zone membership,
-pending decisions, targets and continuations remain intact. Individually revealed cards and cards
-carrying runtime state are pinned. With a known decklist it samples from `decklist − seen`;
-otherwise it permutes the existing hidden multiset. That fallback removes exact hand and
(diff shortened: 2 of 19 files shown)


wingedsheep:
Superseded by #2156, which carries this work forward — thanks for it. The core idea is unchanged and intact: `Visibility.isCardIdentityVisibleTo` as the one answer to "may this perspective know this card's identity?", with Gym, the client, the deck tracker, server decision presentation and AI determinization all delegating to it instead of keeping their own approximations.

Reopening as a new PR rather than pushing here because `main` moved three times underneath this branch, twice landing overlapping work that needed real conflict resolution rather than a rebase:

- **#2153** extracted `HiddenSlotRewrite` and rewrote the same `Determinizer.hiddenCards` body. Resolved onto main's structure — `isCardIdentityVisibleTo` subsumes its `visibleTop`, `handHidden` and `!isCardRevealedTo` clauses exactly. Its `runtimeBlockers` also pins any slot carrying `RevealedToComponent`, which makes the reveal-preservation in `sample()` unnecessary.
- **#2154** reworked `StateDigest` and made the observation produce the facts it hashes.
- **#2155** closed the face-down Gym leak via a new `identityVisibleTo` in `FaceDownNaming`. #2156 keeps all of its structure — every `identityHidden` branch, the `nameVisibleToAll` masking, the projected-name fallback, ability-first stack kind inference, the CR 208.3 power/toughness gating — and swaps only the source of the answer, since `identityVisibleTo` is controller-only and misses explicit reveals, Lens of Clarity, may-play exile grants and foretell.

Review findings addressed on top:

- **CR 708.5 is scoped to the two zones it names.** "You can't look at face-down cards in any other zone" — but `playerWhoMayLookAtFaceDown` falls back through owner, so in exile it returned the owner, and the `Zone.EXILE && hasMayPlayFor` branch was dead for owned cards. Exile now has no controller baseline; a grant is the only way under such a card. Foretell is explicit alongside the may-play grants, since CR 702.143a says "that player may look at that card as long as it remains in exile" and `ForetellCardHandler` stamps `FaceDownComponent` purely to mask it from opponents.
- **A `Zone`-keyed overload** so callers stop synthesizing a `ZoneKey` whose owner defaulted to the viewer.
- **The ingredient queries are private** (`isCardRevealedTo`, `hasLookAtFaceDownCreatures`, `revealsTopOfLibraryPublicly`, `hasLookAtTopOfLibrary`) and `nameVisibleTo` is gone — handing those back out is how the next local approximation gets built.
- **Tests split by layer**, plus coverage for the spectator gating in `isZoneVisibleTo` and the exile cases above.

One thing worth recording: the observation in the description that Gym was collapsing partial knowledge to "zone hidden" was right, and it turned out to be the shallower half of the problem. The face-down direction — a public zone not making the identity underneath it public — is where the actual leaks were.

PR #2224, opened 2026-09-04
Expose typed entity references in actions and responses

huxinq:
Engine integrations that project or rewrite actions need to know which payload
values are entity references. Today they have to mirror every `GameAction` and
`DecisionResponse` shape by hand, or search serialized strings and risk treating
an ordinary decision ID, counter name, or `AbilityId` as an entity.

A cast action shows why this becomes fragile: one payload can reference the
spell, targets, mana sources, convoke creatures, damage recipients, and
entity-valued map keys. Repeated references matter too, because a transforming
adapter needs every location rather than a deduplicated set.

## What changes

`TypedEntityReferences` walks the registered serializers for `GameAction` and
`DecisionResponse` and returns every typed `EntityId` occurrence in serializer
order. Each occurrence includes a structural path through record fields,
collection indices, map key/value roles, and polymorphic payloads.

The existing in-flight reference projection now uses the same traversal. This
gives the engine one answer for which serialized values are entity references,
while preserving the materializer's set-oriented view.

Traversal is fail-closed: serializer failure or an unknown JSON-only payload is
reported as `Incomplete`, rather than silently appearing to contain no entity
references.

## Utility

A viewer-facing adapter can serialize an action normally, replace only the
reported entity leaves with safe aliases, and leave equal non-entity strings
untouched. The engine owns exhaustive schema traversal; the adapter still owns
visibility, aliasing, legality, and policy meaning.

This also removes duplicate schema knowledge from in-flight hidden-world
checks. Adding a new sealed action or response shape now reaches one generic
reference authority instead of requiring each consumer to rediscover its
payload.

## Verification

Run against upstream `b13310723a225fb81d72238c923e579128cd082c`:

- `TypedEntityReferencesTest`: 5 passed
- `InFlightEntityReferencesTest`: 3 passed
- `SerializationPolymorphicRegistrationTest`: 6 passed
- `HiddenWorldMaterializerTest`: 13 passed on the combined prospective tip
- `git diff --check`: clean

The structural paths describe serializer structure. A flattened polymorphic
JSON consumer may need to omit the explicit polymorphic-payload edge when
mapping a path onto encoded object fields.

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/core/InFlightEntityReferences.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/core/InFlightEntityReferences.kt
@@ -7,7 +7,9 @@ import com.wingedsheep.sdk.scripting.values.DynamicAmount
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.PolymorphicSerializer
 import kotlinx.serialization.SerializationStrategy
+import kotlinx.serialization.descriptors.PolymorphicKind
 import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.descriptors.StructureKind
 import kotlinx.serialization.encoding.AbstractEncoder
 import kotlinx.serialization.encoding.CompositeEncoder
 import kotlinx.serialization.encoding.Encoder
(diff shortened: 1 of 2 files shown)


wingedsheep:
Followed up in https://github.com/wingedsheep/argentum-engine/pull/2229

PR #2229, opened 2026-09-04
Expose typed entity references in actions and responses

wingedsheep:
Supersedes #2224 (by @huxinq), rebased onto current `main` with the review
cleanups applied.

## What this does

The engine could already tell which values inside a serialized payload are entity
references rather than ordinary strings, but only for persisted in-flight state,
and only as an unordered set. This generalises that serializer-driven traversal so
it also answers the question for `GameAction` and `DecisionResponse`, and so each
answer carries *where* in the payload the reference sat — map key vs. value, list
position, repeated occurrences kept distinct — instead of collapsing to membership.

A cast action is why the location matters: one payload can reference the spell,
its targets, mana sources, convoked creatures, damage recipients, and
entity-valued map keys, and a transforming adapter needs every occurrence rather
than a deduplicated set.

Traversal stays fail-closed. A serializer failure or an unknown JSON-only payload
is reported as `Incomplete`, never as an empty reference list.

## Changes on top of #2224

- **Collapsed the duplicate `Projection` hierarchy.** `InFlightEntityReferences`
  no longer carries its own structurally identical `Complete`/`Incomplete` pair
  plus a mapping `when`; it delegates to `TypedEntityReferences.project` and
  hidden-world consumers read the set view off
  `TypedEntityReferences.Projection.Complete.entityIds`. That's one fewer type to
  keep in sync, and drops ~30 lines of pure adapter.
- **Split the file.** `TypedEntityReferences` gets its own file rather than
  sitting first inside `InFlightEntityReferences.kt`.
- **Documented the map-ordinal caveat** on `PathSegment.MapEntry`. The KDoc
  already told a JSON consumer to drop `PolymorphicPayload`; it didn't say that a
  `Role.VALUE` occurrence carries no key and can only be located by ordinal, so
  resolving one is sound only while the consumer's document preserves
  serialization order. A `Role.KEY` occurrence *is* the encoded field name and has
  no such constraint.
- Import nit in the test.

## Open question for a follow-up

`action()`, `response()`, `Occurrence` and `PathSegment` still have no production
consumer — the viewer-facing aliasing adapter that motivates them isn't written
yet. Worth deciding when that lands whether the adapter actually wants *locations*
or wants *substitution*: if it's substitution, a rewriting encoder that maps
`EntityId` leaves through a caller-supplied alias function does the same job with
no path vocabulary and no dependence on JSON key order. Keeping the path shape for
now since it's what #2224 proposed and it's the more general of the two.

## Verification

- `:rules-engine:test` — green (full module suite, 3m13s).
- `TypedEntityReferencesTest` 5/5, `InFlightEntityReferencesTest` 3/3,
  `HiddenWorldMaterializerTest` 13/13, `HiddenSlotRewriteTest` 4/4.

The tests cover a maximal `CastSpell` with 12 distinct ids and one id repeated
across 7 payload positions; a nested `SubmitDecision` with `Map<EntityId,
List<EntityId>>` in both directions; an `ActivateAbility` whose `AbilityId` is
byte-identical to an `EntityId` and correctly isn't reported; and both fail-closed
paths. The first three additionally re-encode with the engine `Json` and assert
every reported path resolves to the expected id in the real output, so the path
vocabulary is proven against actual JSON rather than a hand-written expectation.

https://claude.ai/code/session_01P9F4SwiaWyQLPMsQzfvacH

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/core/InFlightEntityReferences.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/core/InFlightEntityReferences.kt
@@ -1,155 +1,32 @@
 package com.wingedsheep.engine.core
 
 import com.wingedsheep.engine.state.ComponentContainer
-import com.wingedsheep.sdk.model.CharacteristicValue
-import com.wingedsheep.sdk.model.EntityId
-import com.wingedsheep.sdk.scripting.values.DynamicAmount
-import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.PolymorphicSerializer
-import kotlinx.serialization.SerializationStrategy
-import kotlinx.serialization.descriptors.SerialDescriptor
-import kotlinx.serialization.encoding.AbstractEncoder
-import kotlinx.serialization.encoding.CompositeEncoder
-import kotlinx.serialization.encoding.Encoder
-
-@OptIn(ExperimentalSerializationApi::class)
-private val entityIdSerialName = EntityId.serializer().descriptor.serialName
-
-@OptIn(ExperimentalSerializationApi::class)
-private val characteristicValueSerialName = CharacteristicValue.serializer().descriptor.serialName
 
 /** Internal seam for testing consumers' fail-closed response to projection errors. */
 internal interface InFlightReferenceProjector {
-    fun project(stackObject: ComponentContainer): InFlightEntityReferences.Projection
-    fun project(decision: PendingDecision): InFlightEntityReferences.Projection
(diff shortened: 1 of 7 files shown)


huxinq:
One downstream data point: MTGallium’s current adapter already has essentially this use case. It serializes actions/responses normally, then rewrites only typed EntityId positions to perspective-safe semantic references.

So for this consumer, substitution is the useful abstraction; paths are implementation detail. One wrinkle is that the mapping is intentionally non-injective, so rewriting map keys must preserve collisions/multiplicity rather than overwrite them.

A good follow-up test would be whether the upstream API can replace MTGallium’s hand-maintained typed-reference walker without changing its canonical semantic payloads.

PR #2238, opened 2026-09-05
Avoid per-card entity-map copies when constructing hidden worlds

huxinq:
Hidden-world construction currently copies the entire entity map for each rewritten card. A world with dozens of hidden slots therefore rebuilds the same map dozens of times before any caller can observe the result, adding allocation to repeated search sampling.

In MTGallium's earlier strict known-deck sampler, the same batching strategy produced **3.45× as many samples per unit of thread CPU**, with **69% less allocation**:

| Mean cost per sampled world | Per-card copies | Batched copies | Reduction |
| --- | ---: | ---: | ---: |
| Thread CPU time | 306.95 µs | 88.85 µs | 71.1% |
| Allocated bytes | 1,025,524 | 316,209 | 69.2% |

This comparison used one fixed initial state in one JVM, 1,000 warmup calls per variant, then 10 old/new/new/old rounds with 100 calls per batch: 2,000 measured calls per variant. All 64 parity-check worlds matched exactly. These measurements illustrate the benefit in the earlier sampler; they measure local sampling cost, and have not been repeated on this upstream port or established as an end-to-end search speedup.

Batch the replacements into a private map and publish one immutable state after the batch: once per player in `Determinizer`, and once per request in `HiddenWorldMaterializer`. A container-level overload of `HiddenSlotRewrite.rewrite` keeps the existing card reconstruction and controller-preservation logic shared with the single-slot API.

Sampling choices, RNG use, entity iteration order, and library shuffling retain their existing behavior. Explicit materialization still validates against the original state in sorted slot order and returns the first refusal without publishing a partial world. Empty assignments retain the original entity map after the in-flight completeness check.

Verification: all 28 tests passed across the three focused suites on `55674a017f`. All 14 materializer tests passed again after the empty-assignment fast path was added in `47edcee60d`, including map reuse and unchanged refusal for an incomplete in-flight graph.

```sh
just test-class DeterminizerInvariantsTest :rules-engine:test \
  --tests '*HiddenSlotRewriteTest' --tests '*HiddenWorldMaterializerTest'
```

The sampler regression matches a fingerprint captured from upstream `12589ae4a2` across 64 cases covering identity permutation, known decklists, both viewers, and 16 seeds. The materializer regressions cover input and returned-world immutability, entity iteration order, and refusal after an earlier valid assignment; the existing suites exercise visibility, in-flight pins, controller absence, slot preservation, and future RNG installation.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
@@ -92,10 +92,13 @@ class Determinizer internal constructor(
             }
 
             if (definitions.size != hidden.size) continue
+            // No caller observes intermediate identities. Copy once for this player and publish
+            // only after all replacements, without mutating an input or previously sampled world.
+            val sampledEntities = sampled.entities.toMutableMap()
             for ((entityId, cardDef) in hidden.zip(definitions)) {
-                val old = sampled.getEntity(entityId) ?: continue
+                val old = sampledEntities[entityId] ?: continue
                 val ownerId = old.get<OwnerComponent>()?.playerId ?: opponentId
-                sampled = HiddenSlotRewrite.rewrite(sampled, entityId, cardDef, ownerId)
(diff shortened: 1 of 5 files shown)


huxinq:

PR #2239, opened 2026-09-05
Reuse validated card containers during hidden-world materialization

huxinq:
Hidden-world materialization builds each replacement card twice: first to validate the factory-generated definition ID against the registry, then again to install the card. The first container is already suitable for installation once the source slot's controller semantics are applied.

Reuse that validated container through an internal `HiddenSlotRewrite` overload. The existing definition-based overload delegates to the same controller-preservation logic, including retaining an absent controller. Registry checks, failure order, and atomic publication remain unchanged.

Depends on #2238 for the container-level rewrite API. Until that PR merges, the main-based diff includes its changes. The [follow-on diff](https://github.com/huxinq/argentum-engine/compare/47edcee60d9c25895d00bcf43911f99f26e4e5a7...e7746fcd070cb412cce9d3d22da8ea8c22a9d37d) isolates this optimization: three files, 13 additions and 6 deletions.

A matched benchmark against the already-batched materializer showed:

| Assigned slots | CPU time per materialization | Allocated bytes per materialization |
| --- | --- | --- |
| 8 | 12.692 → 10.551 µs (**16.9% less**) | 43,902 → 32,477 (**26.0% less**) |
| 104 | 170.674 → 145.051 µs (**15.0% less**) | 546,136 → 397,624 (**27.2% less**) |

These are two synthetic positions in one JVM, measured with current-thread CPU and allocation counters and coverage disabled. Each fixture used 1,000 warmup calls per variant and ten old/new/new/old rounds of 100 calls per batch (2,000 measured calls per variant). All ten rounds favored reuse for both fixtures, and all 128 paired full results matched. Benchmark treatment: `97e350764b`, baseline: `55674a017f`; these are local materialization costs, not end-to-end search timings.

The three focused suites (`HiddenWorldMaterializerTest`, `HiddenSlotRewriteTest`, `DeterminizerInvariantsTest`) passed for the reuse change. All 14 materializer tests passed again after integrating the parent PR's empty-request fast path. Added assertions cover the factory's canonical definition ID and controller presence; existing tests cover controller absence, definition-derived components, and atomic refusal.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
@@ -92,10 +92,13 @@ class Determinizer internal constructor(
             }
 
             if (definitions.size != hidden.size) continue
+            // No caller observes intermediate identities. Copy once for this player and publish
+            // only after all replacements, without mutating an input or previously sampled world.
+            val sampledEntities = sampled.entities.toMutableMap()
             for ((entityId, cardDef) in hidden.zip(definitions)) {
-                val old = sampled.getEntity(entityId) ?: continue
+                val old = sampledEntities[entityId] ?: continue
                 val ownerId = old.get<OwnerComponent>()?.playerId ?: opponentId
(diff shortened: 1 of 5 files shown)


huxinq:

PR #2240, opened 2026-09-05
Index zone membership for multi-card hidden-world assignments

huxinq:
Materializing several hidden cards currently rescans every zone list for each assignment, then scans the matching lists again to count occurrences. A large hidden-world request repeats this work for nearly every card in the game.

Build a request-local index of zone occurrences for requests with more than two assignments. Retain one entry per occurrence, including duplicate IDs within a zone, so malformed states receive the same missing, duplicate, and unsupported-zone refusals. Validation still chooses the first obstruction in sorted assignment order. One- and two-card requests retain the original scan to avoid index setup cost; empty requests keep their existing fast path.

Depends on #2239 (which builds on #2238). Until those changes merge, the main-based diff includes its parents. The [follow-on diff](https://github.com/huxinq/argentum-engine/compare/e7746fcd070cb412cce9d3d22da8ea8c22a9d37d...27e95f23e4bd355211a8b986a9e846e388b99ddf) isolates this optimization. The benchmark baseline includes the parent change so these measurements isolate the additional zone-lookup optimization.

A matched in-JVM benchmark of `27e95f23e4` against `e7746fcd07` produced:

| Assigned / available hidden slots | CPU µs, before → after | Allocated bytes, before → after |
| --- | ---: | ---: |
| 8 / 8 | 12.139 → 8.964 | 31,325 → 29,421 |
| 104 / 104 | 143.412 → 81.400 | 382,648 → 355,080 |
| 1 / 104 | 2.863 → 2.930 | 9,560 → 9,560 |
| 2 / 104 | 4.254 → 4.370 | 14,112 → 14,112 |
| 3 / 104 | 5.775 → 5.275 | 17,400 → 18,552 |
| 4 / 104 | 6.934 → 5.785 | 21,040 → 21,888 |
| 8 / 104 | 12.441 → 9.190 | 35,464 → 35,096 |

The dense 104-card case used **43.2% less thread CPU and 7.2% less allocation**. Small requests show the tradeoff: one/two assignments keep allocation unchanged and timed about 2–3% slower in this run; three/four improve CPU while allocating about 4–7% more. These are synthetic materialization measurements, not end-to-end search timings.

Each fixture used 1,000 warmup calls per variant followed by ten old/new/new/old rounds with 100 calls per batch (2,000 measured calls per variant), using thread CPU/allocation counters with coverage instrumentation disabled. All **448 full-result parity comparisons** matched.

Verification: all **15 `HiddenWorldMaterializerTest` tests passed**. The new malformed-state regression checks missing membership, duplicate IDs in one zone, duplicate IDs across zones, unsupported zones, and refusal ordering on both indexed and direct-scan paths.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
@@ -92,10 +92,13 @@ class Determinizer internal constructor(
             }
 
             if (definitions.size != hidden.size) continue
+            // No caller observes intermediate identities. Copy once for this player and publish
+            // only after all replacements, without mutating an input or previously sampled world.
+            val sampledEntities = sampled.entities.toMutableMap()
             for ((entityId, cardDef) in hidden.zip(definitions)) {
-                val old = sampled.getEntity(entityId) ?: continue
+                val old = sampledEntities[entityId] ?: continue
                 val ownerId = old.get<OwnerComponent>()?.playerId ?: opponentId
-                sampled = HiddenSlotRewrite.rewrite(sampled, entityId, cardDef, ownerId)
+                sampledEntities[entityId] = HiddenSlotRewrite.rewrite(old, cardDef, ownerId)
             }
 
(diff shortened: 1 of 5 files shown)


huxinq:

PR #2241, opened 2026-09-05
Analyze in-flight pins once per sampled position

huxinq:
`Determinizer` analyzes the same source position's in-flight references once for every player. A two-player sample therefore traverses a paused decision and its continuations twice before reconstructing hidden cards.

Compute that analysis once per sample and share the resulting pin set across players. Incomplete analysis still returns the original state without sampling, and an empty turn order still bypasses analysis. Candidate selection, visibility, reconstruction, and RNG ordering retain their existing behavior.

Depends on #2238, which batches hidden-card reconstruction. Until it merges, the main-based diff includes its changes. The [follow-on diff](https://github.com/huxinq/argentum-engine/compare/47edcee60d9c25895d00bcf43911f99f26e4e5a7...5fee28bec7ccd0efacc9bdae6d0eba524e65899f) isolates this optimization. The comparison below measures the additional effect on that already-batched sampler.

### Measured example

A short matched comparison used two-player synthetic positions with 8 or 104 hand/library slots, both with ordinary state and paused Mind Rot resolution. It alternated viewers and compared the previous sampler at `47edcee` with this change at `5fee28b`. The baseline implementation ran alongside the candidate in the same JVM, with coverage instrumentation disabled.

| Position | CPU per sample, before → after | Allocated bytes, before → after |
| --- | ---: | ---: |
| Ordinary, 8 slots | 8.651 → 8.614 µs (−0.4%) | 23,227 → 23,137 |
| Ordinary, 104 slots | 83.471 → 87.923 µs (+5.3%) | 340,024 → 339,928 |
| Paused Mind Rot, 8 slots | 26.301 → 15.806 µs (−39.9%) | 87,274 → 52,410 (−39.9%) |
| Paused Mind Rot, 104 slots | 101.240 → 94.172 µs (−7.0%) | 403,720 → 369,304 (−8.5%) |

Each fixture used 1,000 warmup calls per variant followed by ten old/new/new/old rounds of 100 calls per batch: 2,000 measured calls per variant. CPU is current-thread CPU time; allocation uses JVM thread allocation counters. The ordinary 104-slot mean includes a slow candidate batch despite the candidate having lower aggregate CPU in nine of ten rounds. These measurements support reduced work on the paused fixtures; they do not establish an ordinary-position or end-to-end search speedup. This exercises Argentum's built-in sampler, not MTGallium's explicit materializer path.

### Verification

All 12 `DeterminizerInvariantsTest` tests pass. New regressions require one analysis across players, one incomplete analysis with unchanged output, and no analysis for an empty turn order. Existing coverage includes both sampling models, both viewers, sixteen seeds, paused references, and entity/library ordering. The benchmark additionally checked full state equality and entity iteration order for 64 pairs per fixture (256 pairs total), with no failed or excluded samples.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
@@ -92,10 +92,13 @@ class Determinizer internal constructor(
             }
 
             if (definitions.size != hidden.size) continue
+            // No caller observes intermediate identities. Copy once for this player and publish
+            // only after all replacements, without mutating an input or previously sampled world.
+            val sampledEntities = sampled.entities.toMutableMap()
             for ((entityId, cardDef) in hidden.zip(definitions)) {
-                val old = sampled.getEntity(entityId) ?: continue
+                val old = sampledEntities[entityId] ?: continue
                 val ownerId = old.get<OwnerComponent>()?.playerId ?: opponentId
-                sampled = HiddenSlotRewrite.rewrite(sampled, entityId, cardDef, ownerId)
+                sampledEntities[entityId] = HiddenSlotRewrite.rewrite(old, cardDef, ownerId)
             }
 
             val libraryKey = ZoneKey(opponentId, Zone.LIBRARY)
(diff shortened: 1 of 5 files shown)


huxinq:

PR #2242, opened 2026-09-05
Reduce library membership scans during hidden-card sampling

huxinq:
`Determinizer` repeatedly searches a library list while classifying hidden cards, then builds a hidden/library intersection before shuffling. As libraries grow, these repeated membership scans do unnecessary work.

Use a local set for library membership during classification. For shuffling, use the hidden-ID set directly: both consumers already iterate the library, so intersecting with it first is redundant. The original lists still determine candidate order, shuffle inputs, and reconstruction order, including duplicate occurrences.

Depends on #2238's batched sampler and remains independent of #2241's in-flight analysis optimization. Until its parents merge, the main-based diff includes their changes. The [follow-on diff](https://github.com/huxinq/argentum-engine/compare/47edcee60d9c25895d00bcf43911f99f26e4e5a7...a554b885d049767918d91f2337f507475a3fec02) isolates this optimization.

### Measured example

A short comparison ran the baseline sampler at `47edcee` alongside the candidate at `a554b88` in one JVM with coverage disabled. It used two-player synthetic positions with 8 or 104 hand/library slots, ordinary state or paused Mind Rot resolution, and alternating viewers.

| Position | CPU per sample, before → after | Allocated bytes, before → after |
| --- | ---: | ---: |
| Ordinary, 8 slots | 9.886 → 10.375 µs (+4.9%) | 23,598 → 23,992 (+1.7%) |
| Ordinary, 104 slots | 88.321 → 80.752 µs (−8.6%) | 340,144 → 342,832 (+0.8%) |
| Paused Mind Rot, 8 slots | 31.206 → 31.597 µs (+1.3%) | 88,606 → 88,781 (+0.2%) |
| Paused Mind Rot, 104 slots | 104.382 → 92.327 µs (−11.5%) | 404,128 → 406,792 (+0.7%) |

The larger fixtures used less CPU at a small allocation cost; the small fixtures regressed. These are current-thread CPU and allocation measurements for these positions, not an end-to-end search speedup. Each fixture had 1,000 warmup calls per variant and ten old/new/new/old rounds of 100 calls per batch, totaling 2,000 measured calls per variant. Larger-fixture CPU favored the candidate in nine of ten ordinary rounds and all ten paused rounds; the ordinary mean includes one slow candidate round.

### Verification

All 11 `DeterminizerInvariantsTest` tests pass, including the existing 64-case sampling fingerprint and a new regression preserving duplicate library entries and repeated player traversal. The benchmark checked full state equality, entity iteration order, and source preservation for 64 pairs per fixture (256 total). All fixtures completed without exclusions. The measurement isolates this change from the other performance PRs.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
@@ -92,10 +92,13 @@ class Determinizer internal constructor(
             }
 
             if (definitions.size != hidden.size) continue
+            // No caller observes intermediate identities. Copy once for this player and publish
+            // only after all replacements, without mutating an input or previously sampled world.
+            val sampledEntities = sampled.entities.toMutableMap()
             for ((entityId, cardDef) in hidden.zip(definitions)) {
-                val old = sampled.getEntity(entityId) ?: continue
+                val old = sampledEntities[entityId] ?: continue
                 val ownerId = old.get<OwnerComponent>()?.playerId ?: opponentId
-                sampled = HiddenSlotRewrite.rewrite(sampled, entityId, cardDef, ownerId)
+                sampledEntities[entityId] = HiddenSlotRewrite.rewrite(old, cardDef, ownerId)
             }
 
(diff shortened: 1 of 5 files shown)


huxinq:

PR #2243, opened 2026-09-05
Reuse source-definition defaults within materialization requests

huxinq:
`HiddenWorldMaterializer` reconstructs a source card's factory defaults for every assigned slot to check for runtime components. Multiple copies of the same card under the same owner repeat that construction even though their expected defaults are identical.

Reuse those defaults within one materialization request, keyed by the exact registered source-definition ID and owner. Every slot still resolves its definition and compares its own components against the defaults; a runtime blocker on one copy still refuses the request at that slot. The cache ends with the call, so the next request observes registry updates. Replacement construction and validation remain separate.

Depends on #2240's zone lookup optimization and measures the additional effect on that materializer. Until its parents merge, the main-based diff includes their changes. The [follow-on diff](https://github.com/huxinq/argentum-engine/compare/27e95f23e4bd355211a8b986a9e846e388b99ddf...3a3183f82d9c4d499f3e9379214dda0b16dd5e48) isolates this optimization.

### Measured example

A short matched comparison ran the baseline at `27e95f2` and this change at `3a3183f` in one JVM with coverage disabled. Synthetic two-player positions had 8 or 104 hand/library slots and dense or sparse assignments. The repeated-definition cases use Grizzly Bears in hands and Forest in libraries; the distinct cases give each source slot a separately registered synthetic definition.

With 104 assignments and four source-definition/owner combinations, expected-container construction falls from 104 calls to four. That fixture used **21.0% less CPU and 29.9% less allocation**. Cache misses add overhead, visible in the distinct-definition and single-assignment cases:

| Assigned slots | Source definitions | CPU per request, before → after | Allocated bytes, before → after |
| --- | --- | ---: | ---: |
| 8 of 8 | Repeated | 11.570 → 10.257 µs (-11.3%) | 30,573 → 25,966 (-15.1%) |
| 8 of 8 | Distinct | 6.666 → 6.908 µs (+3.6%) | 30,632 → 31,416 (+2.6%) |
| 104 of 104 | Repeated | 76.163 → 60.170 µs (-21.0%) | 370,056 → 259,464 (-29.9%) |
| 104 of 104 | Distinct | 82.251 → 82.865 µs (+0.7%) | 371,656 → 382,104 (+2.8%) |
| 1 of 104 | Repeated | 3.014 → 3.115 µs (+3.4%) | 9,731 → 9,957 (+2.3%) |
| 1 of 104 | Distinct | 2.777 → 2.823 µs (+1.7%) | 9,704 → 9,928 (+2.3%) |
| 2 of 104 | Repeated | 4.137 → 3.990 µs (-3.6%) | 14,400 → 13,192 (-8.4%) |
| 2 of 104 | Distinct | 4.205 → 4.308 µs (+2.5%) | 14,400 → 14,704 (+2.1%) |

Each fixture had 1,000 warmup calls per variant, then ten old/new/new/old rounds with 100 calls per batch: 2,000 measured calls per variant. CPU and allocation are current-thread JVM counters. These measurements establish the tradeoff for these synthetic fixtures; they do not measure typical deck composition or end-to-end search performance.

### Verification

All 31 focused tests pass across `HiddenWorldMaterializerTest`, `HiddenSlotRewriteTest`, and `DeterminizerInvariantsTest`. New regressions exercise repeated definitions under different owners, a later revealed copy refusing after an earlier valid copy, and a registry overlay changing a definition between requests. The public blocker API retains its definition/owner form; the overload accepting factory defaults is internal to the engine.

The benchmark checked full result equality, entity iteration order, and source preservation for 64 pairs per fixture (512 total). All eight fixtures completed, without failed or excluded samples.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
@@ -92,10 +92,13 @@ class Determinizer internal constructor(
             }
 
             if (definitions.size != hidden.size) continue
+            // No caller observes intermediate identities. Copy once for this player and publish
+            // only after all replacements, without mutating an input or previously sampled world.
+            val sampledEntities = sampled.entities.toMutableMap()
             for ((entityId, cardDef) in hidden.zip(definitions)) {
-                val old = sampled.getEntity(entityId) ?: continue
+                val old = sampledEntities[entityId] ?: continue
                 val ownerId = old.get<OwnerComponent>()?.playerId ?: opponentId
-                sampled = HiddenSlotRewrite.rewrite(sampled, entityId, cardDef, ownerId)
+                sampledEntities[entityId] = HiddenSlotRewrite.rewrite(old, cardDef, ownerId)
             }
 
             val libraryKey = ZoneKey(opponentId, Zone.LIBRARY)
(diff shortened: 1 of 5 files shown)


huxinq:

PR #2245, opened 2026-09-05
Assemble triggered-ability grants without repeated list copies

huxinq:
`TriggerAbilityResolver` combines thirteen sources of granted abilities by repeatedly concatenating lists. Each concatenation copies the accumulated grants, including empty-list allocations when no grants apply. Both ordinary trigger lookup and the lookup using indexed providers take this path.

Collect the already-resolved lists once with `buildList`, in their existing order. Duplicate grants remain duplicate abilities. When no grants apply, assembly retains the existing base list; text replacement still runs afterward. Provider lookup, duration checks, and filtering happen at the same points as before.

### Measured example

The baseline at `12589ae4` and candidate at `22c7d8a` ran side by side in one JVM with coverage disabled. Eight synthetic fixtures exercised regular and indexed lookup across eight target creatures. Battlefield indexes were prebuilt and projection was warm, matching callers that reuse them across lookups.

| Lookup path | Abilities per target | CPU per lookup, before → after | Allocated bytes, before → after |
| --- | --- | ---: | ---: |
| Regular | 0 base, 0 temporary, 0 static | 0.962 → 0.646 µs (−32.8%) | 1,278 → 966 (−24.4%) |
| Indexed providers | 0 base, 0 temporary, 0 static | 0.585 → 0.391 µs (−33.2%) | 936 → 608 (−35.0%) |
| Regular | 1 base, 0 temporary, 0 static | 0.570 → 0.399 µs (−30.0%) | 1,208 → 896 (−25.8%) |
| Indexed providers | 1 base, 0 temporary, 0 static | 0.492 → 0.391 µs (−20.4%) | 880 → 552 (−37.3%) |
| Regular | 1 base, 2 temporary, 1 static | 1.497 → 1.012 µs (−32.4%) | 2,090 → 1,200 (−42.6%) |
| Indexed providers | 1 base, 2 temporary, 1 static | 1.288 → 0.906 µs (−29.7%) | 1,789 → 910 (−49.1%) |
| Regular | 1 base, 8 temporary, 1 static | 1.271 → 0.884 µs (−30.5%) | 2,624 → 1,240 (−52.7%) |
| Indexed providers | 1 base, 8 temporary, 1 static | 0.648 → 0.609 µs (−5.9%) | 2,376 → 936 (−60.6%) |

Each fixture used 10,000 warmup calls per variant, then ten old/new/new/old rounds with 500 calls per batch (10,000 measured calls per variant). CPU and allocation use current-thread JVM counters. Allocation was lower in every round; CPU was lower in every round for seven fixtures and in seven of ten rounds for the indexed base-only fixture. These measurements describe ability lookup with reused indexes and warm projection; they do not measure complete event processing or gameplay speedup.

### Verification

All 33 tests pass across `TriggerAbilityResolverTest`, `TriggerDetectorBatchTriggerTest`, and `EntityMatchesAttachmentScenarioTest`. The new regression checks both lookup paths for base-list identity without grants and exact base → duplicate temporary → static ability order with grants. Existing tests exercise downstream batch triggers and attachment-granted execution.

The benchmark compared complete ordered ability lists for 64 calls per fixture, cycling through the eight targets, and checked source preservation. All eight fixtures completed without failures or exclusions. This PR is based directly on upstream `12589ae4` and is independent of the projection and hidden-world optimization PRs.

### DIFF
--- a/rules-engine/src/main/kotlin/com/wingedsheep/engine/event/TriggerAbilityResolver.kt
+++ b/rules-engine/src/main/kotlin/com/wingedsheep/engine/event/TriggerAbilityResolver.kt
@@ -125,10 +125,21 @@ class TriggerAbilityResolver(
         // gates it, the printed KeywordAbility.Numeric supplies N.
         val renownAbilities = getRenownTriggeredAbilities(entityId, cardDefinitionId, state)
 
-        val allGranted = grantedAbilities + staticGrantedAbilities + attachedGrantedAbilities +
-            selfGrantedAbilities + wardAbilities + flankingAbilities + ringBearerAbilities +
-            suspendAbilities + paradigmAbilities + siegeAbilities + vanishingAbilities +
-            fabricateAbilities + renownAbilities
+        val allGranted = buildList {
+            addAll(grantedAbilities)
+            addAll(staticGrantedAbilities)
+            addAll(attachedGrantedAbilities)
+            addAll(selfGrantedAbilities)
+            addAll(wardAbilities)
+            addAll(flankingAbilities)
+            addAll(ringBearerAbilities)
+            addAll(suspendAbilities)
+            addAll(paradigmAbilities)
+            addAll(siegeAbilities)
+            addAll(vanishingAbilities)
+            addAll(fabricateAbilities)
+            addAll(renownAbilities)
+        }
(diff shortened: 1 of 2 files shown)


huxinq:

PR #2246, opened 2026-09-05
Reuse the component type index when validating hidden slots

huxinq:
`HiddenSlotRewrite.runtimeBlockers` rebuilds a class-to-component map for every validation even though the expected factory container already holds that index. It then creates intermediate lists to exclude card identity and matching components before collecting blocker names.

Read the existing component index and collect blocker names in one pass. The lookup still uses each source component's runtime Java class, `CardComponent` remains excluded, and the result keeps the same sorted names. The container's internal map supplies this lookup without a new public API.

This relies on factory-built expectations: both production paths obtain them from `CardEntityFactory.create`, whose insertions use runtime-class keys. `ComponentContainer` does not guarantee that invariant for arbitrary maps or widened `with<T>()` insertions. Source containers are still enumerated by value, so their stored keys do not affect the comparison.

Depends on #2243's internal blocker overload for reusing factory defaults. Until its parents merge, the main-based diff includes their changes. The [follow-on diff](https://github.com/huxinq/argentum-engine/compare/3a3183f82d9c4d499f3e9379214dda0b16dd5e48...d39360ea616754c5fe23b636ba63573932cc3b71) isolates this optimization. The comparison below measures this further allocation reduction.

### Measured example

The baseline at `3a3183f` and candidate at `d39360e` ran side by side in one JVM with coverage disabled. The baseline retained both its materializer and its blocker helper, so it did not share the optimized helper. Eight synthetic two-player materialization fixtures exercised repeated and distinct source definitions with dense and sparse assignments.

| Assigned slots | Source definitions | CPU per request, before → after | Allocated bytes, before → after |
| --- | --- | ---: | ---: |
| 8 of 8 | Repeated | 9.402 → 8.943 µs (−4.9%) | 25,585 → 22,953 (−10.3%) |
| 8 of 8 | Distinct | 7.530 → 7.044 µs (−6.5%) | 31,416 → 28,664 (−8.8%) |
| 104 of 104 | Repeated | 60.020 → 55.282 µs (−7.9%) | 259,464 → 226,088 (−12.9%) |
| 104 of 104 | Distinct | 83.537 → 77.592 µs (−7.1%) | 382,104 → 346,328 (−9.4%) |
| 1 of 104 | Repeated | 3.059 → 3.014 µs (−1.5%) | 9,955 → 9,610 (−3.5%) |
| 1 of 104 | Distinct | 2.979 → 2.937 µs (−1.4%) | 9,928 → 9,584 (−3.5%) |
| 2 of 104 | Repeated | 4.297 → 4.177 µs (−2.8%) | 13,192 → 12,528 (−5.0%) |
| 2 of 104 | Distinct | 4.441 → 4.340 µs (−2.3%) | 14,704 → 14,016 (−4.7%) |

Each fixture had 1,000 warmup calls per variant and ten old/new/new/old rounds of 100 calls per batch: 2,000 measured calls per variant. CPU and allocation use current-thread JVM counters. All eight aggregate means improved; both dense 104-slot cases used less CPU and allocation in every round. The smaller CPU differences are less conclusive: several sparse cases had two slower rounds. These are single-JVM synthetic materializer measurements, not an end-to-end gameplay or search speedup.

### Verification

All 32 focused tests pass across `HiddenSlotRewriteTest`, `HiddenWorldMaterializerTest`, and `DeterminizerInvariantsTest`. The new regression checks that a printed card-name difference is ignored, matching definition-derived Madness remains safe, and a changed Madness cost plus a reveal produces the same sorted blocker names.

The benchmark checked full result equality, entity iteration order, and source preservation for 64 pairs per fixture (512 total). All eight fixtures completed without failures or exclusions.

### DIFF
--- a/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
+++ b/ai/src/main/kotlin/com/wingedsheep/ai/engine/hidden/Determinizer.kt
@@ -92,10 +92,13 @@ class Determinizer internal constructor(
             }
 
             if (definitions.size != hidden.size) continue
+            // No caller observes intermediate identities. Copy once for this player and publish
+            // only after all replacements, without mutating an input or previously sampled world.
+            val sampledEntities = sampled.entities.toMutableMap()
             for ((entityId, cardDef) in hidden.zip(definitions)) {
-                val old = sampled.getEntity(entityId) ?: continue
+                val old = sampledEntities[entityId] ?: continue
                 val ownerId = old.get<OwnerComponent>()?.playerId ?: opponentId
-                sampled = HiddenSlotRewrite.rewrite(sampled, entityId, cardDef, ownerId)
+                sampledEntities[entityId] = HiddenSlotRewrite.rewrite(old, cardDef, ownerId)
             }
 
             val libraryKey = ZoneKey(opponentId, Zone.LIBRARY)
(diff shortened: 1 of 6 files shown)


huxinq:

